PR #53, opened 2021-03-04
Add application model conceptual document

rynowak:
Fixes: #13


AaronCrawfis (changes requested):
Overall looks great. Eventually we may need to split into a couple pages but for now this looks good.

PR #66, opened 2021-03-09
Update workflow to only run when docs is updated

AaronCrawfis:


rynowak (approved):
wow, this is an impressive PR to do in a meeting :shipit:

PR #69, opened 2021-03-09
Remove use of service principals

rynowak:
Fixes: radius-project/radius#40
Fixes: radius-project/core-team#996

This change removes service principals from our environment creation
step and replaces them with managed identity.

There's a pretty dramatic simplification of the environment setup code
here from SP removal, which was the root cause for a bunch of complexity
like retries.

This also upgrades all of the dependencies. We needed an upgrade to the
go-autorest package to resolve issues around managed identity.

Also removed the SSH keys from the ARM template for k8s. That isn't
required and was another source of non-idempotency.


kachawla (approved):
looks good to me!

vinayada1 (approved):
e2e-tests are passing!

PR #72, opened 2021-03-11
Adding delete environment operation

kachawla:
Deletes environment by deleting associated resource group and all its resources.

Related issue: https://github.com/Azure/radius/issues/12


rynowak (approved):
Great 👍

PR #73, opened 2021-03-11
Incorportate example applications into docs

AaronCrawfis:
Created custom rad shortcode and incorporates examples into docs


rynowak (approved):
I'm happy to see this get merged as-is. 

I'll do some cleanup on the samples/tests separately. The issue is that not all of these are *known good*. Some of them include Lima (not installed by our env setup) and some of them are just plain outdated. These have kinda been a dumping ground for me for whatever sample code I want to have on hand 🤣

PR #87, opened 2021-03-13
Add troubleshooting page

AaronCrawfis:
Closes #38


rynowak (approved):
Happy to take this in as-is. I think over time we'll replace some of the problems here with the ones people commonly encounter. 

A lot of the stuff mentioned here is like ... wierd errors that no one will see again. Sorry to break to you @msfussell but you are cursed.

PR #118, opened 2021-03-18
Add service bus to control plane

vinayada1:


rynowak (changes requested):
I left some small comments here and there but this mostly looks like what I'd expect. Love it!

rynowak (approved):

PR #135, opened 2021-03-25
Add some basic schemas

rynowak:
I spend some time during meetings putting together schemas for a few
types we accept today. The idea is that we'll use these schemas for:

- Bicep intellisense
- SDK generation
- Validation (of payloads)
- Generating types for our code

We'll need to really *scale up* our ability to work with new data types,
and probably invest in a lot of our own logic and tooling around these
schemas.

I wanted to put something basic out there that establishes the pattern
so we can maintain these and add more as we light up new data types.

The work to start using these is scheduled to start relatively soon, and
I think we're at the tipping point where we should think about becoming
more data driven.


kachawla (commented):
Would it make sense to write a short document around why we need these schemas, how/where they are consumed etc? This will help keeping context around for anyone who is interested in knowing more or who joins the team at a later point.

kachawla (commented):
I already started the review so just posting a few clarifying general questions that I had, we can discuss them during the sync tomorrow if that's easier

kachawla (approved):

AaronCrawfis (approved):

PR #138, opened 2021-03-26
Bootstrap our code generation via OpenAPI

rynowak:
Adds Makefile support. Run `make generate` to update our generated
client and test mocks.

Note: it's conventional in Go to check in generated code.

Adds a basic OpenAPI schema for our Application resource.


kachawla (approved):
Didn't really go through everything since we touched upon most of it during the sync earlier. We can work through the changes as we see them while using it.

PR #141, opened 2021-03-29
Deploy RP as part of deployment tests

rynowak:
Fixes: #83

This change reworks our build pipeline so that we build binaries and
containers prior to running deployment tests. This allows us to use
those assets for deployment tests.

In addition to this, we also build and push the image for the RP as part
of the RP workflow so that we can use it in tests.

This change will use the `az` cli to update the webapp of the deployment
test environment prior to running tests. We don't need to revert this
because the next PR will also update the image to its own copy.


vinayada1 (approved):

AaronCrawfis (approved):
Docs look good

PR #160, opened 2021-04-02
Adding command to allow application deletion

kachawla:
Adding command to allow application deletion https://github.com/Azure/radius/issues/158. Also added temporary workaround to handle the bug that was surfaced with cache removal: https://github.com/Azure/radius/issues/159.


rynowak (approved):
Awesome 👍

PR #177, opened 2021-04-07
Removing old custom rad client implementation and unused code caught by go linter

kachawla:
Now that application list and get implementation has moved to autogenerated client, the old custom client is no longer needed.

Also go linter complained [here](https://github.com/Azure/radius/pull/167) about some unused code in docgen implementation, removing that.


AaronCrawfis (approved):
docgen looks good to me

rynowak (approved):

PR #185, opened 2021-04-08
Add Azure Key vault support

vinayada1:
1. Create a KV
2. Create a managed identity per container
3. Grant KV access to the MI using RBAC
4. Create pod identity binding the MI to a pod label
5. Modify container k8s to use the aadpodidentity label

Fixes #125 , #334


rynowak (changes requested):
It's awesome to see this come together E2E. There are some design challenges here that we should talk through.

rynowak (commented):
This is very much on track, just a few smaller comments here to address.

AaronCrawfis (changes requested):

rynowak (approved):

AaronCrawfis (approved):

PR #193, opened 2021-04-09
Add install script

vinayada1:


AaronCrawfis (changes requested):
Scripts look good! Left two comments about switching to codetabs and placing in the install-cli page instead of creating a new page

PR #230, opened 2021-04-14
Use azure storage for download links

rynowak:
Fixes: #213

Download links aren't working because they are based on a private repo
right now. Fixing this to point to our public storage account.


AaronCrawfis (approved):

kachawla (commented):
Will it expose radius to be randomly discovered through the public link?

kachawla (approved):

PR #301, opened 2021-04-23
print env URL during some CLI commands. stop printing "Using config file: __" 

emily-potyraj:
1. Print info about accessing env URL during the following commands:  
- rad env init azure
- rad env switch
- rad deploy
2. Stop printing `Using config file: __` every time we read the config file via initConfig. Closes: radius-project/radius#254, radius-project/core-team#796
3. During `rad deploy`, print which env is being used. 

Not sure whether we prefer to add error handling inside GenerateEnvUrl (in utils.go) or in the other files where GenerateEnvUrl is called - or both. 
_____

Sorry for any confusion about SUPPORT.md being in this PR. Until that EOL change gets merged, git makes it a mandatory change for me.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,21 +1,7 @@
-<<<<<<< HEAD
-=======
-# TODO: The maintainer of this repo has not yet edited this file
-
-**REPO OWNER**: Do you want Customer Service & Support (CSS) support for this product/project?
-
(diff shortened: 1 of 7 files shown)


rynowak (commented):
Left some minor feedback, overall looks to be on track 👍

AaronCrawfis (approved):

rynowak (approved):
Awesome!

PR #314, opened 2021-04-27
Fix issue# 309 for tests to work

vinayada1:
1. Resource groups with upper case names cannot be found while merging env credentials
2. In the deploy tests, after logging in with the service principal, the default sub is chosen. After changing the test subscription to a different one, need to set the subscription

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -138,6 +138,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GOVER: '^1.16.0'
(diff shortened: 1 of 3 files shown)


kachawla (commented):
I think we as a team can do better at commit messages and PR description. In general it's a better practice to add more descriptive commit messages, so that it's easy to get context when looking at commit history in the repo. Issue 309 doesn't really provide much information unless I go and look at the issue :)

kachawla (commented):
Just to understand the bug better - none of the commands that take env name as input would currently work if the name contains an upper  case letter?

rynowak (approved):

PR #338, opened 2021-05-04
fixes delete does not invoke delete for workloads

vinayada1:
rad application delete does not delete the individual workloads. This was introduced by a regression.
Fixes #325


kachawla (commented):
This was quite subtle. Curious to know more on how setting the status fixes call to delete? I can sync with you offline if that's easier

rynowak (approved):
Approving this so we can get it in. 

This is the area that we'll be covering as part of #305 and similar items. Normally I'd want to ask for a test change, but I'm not sure there's a super easy way to add a test for this today.

We'll be revisiting this whole area soon.

PR #379, opened 2021-05-18
Refactor: Remove component revision tracking

rynowak:
Part of AppModel v2 work

I found more vestigial features when working through the AppModel v2
changes. Previously we supported a Deployment object explicitly
specifying a revision of a component. This would allow you to target any
version of a component's lineage and deploy the version specifically.

This is holdover from a previous era of the AppModel where a component
was a defintion and a Deployment was an application. The idea was that
you could define a component once and deploy it multiple times with
different parameters. Note that we no longer have the idea of parameters
:). This is truly unused.

The actual requirement here is that we can easily compare the content of
a component to the deployed version so we can tell if it is the same -
this is much simpler than what was here.

----

So removing this old/unused feature simplifies the database code for
components as well as all of the code that accesses the database layer
of the system.

I thought it make sense to work through this as a self-contained change
(mostly deletions) separately from the actual AppModel v2 change.

### DIFF
--- a/mocks/inmemory_db.go
+++ b/mocks/inmemory_db.go
@@ -11,7 +11,6 @@ import (
 
 	db "github.com/Azure/radius/pkg/curp/db"
 	resources "github.com/Azure/radius/pkg/curp/resources"
-	revision "github.com/Azure/radius/pkg/curp/revision"
 	"github.com/golang/mock/gomock"
 )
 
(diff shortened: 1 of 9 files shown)


jkotalik (approved):
Lots to learn, but the rest of the change seems good.

kachawla (approved):

PR #384, opened 2021-05-22
Refactor examples to into components

AaronCrawfis:
- Rename examples into components
- Move into docs folder instead of root
- Remove all old examples
- Add structure for Azure and Dapr
- Write docs for table storage
- Update KeyVault docs
- Update ServiceBus example
- Add Dapr sidecar docs


rynowak (changes requested):
Deployment tests need to be updated, they refer to these files.

emily-potyraj (commented):
This is some great much-needed cleanup! 

btw, looks like the inbound-route work just got merged. Might want to scoop up the update for `examples/inbound-route/index.md`'s path in this PR, too.

jkotalik (commented):
I like the direction we are moving here with the docs!

PR #429, opened 2021-05-25
Use resource group name for env name default

jkotalik:
Fixes https://github.com/Azure/radius/issues/294

Updates name after acquiring the resource name.


emily-potyraj (commented):
> The odd thing is there are two logs to `New Environment with ...` in the code. One gets logged when doing interactive, but two get logged without interactive. Also, one has the environment name and the other doesn't.

It might still be useful to highlight to the user whether we've gone down the "create env" or "connect to env" path.

rynowak (approved):

PR #446, opened 2021-05-26
 Add unmanaged-resource support for servicebus

rynowak:
Part of: #414

This change implements support for 'unmanaged' resources in Radius,
starting with ServiceBus and Dapr.

The principle here is that when a user specifies a resource id, we can
handle this during the 'rendering' phase and break it down into parts.
This way the 'handler' has a relatively simple job, and we do the
validation up front.


vinayada1 (approved):

kachawla (approved):

kachawla (commented):
Looks great!

PR #448, opened 2021-05-27
Adds `rad application switch` and updates all CLI commands

jkotalik:
Fixes https://github.com/Azure/radius/issues/348.
Fixes https://github.com/Azure/radius/issues/349.

I haven't added "add a prompt to user: "Are you sure you want to delete $application from $env?" (Y/N)" for `rad application delete` as I wanted to make sure we are okay prompting for user input without the `-i` flag. At the very least, we should add a way to bypass this without requiring user input for scripting scenarios.

I'm currently logging a message for switching like:
```
Switching default application from webapp to dapr-hello
```

Finally, is there any documentation I need to add for the new command and changes or will everything be autogenerated?

### DIFF
--- a/.github/workflows/create-environment.yaml
+++ b/.github/workflows/create-environment.yaml
@@ -76,7 +76,7 @@ jobs:
         run: |
           go run cmd/cli/main.go \
             env init azure \
-            --name ${{ steps.generate-resource-group-name.outputs.result }} \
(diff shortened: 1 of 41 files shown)


rynowak (commented):
Code seems to generally do good things. Some comments/questions about behaviours.

rynowak (approved):

PR #455, opened 2021-05-28
init commit for output resources

vinayada1:


rynowak (changes requested):
Just one more small thing :)

rynowak (approved):

PR #474, opened 2021-06-02
Add completions to rad cli via `rad completion`

jkotalik:
Fixes https://github.com/Azure/radius/issues/256

The design I went with follows what `dapr` does today and has a top level command for completions.

image

We could consider adding this to the installation script for radius as well.


vinayada1 (approved):

kachawla (approved):
+1 to adding it to installation script, assuming with current setup this would need to be manually triggered through rad completion?

rynowak (approved):

PR #488, opened 2021-06-04
doc updates for v0.2.0

emily-potyraj:
- fixes a couple straggler references to the old CLI format
- adds language to the bindings language to attempt to answer questions users might have about them
- Radius-managed v. user-managed examples: add code directly to the page for users to read (instead of just downloadable). 
- minor language nit-picks 
- rename state store component in examples so there's no dash character in the name ("inventory-store" --> "inventorystore") 
  - "Code="DeploymentFailed" ... "inventory-storedf40386cf is not a valid storage account name. Storage account name must be between 3 and 24 characters in length and use numbers and lower-case letters only." 

btw, we've got 3 sections that describe Bindings functionality. I attempted to lean into explaining bindings in the framework of each of those 3 contexts.   
- /concepts/overview-concept/#bindings 
- /app-model/components-model/#binding-configuration
- /app-model/bindings-model/


jkotalik (approved):
Your writing is 🔥

AaronCrawfis (changes requested):

AaronCrawfis (changes requested):

AaronCrawfis (changes requested):

jkotalik (approved):

rynowak (approved):
:shipit: after addressing comments

AaronCrawfis (approved):

PR #501, opened 2021-06-07
Add more logging where use az cli in CI

rynowak:
Fixes: #457

We've encountered failures in environment create/delete that don't give
enough information to investigate. There's likely more error info
available in these cases that we can't see just from the CLI. Adding
`--debug` means the CLI will log request/response payloads.


kachawla (approved):
Not blocking for this PR - if you have an example on what additional information verbose throws out, I'd be interested in that. Also if verbose solves all our problems or we need more information sent from Radius RP as well?

jkotalik (approved):

PR #503, opened 2021-06-07
 Refactor CLI to separate business logic from clients

jkotalik:
For part of https://github.com/Azure/radius/issues/495

This refactors all application, component, deploy, and deployment commands to use a client abstraction, allowing for an easy addition of k8s.

- Client functions doesn't include updating config. For example, `rad application delete` will remove the defaultapplication if present. This functionality isn't done by the client and rather is still in the CLI.
- I can see us changing the client return types for list/show to return strings instead of printing in the function.
- Doesn't modify env functions for now, I think the only commands that calls out to azure are `env init azure` and `env delete`. Env delete would be the only one we'd want a client abstraction for, and we can probably live with an if check in code for now.

### DIFF
--- a/cmd/cli/cmd/appDelete.go
+++ b/cmd/cli/cmd/appDelete.go
@@ -8,15 +8,10 @@ package cmd
 import (
 	"fmt"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/armcore"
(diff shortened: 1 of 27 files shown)


rynowak (commented):
Pretty cool, left some feedback

rynowak (approved):

PR #553, opened 2021-06-14
Support Component type in kubernetes controller

jkotalik:
- Supports Component Custom Resource Definition
- Adds test for Component to verify it can be applied and obtained


rynowak (approved):
:shipit: after addressing minor feedback. Discussed a lot of the outstanding things offline and they will move to followups.

PR #561, opened 2021-06-15
Remove radtest and replace with rad

rynowak:
Fixes: #330

This change makes it possible to use `rad` for all the functionality
that `radtest` previous provided. This change adds support for an
undocumented `localrp` kind of environment that is special cased for our
development purposes. I say undocumented, but it's actually documented
in the docs for development :)

The major investment here is to add the new environment and implement
the diagnotics, deployment, and management clients for it. Only the
deployment client is really special - this uses a CLI-driven interpreter
for a subset of ARM template functionality. We'll be using this for K8s
also for now and replacing this in the longer term.


jkotalik (commented):
So far I think this is good. I'll let you finish up any last pieces.

vinayada1 (approved):

PR #566, opened 2021-06-16
Logging changes using zap - logr

vinayada1:


jkotalik (changes requested):
Have we considered using the SugarLogger instead of Logger? https://pkg.go.dev/go.uber.org/zap#SugaredLogger. Seems like it is slightly less performant, but gives us some useful methods for logging: https://github.com/uber-go/zap#quick-start

rynowak (commented):
This is awesome! One open question about the use of `Info` vs `Error`

jkotalik (approved):
:shipit:

PR #607, opened 2021-06-23
Use explicit parallelism for integration tests

rynowak:
By default `go test` will use the GOMAXPROCS as the number of tests to run in parallel, which is likely to be a small number in an environment like GH actions (usually number of hw processors accessible to the VM).

Our tests are highly I/O bound, and spend most of their time waiting on remove operations. Therefore it really makes sense for us to oversubscribe when oversubscribing on CPU-bound tests would not be.

I've seen some timeouts recently, and I'm not sure we're executing tests in a sufficiently parallel way. For now this is just testing to see if it makes an improvment.

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -11,7 +11,7 @@ test: ## Runs unit tests, excluding kubernetes controller tests
 
 .PHONY: test-integration
 test-integration: ## Runs integration tests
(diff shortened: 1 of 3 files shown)


kachawla (approved):
Looks good except the docs changes that probably shouldn't be here?

jkotalik (approved):

PR #616, opened 2021-06-24
tolerate 404 during delete

vinayada1:
Fixes: #483


kachawla (commented):
I'm not completely clear on what would cause us to get into this situation, do we have any visibility into that? Either ways, I would suggest at least logging when this happens so that we can track this. Let me know your thoughts.

nit - I probably have mentioned this in previous PRs, but in general making commit messages more informative helps long term for anyone to be able to get context from repo history without the need to look into the actual code diff. In this case for example I would call out tests specifically in the commit message. Let me know if you think we should discuss this as a team, happy to get more inputs.

PR #630, opened 2021-06-25
Enhancements to rad shortcode

rynowak:
Part of: radius-project/core-team#1026

This change enhances the `rad` shortcode for code snippets to allow:
- Including a subregion of code in the output instead of a whole file
- Doing replacements to abbreviate regions of the code

This adds up to use being able to use fully-valid `.bicep` files that
can be validated, but still have a lot of control over what shows up in
the docs.


AaronCrawfis (approved):
This is AWESOME!

PR #654, opened 2021-06-30
Remove back-compat code for 0.2 environments

rynowak:
Fixes: #565

This change removes code that handles the back-compat cases, basically
the way that Radius environments looked and worked when they were a
single resource group instead of two.


jkotalik (approved):
🚢 🇮🇹

kachawla (approved):

PR #699, opened 2021-07-12
Health monitoring - Part1

vinayada1:
Covers the following items in radius-project/core-team#765 
 - [x] Data model
- [x] Changes to CLI commands to show the health and provisioning state
- [x] Infra for output resources to check health periodically and save in DB
- [ ] Logic for each output resource to compute health
- [ ] State machine to compute component and application health
- [ ] Record application health
- [ ] Health dashboard
- [ ] Tests


rynowak (commented):
Very cool to see this getting started 👍

rynowak (commented):
Would like to see some unit tests as well for the infra pieces like the health monitor :) Basically all of the things that don't need to touch cloud resources.

This is good progress 👍

jkotalik (approved):

PR #751, opened 2021-07-19
Make HTTPS required for custom RP

rynowak:
Fixes: #572

This will configure app service to block HTTP and require HTTPS. This is
beneficial since we DO have a cert as part of app service, so we have no
need for non-encrpyted HTTP.

Fixes an Azure Security Center warning as well.


jkotalik (approved):
If we really do have a cert 😄 .

Do we need to worry about the lifetime of that cert, or is it managed elsewhere?

vinayada1 (approved):
:)

PR #798, opened 2021-08-01
Add support for generic MongoDB

rynowak:
Fixes: radius-project/core-team#1278

This change adds support for a new 'generic MongoDB' component kind
implemented as:

- CosmosDB MongoDB on Azure
- A basic dev configuration on Kubernetes (stateful set)

We're going to change the tutorial to use the generic MongoDB component
so that it's covered on Azure. We have good tests in place for Cosmos
already and the new Azure functionality reuses much of the existing
implementation.

Includes E2E tests for Kubernetes.


jkotalik (approved):
👍

PR #821, opened 2021-08-05
Implement dependency ordering for output resource

kachawla:
Implement DependencyItem interface for output resources. Using local ID as the unique identifier for output resource.


jkotalik (approved):

rynowak (approved):
The code changes look fine 👍

PR #822, opened 2021-08-05
Log output from k8s pods/containers

jkotalik:
Fixes https://github.com/Azure/radius/issues/710
- Captures output from all pods that start, including the radius controller, and uploads them as artifacts
  - Only runs the radius controller log capture once for all test runs
  - Run watch on each application name so each test logs


rynowak (approved):
This is super cool 👍

tcnghia (approved):
Only one minor comment. I think it is probable that `stream.Read` will return -1 or something when err happens but it's best not to rely on that. Feel free to check in after addressing that.

This is awesome. Thanks a lot! I am sure we would save many frustrating hours later when debugging.

tcnghia (approved):

PR #860, opened 2021-08-12
Add additional error codes for `NotFound` and `Conflict`

tcnghia:
Part of #846.

As suggested in https://github.com/Azure/radius/issues/846#issuecomment-896425502, this PR adds the codes `NotFound` and `Conflict` and makes sure our error response uses these in the appropriate scenario. This will help with fixing #690 and #846.


jkotalik (approved):
Approved pending maybe a rename.

PR #869, opened 2021-08-16
Adds rabbitmq component to kubernetes

jkotalik:
- Adds rabbitmq component to kubernetes
- Small refactor to move findDeployment/Service to the kubernetes test package

I did a good amount of research to see if we could create a generic AMQP component type which would deploy azure service bus on azure and rabbitmq on kubernetes. But from my initial attempts, it looks like the connection strings for service bus aren't compatible with generic AMQP clients. I reached out to the service bus team to follow up on that idea, but their initial response made it seem like it wasn't possible.


rynowak (approved):
Test code looks much better 👍

PR #902, opened 2021-08-18
Process deployments in dependency order. Integrating with Cosmos MongoDB

kachawla:
* Update deployment processor to deploy resources in dependency order during component create or update.

* Integrate with CosmosDB Mongo deployment -
   * Update renderer to return two separate output resources for account and database.
   * Implement a new handler for CosmosDB mongo account creation, and corresponding updates to mongodb handler.
 
* No updates to order of deletion of resources yet, since we just ignore 404 if cosmos account is deleted before the database. This should be updated once we run into a use case that strictly requires dependent resources to be deleted first. Ordering for deletion can be achieved by including dependencies in db.OutputResource.


jkotalik (approved):
Pending some minor refactors and a few outstanding questions, looks good!

rynowak (approved):
Another step forward! :shipit: after addressing comments from others

PR #966, opened 2021-08-27
Remove mocks/ directory

rynowak:
Part of: radius-project/core-team#1270 

This change removes the top level mocks directory and puts the relevant
mocks in the package where they belong. The files are named with `test_`
so they will not be compiled during a standard build.

Also used the 'fake' client from controller-runtime in replacement of
our own mock.


jkotalik (approved):

kachawla (commented):
Does this need to be deleted?https://github.com/Azure/radius/blob/main/pkg/health/mocks/mock_healthhandlers.go

kachawla (approved):

PR #968, opened 2021-08-27
Container render/deployment updates

kachawla:
Move deployments from renderer for container to deployment processor -> handler by leveraging dependencies between output resources generated by the renderer.

Fixes - https://github.com/Azure/radius/issues/963 and https://github.com/Azure/radius/issues/497

Deletion of resources is currently bundled in one handler for multiple resources, I'll be addressing that as a separate PR to limit the scope of this change.


jkotalik (approved):
LGTM

PR #974, opened 2021-08-31
Add health checks support for K8s resource with Azure RP

vinayada1:


vinayada1 (commented):
/create-environment

rynowak (approved):
Let's fix whatever's causing the tests to fail and :shipit:

PR #977, opened 2021-09-01
Clean up 404 errors

jkotalik:
Fixes https://github.com/Azure/radius/issues/690

Drafting because I want people's opinions on what the error should look like.

image

I personally think this should be terser. I think `Error: Application 'app-name' was not found` would be best rather than having the nested lines. But this logic was already being used for `rad deployment show`, so I wanted to get opinions before moving forward.

@tcnghia @AaronCrawfis your thoughts would be appreciated here 😄

### DIFF
--- a/pkg/azure/radclient/errors.go
+++ b/pkg/azure/radclient/errors.go
@@ -22,7 +22,7 @@ type RadiusError struct {
 
 // Error implements the error interface for the RadiusError type.
 func (e *RadiusError) Error() string {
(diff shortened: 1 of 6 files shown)


kachawla (commented):
@jkotalik I didn't fully follow this so wanted to clarify -

`I personally think this should be terser. I think Error: Application 'app-name' was not found would be best rather than having the nested lines. But this logic was already being used for rad deployment show, so I wanted to get opinions before moving forward.`

Do you mean we should decide between having both error code and message in the response, and just the message?

tcnghia (approved):

PR #993, opened 2021-09-03
Aggregate component status based on output resources

vinayada1:
Aggregate the component status and Provisioning state based on the state of the output resources


jkotalik (approved):

tcnghia (commented):
Is it easy to add a unit test for this?

kachawla (approved):

tcnghia (approved):
Thanks for making the change and adding the test.

rynowak (approved):

PR #1003, opened 2021-09-07
Tool to generate autorest-compatible OpenAPI spec from our JSON Schema

tcnghia:
Part of #887 

I've tested this by generating a new OpenAPI spec using this tool. I will include that generated code in the next PR.

### DIFF
--- a/cmd/autorest-schema-gen/main.go
+++ b/cmd/autorest-schema-gen/main.go
@@ -1,0 +1,65 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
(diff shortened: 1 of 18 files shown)


jkotalik (commented):
The PR itself looks fine, but a few thoughts:
- When we move to app model v3, it is expected that we will need to edit this code to adopt it I assume.
- I think a high level summary of the specific changes that are made (oneOf -> AllOf with discrim field, for example) will help with comprehension.

rynowak (approved):
This is really surprisingly clean. I can already visualize how I'm going to integrate the appmodel v3 changes 😻 

I like the idea of us building up more schema-related tooling.

PR #1013, opened 2021-09-11
Fix Go version to 1.16.8.

tcnghia:
It seems that Go 1.17 causes our generated code to differ from 1.16 and causing `make generate` check to fail.

Fixing this for now before 1.17 becomes more widely adopted in the team.


jkotalik (approved):
Let's file a follow up issue to revert this.

PR #1017, opened 2021-09-14
Add executable CRD

karolz-ms:
I am still working on some tests but I thought it would be good to get the discussion started.

I had some conversation with Justin about this eventually morphing into a custom controller for the K8s standard Deployment kind. For the current proof-of-concept work I went with implementing Executable as a separate CR.


jkotalik (commented):
This is awesome! Some initial thoughts.

rynowak (approved):
This is great progress. I want it now 😍  - can you get a signoff from @jkotalik as well - I didn't review the controller code in depth, I figured you just need one person to do that.

jkotalik (approved):

PR #1019, opened 2021-09-16
Add v3 schema support

rynowak:
This change adds support for schema validation in the v3 code path as
well as:

- Updated v3 schemas (about half of them)
- Added a 'manifest' format we can process for the list of our supported
  resource types
- Improved testing framework for our schemas and validations
- Improved algorithmic complexity of schema loading from O(N^2) -> O(N)
- Plumbing and tests for all of the above


tcnghia (approved):
This is great. Thanks a lot!

jkotalik (approved):

kachawla (approved):

PR #1022, opened 2021-09-17
Add new CRD types for app model v3

jkotalik:
This is just CRDs for all of the new types.

Big change is coming up next.

All types are versioned so this shouldn't break any existing scenarios at the moment.


rynowak (commented):
Good start, I pointed out a few things that are minor inconsistencies.

rynowak (approved):

PR #1027, opened 2021-09-19
OpenAPI codegen tool generates the boilerplate and the API paths

tcnghia:
Since there are many paths in V3, we will need to auto-generate them. This PR expands the OpenAPI codegen tool to do so.

Part of #887


rynowak (dismissed):
This is on track, but still missing a few things. let's make sure that validate the the outputs work with autorest.

This should also get wired up to the build so that we validate it on every PR.

rynowak (approved):

PR #1029, opened 2021-09-19
Adding generator for the Custom RP manifest

rynowak:
This change adds a generator that will generate the Custom RP manifest
from our 'resource-types.json' file. This automates one of the steps we
need to make the RP work end to end for a new resource type.

After this change, the actual deployment template we use
'rp-full.json' becomes a generated asset. Modify 'rp-full.input.json'.

This approach is preferrable because we'll be able to see an verify the
diff in the manifest when a PR occurs. I also considered an approach of
generating parameters to pass into the deployment when creating an
environment, but the fact that this relies on a file in plaintext rather
than a magic CLI behavior seemed valuable.

### DIFF
--- a/build/generate.mk
+++ b/build/generate.mk
@@ -6,7 +6,15 @@
 ##@ Generate (Code and Schema Generation)
 
 .PHONY: generate
-generate: generate-radclient generate-go generate-k8s-manifests generate-controller ## Generates all targets.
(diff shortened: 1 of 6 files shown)


tcnghia (approved):
LGTM. Thanks for the PR, it clarified a bunch of things for me in  https://github.com/Azure/radius/pull/1027

PR #1036, opened 2021-09-20
Add the rest of components into schema V3

tcnghia:
This is a follow up of https://github.com/Azure/radius/pull/1019 to add the remaining component types into schema.


kachawla (approved):

rynowak (commented):
Showed up late but this is awesome! 👍

PR #1042, opened 2021-09-21
Skip `deploy_tests` and `kubernetes_tests` when we only change docs

tcnghia:
Will save some Azure resources for other workflows and also expedite small doc changes.


rynowak (commented):
If it's this easy then that's super cool. Excelsior!

rynowak (approved):
Thanks, this is awesome!

PR #1048, opened 2021-09-22
Make sure working dir (from Spec) is used by Executable

karolz-ms:
I also changed the interface to process executor to use `exec.Cmd` as the input parameter--the list of `StartProcess()` method was getting too long, and would be even longer when we add stdout/stderr tracking. This is a change that @rynowak suggested in the previous PR.


jkotalik (approved):
Some small nits

PR #1060, opened 2021-09-24
Implement model and deployment processor for app model v3

kachawla:
Implement model and deployment processor for app model v3. Will be following up with tests in a separate PR. This will unblock Ryan to update with fixes for the bugs that he found during end to end testing.


rynowak (commented):
Looking great overall, some minor comments and waiting on tests

rynowak (approved):
Let's do it!

I'll follow up with a smaller PR that gets one of our end-to-ends working.

PR #1070, opened 2021-09-26
Add `rad resource list/show` for V3 resources

tcnghia:
Part of #1043


rynowak (dismissed):
I think there's a little confusion still here about positional args, we should clean that up.

jkotalik (approved):

rynowak (approved):

PR #1085, opened 2021-09-27
Fix persistence of properties

rynowak:
When we process a resource we allow the 'handler' to persist some state
that will be provided to the handler again when the resource is visited
again (updates/deletes). This is how many of the older resource handlers
communicate about the identity of the resource being updated/deleted,
whether it exists, etc.

This is broken right now for AppModelv3 because of the way Unmarshaling
works in the Mongo driver. We modeled the data as `interface{}` and that
results in us getting a bson-specific data type when it comes back out
of the database. If we model this as `map[string]string{}` (which is a
more accurate reflection of what we use it for) then it will work
successfully.

This isn't an issue in V1 because we process resources as a batch.


tcnghia (approved):
Thanks for fixing this

kachawla (commented):
Looks good, quick question about a comment that was unclear.

PR #1100, opened 2021-09-28
Forgot about those labels

jkotalik:
Need to sync with @tcnghia about what this needs to solve.
Fixes https://github.com/Azure/radius/issues/1097

I created the annotations for the k8s resource, but forgot to set it.


tcnghia (approved):
Thanks for making this change

PR #1107, opened 2021-09-29
Enable deletion of apps in v3 tests

rynowak:
Now that the CLI functionality is in for deletion, we can delete again!


jkotalik (approved):
If this works 😄

PR #1109, opened 2021-09-29
Add `rad applicationV3 delete` for Kubernetes

tcnghia:
Part of #1089


jkotalik (approved):
Please include in #1097 to fix rad application delete as well.

I don't want this code to ship 😄

PR #1111, opened 2021-09-29
Copy the error unfold logic to use radclientv3 package

tcnghia:
This will help unwrap the DownstreamError better.

Part of https://github.com/Azure/radius/issues/1091


jkotalik (approved):
At some point, we probably can just start deleting the V3 versions and delete the V1. Let's make sure to get all of the renames.

PR #1112, opened 2021-09-29
Fix OpenAPI spec with 202 code for async Delete

tcnghia:
Part of #1043


rynowak (approved):
Thanks, this should fix delete (for real this time 🤣 )

PR #1119, opened 2021-09-29
changes for readiness/liveness probes

vinayada1:
Set the k8s spec as per the readiness/liveness settings in the bicep file


rynowak (commented):
Mostly looks good. Should be checking those serialization errors.

Can we add this to one of the E2E tests?

rynowak (approved):

PR #1150, opened 2021-10-03
Delete AppModel V3 RP, DB, and Deployment

rynowak:
This change deletes the AppModel v3 support spanning from the HTTP layer
back to the deployment processor.

The following places I was able to reorganize/rename code artifacts in a
cleaner way after removing some of the duplication:

- Database types
- Application model types


vinayada1 (approved):
Left minor comments

PR #1151, opened 2021-10-04
Generic ARM handler PoC

rynowak:
This is a work-in-progress PoC of replacing out existing 'handler' code
with a single generic ARM resource handler. The benefit of this is
that gives us a single good thing to harden and improve and moves more
of our functionality into the declarative layer.

This is incomplete because there are still a few design problems left to
solve, namely about how dependencies and how data flows between multiple
output resources of a Radius Resource.

Consider the following - this can solve 'how do you create service' in a
generic/declarative way. It needs more functionality to solve 'how do
you get the connection string into the Dapr manifest'. I have some ideas
here but wanted to start a discussion before getting too far with it.

### DIFF
--- a/pkg/handlers/arm.go
+++ b/pkg/handlers/arm.go
@@ -1,0 +1,293 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


kachawla (commented):
Overall direction here looks ok to me.

PR #1166, opened 2021-10-05
Rename radclientv3, open-api-specs-v3 directories

tcnghia:


rynowak (approved):
Found a few minor things that could still be cleaned up. Feel free to ignore if it makes sense to handle those separately.

PR #1217, opened 2021-10-07
Add support for SQL Azure

rynowak:
This change adds support for user-managed SQL Azure. Somewhat
differently from our other resources we don't handle usernames and
passwords here or connection strings. This is because SQL Azure requires
the user to manage these things, so we don't have a way to access them
from ARM.


jkotalik (approved):
Looks as I would expect

kachawla (approved):

PR #1226, opened 2021-10-08
Add parameter support to rad deploy

rynowak:
Fixes: #1216

This change adds support for passing parameters to the `rad` CLI. The
structure of this is similar to how it's done in the azure CLI for
bicep/ARM deployment.


tcnghia (commented):
Super cool and useful. Thanks

jkotalik (approved):

PR #1239, opened 2021-10-08
Fixes services to include app names for rabbit and gateway

jkotalik:
Found when verifying eshop


amanbha (approved):

rynowak (approved):
You can safely ignore the Bicep validation issue, that's being fixed separately and doesn't block merging.

kachawla (approved):

PR #1272, opened 2021-10-14
Retrieve role assignment information on 409

kachawla:
Fixes https://github.com/Azure/radius/issues/704


kachawla (commented):
We don't seem to have unit tests for handlers at this point. Let me know if I missed anything that's existing, else I'll create a separate task tracking handler test coverage as I would prefer not to increase scope of this bug fix in this milestone.

tcnghia (approved):

PR #1277, opened 2021-10-14
Reduce the build time by building outside Docker

tcnghia:
This is the alternative approach to #1245, also achieving a similar speedup.


jkotalik (approved):

rynowak (approved):
Awesome 👍

PR #1283, opened 2021-10-18
Add kcp and radiusd to binaries build

karolz-ms:
These will be required for local development environment, in particular, `env start` and `env stop` commands


jkotalik (approved):
Pending fixing the build issues you are hitting, LGTM.

PR #1307, opened 2021-10-20
Use autorest.go 4.0.0-preview29

tcnghia:
This is to pick up the correct unmarshaller for `ContainerComponent.Volumes`. 

This should have been a small change but there is some large refactor going on in `azcore`.


rynowak (commented):
Code changes look fine, I want to understand more about the module version -- I'll follow up offline.

jkotalik (approved):

PR #1328, opened 2021-10-22
Gateway API support

jkotalik:
Fixes https://github.com/Azure/radius/issues/1180


rynowak (commented):
This is looking good overall. Do we need to split off parts of this to get it to go through the CI?

Updating environment setup and schemas don't take effect on your PR. It might to be good to split off those pieces so we don't have to #YOLO the whole thing.

rynowak (approved):

PR #1351, opened 2021-10-29
handle volume changes for k8s 

vinayada1:


rynowak (commented):
Looks like this is on the right track in general, I had some questions about how decoupled this functionality is meant to be.

jkotalik (approved):
Pending outstanding comments.

PR #1355, opened 2021-10-30
Refactor controller entry points

rynowak:
This change refactors our controller entry points
to make them more easily consumable from tests and executables.


karolz-ms (approved):
nice

jkotalik (approved):

PR #1408, opened 2021-11-03
fix field selector to filter out the resource

vinayada1:
# Description

rad resource show uses a field selector. This did not have an application name prefixed due to which there was no resource in the filtered list
## Issue reference
#1407 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


tcnghia (approved):

kachawla (commented):
Are there any tests that should be updated? If not, can we add a test for this?

PR #1458, opened 2021-11-13
Add `make install` target for convenience

rynowak:
This change adds a new makefile target to 'install' our bits as a local
working copy. I've been using this to iterate quickly on the local dev
scenario that uses `radiusd` but its generally useful as well for
updating `rad` locally.


tcnghia (approved):
Just a few nits but LGTM

PR #1460, opened 2021-11-14
Unmanaged Redis on Kubernetes

tcnghia:
# Description

Support unmanaged Redis on Kubernetes.

## Issue reference

Part of radius-project/core-team#406 : In this step we scrape out the rendered secrets. Then we use them, together with other connection-provided properties, in the unmanaged `RedisComponent`'s `Status` as `ComputedValues`.

These `ComputedValues` are then handled by the existing connection-handling logic of `ContainerComponent` and thus will be rendered correctly in the pods as `SecretRef` env vars.

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,4 +56,5 @@ require (
 	rsc.io/letsencrypt v0.0.3 // indirect
 	sigs.k8s.io/controller-runtime v0.9.2
 	sigs.k8s.io/gateway-api v0.3.0
+	sigs.k8s.io/yaml v1.2.0
 )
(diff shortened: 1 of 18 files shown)


rynowak (dismissed):
Would also like to see some tests in test/integration/kubernetes that cover the new mechanics of the deployment template controller:

- Secret getting created 
- Secret getting updated
- Secret getting deleted when 'owned' resource is deleted

In the past we've really undertested this kinds of things and had bugs. 

Would also like to discuss more if we should change the structure of our K8s objects so they can just secretRef. Right now a user using kubectl/yaml doesn't need to interact with this new 'secrets' mechanism which means that the layering is off.

jkotalik (approved):

vinayada1 (approved):

PR #1504, opened 2021-12-01
Rewrite authoring user guides

AaronCrawfis:
- Page on Bicep
- Page on Modeling infra
- Page on modeling app
- Page on Breaking into modules
- Page on templating


jkotalik (commented):
I think we should separate these user guide flows up into separate tasks. Most of these seem distinct.

emily-potyraj (commented):
Great work! Thanks for driving it.

emily-potyraj (approved):

jkotalik (approved):

PR #1546, opened 2021-12-09
Add docs for Azure resource connections

kachawla:
# Description
Documentation for https://github.com/Azure/radius/issues/1321

## Issue reference

Please reference the issue this PR will close: https://github.com/Azure/radius/issues/1545


AaronCrawfis (changes requested):
Looking good! Couple comments. Also:

- Should we rename `role` to `roles` since it's an array?
- Can the `role` field accept a GUID for a custom role definition ID in addition to the role names?

AaronCrawfis (approved):

PR #1569, opened 2021-12-15
Add support for version command

asilverman:
# Description

I added a new command called version and used it to support both itself and the --version flag, to do so I also added a new `ListFormatter` that can be used with the `--output` flag and can be leveraged more broadly to  print to console not only the version but other resources as well (in the spirit of code-reuse)

## Issue reference

Please reference the issue this PR will close: #1323 

## Checklist

* [x] Code compiles correctly
* [x] Adds necessary E2E tests for change - e2e doesn't seem applicable
* [x] Unit tests passing
* [?] Extended the documentation / Created issue for it - this is probably something I need some help with, perhaps it makes sense to add the docs myself or add an issue, need some guidance on it


jkotalik (approved):
Pending feedback, LGTM

jkotalik (approved):

PR #1576, opened 2021-12-16
Change structure of scaffolded app

rynowak:
Fixes: radius-project/core-team#448

This change updates the structure we generate by default for a new
scaffolded application, and also how we locate rad.yaml.

The primary change here is to look for rad.yaml in the user's current
directory rather than at `./rad/rad.yaml`. This has two nice benefits:

- It's more obvious and matches other tools better
- It's more obvious what relatives paths in rad.yaml do

In addition to these we're using the convention of `iac/` for where we
place the bicep files. This matches what `az dev` is doing.


AaronCrawfis (approved):
:shipit:

PR #1600, opened 2021-12-30
Use sub and resource group for unique names

rynowak:
Fixes: radius-project/radius#1599

This change adds the subscription id as a seed value for generating
unique names in our environment deployment template. Previously we were
just using the environment name (same as the resource group) but not
considering the subscription. The result is that environments with the
same name across different subs would collide.

### DIFF
--- a/deploy/rp-full.input.json
+++ b/deploy/rp-full.input.json
@@ -43,7 +43,7 @@
         },
         "siteName": {
             "type": "string",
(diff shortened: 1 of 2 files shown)


asilverman (approved):
LGTM 🚢

jkotalik (approved):

kachawla (approved):

PR #1606, opened 2022-01-03
Fix handling of resource deletion in controller

rynowak:
This change fixes the handling of 'not found' reconcile requests. When a
resource is being deleted it will trigger a reconcile, but the
underlying resource is gone. This is for the controller to clean up any
resources it needs to (GC). We don't need this so the right thing to do
is just return success.

Right now we report these 'not founds' as errors which causes them to
loop.


asilverman (approved):
LGTM 🚢

jkotalik (approved):

kachawla (approved):

PR #1655, opened 2022-01-06
Don't wait for all pods when application delete is running

jkotalik:
# Description

Don't wait for all pods to be removed from a namespace before a test completes

## Issue reference

Fixes https://github.com/Azure/radius/issues/1243


youngbupark (approved):
LGTM.

For my learning, do we have the tear down operation which deletes all pods at the end of test ?

kachawla (approved):

PR #1663, opened 2022-01-06
comment out test that is causing flakiness in ci build

asilverman:
# Description

commenting out broken tests as a result of azure policy changes, likely this test will end up being removed once support for managed resources is dropped from rad.

## Issue reference

Please reference the issue this PR will close: #1661

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


kachawla (commented):
> likely this test will end up being removed once support for managed resources is dropped from rad.

This is a new test that was added to test connection to non-radius azure connections, so this is unrelated to managed resources, let's not comment it out. The error listed in https://github.com/Azure/radius/issues/1661 is related to hitting free tier/account limits for TextTranslation service `Only one free account is allowed for account type 'TextTranslation'."}]`, you can update the bicep to create a storage account or another resource type with more lenient quota. Feel free to assign it to me if that helps and I can update it.

kachawla (approved):

PR #1673, opened 2022-01-07
Add provider model for ARM processor

rynowak:
This change adds a 'provider' model for our ARM processor so we can
customize the handling of different resource families. This is a mirror
of how the actual deployment engine is implemented and will make it easier for
us to extend.

This also adds support for nested deployments (modules) in the 'local' execution model of the deployment client. We didn't really have any good tests for this code path so I added some pretty expansive new integration tests.


tcnghia (commented):
My first batch of comment, more coming

jkotalik (approved):
Great tests

tcnghia (approved):
The rest are test files so I only have the original comments.

PR #1695, opened 2022-01-09
Add support for 'local dev' environment

rynowak:
This change introduces a 'local dev' environment concept for `rad` to
manage. Really this is just another variant of Kubernetes, what makes it
special is that we're curating the experience and providing some
lifecycle management inside `rad`.

The 'hard parts' of using Kubernetes as a local dev environment are
related to all of the choices on needs to make that are outside of the
scope of Radius:

- What ingress controller and how to set it up
- Where do I push my docker images (local registry)
- Understanding and accomodating behavior differences in networking

If we manage the lifecycle of the cluster then we can provide a curated
experience for all of these needs and leave less for the user to do.

The new functionality here includes:

- Creating/destroying the K3d cluster
- Starting/stopping the K3d cluster
- Reporting status
- Installing our bits with the correct configuration

This first cut won't include deploying any applications, just the
infrastructure management. There are a few small issues to fix blocking
those scenarios that will be done separately.


karolz-ms (approved):
Super nice, just a couple minor questions/suggestions

jkotalik (approved):

PR #1732, opened 2022-01-14
Add workaround for proxy in front of API Server

rynowak:
Fixes: radius-project/core-team#576 

One challenge of working with K3d (or similar) setups is that there is a
proxy in front of the API Server (which is itself a proxy), but only one
of these proxies is injecting Forwarded headers.

This is a problem for us because the ARM-RPC protocol which we have
to use on Azure requires an Absolute and Fully-Qualified URLs. Without
the original scheme/host that is impossible for us.

The fix here is to handle the Location header ourselves on the client
and fix it up. This is a temporary fix and will only be needed while
we're still using ARM-RPC as a protocol.


karolz-ms (approved):
Thank you!!! Looks great, just a couple minor suggestions.

PR #1746, opened 2022-01-19
Override Hugo partial for generating page link

AaronCrawfis:
Fixes radius-project/core-team#472

Looks more complicated than it is. I just edited line 15 from https://github.com/google/docsy/blob/master/layouts/partials/page-meta-links.html and removed line 42


jasonviviano (approved):
LGTM

PR #1757, opened 2022-01-23
Perform build in parallel for rad deploy

rynowak:
This change adds build parallelism (and colorized output) to rad app
deploy and friends. This is a significant improvement in performance
that scales nicely, especially given than a "no op" container build
and push takes at least 5-10 seconds. For the container apps sample I've
been using this is a ~20s improvement for me when all of the images are
up to date already.

The bulk of the work here is in adjusting the output to be color-coded
and prefixed. I was able to find some good examples to build on here.
The principle is that we implement infra for color-coded and named
output streams, and then layer an `io.Writer` implementation on top.

# Description

_Please explain the changes you've made._

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/pkg/cli/builders/docker.go
+++ b/pkg/cli/builders/docker.go
@@ -55,11 +55,19 @@ func (builder *dockerBuilder) Build(ctx context.Context, options Options) (Outpu
 		"-t", input.Image,
 	}
 	cmd := exec.CommandContext(ctx, "docker", args...)
-	cmd.Stdout = options.Stdout
-	cmd.Stderr = options.Stderr
+	writer := options.Output.Writer()
+	cmd.Stdout = writer
+	cmd.Stderr = writer
 
-	fmt.Printf("running: %s\n", cmd.String())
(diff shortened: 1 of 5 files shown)


karolz-ms (changes requested):
Parallel build 😍 
Prefixing each builder output for easy differentiation 👍 
Colorizing build output 😞 

My experience with using colors in output from CLI utilities has been almost always negative. I prefer light background for terminal windows and code editors; the color palette is almost always optimized for dark background, resulting in unreadable combinations such as light-yellow font on off-white background. Not to mention different kinds of color-blindness, which makes combinations like red-over-dark green totally invisible. That is why I tend to turn off colorized output for all utilities that have such option and avoid altogether those that don't.

So, what I suggest, as a minimum, there should be a command-like option to turn off color for an invocation of rad CLI. What I would like much, much more (and would always set) is some sort of .rc file where I could set "never use color" option for all rad CLI invocations.

PR #1758, opened 2022-01-23
Add rad app run command

rynowak:
This change adds a `rad app run` command which is very similar to deploy
except it blocks and outputs logs from the apps containers.

This command is a 'hub' for more features we're planning to build as
part of the local dev epic. For right now it just does logs but we plan
to integrate more tools over time.

Currently this integrates `stern` as the log streaming tool, and will
output installation instructions if you don't have it installed.
Longer-term we might choose to bring the functionality we want inside
Radius. For now we're building the local dev features incrementally.


karolz-ms (approved):
Nice!

PR #1803, opened 2022-01-26
Disable dependabot

AaronCrawfis:
Dependabot is broken in its current state. Removing until we have an end-to-end solution and process that handles:
- Secrets
- Review process


amanbha (approved):

jkotalik (approved):
If it's broken

PR #1915, opened 2022-02-10
Remove managed: true from code

jkotalik:
# Description

- Removes `managed: true` from
  - handlers
  - renderers
  - outputresources
  - database
  - rest representation
- Removes kind: any from dapr types because that implied managed: true

- Areas I'm concerned about
  - Confirming we remove all properties from types that were only used with managed: true

## Issue reference

Part 0 of https://github.com/project-radius/radius/issues/1901
Fixes https://github.com/project-radius/radius/issues/1500 (minus docs and tutorials which will be in the docs repo)
Fixes https://github.com/project-radius/radius/issues/758

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing
* [x] Extended the documentation / Created issue for it (issue/PR in flight already https://github.com/project-radius/docs/pull/22)


kachawla (commented):
Can remove this as well https://github.com/project-radius/radius/blob/main/pkg/handlers/namegenerator.go#L30-L57

kachawla (approved):

PR #1932, opened 2022-02-14
Install controllers for Azure environments

rynowak:
Part of: radius-project/core-team#542

This change installs the Radius helm chart in Azure environments. We'll
be adding runtime functionality to this chart, and in order to take a
dependency on it we need it to exist in all of our environments.

Parts of this change:

- Produce an 'edge' build on pushes to main (0.42.42-dev)
- Install our chart as part of the environment creation script
- Update the AKS cluster as part of functional tests

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -165,7 +165,7 @@ jobs:
     name: Helm chart build
     needs: ['images']
     runs-on: self-hosted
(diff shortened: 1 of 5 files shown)


jkotalik (approved):
Approved if either CI/CD is verifying the initialize script or if you manually ran it once.

PR #1978, opened 2022-02-25
schema changes for showing health/provisioning status as a generic property of a resource

vinayada1:
# Description

Adding health and provisioning status to Resource which is used by clients such as VSCodeplugin. With this, the client code does not need to know the resource type to extract the status.

#1955 

Please reference the issue this PR will close: #1955 
## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):

bjoginapally (approved):

kachawla (commented):
I'm a bit confused here. Over the wire format does not have this status as top level property so these fields get copied over as a part of properties field today. Are you planning to change rest format for resource with this change?

PR #2062, opened 2022-03-10
Adding changes to dynamically load config file when there is a change

vishwahiremat:
# Description

Added changes to dynamically load the config when config.yaml is edited.
This resolves the issue of config file getting overwritten when 2 commands are editing config.yaml file simultaneously.

Tested it manually, here are the steps:
1.  Deleted an existing azure environment.
2. While its being deleted, created a kubernetes environment, it gets created and added in the config file
3. Once azure environment is deleted, confirmed that kubernetes environment is not deleted on config file.

 
## Issue reference

Please reference the issue this PR will close: radius-project/core-team#555 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):
Pending feedback, I think this is looking good!

Good find on how we need the directory created before obtaining a file lock.

PR #2083, opened 2022-03-16
Standardizing -n for `rad env init` for naming environments

mhaveliwala1:
# Description

Currently, the behavior for naming environments is different across `rad env init` for kubernetes, dev, and azure. This PR aims to standardize that behavior, using -n to name environments across the board.

## Issue reference

Please reference the issue this PR will close: #_2073_

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -322,7 +322,7 @@ jobs:
       run: |
         export PATH=$GITHUB_WORKSPACE/dist:$PATH
         mkdir ~/.rad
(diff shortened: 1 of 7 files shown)


kachawla (commented):
The [initial reasoning](https://github.com/project-radius/radius/issues/278) for current convention for this environment name flag was to standardize it across different commands, for example `-e` is used here as well https://github.com/project-radius/radius/blob/main/cmd/rad/cmd/application.go#L22.

Could you please get inputs from @emily-potyraj / @AaronCrawfis on this, and if this should apply to other commands as well if we decide to go with this new convention?

jkotalik (approved):

PR #2085, opened 2022-03-16
Dynamic configuration of app model based on arm credentials

vinayada1:
# Description

Dynamic configuration of app model based on arm credentials
radius-project/core-team#565 

Please reference the issue this PR will close: radius-project/core-team#565 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


rynowak (commented):
This looks like it's overall headed in the right direction 👍  I think there's an opportunity to refactor some things we should take advantage of.

rynowak (approved):
This is great! I'd like to see if we can clean up some of the magic strings we're using (the `resourcekinds` package), but that should be easier after this is merged.

PR #2100, opened 2022-03-22
Fixing some cases where we should return an err rather than nil

jkotalik:
# Description

Found when investigating https://github.com/project-radius/deployment-engine/issues/49. We need to return errs in some places here.

I'd love to find some linter that handles this case, I searched for one and nothing quite handles it (probably because there are cases where you'd like to return nil, but still quite annoying).


kachawla (approved):
Do we have a backlog issue for unit test coverage for cli?

rynowak (approved):

amanbha (approved):

PR #2115, opened 2022-03-24
Dynamic query for resources in K8s validation

mhaveliwala1:
# Description

Changed K8s validation to check for namespace and labels of resources, instead of just GVR/GVK mapping. This should avoid some false positive tests we may have been seeing.

## Issue reference

Please reference the issue this PR will close: #_1977_

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):
Asked offline to do some quick validation to make sure negative cases cause CI/CD to fail (ex add a pod that doesn't exist and let the tests timeout)

LGTM otherwise.

PR #2119, opened 2022-03-24
Skip localdev integration tests except for PRs

rynowak:
This change disables the local dev integration tests for tags and pushes
to main. We don't generally run tests in those cases since we run as
part of the PR.

This is causing failures right now when a push to main happens, and
using extra cycles. The failures are because the publish assets phase
cleans up the build artifacts before the tests start, leading to
failures with downloading the needed artifacts.


jkotalik (approved):
That's very funny 😄

vinayada1 (approved):

PR #2123, opened 2022-03-28
Remove Custom RP from Azure environment scenario

rynowak:
Fixes: https://github.com/project-radius/radius/issues/2018
Fixes: https://github.com/project-radius/radius/issues/2019

This change removes the use of Custom Resource Provider from our Azure
environments. Instead the Deployment Engine and RP are hosted on AKS,
and our client code is updated to to talk to them. The DE and RP running
in AKS use a managed identity to talk to ARM and so there's no loss of
Azure functionality.

This is a step forward in converging our control-planes in that both the
Azure-hosted-environment and Kubernetes-hosted-environments are
converged -- they both use the DE and RP hosted in Kubernetes. This
means that the Kubernetes controllers for Radius are obsolete and can be
removed in the next step.

The major milestones here are threefold:

- We're able to delete the controllers and have a single implementation
- We're able to evolve the Radius API shape to match the proposed design
  (CustomRP was a blocker)
- We're able to evolve the Radius API shape to remove Azure concepts
  like subscriptions from the self-hosted version.

Mostly of the truly impactful changes have been done in previous PRs so
this is just a string of 'glue' changes to get the E2Es working.


jkotalik (approved):
Once this is working, LGTM

vinayada1 (approved):
This is awesome!! Had a question....would help to understand the reasoning.

PR #2171, opened 2022-04-07
change deployment engine image tag

mhaveliwala1:
Changing what image radius uses for the deployment engine from the latest tag to a stable tag for now, so that changes in the deployment engine don't cause a regression in radius.


jkotalik (approved):
For everyone else, this is quick fix to make it so we don't break the world every time we make a change to the deployment engine. Instead, we will make incremental PRs.

Long term, we'll come up with a semantic versioning plan here.

PR #2261, opened 2022-04-27
Using release version instead of latest for radius-rp tag

willdavsmith:
* Fixing issue where `rad` would always use latest version of `radius-rp` image.
* Removing container image push to latest on git tag.

Fixes radius-project/core-team#621 
Fixes radius-project/radius#2143


jkotalik (approved):
I think we can use some work in transparency around versions as well as validation. May be time to setup some CI/CD tests for environment creation.

PR #2316, opened 2022-05-06
Improve error handling for invalid config

rynowak:
This change improves the error handling/reporting for cases where you
have an invalid CosmosDB configuration. Previously the error was a panic
with "Failed to create request headers: illegal base64 data at input
byte 3" as the only hint to the problem.

This change removes the panic (now it's just an error) as well as adding
more contextual information using wrapping along the failing code path.

Additionally I improved the logging behavior where panics get handled,
previously we were not logging here, and so there was no way to
determine the call stack the caused the failure.

### DIFF
--- a/pkg/corerp/dataprovider/factory.go
+++ b/pkg/corerp/dataprovider/factory.go
@@ -7,6 +7,7 @@ package dataprovider
 
 import (
 	context "context"
+	"fmt"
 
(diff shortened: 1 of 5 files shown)


youngbupark (approved):
/lgtm

jkotalik (approved):

PR #2318, opened 2022-05-07
Remove response body from deletes

rynowak:
Removing an extra message from resource deletion. It's not a great idea
to return a string (vs an object) as JSON as it cannot be
evolved/versioned over time.

Additionally making some small improvements to tests.


youngbupark (approved):
/lgtm

PR #2336, opened 2022-05-10
Proposes the use of ADRs as a means to enhance collaboration and transfer context

asilverman:
# Description

I am creating the initial proposal to discuss the use of ADRs in our team and describes the process to use them effectively moving forward


youngbupark (approved):
/lgtm

PR #2344, opened 2022-05-12
remove metadata from bicep output

asilverman:
# Description

Removes metadata key from bicep build output so that bicep version no longer breaks the build. 

The `metadata` section in ARM-JSON is non-semantic https://github.com/project-radius/radius/blob/main/deploy/rp-full.json
 
They just use this for telemetry purposes like tracking version usage across Bicep releases. It doesn't affect the behavior of the deployment engine.


rynowak (approved):
awesome 👏

jkotalik (approved):

PR #2364, opened 2022-05-17
Install Open Service Mesh as part of rad env init kubernetes

ruokun-niu:
# Description

Open Service Mesh (OSM) is a service mesh that supports traffic splitting between two or more containers. Previously, to use OSM, a common approach is to install the OSM CLI and execute `osm install`. 

As a part of [Tommy's internship project](https://github.com/project-radius/radius/issues/2351), OSM will be used for performing traffic-splitting and we would like to have OSM installed when we create an environment in a Kubernetes cluster. In other words, when the user execute `rad env init kubernetes` in the terminal, OSM should be installed automatically, which eliminates the need to install the OSM CLI.

To achieve this, since the OSM chart can be installed directly via the Helm CLI, we will create an OSM client that is similar to this [radius client](https://github.com/project-radius/radius/blob/main/pkg/cli/helm/radiusclient.go). The installation function in this client will be called when the user execute `rad env init kubernetes`, along with several other necessary installations. 

### Key edits in this PR: 

- **Created osmclient.go** This file contains several functions regarding the installation of OSM in the environment. `ApplyOsmHelmChart()` will obtain the helm config and helm chart for OSM and call `runOsmHelmInstall()` to perform the actual installation of OSM if it is not previously installed in the cluster. `runOsmHelmUninstall()` is responsible for uninstalling OSM when the user performs `rad env delete [flags]`
- **Edited cluster.go** Functions in this file are responsible for calling the respective functions for installing/uninstalling tools. For OSM installation, the version is specified with a default value of "0.11.1". `InstallOnCluster()` will call `ApplyOsmHelmChart()`(defined in osmclient.go) with this version and OSM should be automatically installed as a part of the initialization. Similarly, uninstalling the OSM will go through a similar process. 

### Initial testing:

- After running `rad env init kubernetes` on a new Kubernetes cluster and then running `kubectl get pods -n radius-system`, one can see that the OSM-related pods are installed and running.
- Performed a single test on the traffic splitting capabilities by following this [tutorial](https://release-v1-1.docs.openservicemesh.io/docs/getting_started/install_apps/). Traffic splitting was carried out smoothly and no error was encountered
- Using the initialized cluster, a simple deployment (`rad deploy`) was created with a simple bicep file that contains a container, HTTPRoute, gateway and a MongoDB database. No error was encountered.

## Issue reference

Please reference the issue this PR will close: [#2361](https://github.com/project-radius/radius/issues/2361)


willdavsmith (changes requested):
Awesome!

ruokun-niu (commented):
Modified the files based on the comments

willdavsmith (approved):
LGTM

PR #2393, opened 2022-05-19
change de tag to latest

mhaveliwala1:
# Description

This PR changes the deployment engine version that radius environments use from being a hardcoded commit hash to being `latest`. We are able to do this since we have implemented some radius validation in the deployment engine repo [here](https://github.com/project-radius/deployment-engine/blob/main/.github/workflows/validate-rad.yaml), which means that we can pull the latest deployment engine with some confidence that it won't break radius. This change should empower us to use the most recent deployment engine work in radius without having to create a PR whenever we want to get the newest version.

## Issue reference

Please reference the issue this PR will close: #2392

### DIFF
--- a/deploy/Chart/values.yaml
+++ b/deploy/Chart/values.yaml
@@ -1,7 +1,7 @@
 global:
   engine:
     image: radius.azurecr.io/deployment-engine
-    tag: 291be62f27792baef9af3f596c8f0a794659cf97
+    tag: latest
   extensibility:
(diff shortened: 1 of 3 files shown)


jkotalik (approved):
Pending passing legs and remove the changes to go.mod and go.sum files.

bjoginapally (approved):

amanbha (approved):

PR #2416, opened 2022-05-23
Combine UCP and Radius versions of the store

rynowak:
This change combines our two versions of the store abstraction and
updates to use the canonical one everywhere.

There were a few changes to the contract as part of this:

- UCP version returns a QueryObjectResult from Query
- Radius version updates the object in place in Save
- UCP version uses strings instead of resource IDs as parameters

Overall the functionality should be the same. We have good tests in
place which failed when I updated these sites, so I was able to fix
regressions.


vinayada1 (commented):
Looks good otherwise.

vinayada1 (approved):

PR #2427, opened 2022-05-24
Make data store configurable for RP and worker

rynowak:
This change makes the data store configurable for the CoreRP and Core
async worker using our existing configuration files. Now this will
support:

- ETC.d (in memory) for dev/test
- APIServer (Kubernetes) for dev/test and production self-host

See the configuration files for examples of how to configure this.


shalabhms (approved):
/lgtm. Thanks .

PR #2533, opened 2022-06-07
Disable subscription API for self hosted mode

youngbupark:
# Description

E2E test is failed because of subscription api is enabled at `/` path. This is to disable subscription api endpoint.

## Issue reference

Please reference the issue this PR will close: #2535

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


ytimocin (approved):
We have some routes that are only available if it is self hosted and some only if it is not self hosted. I think we can make a list of those routes.

PR #2550, opened 2022-06-09
Add validation for name field in UCP resource group and planes

vinayada1:
# Description

Add validation for name field in resource group and planes

#2548 

Please reference the issue this PR will close: #2548 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


nithyatsu (commented):
I had originally the plane name and type in body, but Ryan had corrected that, since we should not have the fixed part in body. So it now just has the properties. radius / local cannot be edited once created, so they are inferred from just the path.

PR #2561, opened 2022-06-09
Add resource list command for connector rp

bjoginapally:
# Description

Since the existing management client is being used by other commands. We will have to create a new management client with root scope instead of subscription Id and resource group. This PR attempts to create a new client, Naming of the client can be up for debate. Since we're creating 1P service now, thought of naming the creation function with "FirstPartyService". I and nithya can parallelize the cli commands by adding the required functions into ucp_management.go

Sample screenshot:
Screen Shot 2022-06-16 at 3 59 10 PM

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


nithyatsu (commented):
ucp_management under azure package does not sound right.. Should it not be under a kubernetes package?

nithyatsu (commented):
I will use the ARMUCP management client in kubernetes.go, change the struct to have a Scope instead of resourcegroup and subscription.

kachawla (commented):
Can you please add more details in the description in the description on what you mean by parallel management framework for ucp? It isn't clear what we are trying to achieve here.

jkotalik (dismissed):
I think this needs to be an API on the server side. Is that currently designed? Do we need to add an API spec for that?

kachawla (commented):
Could you please paste output from local testing in the description? Will help making sure it aligns with expected user experience described in https://github.com/project-radius/radius/issues/2297

rynowak (approved):

shalabhms (approved):

nithyatsu (commented):
@bjoginapally  I was rebasing the env list and ran into a couple of questions. Sorry for not bringing it up earlier. 

Should the below not change for ucp to be just the resource group associated with the environment? ex: 

RootScope:"/ResourceGroups/"+e.UCPResourceGroupName,

@vinayada1 @asilverman @jkotalik That would map to env.UCPResourceGroupName correct?

currently in kubernetes.go: 
return &ucp.ARMApplicationsManagementClient{
		EnvironmentName: e.Name,
		Connection:      connection,
		RootScope:       "/Subscriptions/" + e.SubscriptionID + "/ResourceGroups/" + e.ResourceGroup,
	}, nil
}

PR #2611, opened 2022-06-16
MAINT: Allow setting the path for test execution when using vscode codelens functionality

asilverman:
# Description

Sometimes you want to debug/run your go tests on radius on the latest built version of rad, however because the current implementation is using `LookupPath` to find `rad` and you may have `rad` GA installed in your system you end up with the computer executing your tests on a different bit than you intended.

With this change you can set your vscode settings.json file to contain:
```json
{
 "go.testEnvVars": {
        "RAD_PATH": "${workspaceFolder}/dist/linux_amd64/release"
    },
}
```
And now the intended bit will be used for debug...

See example:
If you want to debug the test below using the codelens link (yellow circle) and want the rad executable to be resolved from your `dist/linux_amd64/release` folder, then you set the JSON setting above in `settings.yaml`
image


jasonviviano (approved):
Created a task to track the documentation of this new behavior, LGTM if @jkotalik  thinks the code changes are good.

jkotalik (approved):

PR #2622, opened 2022-06-17
Container Renderer to use new data model type, leveraging the existing renderer

shalabhms:
# Description

Adding container renderer leveraging the existing container renderer to work with new RP data models.
This change list is to update the interface and implement Container resource with the new interface.
Container controller will be then calling this renderer to create K8S constructs.

## Issue reference

Please reference the issue this PR will close: https://github.com/project-radius/core-team/issues/234

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing

### DIFF
--- a/pkg/corerp/api/v20220315privatepreview/container_conversion.go
+++ b/pkg/corerp/api/v20220315privatepreview/container_conversion.go
@@ -452,18 +452,18 @@ func fromExtensionClassificationDataModel(e datamodel.ExtensionClassification) E
 func toHealthProbeDataModelProperties(h HealthProbeProperties) datamodel.HealthProbeProperties {
 	return datamodel.HealthProbeProperties{
 		Kind:                to.String(h.Kind),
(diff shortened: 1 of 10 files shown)


jkotalik (commented):
I think we need to think about how we invoke the functions that we do in https://github.com/project-radius/radius/blob/main/pkg/radrp/backend/deployment/deploymentprocessor.go. I chatted with @lakshmimsft about this briefly and I think we may need to be slightly coordinated on it.

ytimocin (approved):
LGTM
nit: `render_test.go` can be simplified by using test cases and grouping similar tests. Some variables can be moved to the top of the file and shared.

bjoginapally (approved):

rynowak (approved):

jkotalik (approved):

youngbupark (approved):

kachawla (approved):

PR #2643, opened 2022-06-21
Generic client generation

bjoginapally:
# Description

Core and connector clients are strongly typed. This is an attempt to loosely couple the clients from resource types data model and their endpoints

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


youngbupark (dismissed):
This should not work for ARM. In ARM, each resource type must be defined and registered separately. we may be able to share the request and response definition with the different resource type names. Such swagger api definition will not be accepted.

rynowak (approved):

PR #2654, opened 2022-06-22
rad env ucp support 

nithyatsu:
# Description

changes to support rad env list in ucp env.

NOTE: currently I have hard coded the scope value as RootScope: "/ResourceGroups/dummy". It should be accessible from "env" as env.resourceGroup once rad env creation is working for ucp enabled scenario

## Issue reference

Please reference the issue this PR will close: #2609 

## Checklist

* [X] Code compiles correctly
* [ ] Unit tests passing


jkotalik (changes requested):
Do we have mock clients we can use for testing? I think a test where we mock the response from the Core RP would allow us to test that our environment structure parsing works.

PR #2716, opened 2022-06-28
Configure TargetPort values for HttpRoutes with TrafficSplit configurations

ruokun-niu:
# Description

When we attempt to create a Kubernetes service from a HttpRoute that has traffic-splitting properties configured in it, we need to make sure that the TargetPort of this service matches the port number of the destination services. I have modified the renderer for HttpRoute to achieve this. Errors will be thrown if we can not find/use a valid port number. 

Please reference the issue this PR will close: radius-project/core-team#682 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):
This looks generally correct. I assume a separate PR for the namespace patching change?

PR #2732, opened 2022-06-29
Configure osm namespace <add> using k8s API

ruokun-niu:
# Description

Added a label and annotation in PatchNamespace(). This essentially achieves `osm namespace add <namespace>` for a particular namespace.

One thing to note is that with this implementation, we are adding all of the namespaces to the service mesh (if I understood things correctly). I am unsure if this is a desired behavior. We can modify the file by adding a conditional so that only the desired namespace(s) will be added to mesh 

## Issue reference

Please reference the issue this PR will close: #2729 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):
This looks like what I would expect!

PR #2734, opened 2022-06-29
Move /pkg/queue to /pkg/ucp and refactor queue implementation

youngbupark:
# Description

> I will create another PR to update helm chart to use CRD based queue.

* Move /pkg/queue to /pkg/ucp
* Reimplement inmemory queue with new client interface
* Add apiserver option
* Created the shared queue test and remove the duplicated test codes.

## Issue reference

Please reference the issue this PR will close: #2733

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


rynowak (approved):
One minor piece of feedback about the config, but I don't feel super strongly.

ytimocin (commented):
One thing I am trying to understand is that, why should the queue implementation live under ucp? I understand that we will have a few different implementations of the queue, but the one that we will use for async operations, do they belong to UCP?

In my understanding, queue will take care of deploying UCP and ARM resources. Isn't that right? And, for example, in Core RP we have some async endpoints now. Those will be handled by the queue. Why should they be processed by a UCP queue?

shalabhms (approved):

PR #2744, opened 2022-06-29
Remove rad env init azure support

rynowak:
Fixes: radius-project/core-team#679

This change removes `rad env init azure` and all of the supporting infra
for that command. Additionally this disables the "Azure" leg of our
tests for the OLD CUSTOM RP IMPLEMENTATION since that relied on `rad env
init azure`.

I have remove all of the references I can find in the repo, and will
follow up separately with:

- The architecture doc (very out of date)
- The docs repo


jkotalik (approved):
LGTM, I think we can maybe delete more 🔥

PR #2787, opened 2022-07-01
Add queue messages to permissions

jkotalik:
Fixes https://github.com/project-radius/radius/issues/2786


youngbupark (approved):
nice catch.

PR #2812, opened 2022-07-06
Separate install and env commands

rynowak:
Fixes: #2735

This change separate the `rad env install/uninstall` commands for the
environment concept.

- It is now possible to install without creating an environment
- It is now possible to uninstall without having created an environment
- `rad env init ...` will still install if necessary

See the issue for design notes.

In general this PR is about uncoupling things that do not need to be
coupled. Along the way I found a lot of things that could be simplified
and made less monolithic. I also leaned into code re-use where we're
trying to standardize behaviors. I found places where we have an
uncomfortable amount of code in a top level CLI command and agressively
turned those things into members of `pkg/`.


jkotalik (approved):
Just to confirm, how does this PR avoid the issue of having rad env init clobber existing configuration (like azure stuff). If I do `rad install --provider-azure...` and then do `rad env init`, the call to `rad env init` would noop the installation, right? Only if --reinstall is set would it recreate it.

PR #2850, opened 2022-07-08
remove corerp prefix from validation

mhaveliwala1:
Contributes to radius-project/core-team#709. Removes `corerp-*` prefix from the Bicep validation and uses `import radius as radius` to parse files that use corerp instead. Will remove this check after customrp is removed.


willdavsmith (approved):
Looks great, can we do this for docs as well?

PR #2867, opened 2022-07-12
Add support for workspaces

rynowak:
This change adds support for 'workspaces' as the client-side concept for
configuration in Radius. The workspace encapsulates:

- A connection to a radius install
- (optional) a scope for radius resources
- (optional) an environment id

Workspaces are a replacement for the v0.11 **client-side** functionality
associated with environments. Now that environments are a
**server-side** concept, we need different terminology for the
configuration file. If you understand what a kubernetes context is then
you're positioned well to understand workspaces.

Workspaces come with the standard CRUDL commands for manipulation. There
are two user journeys we think about for creating workspaces:

- (easy) rad env init ... will create a new workspace if necessary
- (manual) rad install + rad workspace init

Users should only need to know about workspaces when they need more than
one. eg: for a new user doing a tutorial, a workspace is an implementation
detail.

Since the scope + environment concepts are optional, we plan to allow
users to provide them at the command line everywhere that they matter.
The workspace is just a convenient way to save this data. Most users
will end up with a saved scope + environment.

closes https://github.com/project-radius/radius/issues/2757


jkotalik (approved):
Looks generally good to me with some minor questions.

bjoginapally (approved):

PR #2926, opened 2022-07-14
Fix rad resource logs and rad resource expose

rynowak:
Fixes: #2148

This change updates the rad resource logs and rad resource expose
commands to the new resource model.


youngbupark (approved):
aaaarg another case sensitive problem.

kachawla (approved):

PR #2948, opened 2022-07-15
Adding resource + listSecrets to routes

ytimocin:
# Description
When a request with a custom action comes in, the API Validator cannot validate the request because resource type is returned as resource+customAction (for example; Applications.Connector/mongoDatabases/listSecrets). We need to make sure that we don't return the custom action when getting the type for a resource.

## Issue reference
Please reference the issue this PR will close: #2947 

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


youngbupark (changes requested):

jkotalik (approved):

youngbupark (approved):
Please add Type() test. Other than this, LGTM. Thanks for this change.

PR #2969, opened 2022-07-15
Fix handling of different workspace cases

rynowak:
Fixes: radius-project/core-team#733

Fixes: radius-project/core-team#735

This change fixes the handling of different casing for workspace names
for radius-project/core-team#733. Unfortunately viper lowercases all keys for storage, but it
**SILENTLY TOLERATES**  duplicates that differ by case. This means that
we need to be consistent in using lowercase everywhere we
programmatically interact with the workspace name. I did my best to add
verification for this, so we can avoid introducing more bugs in the
future.

For radius-project/core-team#735 we had the wrong logic on a boolean check - the logic for
determining if a workspace was reusable was inverted. Oops.

# Description

_Please explain the changes you've made._

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/cmd/rad/cmd/appSwitch.go
+++ b/cmd/rad/cmd/appSwitch.go
@@ -56,7 +56,7 @@ func switchApplications(cmd *cobra.Command, args []string) error {
 		}
 
 		workspace.DefaultApplication = applicationName
-		section.Items[workspace.Name] = *workspace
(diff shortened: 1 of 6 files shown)


ytimocin (approved):
Do we need to have tests for this change?

PR #2993, opened 2022-07-18
Updating computed values across the board

jkotalik:
There are a lot of issues with how we handle computed values:
- Don't set them from all connectors
- Some computed values are not in json, this should be fixed now.


ytimocin (commented):
Unit tests must be updated

kachawla (approved):

PR #3023, opened 2022-07-19
Trying stuff to fix make generate

jkotalik:
I don't see any recent changes that should have affected this, going to do some binary searching I guess?


ytimocin (commented):
There was a commit to the autorest repo 5 hours ago. Maybe instead of latest we should use a specific version.

youngbupark (approved):

PR #3043, opened 2022-07-20
Log errors on resource deletion rather than failing delete

jkotalik:
Seeing what is returned

Will fix https://github.com/project-radius/radius/issues/3017 (once we figure out what's going on)


bjoginapally (dismissed):

youngbupark (changes requested):
We should not skip the resource deletion for failures, which is workaround and functionally incorrect. @vinayada1 created PR to fix real root cause soon.

PR #3165, opened 2022-07-26
Grant Bicep extensiblity permission on everything

rynowak:
Fixes: radius-project/core-team#806

This change grants Bicep extensibility maximal permissions on the
cluster. Right now we have a more limited set of permissions set which
lead to spotty limitations and failures.

This is ultimately something we'd like to review and come up with a more
nuanced policy. I'm adding this to our design queue to review.


jkotalik (approved):

ytimocin (approved):
Do we want to have a follow-up issue to remove the admin rights in future?

PR #3236, opened 2022-07-29
Removing unused output resources of the same resource

ytimocin:
# Description
Imagine deploying the same resource twice with different properties. We have to delete the old output resources that are not used by the new version of the resource.

What if the same output resource is being used by two different resources and one ends up deleting that output resource?

~~Note: This doesn't remove the secrets. Will continue investigating.~~ It removes all resources that are not used by the new version of the main resource.

## Issue reference
Please reference the issue this PR will close: radius-project/core-team#756 

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing

### DIFF
--- a/pkg/armrpc/api/v1/types.go
+++ b/pkg/armrpc/api/v1/types.go
@@ -160,6 +160,10 @@ type ResourceStatus struct {
 	OutputResources []outputresource.OutputResource `json:"outputResources,omitempty"`
 }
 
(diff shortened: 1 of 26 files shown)


jkotalik (approved):
I think my comment will simplify the logic of finding the diff, otherwise code LGTM

vinayada1 (approved):

youngbupark (approved):

kachawla (approved):

PR #3201, opened 2022-07-28
Add test coverage for cli commands

bjoginapally:
# Description

issue: https://github.com/project-radius/radius/issues/3170
Adding functional test coverage for rad commands. This is part of two PRs. The next will be dealing with deletion scenarios

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Adds necessary unit tests for change
* [ ] Adds necessary E2E tests for change
* [ ] Unit tests passing
* [ ] Extended the documentation / Created issue for it

PR #3243, opened 2022-07-29
Update CLI docs to reflect change from "Container" to "containers"

AaronCrawfis:
# Description

Based on https://github.com/project-radius/radius/pull/3201/files#diff-201dafbce4b1898d58153651f27250c637b4324fd3138087f869e070ac980017 we need to update all our CLI docs to use "containers" instead of "Container"


bjoginapally (approved):
Thanks for updating the docs

PR #3271, opened 2022-08-01
Add debug command that will capture all logs from the radius control plane as a zip

jkotalik:
# Description

Adds command called `rad feedback` (name pending) which will capture all logs from the radius control plane and put it in a zip folder. We can then update our bug instructions to mention "run `rad feedback` and upload logs to this issue".

## Issue reference

Part of https://github.com/project-radius/radius/issues/1324
Part of https://github.com/project-radius/docs/issues/226

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/cmd/rad/cmd/debuglogs.go
+++ b/cmd/rad/cmd/debuglogs.go
@@ -55,7 +55,6 @@ func debugLogs(cmd *cobra.Command, args []string) error {
 
 	c := connection.(*workspaces.KubernetesConnection)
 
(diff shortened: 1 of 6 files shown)


rynowak (commented):
It's really cool to see some traction on this!. 

One of the things we have to keep top of mind is **privacy**. We need to make sure the user understands that they are responsible for reviewing/disclosing whatever is in the logs, and that they don't just **trust us**.

I think we should also make sure we're only collecting the logs we need for Radius (not dapr, not contour).

willdavsmith (approved):

PR #3289, opened 2022-08-02
error message prompting for a valid env before deployment

nithyatsu:
# Description

rad deploy should print a suitable error if deploying without an environment.

## Issue reference

Please reference the issue this PR will close: radius-project/core-team#832

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


jkotalik (approved):
I think we should file a follow up to potentially confirm the environment resource exists before providing it as a parameter.

PR #3408, opened 2022-08-22
Check if Service Error conversion results in nil

jkotalik:
# Description

This fixes two bugs with our error presentation logic for presenting ARM errors to the user:

- The unfold logic didn't handle cases where Unmarshal returned nil for the details, which was happening in original issue's case. With the nil check, we would have at a minimum seen at least an error saying an unformatted error.
- Adds logic to Unmarshal both ErrorResponse and ErrorDetails. I added an in-depth description for why this logic was added in code-comments as it's a bit nuanced.

## Issue reference

Fixes: https://github.com/project-radius/radius/issues/3391

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/pkg/azure/clients/unfold.go
+++ b/pkg/azure/clients/unfold.go
@@ -89,43 +89,9 @@ func UnfoldErrorDetails(d *radclient.ErrorDetail) *radclient.ErrorDetail {
 		return &new
 	}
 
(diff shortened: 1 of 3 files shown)


ytimocin (approved):
lgtm

PR #3447, opened 2022-08-27
Remove more old code and consolidate common things

rynowak:
This change removes more old code (health, old API types, old id parser)
and consolidates things (error codes, rest return types, API types that
are still relevant).

For this more part this change can be understood as moving 'common'
ARM-RPC api types into the pkg/armrpc folder.


youngbupark (approved):
Thanks for this change!

ytimocin (approved):

PR #3459, opened 2022-08-30
Use generic version of list and get operation controller for Applications.Core

youngbupark:
# Description

This PR defines the generic version of List and Get operation controller and replaces List and Get operation controller of CoreRP.

> Note: I will refactor CoreRP first and then apply the same pattern to Connector RP

## Issue reference

#3430 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


rynowak (approved):
Awesome 😻

kachawla (commented):
Thanks for making this change, so much cleaner now! Should this be applied to delete controllers as well?

PR #3502, opened 2022-09-03
Remove coupling to one subscription/resourceGroup

rynowak:
This change makes our use of Azure SDKs more flexible for a variety of
Azure scenarios. In old CustomRP model for Radius, we configured the
control plane with a single 'destination' resource group to use for
resources. This limitation was mostly to keep things simple while we
were running Radius as a 'user-mode-PaaS' in the user's subscription.
This choice leaked into all the code, where we ignore the actual
subscription/resourceGroup on the resources we were working with in
favor of our single value for each.

There's no good reason for this coupling, and the correct thing to do in
every case is parse the resource IDs of the resources we work with so we
can correct scopes. For code hygiene we should run, not walk, away from
the old model :).

This change fixes the way we handle scopes to be data-driven by the
actual resources, and removes the settings from the Helm chart that we
used to rely on.

The one part of the code that can't easily be updated is the handling of
managed identity and pod identity. Getting these functional again (they
aren't currently supported or working) will require the *user* to
provide this information as part of the environment configuration. For
now I've left a hefty comment in place to make it clear what to do if we
resurrect this code.

### DIFF
--- a/deploy/Chart/charts/appcore-rp/templates/appcore-rp.yaml
+++ b/deploy/Chart/charts/appcore-rp/templates/appcore-rp.yaml
@@ -107,8 +107,6 @@ stringData:
   # Enable ARM without explicit credentials
   SKIP_ARM: 'false'
   ARM_AUTH_METHOD: 'ManagedIdentity'
-  ARM_SUBSCRIPTION_ID: {{ .Values.global.rp.provider.azure.subscriptionId }}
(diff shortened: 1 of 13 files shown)


ytimocin (approved):
Would it be valuable to add a functional test that creates resources in different subscriptions and resource groups?

youngbupark (approved):

jkotalik (approved):

PR #3507, opened 2022-09-06
fix destructive changes to users PATH and registry

asilverman:
# Description

* Fixes destructive changes to users PATH variable
* Adds tests to the `install.ps1` script to verify the changes in the PATH variable

Fixes: #3355


danieljurek (approved):
The installer env variable updating changes LGTM. 

I recommend setting up some test scenarios in a CI pipeline of sorts to ensure that changes don't regress installers.

youngbupark (approved):

PR #3509, opened 2022-09-07
Add provider command for managing cloud config

rynowak:
Fixes: #3436

This change adds a new top level command (`provider`) for manipulating the settings of cloud providers separate from the Radius install. This new command has really limited support (single instance of each provider, and only supports Azure) based on what the backend can do in v0.13. We have some big plans for this in the future, but we are URGENTLY decoupling the configuration of Azure credentials from Radius installation based on our experiences with v0.12.

The actual functionality of these commands is being working on by @bjoginapally separately. This change adds the commands as 'hidden' and they can become visible when the 'client' part of this change is done, or not at all for v0.12 if we miss the train.


ytimocin (approved):
other than some small nits LGTM

jkotalik (approved):

bjoginapally (approved):

PR #3516, opened 2022-09-08
`rad env create`

sk593:
# Description

need to add unit tests and resource group checks and clean up some code but basic implementation is done

## Issue reference

Fixes: #3433

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


AaronCrawfis (changes requested):
Couple small description updates.

AaronCrawfis (changes requested):
There's an edge case I'm hitting when `config.yaml` does not have a scope set (got into this state by running `rad workspace create`):

```yaml
workspaces:
  default: aacrawfi
  items:
    aacrawfi:
      connection:
        context: aacrawfi
        kind: kubernetes
```

This leads to:

```bash
PS D:\Repos\project-radius\radius\cmd\rad> .\rad.exe env create testenv
Error: '' is not a valid resource id
```

Is there a way we can make this error message a little more useful? Or check that a scope is set in the users workspace. I would expect this to be:

```bash
PS D:\Repos\project-radius\radius\cmd\rad> .\rad.exe env create testenv
Error: no resource group specified or set as default. Specify a resource group with '--group' and try again.
```

ytimocin (approved):
other than Aaron's comments on documentation lgtm

AaronCrawfis (approved):

PR #3535, opened 2022-09-09
Add secret hash to deployment annotations

rynowak:
Fixes: radius-project/core-team#874

This root cause of this bug is that the content of the Kubernetes Deployment does not change in this case. We store the "connection values" in Kubernetes secrets and reference those secrets from the deployment. However, updating the secret DOES NOT cause a new rollout of the deployment/replicaset.

We've probably had this bug for a very long time, and now that we're doing a better job testing we just happened to notice.

The fix here is to hash the secret data and add it as an annotation on the deployment resource. This is the standard solution for these cases, and is used by a number of other projects.

### DIFF
--- a/pkg/corerp/renderers/container/render.go
+++ b/pkg/corerp/renderers/container/render.go
@@ -7,6 +7,7 @@ package container
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


shalabhms (approved):
Thanks Ryan for making this change !

kachawla (approved):

youngbupark (approved):

jkotalik (approved):

mishrapratikshya (approved):

PR #3705, opened 2022-09-14
rad group commands

nithyatsu:
# Description

Support rad group commands compliant with new cli framework and using ucp generated clients. 
Also brings in UCP swagger + auto generated clients since its a dependency from https://github.com/project-radius/radius/issues/3510

## Issue reference
Fixes: #3435 

## Checklist

* [X] Code compiles correctly
* [x] Unit tests passing


AaronCrawfis (changes requested):
Added some comments on CLI descriptions. Biggest question is if we should refer to them as "RAD resource groups" or just "resource groups", given they're a UCP concept and not a Radius concept (using the definition of Radius being the RPs). We can still refer to them as "RAD" or "Radius" resource groups if we want to refer to Radius as the project & entire control plane.

youngbupark (changes requested):

bjoginapally (approved):

AaronCrawfis (changes requested):
Couple notes on text

AaronCrawfis (approved):

youngbupark (approved):
I left two comments. I do see this is missing comments for public functions. I strongly recommend to add comments for public functions (it is fine to not have it in test)

PR #3942, opened 2022-09-28
Add specialized ID parsing functions

rynowak:
This change adds more functions for parsing different kinds of resource IDs as well as cleaning up the semantics for how we detect different kinds.

New code should use ParseResource or ParseScope if possible. Parse is very general and doesn't provide a lot of guardrails. Only very very general-purpose code should use Parse (eg: a lot of places in UCP, or middleware for logging/metrics). The vast majority of code should use ParseResource.

I've updated all of the existing callsites for Parse to use a more specialized function if possible. I've also added a comment as justification in the places that have a non-obvious reason to use Parse.

### DIFF
--- a/cmd/rad/cmd/appStatus.go
+++ b/cmd/rad/cmd/appStatus.go
@@ -62,7 +62,7 @@ func showApplicationStatus(cmd *cobra.Command, args []string) error {
 	}
 
 	for _, resource := range resourceList {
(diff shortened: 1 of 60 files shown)


ytimocin (approved):

youngbupark (approved):
/lgtm

PR #3943, opened 2022-09-28
Fix context name

AaronCrawfis:
# Description

https://github.com/project-radius/radius/commit/b0cb333a69c0e1f4c42235950e80ac2a5ce54697 changed the prompt for selecting a kubeconfig context, and appended `(current)` to the name of the current context before inserting it into the context array.

The returned context is pulled from the context array, so now instead of returning the name of the default context, we're returning the appended string instead, which causes issues when the client tries to use the context.

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


ytimocin (approved):
Do we need to add a test for the order of the configs? Other than that LGTM.

PR #3975, opened 2022-09-30
Recipe handler/renderer updates

kachawla:
# Description

* Server/client error separation; logging; consolidating duplicated code
* Move Azure resource type to secrets reference closer to map it to secrets action - recipe output can contain multiple resources but secrets action can only be performed only on one specific resource type. For example for Cosmos DB based Mongo connector, listConnectionString is allowed on Azure Cosmos Account instead of Cosmos DB resource type.

## Issue reference

https://github.com/project-radius/radius/issues/3957

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


ytimocin (approved):
left a few comments and questions but lgtm in overall

PR #4420, opened 2022-10-28
Fix CLI argument validation

rynowak:
There were some gaps in how we were validating negative cases for CLI parsing. I found this while writing some new tests, and I was surprised when my negative tests didn't fail correctly.

Since the validation for the validation was invalid, it was possible for one of our negative tests to return a false-positive.

Fortunately NONE of our existing tests were affected. The new tests I was writing were the only ones hitting the validation problem.


nithyatsu (approved):

bjoginapally (approved):
Do you think it would be much easier if we have an error in the test case struct instead of ExpectedValid. That way we can directly check if the test error is not nil and check for equalness.

PR #4421, opened 2022-10-28
Refactor workspace commands and add tests

rynowak:
This change refactors the 4 remaining workspace commands to the updated structure and adds tests for the current behavior.

Separately from this PR I am making some changes to workspaces and so this seemed like a worthwhile step to take first to ensure we don't break anything. The behavior of these commands will change in some subtle ways as part of radius-project/core-team#999.

I successfully resisted the urge to refactor how we're handling configuration, so it continues to grow organically for now.


ytimocin (approved):
looks neat

bjoginapally (approved):

PR #4430, opened 2022-10-29
Add location and tag support to Resource Groups

rynowak:
This change fixes bugs in the handling of location and tags for resource groups. As-is these fields were not handled by the converter and so would be omitted from the response even if they were specified on PUT.

This change also specifies a location (my original goal) in all of the places where create resource groups. Turns out tht we have a bug where location is not validated #4429.

This change also adds these fields to the relevant tests and and does some minor code cleanup. Many of these tests were using raw HTTP bodies or alternate data-model types that allowed us to drift from what's expected.


ytimocin (approved):
a few nits but other than them LGTM

vinayada1 (approved):

ytimocin (approved):
We should have more consts like this! Thanks!

PR #4435, opened 2022-10-31
Remove deprecated commands

rynowak:
This change removes CLI functionality that we deprecated during private release. We don't have any active plans to bring this functionality back in it's current form.

Why do this now? Well we've been steadily cleaning up and adding tests for our CLI commands. This removes EIGHT commands that we don't care about updating or adding tests for.

Secondly, we'll be revisiting *some* of this functionality, but not in exactly the same shape or form. It will be easier to review as a "net new" than as a refactor of code that no one is familiar with.


bjoginapally (approved):

ytimocin (approved):
should the docs on the Radius web site be updated?

kachawla (approved):

vinayada1 (approved):

PR #4464, opened 2022-11-02
Add Identity to Containers resource and remove system identity support

youngbupark:
# Description

Workload identity (aka Federated identity) is supported at cluster-level. It means setting workload identity at containers or volume resource doesn't make sense. Instead, we will populate `identity` property in Container resource while creating Containers resource using environment-level identity infomation.

- Remove identity property in Volume resource
- Populate Identity as readonly in Container
- Remove `azure.com.systemassigned` because we want to support per-container managed identity
- Remove Resource in Identity because PUT container creates its own managed identity.
- More cleanup

## Issue reference

Fixes: #4394 

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


sk593 (approved):

shalabhms (approved):

kachawla (commented):
LGTM

PR #4448, opened 2022-11-02
MBCP - updates to container,env,app datamodel

lakshmimsft:
Updates to container,application,environment datamodel to account for KubernetesMetadataExtension

Note: some overlap with PR https://github.com/project-radius/radius/pull/4443/files
which we have now resolved

#4442 

Fixes: #4442

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Adds necessary unit tests for change
* [ ] Adds necessary E2E tests for change
* [ ] Unit tests passing
* [ ] Extended the documentation / Created issue for it

PR #4469, opened 2022-11-03
Updating datamodel to use ContainerExtension types

jkotalik:
# Description

Fixes datamodel for Container/Application/Enviornment for Extensions. Forgot to add the right type as the array type.

Again the place we are starting from is that Application Extensions are separate from Environment extensions, etc. We are assuming that it is a bad user experience to be able to specify the DaprExtension at the application layer that noops or throws at deploy time rather than at compile time.

## Issue reference

Part of https://github.com/project-radius/radius/issues/4442, https://github.com/project-radius/radius/pull/4448 will finish the fight

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


rynowak (approved):
LGTM - would be good to get a signoff from @ytimocin or @youngbupark  as well.

PR #4480, opened 2022-11-04
MAINT: Make sure to validate biceps files of all `project-radius` repos to prevent drift

asilverman:
# Description

Runs the bicep validation on all the repos that need to be updated to prevent drift

## Issue reference

Fixes: #4361


AaronCrawfis (dismissed):
Love it! Is the expectation that all docs/samples are updated and PRs merged before the radius repo PR can be merged?

Does this work as expected when there's a change to schema that would result in a chicken/egg situation where radius depends on docs which depends on radius?

vishwahiremat (approved):

PR #4492, opened 2022-11-06
Add logic to `rad init` for existing environments

rynowak:
Part of: https://github.com/project-radius/radius/issues/4303

This change adds support to `rad init` to work with existing environments. Currently `rad init` assumes that you will always be creating a new environment which isn't usual in practice. The common cases are:

- Setting up a cluster for the first time
- Connecting to an existing environment 

If users select 'reinstall to update cloud providers' we will not ask about existing environments. We need to modify/create an environment to do that.

If there are no existing environments we will also not ask about it.


bjoginapally (approved):
LGTM, thanks for adding the mock function

PR #4679, opened 2022-11-17
Add CLI log-streaming infrastructure

rynowak:
This change adds infrastructure for streaming logs for a *whole application* at a time. This is using a project called Stern as a library. This will be used in the upcoming `rad run` command, and is being PRed separately to make the changes more reviewable.

This change doesn't include a lot of our own business logic, and just includes the boilerplate to use Stern (a lot of config) as well as an interface and mocks so we can write tests.


ytimocin (approved):
neat!

kachawla (approved):

PR #4707, opened 2022-11-23
Trying self hosted runners to see if they work

jkotalik:
To unblock CI/CD


willdavsmith (approved):
thanks for doing this

youngbupark (approved):

PR #4841, opened 2022-12-08
Introduce TestSecretResource helper in corerp e2e test

youngbupark:
# Description

The old test secret injection is designed only for `default` namespace. Since we support app-scoped namespace, we need to have a way to specify target namespace.

This change introduces TestSecretResource helper for initialResources arg of NewCoreRPTest() to set up the test secret instead of having additional code and arguments to handle test secret.

## Issue reference

Fixes: #4839 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing

### DIFF
--- a/test/functional/corerp/cli/cli_test.go
+++ b/test/functional/corerp/cli/cli_test.go
@@ -342,8 +342,6 @@ func Test_CLI(t *testing.T) {
 	template := "testdata/corerp-kubernetes-cli.bicep"
 	name := "kubernetes-cli"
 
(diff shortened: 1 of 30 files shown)


ytimocin (approved):
The issue I ran into while playing around with secrets is that in the test pipeline, when we are running the tests in a parallel way, the secrets with the same name clashed sometimes. Like a test may be trying to delete a secret that is being used by another test therefore making that test fail. We need to make sure that we have **unique names** for the secrets like other resources we have.

PR #4856, opened 2022-12-13
Validate Dapr install in Dapr links

rynowak:
Part of: #2952

This change validates the installation status of Dapr inside links RP for Dapr resources. This is step one in removing Dapr from Radius' default installation. After we have a good experience with Dapr missing, we can remove it.

This validation is only applied on PUT calls to be more friendly to API consumers. We still allow clients to get and list dapr link resources without Dapr installed.

The pattern I followed is similar to the validation we perform for the AKV CSI driver.


ytimocin (approved):
In future we can have a BaseDaprController and move the shared logic to that Controller.

PR #4857, opened 2022-12-13
Remove Dapr from Radius installation

rynowak:
Fixes: #2952

This change removes Dapr from Radius' installation, and makes it the user's responsibility. Previously in #4852 we added checks to all of the Dapr links that will prevent creation of those resources without Dapr installed. So we're now ready to remove it from our installation code.


ytimocin (commented):
lgtm

youngbupark (approved):

ytimocin (approved):

jkotalik (approved):

PR #4861, opened 2022-12-14
Introduce 'required features' on functional test

rynowak:
This change adds the ability for a functional test to declare "required features" like a CSI driver or Dapr. Then the test will detect the installation of those features and skip itself if appropriate.

I've applied to the existing tests that have a dependency on either CSI drivers or Dapr.


youngbupark (approved):
👍

PR #4873, opened 2022-12-20
Use generic GET/LIST controllers for link RP

rynowak:
This change replaces the handwritten controllers from linkrp with the generic GET and LIST controllers that are already in use by the corerp.

The main changes required to replace the functionality with the generic implementation are superficical changes needed to make interfaces conform. See #4872 for an example of these changes with commentary about what's being done.

The most substantial change in this PR is to simplify the extender resource datamodel and fix bugs in tests. The extender had separate data model types for requests and responses, which doesn't follow the pattern we've established. This change removes the 'response' datamodel and updates the converter for the 'request' datamodel to omit secrets. This is the established convention we use for other link resources.

Additionally there were bugs in the extender conversion tests, either with tests asserting the wrong things or not covering the secret functionality of the data model. These tests have been improved.


youngbupark (approved):
lgtm

snehabandla (approved):

PR #4880, opened 2022-12-23
Add in-memory integration tests for UCP

rynowak:
This change adds in-memory integration tests for the UCP as a precursor to adopting the generic controllers. These tests use a real data-store (ETCD) rather than mocking a database.

Most of the testing we've done for
UCP uses the controller interface, and since we're deleting the handwritten UCP controllers these tests would be invalid. UCP's use case are unique enough that they we want some overall coverage in addition to what our functional tests provide. Just going through the motions of writing the tests uncovered several bugs.

Most of the work in this PR was put into tuning our ETCD infrastructre to support use in integration tests. When I started writing these tests I really quickly ran into resource limitations (likely ulimit) that caused the tests to hang or run extremely slowly. Go leak detector to the rescue!

There were two major changes needed:

- We need to create fewer ETCD clients and ensure they are cleaned up as part of each test.
- We were calling the wrong 'stop' function and as a result ETCD was not fully shutting down.

This also was my opportunity to update `hosting.AsyncValue` to use generics.


ytimocin (approved):
left some comments

PR #4905, opened 2023-01-04
Set location for Link RP service

rynowak:
Link RP has a bug where URLs for async operations are wrong because they are missing the location and output a blank URL segment. We've seen this bug in the past and it's NASTY to try and diagnose if you don't already know the cause.

The root cause is that we forgot to set a field. Lame.

The only async operation in Link RP right now is the DELETE for Mongo. I'm not 100% certain why we haven't found the bug before now. The deployment engine definitely chokes on it so when I changed a PUT operation to async it explodes. This kind of bug won't be caught by postman testing because we don't usually do polling when we do that.

### DIFF
--- a/pkg/armrpc/servicecontext/middleware.go
+++ b/pkg/armrpc/servicecontext/middleware.go
@@ -15,6 +15,16 @@ import (
 
 // ARMRequestCtx is the middleware to inject ARMRequestContext to the http request.
 func ARMRequestCtx(pathBase, location string) func(h http.Handler) http.Handler {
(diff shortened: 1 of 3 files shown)


youngbupark (approved):

ytimocin (approved):

kachawla (approved):
It is strange that broken delete went unnoticed. I thought we were exercising delete path in e2e tests - will create a task for this.

mishrapratikshya (approved):

shalabhms (approved):

PR #4930, opened 2023-01-10
Parse Referer header to preserve the casing of URL Path.

youngbupark:
# Description

[ARM RPC spec](https://github.com/Azure/azure-resource-manager-rpc/blob/022630d456f8a92999040b0193929807815e7246/v1.0/common-api-details.md?plain=1#L19) specifies `Referer` header to keep the uri that Client uses. To preserve the original casing, we will use this `Referer` header value. ARMRequestContext parses resource id from this header. This change is also related to #3068 

## Issue reference

#3582 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


ytimocin (approved):

rynowak (commented):
Are you planning to make the change to add Referrer in UCP or is someone else signed up for that?

Edit: we will have to update clients too. UCP actually doesn't know the original hostname (in Kubernetes) due to https://github.com/kubernetes/kubernetes/issues/107435

sk593 (approved):

PR #4939, opened 2023-01-11
Migrating Deployments Client and DeploymentOperations Client

ytimocin:
# Description
Migrating Deployments Client and DeploymentOperations Client.

## Issue reference
Fixes: #issue_number

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Adds necessary unit tests for change
* [ ] Adds necessary E2E tests for change
* [x] Unit tests passing
* [x] Extended the documentation / Created issue for it

PR #4952, opened 2023-01-12
Add SDK package as home for client code

rynowak:
This change introduces a new top-level package 'pkg/sdk' that contains client code for interacting with the UCP & Radius APIs. The reason for the new package is that we need to start calling into UCP from inside our server APIs. We have code today that accesses the database instead of calling the API (eg: link RP reading the database of core rp). We also need to call into the deployment engine from the link rp as part of the recipes work.

The problem is that all of our client code for interacting with the APIs today is part of pkg/cli. Rather than introducing cli references to our RP code, I chose to fix the layering. This is a refactor I've been wanting to do for a while now.

I also used this opportunity to clean up the proliferation of different options and APIs that have sprung up organically. Lots of our existing code futzes around with baseURLs and roundtrippers and other low level details due to the lack of abstraction.

Now there's one abstraction (sdk.Connection), two functions for creating a connection (one for testing and one for Kubernetes), and one way (per generation of SDK) to hook up a generated client to a connection.

This change introduces the new functionality, adds tests where they were missing before and refactors the client code to use the new types. My next step will be to introduce sdk.Connection on the server side and wire it up based on our server side configuration.

### DIFF
--- a/cmd/rad/cmd/debuglogs.go
+++ b/cmd/rad/cmd/debuglogs.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/project-radius/radius/pkg/cli"
 	"github.com/project-radius/radius/pkg/cli/kubernetes"
-	"github.com/project-radius/radius/pkg/cli/workspaces"
 	k8slabels "github.com/project-radius/radius/pkg/kubernetes"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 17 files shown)


ytimocin (commented):
Looks good so far. When this PR is merged, I will rebase [this PR](https://github.com/project-radius/radius/pull/4939) and have to update some of these files with the new SDK. We can discuss further offline.

bjoginapally (approved):

PR #4991, opened 2023-01-20
Add plumbing for UCP connection in RPs

rynowak:
Fixes: AzDo Task 5779 5782

This change adds the settings for creating an sdk.Connection to the host configuration used by our backend services.

This change also fixes the ClusterRole which was using the wrong api group (api.radius.dev was used before we created UCP).

I found a group of constants that should been removed during the previous SDK pr, which was the source of the confusion about the api group. None of these constants are used because all of that code was deleted in the SDK PR.


youngbupark (approved):
one small comment. other than this lgtm.

PR #5050, opened 2023-01-30
Enable value-based recipes

rynowak:
This change enables value-based recipes. We originally set the goal of enabling value-based recipes for Redis, but due the nature of the deployment engine, they are now enabled for any of our links that support recipes.

The downside is that it's pretty complicated (see new comments in deployment processor). Since our existing recipe support is based on resource binding, all of the concepts get mixed together. There's no easy to way to one or the other.

There's also some significant new infrastructure needed to power this in the form of:

- A generic Kubernetes handler that supports any Kubernetes type
- A generic AWS handler (same store)
- Functionality for turning a resource ID into a resource identity

There are many impedance mismatches between the "world of UCP" and the "world of output resource" that we should close down in follow ups.

This PR also incorporates and supersedes radius-project/radius#5031 and radius-project/radius#5044.

### DIFF
--- a/pkg/armrpc/asyncoperation/worker/worker.go
+++ b/pkg/armrpc/asyncoperation/worker/worker.go
@@ -234,6 +234,7 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 			if err != nil {
 				armErr := extractError(err)
 				result.SetFailed(armErr, false)
+				logger.Error(err, "Operation Failed")
 			}
 			w.completeOperation(ctx, message, result, asyncCtrl.StorageClient())
 		}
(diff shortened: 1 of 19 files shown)


kachawla (commented):
Still going through rest of the PR, minor nits and a question on recipe_handler/helper so far.

kachawla (commented):
Done with deployment processor, moving on to renderers now

kachawla (commented):
We will need to add recipedata update in [redis datamodel](https://github.com/project-radius/radius/blob/main/pkg/linkrp/datamodel/rediscache.go#L30) similar to what we do for other supported link types https://github.com/project-radius/radius/blob/main/pkg/linkrp/datamodel/mongodatabase.go#L55
`r.RecipeData = do.RecipeData`

mishrapratikshya (approved):

kachawla (approved):

PR #5426, opened 2023-04-09
Add reporting of unit tests results

rynowak:
This change adds reporting of unit tests to our PRs using an external github action. The action will create a test run summary as well as annoations (errors and warnings) to failing unit tests.

See an example of this in action here:

  https://github.com/project-radius/radius/pull/5422

There are three parts to this PR:

- Producing the jUnit XML file format (gotestsum)
- Working around bug in file/line number reporting
- Processing and uploading results

The main that needs commentary is the integration with gotestsum. This is a tool I've been using for a while now because it provides a better local experience for understanding test failures. This PR integrates it as an OPTIONAL dependency. Our various `make test` commands know how to use gotestsum if its present and fall back to `go test` otherwise. gotestsum is compatible (mostly) with `go test` so the only change in behavior is the reporting.

However, there's a bug where the test report is missing file/line number. The GH action I'm using doesn't understand how to parse the output of `go test` and retrieve this info, but the file format we're using (jUnit XML) supports it. I wrote a basic script to address this. Potentially this could be fixed inside gotestsum.

I looked at a few different github actions and this one seemed like it has the best features and is actively maintained. In general they all work with the jUnit format so we're not getting locked in.

### DIFF
--- a/.github/actions/process-test-results/action.yaml
+++ b/.github/actions/process-test-results/action.yaml
@@ -1,0 +1,68 @@
+name: "Process Test Results"
+description: |
+  Processes a batch of test results including creating a github artifact and annotating the Pull Request (if applicable).
+
+  This requires using 'gotestum' to run tests, which our makefile can do automatically. Install 'gotestsum' using:
+
+    go install gotest.tools/gotestsum@v1.10.0
+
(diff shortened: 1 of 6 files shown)


youngbupark (approved):
💯

PR #5506, opened 2023-05-01
Add patch namespace and split test utils

rynowak:
## Description

This change adds the server-side "patch namespace" functionality to a shared utility package and adds tests for it.

Adding tests triggered a need to refactor our `testutil` package. The problem is that `testutil` references many different abstractions across Radius (like store and secrets) which causes a package reference cycle. Our shared Kubernetes utilities are used by the store, and secret manager.

The fix is to split the Kubernetes testutils functionality into its own package. 'utils' packages need to be really granular in Go or else you end up with cycles like this frequently.

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing

### DIFF
--- a/pkg/cli/kubernetes/kubernetes_test.go
+++ b/pkg/cli/kubernetes/kubernetes_test.go
@@ -11,7 +11,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/project-radius/radius/test/testutil"
+	"github.com/project-radius/radius/test/k8sutil"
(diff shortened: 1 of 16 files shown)


youngbupark (approved):

kachawla (approved):

ytimocin (approved):
There are some warnings -- a rebase might be necessary.

ytimocin (approved):

PR #5516, opened 2023-05-04
added github action to autocut release every 4 wednesdays at midnight

sylvainsf:
# Description

This creates an action that executes every 4 Wednesdays at midnight (late Tuesday night) to autocut the next release.

## Issue reference

https://github.com/project-radius/radius/issues/5513


willdavsmith (approved):
nice! how did you test this?

PR #5521, opened 2023-05-04
Handling dev recipes conflict with user defined recipes

vishwahiremat:
# Description

Changing the behavior of recipe registration when use provides the recipe name that is same as the one defined in dev recipe for the given resource type.

Instead of throwing an error (what we do today), ex:
"recipe name(s) reserved for devRecipes for: recipe with name mongo-azure (linkType Applications.Link/mongoDatabases and templatePath radiusdev.azurecr.io/mongo:1.0)"

we overwrite the recipe with user provided recipe information.

FYI: this behavior would change with the introduction of recipe packs, where dev recipe packs and user defined packs can have recipe with same name for the same type.

## Issue reference

Fixes: #issue_number

## Checklist

* [x] Code compiles correctly
* [x] Unit tests passing


kachawla (commented):
Does this require any doc updates? I don't think this was explicitly called out in the doc, but just checking. @jasonviviano

kachawla (approved):

PR #5579, opened 2023-05-19
Fix for rad deploy should wait for deployment to complete

vinayada1:
# Description

With the introduction of recipes, some deployments are now initiated by LinkRP. The issue was that CoreRP was waiting on a deployment to complete which was signaled by the completion of a linkRP deployment.

https://github.com/project-radius/radius/issues/4022

Fixes: #4022 

## Checklist

* [ ] Code compiles correctly
* [ ] Unit tests passing


rynowak (changes requested):
This looks like the right solution - can you add a test for the case you fixed?

rynowak (approved):

PR #5957, opened 2023-07-28
Clean up sample resource groups and container images in 3 days

youngbupark:
# Description

* Delete dev container images in 2 days (org: 1 week)
* I found the sample repo has the workflow to purge the resource groups. I think it is eaiser to maintain the single workflow to clean up resources in the single place. 

https://github.com/project-radius/samples/blob/v0.23/.github/workflows/purge-test-resources.yaml

## Type of change

### Summary
:recycle::wrench::test_tube:

Updated the purge-test-resources workflow to use the new naming convention for samplestest resource groups. This helps to avoid deleting unintended resources in Azure.

> _`samplestest` names_
> _filter out old resources_
> _autumn cleanup time_

### Walkthrough
*  Update the condition for filtering out resource groups that should not be deleted in the workflow for purging test resources ([link](https://github.com/project-radius/radius/pull/5957/files?diff=unified&w=0#diff-9270fafc31ef4b50d35d1295e9f99f96013853c295c29c92c387492f62cd570dL63-R63))

### DIFF
--- a/.github/workflows/purge-test-resources.yaml
+++ b/.github/workflows/purge-test-resources.yaml
@@ -33,7 +33,7 @@ jobs:
         uses: snok/container-retention-policy@v2
         with:
           image-names: dev/*
-          cut-off: A week ago UTC
+          cut-off: 2 days ago UTC
           account-type: org
           org-name: project-radius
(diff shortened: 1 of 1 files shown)


ytimocin (approved):
Should we remove this from the samples repo?

Issue #6431, opened 2023-10-05
Kubernetes Integration - Helm Support


lakshmimsft:
## Overview of feature request

Add a Kubernetes-centric configuration experience to help enable trivial adoption of Radius for applications that use the Kubernetes application model (Deployment, Service, etc.) for configuration and use the common Kubernetes toolchains for application deployment (Helm, Flux). 

Ref for further Detail: [Design Document](https://github.com/radius-project/design-notes/blob/3d98c6be4c47a4f6a12bc834a22ceae670ab3c92/architecture/2023-10-kubernetes-integration.md)

## Acceptance criteria
1. Use can edit a Helm chart, adding each annotation to each Kubernetes Deployment indicating that Radius should be enabled (radapp.io/enabled: 'true'). Upon merging updates,  the metadata is updated in deployments, which automatically triggers the create of a Radius application and container resources that belong to that application.
1. Users will be able to create recipes using a new CRD type (Recipe) that can be used for any type of resource that supports recipes

## Additional context

[AB#9672](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/9672)

rynowak:
This is done :) - there's a tail of small improvements to make but the features exist and have documentation.

PR #6571, opened 2023-10-25
Recipe/Deployment- validation admission webhook

lakshmimsft:
# Description

Adding a validation webhook for Recipe + placeholder mutating webhook for built in Deployment as part of this effort: https://github.com/radius-project/radius/issues/6431

- This pull request adds or changes features of Radius and has an approved issue #6431 

Fixes: Part of #6431

### Summary
:sparkles::white_check_mark::wrench:

This pull request adds a validating webhook for the Recipe custom resource, which checks the resource type of the Recipe and rejects invalid types. It also adds the necessary code and tests to enable the webhook in both the test and the production environments. The main files affected are `recipe_webhook.go`, `recipe_webhook_test.go`, `service.go`, and `main_test.go`.

> _We summon the power of the webhook_
> _To validate the recipe of doom_
> _We test the server with `WaitForConnection`_
> _We unleash the beast in the test room_

### Walkthrough
*  Implement the validating webhook for the Recipe resource ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-c9129d1a788c077342f4259f45f25de80425f809034bf8a03a941c7e10c41faaR1-R81), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R25-R26), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R39-R47), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R63), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R75), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R94-R145), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b31c03e05358c6faab6e77c96909e14e14eeb0dc09412ede3329625adc197fa9R109-R111))
  * Add the license header and the package declaration for `recipe_webhook.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-c9129d1a788c077342f4259f45f25de80425f809034bf8a03a941c7e10c41faaR1-R81))
  * Add the imports for the admissionv1 and metav1 packages in `main_test.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R25-R26))
  * Define a WebhookInstallOptions struct to hold the webhook server options in `main_test.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R39-R47))
  * Create the validating webhook configuration for the Recipe resource in the test environment in `main_test.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R63), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R94-R145))
  * Assign the webhook install options to a global variable in `main_test.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-77d671434b376fb7c1f0f24a606661bb824d6259e8b3c27fa770977f0a63bbd9R75))
  * Register the validating webhook with the controller manager in `service.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b31c03e05358c6faab6e77c96909e14e14eeb0dc09412ede3329625adc197fa9R109-R111))
* Write the unit tests for the validating webhook for the Recipe resource ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-985c5db6647f07ba029dd36ad47fda7eb315e21c6691f16eafd3b9f3f556f8b7R1-R167))
  * Add the license header and the package declaration for `recipe_webhook_test.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-985c5db6647f07ba029dd36ad47fda7eb315e21c6691f16eafd3b9f3f556f8b7R1-R167))
  * Write two test cases: one for an invalid resource type and one for a valid resource type ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-985c5db6647f07ba029dd36ad47fda7eb315e21c6691f16eafd3b9f3f556f8b7R1-R167))
* Add a utility function to wait for a secure connection to the webhook server in the test environment ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b2622e855cc3c8af44551c69c158d0400b6b2df3e3ebaefef18f7db96c6b66bbL21-R25), [link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b2622e855cc3c8af44551c69c158d0400b6b2df3e3ebaefef18f7db96c6b66bbR286-R306))
  * Add the imports for the crypto/tls, net, and time packages in `util.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b2622e855cc3c8af44551c69c158d0400b6b2df3e3ebaefef18f7db96c6b66bbL21-R25))
  * Implement the WaitForConnection function in `util.go` ([link](https://github.com/radius-project/radius/pull/6571/files?diff=unified&w=0#diff-b2622e855cc3c8af44551c69c158d0400b6b2df3e3ebaefef18f7db96c6b66bbR286-R306))


pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// log is for logging in this package.
+var recipejoblog = logf.Log.WithName("recipe-resource")
```
rynowak, line 34: Is it possible for us to use 'our' logger here? I would feel better if we don't start adopting additional logging packages.

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/validate-radapp-io-v1alpha3-recipe,mutating=false,failurePolicy=fail,sideEffects=None,groups=radapp.io,resources=recipe,verbs=create;update,versions=v1alpha3,name=recipe-webhook.radapp.io,sideEffects=None,admissionReviewVersions=v1
+var _ webhook.Validator = &Recipe{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Recipe) ValidateCreate() (admission.Warnings, error) {
```
rynowak, line 46: Would it be possible for us to use https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator_custom.go#L32 instead? 

This will give us a context, which we can use to get the logger.

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil, nil
+}
+
+// validateRecipeType validates Resource Type to be created.
+func (r *Recipe) validateRecipeType() error {
+	var errList field.ErrorList
+	flPath := field.NewPath("spec").Child("type")
+
+	if !portableresources.IsValidPortableResourceType(r.Spec.Type) {
+		errList = append(errList, field.Invalid(flPath, r.Spec.Type, fmt.Sprintf("invalid resource type %q in recipe", r.Spec.Type)))
```
rynowak, line 72: It would be good if we can list the valid types as well in the error message. Then people can see what to fix.

pkg/controller/reconciler/main_test.go
```diff
@@ -34,6 +36,15 @@ var config *rest.Config
 // scheme holds a reference to the scheme for the test environment.
 var scheme *runtime.Scheme
 
+// webhookInstallOptions holds a reference to the webhook install options for the test environment.
+var webhookInstallOptions *envtest.WebhookInstallOptions
```
rynowak, line 40: It looks like this is only used inside `initializeWebhookInEnvironment`. Do you think we could get away without a global variable? Global variables are scary 👻

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
lakshmimsft: updated

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
lakshmimsft: updated. pls take a look

pkg/controller/reconciler/main_test.go
lakshmimsft: The variable was needed in recipe_webhook_test.go https://github.com/radius-project/radius/blob/c117c5baa63df34a7b36491e69534d9d808dd467/pkg/controller/reconciler/recipe_webhook_test.go#L225 I tried a couple things but perhaps the only way to get rid of it completely is to have the mgr and webhook up and running in for all tests including reconciler tests which I don't think we want.  I limited the number of fields and updated naming of the variable to be obvious and overall am following similar pattern as the reconciler tests. Let me know if okay.

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
lakshmimsft: thank you for the link

pkg/controller/reconciler/main_test.go
rynowak: If we can't avoid it, then we can't avoid it 😢

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,266 @@
+/*
+Copyright 2023 The KEDA Authors
```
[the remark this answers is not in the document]
lakshmimsft: oops! updated.

.vscode/tasks.json
```diff
@@ -63,6 +63,20 @@
 				"showReuseMessage": false,
 				"clear": true
 			}
+		},
+		{
```
lakshmimsft, line 67: Added this to generate certs locally for local run.

.vscode/tasks.json
```diff
@@ -63,6 +63,20 @@
...
+			"label": "generate-certs",
+            "type": "shell",
+            "command": "mkdir -p ~/tls/cert && openssl req -x509 -newkey rsa:4096 -keyout ~/tls/cert/tls.key -out ~/tls/cert/tls.crt -days 365 -nodes -subj \"/CN=localhost\" && export TLSCertDir=~/tls/cert",
+            "problemMatcher": [],
+			"windows": {
+				"command": "if not exist %USERPROFILE%\\tls\\cert mkdir %USERPROFILE%\\tls\\cert && openssl req -x509 -newkey rsa:4096 -keyout %USERPROFILE%\\tls\\cert\\tls.key -out %USERPROFILE%\\tls\\cert\\tls.crt -days 365 -nodes -subj \"/CN=localhost\" && set TLSCertDir %USERPROFILE%\\tls\\cert"
+			}
```
rynowak, line 74: Question about this.... When you fire this up in the debugger, does the webhook actually get called by Kubernetes? How does that work?

rynowak: Usually with Kubernetes webhooks, the only way to **really** test them is to deploy the webhook to the cluster. So what I see most people do is **turn off** the webhook in debugging scenarios.

deploy/Chart/templates/controller/configmaps.yaml
```diff
@@ -41,13 +41,17 @@ data:
...
+      port: 3000
+
     logging:
       level: "info"
       json: true
 
     {{- if and .Values.global.zipkin .Values.global.zipkin.url }}
     tracerProvider:
-      serviceName: "ucp"
+      serviceName: "controller"
```
rynowak, line 53: 👍

deploy/Chart/templates/controller/secret.yaml
```diff
@@ -1,24 +0,0 @@
-{{- $existingSecret := lookup "v1" "Secret" .Release.Namespace "controller-cert"}}
```
rynowak, line 1: Is this being removed because it was added to `webhook.yaml`?

cmd/controller/controller-self-hosted.yaml
```diff
@@ -12,6 +12,8 @@ metricsProvider:
 server:
   host: "0.0.0.0"
   port: 8083
+workerServer:
+  port: 3000
```
rynowak, line 16: I think it would be good if we could list the different ports we're exposing and what they are used for. It's not clear to me from the names what the differences are between `server` and `workerServer`.

I think `metricsProvider` is pretty self-explanatory :)

rynowak: A comment would help a lot, a better name might help even more.

pkg/controller/api/radapp.io/v1alpha3/deployment_webhook.go
```diff
@@ -0,0 +1,56 @@
...
+	logger := ucplog.FromContextOrDiscard(ctx)
+	_, ok := obj.(*appsv1.Deployment)
+	if !ok {
+		return fmt.Errorf("expected a built in Deployment but got a %T", obj)
+	}
+
+	// TODO: Placeholder implementation to be updated.
+	logger.Info("Update Deployment")
+
+	return nil
```
rynowak, line 55: I'm assuming this is all going to change when we implement this :)

pkg/controller/api/radapp.io/v1alpha3/deployment_webhook.go
```diff
@@ -0,0 +1,56 @@
...
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(&appsv1.Deployment{}).
+		WithDefaulter(d).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-radapp-io-v1alpha3-builtindeployment,mutating=true,failurePolicy=fail,sideEffects=None,groups=radapp.io,resources=deployments,verbs=create;update,versions=v1alpha3,name=deployment-webhook.radapp.io,sideEffects=None,admissionReviewVersions=v1
+
+// BuiltInDeployment is a type that implements the mutating webhook function for the type Deployment.
+type BuiltInDeployment struct{}
```
rynowak, line 42: What I was expecting here is for us to implement a custom webhook that handles the built-in `apps/v1 Deployment` Kubernetes type. 

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

I'm good with however you want to sequence the changes. If this is going to be addressed in the next step, that's fine.

pkg/controller/api/radapp.io/v1alpha3/recipe_webhook.go
```diff
@@ -0,0 +1,94 @@
...
+	logger := ucplog.FromContextOrDiscard(ctx)
+	validResourceTypes := strings.Join(portableresources.GetValidPortableResourceTypes(), ", ")
+
+	logger.Info("Validating Recipe Type %s in Recipe %s", r.Spec.Type, r.Name)
+	if !portableresources.IsValidPortableResourceType(r.Spec.Type) {
+		return nil, fmt.Errorf("invalid resource type %s in recipe %s. allowed values are: %s", r.Spec.Type, r.Name, validResourceTypes)
+	}
+
+	return nil, nil
+}
```
rynowak, line 94: This is looking pretty good 👍

deploy/Chart/templates/controller/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
           name: metrics
           protocol: TCP
         {{- end }}
+        - containerPort: 3000
+          name: healthz
```
rynowak, line 52: This is pretty clear btw. By calling the port `healthz` I immediately understand what it's for. Do you think we could the same in our configuration files?

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -0,0 +1,86 @@
...
+  rules:
+  - apiGroups:
+    - radapp.io
+    apiVersions:
+    - v1alpha3
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - deployments
```
rynowak, line 85: I think a wire got crossed here, I was expecting this to handle the built-in Kubernetes `Deployment` resource type: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

We don't need to define our own CRD, we want to provide a webhook for a type that's already defined by Kubernetes.

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,250 @@
...
+		err := client.Create(ctx, &corev1.Namespace{ObjectMeta: ctrl.ObjectMeta{Name: namespace.Name}})
+		require.NoError(t, err)
+
+		// We do not check for error here since error may be captured after this initial call is completed.
+		_ = client.Create(ctx, recipe)
+
+		current := &radappiov1alpha3.Recipe{}
+		require.Eventually(t, func() bool {
+			if err := client.Get(ctx, namespace, current); err != nil {
+				require.EqualError(t, err, expectedError)
```
rynowak, line 67: Can you use `apierrors.IsNotFound` instead? Example: https://github.com/radius-project/radius/blob/9bcf33fa358383a1c4c54b41753bd79265f41aca/pkg/daprrp/datamodel/daprdetector.go#L42

Error strings can change all of the time, because the intended audience is users. Checking for types is much more resiliant.

rynowak: Note: you'll want to add an import alias like:

```
import apierrors "k8s.io/apimachinery/pkg/api/errors"
```

pkg/controller/service.go
```diff
@@ -102,17 +106,22 @@ func (s *Service) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to setup %s controller: %w", "Deployment", err)
 	}
 
-	if s.TLSCertDir == "" {
-		logger.Info("Webhooks will be skipped. TLS certificates not present.")
```
rynowak, line 106: This was intentional by the way. See my comments elsewhere about debugging. It's pretty handy to be able to run the controller without webhooks.

.vscode/tasks.json
lakshmimsft: yes, that makes sense, reverted changes. now passing in "" for cert-dir when debugging.

cmd/controller/controller-self-hosted.yaml
lakshmimsft: Added comment in config file to clarify. The reason for using workerServer was it's description here: https://github.com/radius-project/radius/blob/0077cb8ca4a39ea8bca0d182bfe3f19238a05626/pkg/armrpc/hostoptions/providerconfig.go#L66 , so I thought I could use it.
Let me know if this is okay. Else, I will extend ProviderConfig struct and add, say, a HealthProviderOptions struct and use it instead of WorkerServerOptions.

pkg/controller/api/radapp.io/v1alpha3/deployment_webhook.go
lakshmimsft: If we see ln 34: https://github.com/radius-project/radius/blob/0077cb8ca4a39ea8bca0d182bfe3f19238a05626/pkg/controller/api/radapp.io/v1alpha3/deployment_webhook.go#L34 , this is where we specify that the webhook is for the built-in deployment type. I need to specify a different type on ln 42 as I cannot define new method (Default) on the built-in type

deploy/Chart/templates/controller/webhook.yaml
lakshmimsft: oops! i updated the apiGroups and apiVersions to ensure it is built in deployment type.

pkg/controller/reconciler/recipe_webhook_test.go
lakshmimsft: nice! thank u! updated.

pkg/controller/service.go
lakshmimsft: yes, this makes sense, updated to include it in.

cmd/controller/main.go
```diff
@@ -42,8 +42,8 @@ func main() {
 	config := fmt.Sprintf("controller-%s.yaml", hostoptions.Environment())
 	pflag.StringVar(&config, "config-file", config, "The service configuration file.")
 
-	tlsCertDir := ""
-	pflag.StringVar(&tlsCertDir, "cert-dir", "", "The directory containing the TLS certificates.")
+	tlsCertDir := "/var/tls/cert"
+	pflag.StringVar(&tlsCertDir, "cert-dir", tlsCertDir, "The directory containing the TLS certificates.")
```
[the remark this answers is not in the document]
lakshmimsft: This is a pattern seen used quite often where we define a variable with default value , eg tlsCertDir and then check to see if user has passed in corresponding argument which will override the default value we have set.

.vscode/launch.json
```diff
@@ -40,12 +40,14 @@
             "type": "go",
             "request": "launch",
             "mode": "auto",
-            "preLaunchTask": "Build Radius (all)",
+            "preLaunchTask": "prelaunch",
```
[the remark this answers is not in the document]
lakshmimsft: it was but now reverted those changes

cmd/controller/controller-self-hosted.yaml
rynowak: I think this is fine based on the comment. Thanks.

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -0,0 +1,86 @@
...
+      radapp.io/enabled: "true"
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    caBundle: {{ b64enc $ca.Cert }}
+    service:
+      name: controller
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-radapp-io-v1alpha3-deploymentstate
```
rynowak, line 72: Is this the right path? Have you seen this webhook actually get called?

rynowak (commented):
Almost there 👍

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -0,0 +1,86 @@
...
+  {{ end }}
+
+  {{ if $existingSecret }}tls.key: {{ index $existingSecret.data "tls.key" }}
+  {{ else }}tls.key: {{ b64enc $cert.Key }}
+  {{ end }}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
```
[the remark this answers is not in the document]
lakshmimsft: testing this out..

pkg/controller/api/radapp.io/v1alpha3/zz_generated.deepcopy.go
```diff
@@ -22,9 +22,24 @@ limitations under the License.
...
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new DeploymentState.
+func (in *DeploymentState) DeepCopy() *DeploymentState {
+	if in == nil {
+		return nil
+	}
+	out := new(DeploymentState)
+	in.DeepCopyInto(out)
+	return out
+}
+
```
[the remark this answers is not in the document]
lakshmimsft: updated

deploy/Chart/templates/controller/webhook.yaml
lakshmimsft: updated path is: /mutate-apps-v1-deployment

deploy/Chart/templates/controller/webhook.yaml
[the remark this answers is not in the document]
lakshmimsft: yes, not needed. removed

pkg/controller/api/radapp.io/v1alpha3/deployment_webhook.go
lakshmimsft: removing logging statement since it will currently run for every deployment

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,248 @@
...
+	t.Run("test recipe for invalid type", func(t *testing.T) {
+		recipeName := "test-recipe-invalidtype"
+		namespace := types.NamespacedName{Namespace: defaultNamespace, Name: recipeName}
+		recipe := makeRecipe(namespace, invalidResourceType)
+
+		err := client.Create(ctx, &corev1.Namespace{ObjectMeta: ctrl.ObjectMeta{Name: namespace.Name}})
+		require.NoError(t, err)
+
+		// We do not check for error here since error may be captured after this initial call is completed.
+		_ = client.Create(ctx, recipe)
```
rynowak, line 62: Can you explain this? If the type is invalid, I'd expect this to return an error. Isn't this the thing we're trying to check?

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Environment is created.
+	createEnvironment(radius, "default")
+
+	t.Run("test recipe for invalid type", func(t *testing.T) {
+		recipeName := "test-recipe-invalidtype"
+		namespace := types.NamespacedName{Namespace: defaultNamespace, Name: recipeName}
+		recipe := makeRecipe(namespace, invalidResourceType)
+
+		err := client.Create(ctx, &corev1.Namespace{ObjectMeta: ctrl.ObjectMeta{Name: namespace.Name}})
```
rynowak, line 58: Should we also test an update case?

lakshmimsft: Added on ln 99. Pls review.

pkg/controller/reconciler/recipe_webhook_test.go
lakshmimsft: Thank you! Updated code to catch the error. 
Also pls review new function ln 309: [updateWebhookFailurePolicy()](https://github.com/radius-project/radius/blob/8519fd58a053fb3970287efd01abcfc5b07f63d5/pkg/controller/reconciler/recipe_webhook_test.go#L302) . I'm using it to set webhook FailurePolicy to Fail when running Webhook tests and then setting it back to Ignore after tests are run, so it does not interfere with other reconciler tests.

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,330 @@
...
+			if err := client.Get(ctx, namespace, recipe); err != nil {
+				return err
+			}
+
+			// Update the recipe to have an invalid type
+			recipe.Spec.Type = invalidResourceType
+			return client.Update(ctx, recipe)
+		})
+		// The webhook should reject the update and return a Forbidden error
+		require.True(t, apierrors.IsForbidden(err))
```
rynowak, line 131: per dicussion, this isn't the right return code. We should return an `Invalid` error from the webhook so the right return code shows up here.

rynowak (changes requested):

pkg/controller/reconciler/recipe_webhook_test.go
lakshmimsft: updated pls take a look.

pkg/controller/reconciler/recipe_webhook_test.go
```diff
@@ -0,0 +1,330 @@
...
+		status = waitForRecipeStateDeleting(t, client, namespace, nil)
+		radius.CompleteOperation(status.Operation.ResumeToken, nil)
+
+		// Now deleting of the deployment object can complete.
+		waitForRecipeDeleted(t, client, namespace)
+	})
+
+	// Update the FailurePolicy of the webhook to Ignore after running webhook tests, so that the webhook will not interfere with other tests.
+	failurePolicy := admissionv1.Ignore
+	updateWebhookFailurePolicy(t, webhookConfigName, &failurePolicy)
```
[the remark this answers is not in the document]
lakshmimsft: discussed offline. Added comments for current approach.

rynowak (approved):
:shipit:

PR #6601, opened 2023-10-29
Update README.md - fix grammatrical errors in README.

ayushrakesh:
I have noticed some grammatical mistakes in **README** inside section "**Overview**"

"**in the cloud increasingly difficult**"  should be "**in the cloud is increasingly difficult**"

"**usually focused on compute, to**"  should be  "**usually focused on computing, to**"

Screenshot-

Screenshot (121)


ayushrakesh:
Please review my PR.

lakshmimsft:
@ayushrakesh Thank you for your contribution! I agree with the first change suggested but I'm not sure about the second. The distinction is subtle and intent here is to refer to 'compute' as one part of application infrastructure. I'll request @AaronCrawfis / @Reshrahim for input.

Reshrahim:
> @ayushrakesh Thank you for your contribution! I agree with the first change suggested but I'm not sure about the second. The distinction is subtle and intent here is to refer to 'compute' as one part of application infrastructure. I'll request @AaronCrawfis / @Reshrahim for input.

Thank you for your contribution. For the second change, we would prefer to retain the word "compute" as it represents the abstractions built on the compute layer or on the compute resources. Can you resubmit the PR with a fix for the first typo?

PR #6793, opened 2023-11-16
[wip] Add logs for kubernetes events

sk593:
# Description

Adding logs to track pod failures during schedule func test runs

## Type of change

Fixes: #issue_number

### Summary
🐽🛠

Enhance functional-test workflow to capture pod logs and events on failure. Add `id` to `publish-tf-recipes` step and two new steps to `.github/workflows/functional-test.yaml`.

> _When the recipes fail to publish_
> _We unleash the wrath of `pod logs`_
> _We capture the events of doom_
> _And upload them to the cloud of gloom_

### Walkthrough
* Add an `id` to the recipe publishing step to enable conditional execution of subsequent steps ([link](https://github.com/radius-project/radius/pull/6793/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcR483))

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -16,12 +16,14 @@
 
 name: Functional tests
 on:
-  schedule:
-    # Run every 4 hours on weekdays.
-    - cron: "30 0,4,8,12,16,20 * * 1-5"
-    # Run every 12 hours on weekends.
-    - cron: "30 0,12 * * 0,6"
-  # Dispatch on external events
+  push:
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - main
+      - features/*
+      - release/*
   repository_dispatch:
     types: [de-functional-test]
   workflow_run:
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -16,14 +16,10 @@
 
 name: Functional tests
 on:
-  schedule:
-    # Run every 4 hours on weekdays.
-    - cron: "30 0,4,8,12,16,20 * * 1-5"
-    # Run every 12 hours on weekends.
-    - cron: "30 0,12 * * 0,6"
   # Dispatch on external events
   repository_dispatch:
     types: [de-functional-test]
+  workflow_dispatch:
 
 env:
   # Go version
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -559,6 +560,30 @@ jobs:
             --subscription ${{ secrets.INTEGRATION_TEST_SUBSCRIPTION_ID }} \
             --name ${{ env.AZURE_TEST_RESOURCE_GROUP }} \
             --yes --verbose
+      # TODO add once tested: if: failure() && steps.publish-tf-recipes.outcome == 'failure'
```
rynowak, line 563: Would it be bad if we just always included these logs?

.github/workflows/functional-test.yaml
```diff
@@ -559,6 +560,30 @@ jobs:
...
+        id: get-pod-logs
+        run: |
+          # Create pod-logs directory
+          mkdir -p recipes/pod-logs
+          # Get pod logs and save to file
+          namespace="radius-test-tf-module-server"
+          label="app.kubernetes.io/name=tf-module-server"
+          pod_names=($(kubectl get pods -l $label -n $namespace -o jsonpath='{.items[*].metadata.name}'))
+          for pod_name in "${pod_names[@]}"; do
+            kubectl logs $pod_name -n $namespace > recipes/pod-logs/${pod_name}.txt
```
ytimocin, line 574: We can add a log here like `echo "Pod ${pod_name} logs saved to recipes/pod-logs/"`

.github/workflows/functional-test.yaml
```diff
@@ -559,6 +560,30 @@ jobs:
...
+          done
+          echo "Pod logs saved to recipes/pod-logs/"
+          # Get kubernetes events and save to file
+          kubectl get events -n $namespace > recipes/pod-logs/events.txt
```
ytimocin, line 578: And maybe another log after this line to state that the events are saved?

.github/workflows/functional-test.yaml
sk593: No it wouldn't. We wanted to track recipe upload failures specifically, but we could just track the events whenever they're uploaded

.github/workflows/functional-test.yaml
sk593: There's a line below that logs when all of them are uploaded but will update to this

PR #6796, opened 2023-11-16
Workflow for tests approvals using environments and workflow_run

### LINKED ISSUE #6548: Use GitHub Checks API from our functional test workflow to gate merges

Functional testing action should block merge until functional tests pass on the **exact** commit being merged. The current state of the art is that we rely on the code-reviewer to enforce this.

### Proposed solution

Hey Folks,

I found a good example blog of someone doing basically exactly what we’re trying to with /ok-to-test. This solution is integration with GitHub Checks. For our solution, it means a PR cannot be merged until the functional tests pass on the exact commit (eg: exactly what we want).

You can find the Blog writeup here: https://josh-ops.com/posts/using-github-checks-api/

You should also look at this post because creating a GitHub App is required: https://josh-ops.com/posts/github-apps/ It probably sounds scary to create a GitHub App, but it’s a similar amount of work to creating a PAT. Now that I know more about it, I feel like we should be using GitHub Apps for more things. The reason why we need to create an app is that the Checks API only works correctly from an App. I tried without creating an App first and it's buggy, GH support says not to do that.

I created a working example of this that doesn’t run any tests, so I could prove out the idea. You can find it here: https://github.com/rynowak/actions-testing There’s no permissions limiting /ok-to-test on my repo, so feel free to try it out.

The basic flow is like this:

- /ok-to-test does a repository dispatch to start the functional test workflow (we already have this)
- Inside the functional test workflow we login as the app ([here](https://github.com/rynowak/actions-testing/blob/main/.github/workflows/functional-tests.yaml#L12))
- Create a Check for the functional tests using the app's token ([here](https://github.com/rynowak/actions-testing/blob/main/.github/workflows/functional-tests.yaml#L48))
- We run the functional tests (we already have this)
- When the functional tests finish we update the check with the results ([here](https://github.com/rynowak/actions-testing/blob/main/.github/workflows/functional-tests.yaml#L66))
- One this runs ones on the repo we can select the check in the branch protection menu and mark it as required.

My example of this **does not** do a good job reporting the results of the functional test run. This is something that needs to be addressed for us to use this.

[AB#9889](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/9889)

vinayada1:
# Description

- Remove "ok-to-test" trigger to run functional tests
- Modified github workflow to use "functional-tests" environment which has been configured as:-
MicrosoftTeams-image

## Type of change

Fixes: #6548

### Summary
🆕💾🔒

This pull request adds a new workflow and two new actions to enable manual approval for running functional tests on a PR. It also uses artifacts to pass the PR number between workflows, and fixes a typo in the bug report template. The changes affect the files `.github/workflows/functional-test.yaml`, `.github/workflows/functional-tests-approval.yaml`, `.github/actions/download-pr-data-artifact/action.yaml`, and `.github/actions/save-pr-as-artifact/action.yaml`.

> _We need approval for the functional test_
> _Or we'll face the doom of a broken PR_
> _We save and download the artifact_
> _With `github-script` and `upload-artifact`_

### Walkthrough
*  Enable manual approval for functional tests on PRs ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL26-R30), [link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-336842f9ad3be6e33759449fb701d721641268b444401b71c63a4e371ae91598R1-R15))
  - Add a new workflow `functional-tests-approval.yaml` that saves the PR number as an artifact using `save-pr-as-artifact` action ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-336842f9ad3be6e33759449fb701d721641268b444401b71c63a4e371ae91598R1-R15), [link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-a06147d654903226b1bb5a4635b89fa5f7e0bd2168a3590f05bdbe65c2c0e2d4R1-R17))
  - Modify the trigger events and checkout steps of `functional-test.yaml` to use the PR number artifact and the `workflow_run` event ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL26-R30), [link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL97-R118))
  - Use the `download-pr-data-artifact` action to download and unzip the artifact ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL97-R118), [link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-47021e538b3451866b8d8aa608664649ca716a99718bbaa9472a3af67dcd0610R1-R28))
* Fix typo in bug report body ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL578-L577))
  - Replace the word "here" with a link to the action run in the `bug-report.md` template ([link](https://github.com/radius-project/radius/pull/6796/files?diff=unified&w=0#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcL578-L577))


.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+            return artifact.name == "pr_number"
+          })[0];
+          let download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+          });
+          let fs = require('fs');
+          fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr_number.zip`, Buffer.from(download.data));
```
rynowak, line 25: Looks like the indentation is off here ...

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,26 @@
+name: "Save PR number as artifact"
+description: "Save PR number as artifact"
```
rynowak, line 2: Can you expand on this? Someone looking at this should be able to understand what it does and why it's important.

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,26 @@
+name: "Save PR number as artifact"
```
rynowak, line 1: Is this right? This is the action that downloads and sets the PR number.

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,26 @@
...
+          })[0];
+          let download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+          });
+          let fs = require('fs');
+          fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr_number.zip`, Buffer.from(download.data));
+
```
rynowak, line 26: I think this code should either set the environment variable, or even better return the PR number as an output.

.github/workflows/functional-test.yaml
```diff
@@ -23,7 +23,11 @@ on:
     - cron: "30 0,12 * * 0,6"
   # Dispatch on external events
   repository_dispatch:
-    types: [functional-tests, de-functional-test]
+    types: [de-functional-test]
+  workflow_run:
+    workflows: ['Functional test approval using environments']
+    types:
+      - completed
```
rynowak, line 30: It's kinda nice that with this separation between approval and run, we can still use the scheduled runs and support repository dispatch.

.github/workflows/functional-test.yaml
```diff
@@ -94,22 +98,27 @@ jobs:
...
-      - name: Set up checkout target (repository_dispatch from /ok-to-test)
-        if: github.event_name == 'repository_dispatch'
+      - name: Use custom actions
+        uses: actions/checkout@v3
+      - name: 'Download PR data artifacts'
+        if: github.event_name == 'workflow_run'
+        uses: ./.github/actions/download-pr-data-artifact
+      - name: 'Unzip artifact'
+        if: github.event_name == 'workflow_run'
+        run: unzip pr_number.zip
```
rynowak, line 108: It would be great if this was part of the custom action.

.github/workflows/functional-test.yaml
```diff
@@ -94,22 +98,27 @@ jobs:
             echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
             echo "CHECKOUT_REF=${{ github.ref }}" >> $GITHUB_ENV
             echo "PR_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
-      - name: Set up checkout target (repository_dispatch from /ok-to-test)
-        if: github.event_name == 'repository_dispatch'
+      - name: Use custom actions
+        uses: actions/checkout@v3
+      - name: 'Download PR data artifacts'
+        if: github.event_name == 'workflow_run'
+        uses: ./.github/actions/download-pr-data-artifact
```
rynowak, line 105: There's an alternative syntax for this reference: https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

I don't think it will require the checkout, which might simplify things.

It's kinda hard to reason about the fact that we're checkout out the repo twice, it would be nice to avoid that complication.

.github/workflows/functional-tests-approval.yaml
```diff
@@ -0,0 +1,17 @@
+name: 'Functional test approval using environments'
+on:
+  pull_request:
+    branches:
+      - main
+      - features/*
+      - release/*
+jobs:
+  functional-tests:
```
rynowak, line 9: Can you find a name that makes more sense? This isn't actually running the functional tests.

.github/workflows/functional-tests-approval.yaml
```diff
@@ -0,0 +1,17 @@
+name: 'Functional test approval using environments'
```
rynowak, line 1: Can we simplify this name? This is going to appear in the UI for users. This is the name I chose when we were experimenting to make it distinct.

.github/workflows/functional-test.yaml
vinayada1: I am getting an error without this step. Also, I found this: 
https://github.com/orgs/community/discussions/26579

rynowak: I'm not suggesting you remove that step, and I don't think the discussion you linked is relevant.

You can use a local path as a reference, or you can use a github path.

For this example it would be:

```
uses: radius-project/radius/.github/workflows/download-pr-data-artifact@main
```

If you do it this way, you may not need the checkout. I think it functionally will do the same thing, but it might be less complicated. 

I'll leave this up to you whether it's better or not.

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+      uses: actions/github-script@v6
+      with:
+        script: |
+          let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: context.payload.workflow_run.id,
+          });
+          let matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+            return artifact.name == "pr_number"
```
ytimocin, line 22: Is the "pr_number" the "pr_number" from the outputs? Is there a way to get the artifact just querying with this number so that we don't need to list all and run a filtering function on the result?

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+          })[0];
```
ytimocin, line 23: What if there is no matching artifacts? Can the `[0]` return a failure? I believe an `array out of bounds` error may be likely here.

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+          let download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+          });
+          let fs = require('fs');
+          fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr_number.zip`, Buffer.from(download.data));
```
ytimocin, line 31: Should we check if the **data** exists?

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - name: 'Unzip artifact'
+      shell: bash
+      run: unzip pr_number.zip
```
ytimocin, line 34: Why do we keep using `pr_number`? Can we have more meaningful name here?

.github/workflows/functional-test.yaml
vinayada1: I tried converting this composite action to reusable workflow but it is getting complicated with the download-artifact one. We cannot use the uses in a step...so need to convert the download part to a separate job. Added that as a dependency to the build job which is using the output set but that gives me an error about cyclic dependency: The workflow is not valid. .github/workflows/functional-test.yaml (Line: 80, Col: 12): Job 'build' depends on unknown job 'download_pr_number'. .github/workflows/functional-test.yaml (Line: 301, Col: 12): Job 'tests' depends on job 'build' which creates a cycle in the dependency graph. The trigger also needs to change since a reusable workflow requires a workflow_call trigger. I think it is easier to leave it as is with the composite actions.

.github/actions/download-pr-data-artifact/action.yaml
vinayada1: We are literally saving the PR number as an artifact to pass from one workflow to another

.github/actions/download-pr-data-artifact/action.yaml
vinayada1: we do not have the id

.github/actions/download-pr-data-artifact/action.yaml
vinayada1: This is supposed to be used in conjunction with the save-pr-as-artifact. it's ok for the workflow to fail if that's not the case.

.github/actions/download-pr-data-artifact/action.yaml
vinayada1: we want the workflow to fail if something is not as expected.

.github/workflows/functional-test.yaml
rynowak: I'm not sure I understand what you're saying. 

My suggestion was to reference the reusable action using the repository-syntax instead of cloning the repo and using a relative file-path. Are you saying that's not possible?

vinayada1: I am using a composite action right now: ./.github/actions/download-pr-data-artifact whereas https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow is about reusable workflows. Looks like composite actions require a repo checkout. Therefore, I tried converting the composite action into a reusable workflow.

rynowak: OK thanks, I think I was confused about the terminology.

.github/workflows/functional-tests-approval.yaml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request:
+    branches:
+      - main
+      - features/*
+      - release/*
+jobs:
+  wait_for_approval:
+    runs-on: ubuntu-latest
+    environment: functional-tests
+    steps:
```
vinayada1, line 12: Added an extra step because you cannot use environment with reusable workflow

rynowak: Can you explain more? I'm not sure what you mean. I thought you were writing composite actions?

rynowak (approved):
Very excited to sign off on this 👏

PR #6864, opened 2023-11-28
Fixing the panic for annotations.Configuration

ytimocin:
# Description
Fixing the panic for annotations.Configuration

## Type of change
Fixes: #issue_number

### Summary
🛡️🐛🚀

Fix possible nil pointer dereferences in deployment reconciler. Ensure `annotations.Configuration` is always initialized before accessing its subfields in `pkg/controller/reconciler/deployment_reconciler.go`.

> _`Configuration` /_
> _may be nil, so we guard /_
> _against winter frost_

### Walkthrough
* Initialize configuration annotations field to avoid nil pointer dereferences ([link](https://github.com/radius-project/radius/pull/6864/files?diff=unified&w=0#diff-87a7dfa06c174a6b41b671b2cfffb84c81e481881baec866a026e7dd00db8195L251-R261))

### DIFF
--- a/pkg/controller/reconciler/annotations.go
+++ b/pkg/controller/reconciler/annotations.go
@@ -44,13 +44,29 @@ type deploymentAnnotations struct {
 	// This will be nil if Radius is not enabled for the Deployment.
 	Configuration *deploymentConfiguration
 
-	//ConfigurationHash is the hash of the user-provided configuration.
+	// ConfigurationHash is the hash of the user-provided configuration.
 	// This will be used to diff the configuration and determine if the Deployment needs to be updated.
 	ConfigurationHash string
 
 	// Status is the status of the Deployment (Radius related status).
 	Status *deploymentStatus
 }
+
+// There are 4 cases that is possible based on the previous state and the current state of the Deployment:
+// Case 1: Previous State: Enabled - Current State: Disabled
+// Case 2: Previous State: Disabled - Current State: Enabled
+// Case 3: Previous State: Enabled - Current State: Enabled
+// Case 4: Previous State: Disabled - Current State: Disabled
+//
+// How to understand the previous state:
+// 1. If "radapp.io/status" annotation is set, then the previous state is Enabled.
+//
+// Ways to disable Radius:
+// 1. "radapp.io/enabled" annotation is set to "false".
+// 2. "radapp.io/enabled" annotation is not set.
+//
+// Ways to enable Radius:
+// 1. "radapp.io/enabled" annotation is set to "true".
 
 // deploymentConfiguration is the configuration of the Deployment provided by the user via annotations.
 type deploymentConfiguration struct {
@@ -98,9 +114,9 @@ func readAnnotations(deployment *appsv1.Deployment) (deploymentAnnotations, erro
 		if err != nil {
 			return result, fmt.Errorf("failed to unmarshal status annotation: %w", err)
 		}
-	}
 
-	result.Status = &s
+		result.Status = &s
+	}
 
 	// Note: we need to read and return the configuration even if Radius is not enabled for the Deployment.
 	// This is important so that can clean up previously created connections when Radius is disabled.
@@ -183,4 +199,20 @@ func (annotations *deploymentAnnotations) IsUpToDate() bool {
 	}
 
 	return hash == annotations.ConfigurationHash
+}
+
+// OperationInProgress returns true if there is an operation in progress for the given deployment.
+func (annotations *deploymentAnnotations) OperationInProgress() bool {
+	return annotations.Status != nil && annotations.Status.Operation != nil
+}
+
+// isRadiusEnabled returns true if Radius is enabled for the given deployment.
+func (annotations *deploymentAnnotations) isRadiusEnabled() bool {
+	return annotations.Configuration != nil
+}
+
+// needsCleanup returns true if Radius was previously enabled on the deployment and now is disabled.
+// This means that we need to clean up the resources created by Radius.
+func (annotations *deploymentAnnotations) needsCleanup() bool {
+	return annotations.Configuration == nil && annotations.Status != nil
 }
--- a/pkg/controller/reconciler/annotations_test.go
+++ b/pkg/controller/reconciler/annotations_test.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_readAnnotations(t *testing.T) {
+	testDeploymentStatus := &deploymentStatus{
+		Scope:       "/planes/radius/local/resourceGroups/controller-test",
+		Application: "test-application",
+		Environment: "test-environment",
+		Container:   "test-container",
+		Operation:   nil,
+		Phrase:      deploymentPhraseReady,
+	}
+
+	dsm, err := json.Marshal(testDeploymentStatus)
+	require.NoError(t, err)
+
+	// invalidDeploymentStatus is missing a curly brace at the end of the JSON
+	// so that an unmarshaling error can be triggered.
+	invalidDeploymentStatus := []byte(`{"invalid": "json"`)
+
+	tests := []struct {
+		name        string
+		deployment  *appsv1.Deployment
+		annotations deploymentAnnotations
+		err         error
+	}{
+		{
+			name: "radius-disabled-with-annotation",
+			deployment: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						AnnotationRadiusEnabled: "false",
+					},
+				},
+			},
+			annotations: deploymentAnnotations{
+				Configuration:     nil,
+				ConfigurationHash: "",
+				Status:            nil,
+			},
+			err: nil,
+		},
+		{
+			name: "radius-disabled-empty-annotation-map",
+			deployment: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{},
+				},
+			},
+			annotations: deploymentAnnotations{
+				Configuration:     nil,
+				ConfigurationHash: "",
+				Status:            nil,
+			},
+			err: nil,
+		},
+		{
+			name: "radius-disabled-no-annotations",
+			deployment: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{},
+			},
+			annotations: deploymentAnnotations{
+				Configuration:     nil,
+				ConfigurationHash: "",
+				Status:            nil,
+			},
+			err: nil,
+		},
+		{
+			name: "radius-was-enabled-now-disabled-with-annotations",
+			deployment: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						AnnotationRadiusEnabled:           "false",
+						AnnotationRadiusConfigurationHash: "configuration-hash",
+						AnnotationRadiusStatus:            string(dsm),
+					},
+				},
+			},
+			annotations: deploymentAnnotations{
+				Configuration:     nil,
+				ConfigurationHash: "configuration-hash",
+				Status:            testDeploymentStatus,
+			},
+			err: nil,
+		},
+		{
+			name: "radius-enabled-with-annotations",
+			deployment: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						AnnotationRadiusEnabled:                              "true",
+						AnnotationRadiusConfigurationHash:                    "configuration-hash",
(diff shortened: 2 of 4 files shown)


pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -248,13 +248,17 @@ func (r *DeploymentReconciler) reconcileUpdate(ctx context.Context, deployment *
 		}
 	}
 
+	if annotations.Configuration == nil {
+		annotations.Configuration = &deploymentConfiguration{}
+	}
```
rynowak, line 253: Similar to the last issue we looked at, I'd like to understand the root cause before we start making fixes. 

Do you know what set of conditions are causing this?

It should be possible to add a new integration test in `deployment_reconciler_test.go` that reproduces this issue.

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -478,82 +496,10 @@ func makeDeployment(name types.NamespacedName) *appsv1.Deployment {
...
-		if assert.NotNil(t, annotations.Status) && assert.Equal(t, deploymentPhraseUpdating, annotations.Status.Phrase) {
-			assert.NotEmpty(t, annotations.Status.Operation)
-		}
-	}, deploymentTestWaitDuration, deploymentTestWaitInterval, "waiting for state to be Updating")
-
-	return &annotations
-}
-
-func waitForStateReady(t *testing.T, client client.Client, name types.NamespacedName) *deploymentAnnotations {
+func waitForState(t *testing.T, client client.Client, name types.NamespacedName, phrase deploymentPhrase) *deploymentAnnotations {
```
ytimocin, line 499: Not sure if this is a good in terms of refactoring this code. What do you think @rynowak?

pkg/controller/reconciler/annotations.go
```diff
@@ -52,6 +52,11 @@ type deploymentAnnotations struct {
 	Status *deploymentStatus
 }
 
+// Case 1: Enabled => Disabled
+// Case 2: Disabled => Enabled
+// Case 3: Enabled => Enabled
+// Case 4: Disabled => Disabled
```
rynowak, line 58: I think this comment needs more context to be useful. Imagine if you came back to this in 6 months, how would you know what it's saying?

pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -126,11 +127,22 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	//
+	// If the Configuration is nil, which means that Radius is disabled, and Status is not nil,
+	// then we need to reconcile the deletion of the Deployment.
 	if deployment.DeletionTimestamp != nil || (annotations.Configuration == nil && annotations.Status != nil) {
 		return r.reconcileDelete(ctx, &deployment, &annotations)
 	}
 
-	return r.reconcileUpdate(ctx, &deployment, &annotations)
+	// When do we need to reconcile update?
+	// If Radius is (still) enabled, meaning Status can be nil or not.
```
rynowak, line 140: 👍

pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -249,12 +261,12 @@ func (r *DeploymentReconciler) reconcileUpdate(ctx context.Context, deployment *
 	}
 
 	environmentName := "default"
-	if annotations.Configuration != nil && annotations.Configuration.Environment != "" {
+	if annotations.Configuration.Environment != "" {
```
rynowak, line 264: I'm still waiting to hear what the root-cause is. Did we have a logic error and missed a case we should handle?

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -446,6 +446,33 @@ func Test_DeploymentReconciler_Connections(t *testing.T) {
...
+	name := types.NamespacedName{Namespace: "deployment-disabled", Name: "test-deployment-disabled"}
+	err := client.Create(ctx, &corev1.Namespace{ObjectMeta: ctrl.ObjectMeta{Name: name.Namespace}})
+	require.NoError(t, err)
+
+	deployment := makeDeployment(name)
+	err = client.Create(ctx, deployment)
+	require.NoError(t, err)
+
+	// Deployment will be waiting for environment to be created.
+	createEnvironment(radius, "default")
```
rynowak, line 465: This shouldn't be required actually.

pkg/controller/reconciler/annotations.go
```diff
@@ -44,14 +44,23 @@ type deploymentAnnotations struct {
 	// This will be nil if Radius is not enabled for the Deployment.
 	Configuration *deploymentConfiguration
 
-	//ConfigurationHash is the hash of the user-provided configuration.
+	// ConfigurationHash is the hash of the user-provided configuration.
 	// This will be used to diff the configuration and determine if the Deployment needs to be updated.
 	ConfigurationHash string
 
 	// Status is the status of the Deployment (Radius related status).
 	Status *deploymentStatus
```
ytimocin, line 52: If Status is not provided via annotations even though Radius is disabled, we have no way of knowing if that Deployment had Radius enabled before.

pkg/controller/reconciler/annotations.go
```diff
@@ -98,9 +107,11 @@ func readAnnotations(deployment *appsv1.Deployment) (deploymentAnnotations, erro
 		if err != nil {
 			return result, fmt.Errorf("failed to unmarshal status annotation: %w", err)
 		}
-	}
 
-	result.Status = &s
+		// The reason I moved this here is because if Status is not nil,
+		// then we should set it in the result. Otherwise it should be nil.
+		result.Status = &s
```
ytimocin, line 113: The problem here was that the previous place of this line, which was outside of this **if case**, was causing two different Status: nil and empty object (**deploymentStatus{}**). And in other parts of the code we are checking is Status is nil not if Status is an empty object.

pkg/controller/reconciler/annotations_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := readAnnotations(tt.deployment)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("readAnnotations() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("readAnnotations() = %v, want %v", got, tt.want)
```
rynowak, line 144: Please don't write this style of test comparison. See guidance here: https://github.com/radius-project/radius/tree/main/docs/contributing/contributing-code/contributing-code-writing#appendix-style-guidance-we-dont-follow

pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -678,3 +691,28 @@ func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Secret{}).
 		Complete(r)
 }
+
+// OperationInProgress returns true if there is an operation in progress for the given deployment.
+func (r *DeploymentReconciler) OperationInProgress(annotations deploymentAnnotations) bool {
```
rynowak, line 696: Would it be clearer if these functions were defined on `deploymentAnnotations`? They don't need any state from the reconciler.

pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -678,3 +691,28 @@ func (r *DeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// 	return annotations.Configuration == nil
+// }
+
+// isRadiusEnabled returns true if Radius is enabled for the given deployment.
+func (r *DeploymentReconciler) isRadiusEnabled(annotations deploymentAnnotations) bool {
+	return annotations.Configuration != nil
+}
+
+// radiusEnabledToDisabledCheck returns true if the deployment was previously enabled and is now disabled.
+func (r *DeploymentReconciler) radiusEnabledToDisabledCheck(annotations deploymentAnnotations) bool {
```
rynowak, line 711: Can we come up with a better name for this?

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -446,6 +446,44 @@ func Test_DeploymentReconciler_Connections(t *testing.T) {
 	require.True(t, apierrors.IsNotFound(err))
 }
 
+// Creates a deployment with Radius disabled.
+//
+// Then checks the Events for Disabled-Disabled.
```
rynowak, line 451: I'd like to see the same test for the case where no annotations are present.

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -446,6 +446,44 @@ func Test_DeploymentReconciler_Connections(t *testing.T) {
...
+
+	name := types.NamespacedName{Namespace: "deployment-disabled-disabled", Name: "test-deployment-disabled-disabled"}
+	err := client.Create(ctx, &corev1.Namespace{ObjectMeta: ctrl.ObjectMeta{Name: name.Namespace}})
+	require.NoError(t, err)
+
+	deployment := makeDeployment(name)
+	err = client.Create(ctx, deployment)
+	require.NoError(t, err)
+
+	time.Sleep(time.Second * 5)
```
rynowak, line 464: Can you find a better way than using an explicit sleep? Sleeping for a fixed amount of time introduces bad tradeoffs in tests.

- The test can never go **faster** than the sleeps will allow, so you choose low numbers.
- Eventually you'll encounter a situation where the low number is not long enough, and then the test will be flaky.

PTAL at how the rest of these tests solve this with `require.Eventually`.

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -446,6 +446,44 @@ func Test_DeploymentReconciler_Connections(t *testing.T) {
...
+	// Events: Read Kubebuilder - Events on the Deployment
+	events := &corev1.EventList{}
+	err = client.List(ctx, events)
+	require.NoError(t, err)
+
+	for _, event := range events.Items {
+		require.Equal(t, "NoOperation", event.Reason)
+		require.Equal(t, fmt.Sprintf("No Operation on %s", deployment.Name), event.Message)
+	}
+}
```
rynowak, line 485: We should check the events before **and** after updating the deployment.

pkg/controller/reconciler/annotations.go
ytimocin: And the only way to know that Radius was previously enabled on the Deployment is to check the Status object I guess, right? @rynowak

pkg/controller/reconciler/deployment_reconciler.go
ytimocin: I explained the root cause here: https://github.com/radius-project/radius/pull/6864/files#r1411059357.

pkg/controller/reconciler/deployment_reconciler_test.go
ytimocin: Added

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -446,6 +450,96 @@ func Test_DeploymentReconciler_Connections(t *testing.T) {
...
+		expectedEvent{
+			EventType: "Normal",
+			Reason:    "NoOp",
+			Message:   fmt.Sprintf("Radius is not enabled for %s", deployment.Name),
+			Count:     1,
+		},
+	)
+
+	// Update Labels of the Deployment so that the Reconciler can detect a change.
+	deployment.Labels = map[string]string{"foo": "bar"}
```
ytimocin, line 527: Here there is no annotation set @rynowak.

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -574,6 +668,39 @@ func waitForStateDeleting(t *testing.T, client client.Client, name types.Namespa
...
+
+		found := false
+		for _, e := range events.Items {
+			// If the event is the one we are looking for, check the count.
+			if e.Type == event.EventType && e.Reason == event.Reason &&
+				e.Message == event.Message && assert.Equal(t, event.Count, int(e.Count)) {
+				found = true
+			}
+		}
+		assert.True(t, found)
```
ytimocin, line 700: **Found** must also be True because we are trying to find the Expected Event.

pkg/controller/reconciler/annotations.go
```diff
@@ -44,14 +44,23 @@ type deploymentAnnotations struct {
...
 }
 
+// Ways to disable Radius:
+// 1. "radapp.io/enabled" annotation is set to "false"
+// 2. "radapp.io/enabled" annotation is not set
+
+// Case 1: Radius was Enabled and now it is Disabled.
+// Case 2: Radius was Disabled and now it is Enabled.
+// Case 3: Radius was and is Enabled.
+// Case 4: Radius was and is Disabled.
```
rynowak, line 62: This information is still missing context, it feels a lot more like personal notes than something that will be useful in the future. Someone who was present on the call where we discussed this might get it, but someone reading this in the future definitely won't. 


Can you try adding a few more sentences of explanation?

pkg/controller/reconciler/deployment_reconciler_test.go
```diff
@@ -42,6 +42,10 @@ const (
 	deploymentTestWaitDuration            = time.Second * 10
 	deploymentTestWaitInterval            = time.Second * 1
 	deploymentTestControllerDelayInterval = time.Millisecond * 100
+
+	fetchEventWaitDuration            = time.Second * 10
+	fetchEventWaitInterval            = time.Second * 1
+	fetchEventControllerDelayInterval = time.Millisecond * 100
```
rynowak, line 48: nit: I think you could just use `deploymentTestWaitDuration` and `deploymentTestWaitInterval`. They really have the same purpose.

rynowak: `fetchEventControllerDelayInterval` looks unused, can you remove it?

rynowak (changes requested):
Test changes look right, just a few small things to clean up.

pkg/controller/reconciler/deployment_reconciler_test.go
ytimocin: Done

rynowak (approved):

PR #6969, opened 2023-12-18
Add terraform template path validation

### LINKED ISSUE #6642: Validation for Terraform Recipe Template Paths
### Current State/Problem
Currently, we support only [Terraform registry](https://developer.hashicorp.com/terraform/language/modules/sources#terraform-registry) and [HTTP URLs](https://developer.hashicorp.com/terraform/language/modules/sources#http-urls) as allowed module sources for Terraform recipe template paths. However, we lack a proper validation mechanism for when an incorrect source is given during either recipe registration or deployment. Consequently, when a deployment fails, the reason provided to users is ambiguous resulting into a poor user experience.

### Proposed Solution
To address this, we should utilize [Terraform's module parsing logic](https://github.com/hashicorp/terraform-registry-address/blob/main/module.go#L44) to verify the recipe template path when the recipe `templateKind` is provided as Terraform, at recipe registration stage.

## Acceptance criteria
If an unsupported template path for a Terraform recipe is provided, the recipe registration should fail and return a detailed error message to the user.

[AB#10019](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10019)

sk593:
# Description

This adds validation for terraform template paths. We will only support [Terraform registry](https://developer.hashicorp.com/terraform/language/modules/sources#terraform-registry) and [HTTP URLs](https://developer.hashicorp.com/terraform/language/modules/sources#http-urls) as allowed module sources for Terraform recipe template paths. Module sources outside these will be treated as unsupported and an error will be returned. 

## Type of change

- This pull request adds or changes features of Radius and has an approved issue #6642 

Fixes: #6642 

## Auto-generated summary

copilot:all

### DIFF
--- a/pkg/corerp/api/v20231001preview/environment_conversion.go
+++ b/pkg/corerp/api/v20231001preview/environment_conversion.go
@@ -18,6 +18,7 @@ package v20231001preview
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	v1 "github.com/radius-project/radius/pkg/armrpc/api/v1"
@@ -25,6 +26,7 @@ import (
 	"github.com/radius-project/radius/pkg/kubernetes"
 	types "github.com/radius-project/radius/pkg/recipes"
 
+	tfaddr "github.com/hashicorp/terraform-registry-address"
 	rp_util "github.com/radius-project/radius/pkg/rp/portableresources"
 	rpv1 "github.com/radius-project/radius/pkg/rp/v1"
 	"github.com/radius-project/radius/pkg/to"
@@ -32,7 +34,7 @@ import (
 
 const (
 	EnvironmentComputeKindKubernetes = "kubernetes"
-	invalidLocalModulePathFmt        = "local module paths are not supported with Terraform Recipes. The 'templatePath' '%s' was detected as a local module path because it begins with '/' or './' or '../'."
+	invalidModulePathFmt             = "only Terraform registry and HTTP URLs are supported as module sources with Terraform Recipes. The 'templatePath' '%s' was detected as an invalid module source."
 )
 
 // ConvertTo converts from the versioned Environment resource to version-agnostic datamodel.
@@ -298,10 +300,16 @@ func toEnvExtensionDataModel(e ExtensionClassification) datamodel.Extension {
 func toEnvironmentRecipeProperties(e RecipePropertiesClassification) (datamodel.EnvironmentRecipeProperties, error) {
 	switch c := e.(type) {
 	case *TerraformRecipeProperties:
+		// Radius only supports Terraform registry and HTTP URLs as module sources.
 		if c.TemplatePath != nil {
-			// Check for local paths
-			if strings.HasPrefix(to.String(c.TemplatePath), "/") || strings.HasPrefix(to.String(c.TemplatePath), "./") || strings.HasPrefix(to.String(c.TemplatePath), "../") {
-				return datamodel.EnvironmentRecipeProperties{}, v1.NewClientErrInvalidRequest(fmt.Sprintf(invalidLocalModulePathFmt, to.String(c.TemplatePath)))
+			// We first validate if the template path is a valid Terraform registry module source.
+			_, err := tfaddr.ParseModuleSource(to.String(c.TemplatePath))
+			if err != nil {
+				// If the template path is not a Terraform registry module source, we validate if it is an HTTP URL.
+				_, err = url.ParseRequestURI(to.String(c.TemplatePath))
+				if err != nil {
+					return datamodel.EnvironmentRecipeProperties{}, v1.NewClientErrInvalidRequest(fmt.Sprintf(invalidModulePathFmt, to.String(c.TemplatePath)))
+				}
 			}
 		}
 		return datamodel.EnvironmentRecipeProperties{
--- a/pkg/corerp/api/v20231001preview/environment_conversion_test.go
+++ b/pkg/corerp/api/v20231001preview/environment_conversion_test.go
@@ -142,6 +142,10 @@ func TestConvertVersionedToDataModel(t *testing.T) {
 								TemplatePath: "br:ghcr.io/sampleregistry/radius/recipes/rediscaches",
 								PlainHTTP:    true,
 							},
+							"terraform-recipe": datamodel.EnvironmentRecipeProperties{
+								TemplateKind: recipes.TemplateKindTerraform,
+								TemplatePath: "git::https://example.com/vpc.git",
+							},
 						},
 						dapr_ctrl.DaprStateStoresResourceType: {
 							"statestore-recipe": datamodel.EnvironmentRecipeProperties{
@@ -291,7 +295,11 @@ func TestConvertVersionedToDataModel(t *testing.T) {
 		},
 		{
 			filename: "environmentresource-terraformrecipe-localpath.json",
-			err:      &v1.ErrClientRP{Code: v1.CodeInvalid, Message: fmt.Sprintf(invalidLocalModulePathFmt, "../not-allowed/")},
+			err:      &v1.ErrClientRP{Code: v1.CodeInvalid, Message: fmt.Sprintf(invalidModulePathFmt, "../not-allowed/")},
+		},
+		{
+			filename: "environmentresource-terraformrecipe-unsupported.json",
+			err:      &v1.ErrClientRP{Code: v1.CodeInvalid, Message: fmt.Sprintf(invalidModulePathFmt, "github.com/hashicorp/example")},
 		},
 	}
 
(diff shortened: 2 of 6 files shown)


pkg/corerp/api/v20231001preview/environment_conversion_test.go
```diff
@@ -269,6 +269,88 @@ func TestConvertVersionedToDataModel(t *testing.T) {
 			},
 			err: nil,
 		},
+		{
+			filename: "environmentresource-terraformrecipe-http.json",
```
kachawla, line 273: Isn't this covered by this existing scenario https://github.com/radius-project/radius/blob/9b47d5ea6407e451381fcf8ce971de51f63bae90/pkg/corerp/api/v20231001preview/environment_conversion_test.go#L135

pkg/corerp/api/v20231001preview/environment_conversion_test.go
```diff
@@ -269,6 +269,88 @@ func TestConvertVersionedToDataModel(t *testing.T) {
...
+								TemplatePath: "https://example.com/vpc-module.zip",
+							},
+						},
+					},
+				},
+			},
+			err: nil,
+		},
+		{
+			filename: "environmentresource-terraformrecipe-tfregistry.json",
```
kachawla, line 314: Same here, should this already be covered by this existing scenario? https://github.com/radius-project/radius/blob/9b47d5ea6407e451381fcf8ce971de51f63bae90/pkg/corerp/api/v20231001preview/environment_conversion_test.go#L131

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -299,9 +301,12 @@ func toEnvironmentRecipeProperties(e RecipePropertiesClassification) (datamodel.
 	switch c := e.(type) {
 	case *TerraformRecipeProperties:
 		if c.TemplatePath != nil {
-			// Check for local paths
-			if strings.HasPrefix(to.String(c.TemplatePath), "/") || strings.HasPrefix(to.String(c.TemplatePath), "./") || strings.HasPrefix(to.String(c.TemplatePath), "../") {
-				return datamodel.EnvironmentRecipeProperties{}, v1.NewClientErrInvalidRequest(fmt.Sprintf(invalidLocalModulePathFmt, to.String(c.TemplatePath)))
+			_, err := tfaddr.ParseModuleSource(to.String(c.TemplatePath))
+			if err != nil {
+				_, err = url.ParseRequestURI(to.String(c.TemplatePath))
```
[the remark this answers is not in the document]
sk593: Added

pkg/corerp/api/v20231001preview/environment_conversion_test.go
sk593: Ah yes, removed this and the duplicate tf registry test

kachawla (dismissed):

pkg/corerp/api/v20231001preview/environment_conversion_test.go
```diff
@@ -291,7 +291,11 @@ func TestConvertVersionedToDataModel(t *testing.T) {
 		},
 		{
 			filename: "environmentresource-terraformrecipe-localpath.json",
-			err:      &v1.ErrClientRP{Code: v1.CodeInvalid, Message: fmt.Sprintf(invalidLocalModulePathFmt, "../not-allowed/")},
+			err:      &v1.ErrClientRP{Code: v1.CodeInvalid, Message: fmt.Sprintf(invalidModulePathFmt, "../not-allowed/")},
+		},
+		{
+			filename: "environmentresource-terraformrecipe-unsupported.json",
```
ytimocin, line 297: This will be denied by the `ParseRequestURI` and the other one by the `ParseModuleSource`, right?

pkg/corerp/api/v20231001preview/testdata/environmentresource-terraformrecipe-unsupported.json
```diff
@@ -0,0 +1,25 @@
...
+      "recipes": {
+        "Applications.Datastores/mongoDatabases":{
+          "cosmos-recipe": {
+            "templateKind": "terraform",
+            "templatePath": "github.com/hashicorp/example"
+          }
+        }
+      }
+    }
+  }
```
ytimocin, line 25: nit: reformat may be needed.

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -298,10 +300,16 @@ func toEnvExtensionDataModel(e ExtensionClassification) datamodel.Extension {
...
-			if strings.HasPrefix(to.String(c.TemplatePath), "/") || strings.HasPrefix(to.String(c.TemplatePath), "./") || strings.HasPrefix(to.String(c.TemplatePath), "../") {
-				return datamodel.EnvironmentRecipeProperties{}, v1.NewClientErrInvalidRequest(fmt.Sprintf(invalidLocalModulePathFmt, to.String(c.TemplatePath)))
+			// We first validate if the template path is a valid Terraform registry module source.
+			_, err := tfaddr.ParseModuleSource(to.String(c.TemplatePath))
+			if err != nil {
+				// If the template path is not a Terraform registry module source, we validate if it is an HTTP URL.
+				_, err = url.ParseRequestURI(to.String(c.TemplatePath))
+				if err != nil {
+					return datamodel.EnvironmentRecipeProperties{}, v1.NewClientErrInvalidRequest(fmt.Sprintf(invalidModulePathFmt, to.String(c.TemplatePath)))
+				}
```
ytimocin, line 312: Should this be a separate function? Is this going to be used elsewhere in the code?

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -18,21 +18,23 @@ package v20231001preview
...
+	tfaddr "github.com/hashicorp/terraform-registry-address"
 	rp_util "github.com/radius-project/radius/pkg/rp/portableresources"
 	rpv1 "github.com/radius-project/radius/pkg/rp/v1"
 	"github.com/radius-project/radius/pkg/to"
 )
 
 const (
 	EnvironmentComputeKindKubernetes = "kubernetes"
-	invalidLocalModulePathFmt        = "local module paths are not supported with Terraform Recipes. The 'templatePath' '%s' was detected as a local module path because it begins with '/' or './' or '../'."
+	invalidModulePathFmt             = "only Terraform registry and HTTP URLs are supported as module sources with Terraform Recipes. The 'templatePath' '%s' was detected as an invalid module source."
```
ytimocin, line 37: cc/ @AaronCrawfis

shalabhms (dismissed):

pkg/corerp/api/v20231001preview/environment_conversion_test.go
sk593: They should both error out since neither of these are a TF registry or an HTTP URL

pkg/corerp/api/v20231001preview/environment_conversion.go
sk593: No, this is just going to be used here when we try to register a recipe to an environment. We don't want to deploy the resource/recipe at all if we know the module source isn't supported and this is the first entry point for registering the recipe

pkg/corerp/api/v20231001preview/testdata/environmentresource-terraformrecipe-unsupported.json
sk593: Updated

AaronCrawfis (changes requested):
Given we verified that `git:/` works correctly today, we should probably also support that. I believe @rynowak uses this today as part of AWS demos.

sk593:
> Given we verified that `git:/` works correctly today, we should probably also support that. I believe @rynowak uses this today as part of AWS demos.

[Git repositories](https://developer.hashicorp.com/terraform/language/modules/sources#generic-git-repository) are considered valid HTTP URLs by `ParseRequestURI` so those should still work. I added a test case for this to confirm. [Github URLs](https://developer.hashicorp.com/terraform/language/modules/sources#generic-git-repository), however, are not supported with the current implementation. Do we want to include support for that as well or should we leave it at `git:/`?

AaronCrawfis:
> [Git repositories](https://developer.hashicorp.com/terraform/language/modules/sources#generic-git-repository) are considered valid HTTP URLs by `ParseRequestURI` so those should still work. I added a test case for this to confirm. [Github URLs](https://developer.hashicorp.com/terraform/language/modules/sources#generic-git-repository), however, are not supported with the current implementation. Do we want to include support for that as well or should we leave it at `git:/`? cc/ @AaronCrawfis

Hmm good question. Let's also add GitHub as well, as it's essentially just an alias for git. It should work as well. Thanks!

sk593:
Closing, we thought the initial issue was caused by Radius not supporting all module sources. This doesn't seem to be the case as Git, S3, etc are able to be used so there's no need for additional validation.

Issue #6297, opened 2023-05-24
Bad error message when not logged in to container registry


rynowak:
### Steps to reproduce (required)

- Let your ACR login expire
- Run the following command

```sh
rad bicep publish --file test/functional/corerp/resources/testdata/recipes/test-recipes/dapr-state-store.bicep --target br:rynowak.azurecr.io/test/functional/corerp/recipes/dapr-state-store:latest
```

Any source and destination are fine, as long as you are not logged in to the destination.

### Observed behavior (required)

```sh
Building test/functional/corerp/resources/testdata/recipes/test-recipes/dapr-state-store.bicep...
Error: Failed to publish Bicep file: HEAD "https://rynowak.azurecr.io/v2/test/functional/corerp/recipes/dapr-state-store/manifests/sha256:10131c538619da7466e8ee9fbf573295cecfab81cd46b544bfbe058cd3347cee": POST "https://rynowak.azurecr.io/oauth2/token": response status code 401: Unauthorized
```

### Desired behavior (required)

A better error message, that explains the problem and tells you how to log in. If I were an external customer I wouldn't be sure how to fix this.

[AB#7968](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/7968)

[AB#9502](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/9502)

lechnerc77:
Is this already fixed? I just retried it and the result is: 

```bash
rad bicep publish --file test/functional/samples/testdata/tutorial-environment.bicep --target br:rynowak.azurecr.io/test/functional/corerp/recipes/dapr-state-store:latest
Building test/functional/samples/testdata/tutorial-environment.bicep...
Failed to publish Bicep file "test/functional/samples/testdata/tutorial-environment.bicep" to "rynowak.azurecr.io/test/functional/corerp/recipes/dapr-state-store:latest". Cause: POST "https://rynowak.azurecr.io/v2/test/functional/corerp/recipes/dapr-state-store/blobs/uploads/": response status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information.: [map[Action:pull Name:test/functional/corerp/recipes/dapr-state-store Type:repository] map[Action:push Name:test/functional/corerp/recipes/dapr-state-store Type:repository]].
```

[...]

PR #6974, opened 2023-12-22
Improve error message under 401 condition with Azure ACR

### LINKED ISSUE #6902: Bad error message when showing recipes in ACR
- Login with ACR login
- Publish by running `rad bicep publish`
- Run the following command

```cli
rad recipe show <recipeName> --resource-type <resourceType>
```

```
\"<acrName>.azurecr.io/azure/sql:latest\": HEAD \"https://<acrName>.azurecr.io/v2/azure/<recipeName>/manifests/latest\": GET \"https://<acrName>.azurecr.io/oauth2/token?scope=repository%3Aazure%2F<recipeName>%3Apull\u0026service=<acrName>.azurecr.io\": response status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information."
```

A better error message that explains the problem and tells you how to login to pull. If I were an external customer I wouldn't be sure how to fix this.

[AB#10641](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10641)

gpltaylor:
# Description

Improve error message returned when publishing to a private ACR

## Type of change

Fixes: #6902 #6297

## Auto-generated summary

copilot:all

### DIFF
--- a/pkg/cli/clierrors/types.go
+++ b/pkg/cli/clierrors/types.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package clierrors
 
+import "strings"
+
 // FriendlyError defines an interface for errors that should be gracefully handled by the CLI and
 // display a friendly error message to the user.
 type FriendlyError interface {
@@ -42,7 +44,7 @@ func (e *ErrorMessage) Error() string {
 		return e.Message
 	}
 
-	return e.Message + " Cause: " + e.Cause.Error() + "."
+	return e.Message + " Cause: " + strings.TrimSpace(e.Cause.Error()) + "."
 }
 
 // IsFriendlyError returns true for ErrorMessage. These errors are always handled gracefully by the CLI.
--- a/pkg/cli/cmd/bicep/publish/publish.go
+++ b/pkg/cli/cmd/bicep/publish/publish.go
@@ -168,7 +168,12 @@ func (r *Runner) Run(ctx context.Context) error {
 
 	err = r.publish(ctx)
 	if err != nil {
-		return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q.", r.File, r.Target)
+		if strings.Contains(r.Target, "azurecr.io") && strings.Contains(r.Target, "authentication required") {
+			const azCliHelp = "https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli"
+			return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q. Please login to az cli and az aci detailed here: %q", r.File, r.Target, azCliHelp)
+		} else {
+			return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q", r.File, r.Target)
+		}
 	}
 
 	r.Output.LogInfo("Successfully published Bicep file %q to %q", r.File, r.Target)
--- a/pkg/cli/cmd/bicep/publish/publish.go
+++ b/pkg/cli/cmd/bicep/publish/publish.go
@@ -173,7 +173,7 @@ func (r *Runner) Run(ctx context.Context) error {
 	err = r.publish(ctx)
 	var httpErr *errcode.ErrorResponse
 	if errors.As(err, &httpErr) {
-		message := fmt.Sprintf("Failed to publish Bicep file %s to %s", r.File, r.Target)
+		message := fmt.Sprintf("Failed to publish Bicep file %q to %q", r.File, r.Target)
 		return handleErrorResponse(httpErr, message)
 	} else if err != nil {
 		return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q", r.File, r.Target)
@@ -192,16 +192,16 @@ func handleErrorResponse(httpErr *errcode.ErrorResponse, message string) error {
 	switch httpErr.StatusCode {
 	case http.StatusUnauthorized:
 		if acrInfo == "" {
-			return clierrors.MessageWithCause(httpErr, "%q\nUnauthorized: Please login to %q", message, httpErr.URL.Host)
+			return clierrors.MessageWithCause(httpErr, "%s\nUnauthorized: Please login to %q", message, httpErr.URL.Host)
 		} else {
-			return clierrors.MessageWithCause(httpErr, "%q\nUnauthorized: Please login to %q\n%q", message, httpErr.URL.Host, acrInfo)
+			return clierrors.MessageWithCause(httpErr, "%s\nUnauthorized: Please login to %q\n%s", message, httpErr.URL.Host, acrInfo)
 		}
 	case http.StatusForbidden:
-		return clierrors.MessageWithCause(httpErr, "%q\nForbidden: You don't have permission to push to %q", message, httpErr.URL.Host)
+		return clierrors.MessageWithCause(httpErr, "%s\nForbidden: You don't have permission to push to %q", message, httpErr.URL.Host)
 	case http.StatusNotFound:
-		return clierrors.MessageWithCause(httpErr, "%q\nNot Found: Unable to find registry %q", message, httpErr.URL.Host)
+		return clierrors.MessageWithCause(httpErr, "%s\nNot Found: Unable to find registry %q", message, httpErr.URL.Host)
 	default:
-		return clierrors.MessageWithCause(httpErr, "%q", message)
+		return clierrors.MessageWithCause(httpErr, "%s", message)
 	}
 }
 
--- a/pkg/cli/cmd/bicep/publish/publish.go
+++ b/pkg/cli/cmd/bicep/publish/publish.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -30,6 +31,8 @@ import (
 	"github.com/radius-project/radius/pkg/cli/framework"
 	"github.com/radius-project/radius/pkg/cli/output"
 
+	"net/http"
+
 	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -39,6 +42,7 @@ import (
 	"oras.land/oras-go/v2/registry"
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/errcode"
 	"oras.land/oras-go/v2/registry/remote/retry"
 )
 
@@ -167,11 +171,28 @@ func (r *Runner) Run(ctx context.Context) error {
 	r.Destination = dest
 
 	err = r.publish(ctx)
-	if err != nil {
-		if strings.Contains(r.Target, "azurecr.io") && strings.Contains(r.Target, "authentication required") {
-			const azCliHelp = "https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli"
-			return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q. Please login to az cli and az aci detailed here: %q", r.File, r.Target, azCliHelp)
-		} else {
+
+	var httpErr *errcode.ErrorResponse
+
+	if errors.As(err, &httpErr) {
+		acrInfo := ""
+
+		if strings.Contains(httpErr.URL.Host, "azurecr.io") {
+			acrInfo = "Visit: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli"
+		}
+
+		switch httpErr.StatusCode {
+		case http.StatusUnauthorized:
+			if acrInfo == "" {
+				return clierrors.MessageWithCause(err, "Unauthorized: Please login to %q\nError: %q", httpErr.URL.Host, r.Target)
+			} else {
+				return clierrors.MessageWithCause(err, "Unauthorized: Please login to %q\n%q\nError: %q", httpErr.URL.Host, acrInfo, r.Target)
+			}
+		case http.StatusForbidden:
+			return clierrors.MessageWithCause(err, "Forbidden: You don't have permission to push to %q\nError: %q", httpErr.URL.Host, r.Target)
+		case http.StatusNotFound:
+			return clierrors.MessageWithCause(err, "Not Found: Unable to find registry %q\nError: %q", httpErr.URL.Host, r.Target)
+		default:
 			return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q", r.File, r.Target)
 		}
 	}
--- a/pkg/cli/cmd/bicep/publish/publish.go
+++ b/pkg/cli/cmd/bicep/publish/publish.go
@@ -171,39 +171,38 @@ func (r *Runner) Run(ctx context.Context) error {
 	r.Destination = dest
 
 	err = r.publish(ctx)
-
 	var httpErr *errcode.ErrorResponse
-
 	if errors.As(err, &httpErr) {
-		acrInfo := ""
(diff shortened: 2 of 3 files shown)


pkg/cli/cmd/bicep/publish/publish.go
```diff
@@ -168,7 +168,12 @@ func (r *Runner) Run(ctx context.Context) error {
 
 	err = r.publish(ctx)
 	if err != nil {
-		return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q.", r.File, r.Target)
+		if strings.Contains(r.Target, "azurecr.io") && strings.Contains(r.Target, "authentication required") {
```
rynowak, line 171: Hi @gpltaylor thanks for sending this PR. Apologies for the delay in giving feedback, it's the holiday's here.

I have a few concerns about this approach.

----

First, I'd like to see if we can make the error reporting better for all hosts, not just ACR. ACR is one of the common OCI registries we've seen in use, but in general any OCI registry works.

Secondly, is there a better way to check for authentication errors? Any time we have to do a string comparison on an error message to see what kind of error it is, it's going to be fragile. The project we're relying on (ORAS) can change the error message at any time.

I think the best way to figure this out would be to set a breakpoint, reproduce the problem, and then see what type of error we get back. My guess is that it's going to be one of these: https://github.com/radius-project/radius/issues/6297#issuecomment-1866873796

When you know what type of error is being returned, you can use `errors.Is` to check for it. 


https://github.com/oras-project/oras-go/blob/main/registry/remote/errcode/errors.go#L42

gpltaylor: I agree using string is a poor choice however this error response from ACR has been standard. 
This was a quick fix for ACR and the typical 401 returned. This is why I kept the standard error. 

I will see if I can review the response and see if there is something more strongly typed.

I am still focusing on ACR as this is the problem set. Should I try to make this more generic for other registers?

rynowak: > I agree using string is a poor choice however this error response from ACR has been standard.
This was a quick fix for ACR and the typical 401 returned. This is why I kept the standard error.

> I will see if I can review the response and see if there is something more strongly typed.

If you're able to reproduce this, you should be able to able to handle a more specific error type. That way we can be use it's an authorization problem and not some other kind of problem.

> I am still focusing on ACR as this is the problem set. Should I try to make this more generic for other registers?

We support **any** OCI registry. If you can make generic improvements that would be better. I'm not opposed to printing a nice error for ACR's case as well, but this should get better for any registry.

gpltaylor:
I have added the construct to improve error messaging to a user publishing.
 
There are a number of Container Registers that all act differently and return "stringly" errors. I have included a base for extending this. I have debugged against GCR, ACR and Docker Hub and all return different error messages.

For example google can return the following, or which returns a status 400. oras-go does not provide any way to interprit this message.

```
Failed to publish Bicep file "../../test/myapp/app.bicep" to "gcr.io/my-project/my-new-image:test" Cause: HEAD "https://gcr.io/v2/my-project/my-new-image/manifests/sha256:4629660f49ab61ec5abb96e207cd795c15d95fa0be1647999a9665359f04ad83": GET "https://gcr.io/v2/token?scope=repository%3Amy-project%2Fmy-new-image%3Apull&service=gcr.io": response status code 400: unknown: Service 'containerregistry.googleapis.com' is not enabled for consumer 'project:my-project'..
```

It appears that improving the messaging to the user will require string pattern matching. As this is brittle I have omitted but I have included ACR using a Host match.

However, I have included better messaging for ACR and included more "user friendly" messaging for common errors. 
Using Host and HttpStatus we can extend this quickly in fucture.

I look forward to feedback.

gpltaylor:
polite knock knock @rynowak 
PS Is there a process for PR changes?

pkg/cli/cmd/bicep/publish/publish.go
```diff
@@ -167,8 +171,30 @@ func (r *Runner) Run(ctx context.Context) error {
 	r.Destination = dest
 
 	err = r.publish(ctx)
-	if err != nil {
-		return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q.", r.File, r.Target)
+
```
rynowak, line 174: This is looking a lot better overall 👍 

This code will still need to handle the case where the error is NOT `*errcode.ErrorResponse`. Otherwise cases like a network failure will be reported as success.

You might also want to refactor the error handling for `httpErr` into it's own function for readability.

It might look something like this:

```go
err = r.publish(ctx)
var httpErr *errcode.ErrorResponse
if errors.As(err, &httpErr) {
  return handleErrorResponse(httpErr)
} else if err != nil {
  return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q.", r.File, r.Target)
}
```

```
func handleErrorResponse(err *errcode.ErrorResponse) error {
... // Your current code goes here.
}
```

pkg/cli/cmd/bicep/publish/publish.go
```diff
@@ -167,8 +171,30 @@ func (r *Runner) Run(ctx context.Context) error {
...
+			} else {
+				return clierrors.MessageWithCause(err, "Unauthorized: Please login to %q\n%q\nError: %q", httpErr.URL.Host, acrInfo, r.Target)
+			}
+		case http.StatusForbidden:
+			return clierrors.MessageWithCause(err, "Forbidden: You don't have permission to push to %q\nError: %q", httpErr.URL.Host, r.Target)
+		case http.StatusNotFound:
+			return clierrors.MessageWithCause(err, "Not Found: Unable to find registry %q\nError: %q", httpErr.URL.Host, r.Target)
+		default:
+			return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q", r.File, r.Target)
+		}
```
rynowak, line 197: I tried this out, and this is what it looks for for ACR:

image

As discussed [here](https://github.com/radius-project/radius/issues/6297#issuecomment-1866873796) I think the error message should clearly explain what's failing. I'd suggest bringing back the `Failed to publish Bicep file %q to %q.` as the first line. 

I think you can also remove `Error: %q`. `clierrors.MessageWithCause` will already nicely format the error you give it.

rynowak (changes requested):
This is much improved and on the right track overall. Left some feedback about what the output looks like.

The general error case also needs to be handled before I can merge. Thanks for your help so far 👍

pkg/cli/cmd/bicep/publish/publish.go
gpltaylor: Changed made as requested. 

> This code will still need to handle the case where the error is NOT *errcode.ErrorResponse.
OMG 💯

pkg/cli/cmd/bicep/publish/publish.go
gpltaylor: > I think you can also remove Error: %q. clierrors.MessageWithCause
Done

> As discussed https://github.com/radius-project/radius/issues/6297#issuecomment-1866873796 I think the error message should clearly explain what's failing

You've said this a few times in comments now however, I believe this is done now. 
If I am missing the point please be blunt :)

gpltaylor: @rynowak When refactoring code into a function, should I introduce unit testing for that function? Does a test here, add value?

pkg/cli/cmd/bicep/publish/publish.go
```diff
@@ -167,14 +171,40 @@ func (r *Runner) Run(ctx context.Context) error {
 	r.Destination = dest
 
 	err = r.publish(ctx)
-	if err != nil {
-		return clierrors.MessageWithCause(err, "Failed to publish Bicep file %q to %q.", r.File, r.Target)
+	var httpErr *errcode.ErrorResponse
+	if errors.As(err, &httpErr) {
+		message := fmt.Sprintf("Failed to publish Bicep file %s to %s", r.File, r.Target)
```
rynowak, line 176:
```suggestion
		message := fmt.Sprintf("Failed to publish Bicep file %q to %q", r.File, r.Target)
```

This should use quotes to surround the user-specified details like the filename and destination.

pkg/cli/cmd/bicep/publish/publish.go
```diff
@@ -167,14 +171,40 @@ func (r *Runner) Run(ctx context.Context) error {
...
 	}
 
 	r.Output.LogInfo("Successfully published Bicep file %q to %q", r.File, r.Target)
 	return nil
 }
 
+func handleErrorResponse(httpErr *errcode.ErrorResponse, message string) error {
+	acrInfo := ""
+	if strings.Contains(httpErr.URL.Host, "azurecr.io") {
+		acrInfo = "For more details visit: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli"
```
rynowak, line 189: This is looking much better, here's what it looked like for me when I tried it out.

image

I'd suggest removing the quotes around `message` and `acrInfo` (replace `%q` with `%s`). These are user-facing explanations and it's confusing that they are surrounded with an extra layer of quotes.

rynowak:
> polite knock knock @rynowak PS Is there a process for PR change notification - aka code updated following a first PR? This is not covered here: https://github.com/radius-project/radius/tree/main/docs/contributing/contributing-pull-requests

Apologies, it's the holidays here so I'm a little bit slower, you did the right thing by sending a message.

pkg/cli/cmd/bicep/publish/publish.go
rynowak: > If I am missing the point please be blunt :)

I think this has been addressed now 👍 

> When refactoring code into a function, should I introduce unit testing for that function? Does a test here, add value?

In general yes it's a good thing, the value is that we'll be able to make future changes with confidence.

Give it a try, and let me know if you get stuck or want advice.

gpltaylor:
Happy New Year @rynowak

rynowak:
FYI - since this is your first pull-request to Radius it's asking me to approve actions runs. I've approved these so you can see the build/tests running when you push changes.

The functional tests require approval as well, but there should no be impact on those from your changes. I'll approve that as the last step when this is ready to merge.

pkg/cli/cmd/bicep/publish/publish.go
gpltaylor: I will add some tests and update the PR

pkg/cli/clierrors/types.go
```diff
@@ -42,7 +44,7 @@ func (e *ErrorMessage) Error() string {
 		return e.Message
 	}
 
-	return e.Message + " Cause: " + e.Cause.Error() + "."
+	return e.Message + " Cause: " + strings.TrimSpace(e.Cause.Error()) + "."
```
rynowak, line 47: 👍

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+type MockErrorResponse struct {
+	Method     string
+	URL        *url.URL
+	StatusCode int
+}
+
+// Error implements error.
+func (e *MockErrorResponse) Error() string {
+	// panic("unimplemented")
+	return e.Method
```
rynowak, line 330: Looks like this should be `return e.Error` and you can remove the commented-out code.

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+func (e *MockErrorResponse) GetURL() *url.URL {
+	return e.URL
+}
+
+func (e *MockErrorResponse) GetStatusCode() int {
+	return e.StatusCode
+}
+
+func TestHandleErrorResponse(t *testing.T) {
+	// Test case 1: Unauthorized with ACR info
```
rynowak, line 346: Use [subtests](https://go.dev/blog/subtests) for each of these. Go has a feature for this :)

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+	}
+	message3 := "Failure reason C"
+	expectedError3 := fmt.Sprintf("%s\nForbidden: You don't have permission to push to %q Cause: %q: response status code 403: Forbidden.", message3, httpErr3.URL.Host, "//"+httpErr3.URL.Host)
+
+	// Test case 4: Not Found
+	httpErr4 := &MockErrorResponse{
+		URL:        &url.URL{Host: "myregistry.azurecr.io"},
+		StatusCode: http.StatusNotFound,
+	}
+	message4 := "Faure reason D"
```
rynowak, line 377: Is this meant to say `Failure` instead of `Faure`?

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+		URL:        &url.URL{Host: "myregistry.azurecr.io"},
+		StatusCode: http.StatusInternalServerError,
+	}
+	message5 := "Faure reason E"
+	expectedError5 := fmt.Sprintf("%s Cause: %q: response status code 500: Internal Server Error.", message5, "//"+httpErr5.URL.Host)
+
+	testCases := []struct {
+		httpErr       *MockErrorResponse
+		message       string
+		expectedError string
```
rynowak, line 391: I'd suggest adding a description or name field here. Your comments like `// Test case 5: Other status code` can just become the test name.

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+	}{
+		{httpErr1, message1, expectedError1},
+		{httpErr2, message2, expectedError2},
+		{httpErr3, message3, expectedError3},
+		{httpErr4, message4, expectedError4},
+		{httpErr5, message5, expectedError5},
```
rynowak, line 397: I think this could probably more readable by just moving all of the data inline. 

```go
// Case
{
  name: "...",
  httpErr: &MockErrorResponse{
		URL:        &url.URL{Host: "myregistry.azurecr.io"},
		StatusCode: http.StatusInternalServerError,
	},
  message: "Failure reason5",
  expectedErr: fmt.Sprintf(...),
}
```

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.message, func(t *testing.T) {
+			httpErr := &errcode.ErrorResponse{
+				Method:     tc.httpErr.GetMethod(),
+				URL:        tc.httpErr.GetURL(),
+				StatusCode: tc.httpErr.GetStatusCode(),
```
rynowak, line 405: Is there a reason to use `MockErrorResponse` at all? It looks like we can just use `&errcode.ErrorResponse` directly.

pkg/cli/cmd/bicep/publish/publish_test.go
```diff
@@ -313,3 +317,97 @@ func TestRunner_Validate(t *testing.T) {
...
+			}
+			result := handleErrorResponse(httpErr, tc.message)
+			if result.Error() != tc.expectedError {
+				t.Errorf("Expected error: %s, but got: %s", tc.expectedError, result)
```
rynowak, line 409: Use `require.Equal()`

rynowak (changes requested):
Awesome 👍 thanks for adding tests. Left a few more comments.

gpltaylor:
@rynowak latest updates to the unit tests.

Let me know if there is anything else... Thank you so much for the feedback so far!

rynowak (approved):
Awesome! Thanks @gpltaylor

rynowak:
Merged, thanks @gpltaylor !

PR #6975, opened 2023-12-28
chore: exclude version.yaml from functional tests

lechnerc77:
# Description

This PR contains a proposal on how to avoid unecessary runs of the functional tests if the `versions.yaml` is updated.
Due to the triggers of the functional tests an exclusion of this file is possible if it is triggered by the completion of the workflow `Approve Functional Tests`. The exclusion of the file path is therefore implemented in the workflow definition `functional-tests-approval.yaml`. 

The regular scheduling of the functional tests as well as the external trigger are not impacted by the change

## Type of change

Fixes: #6261 

## Auto-generated summary

copilot:all

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -14,646 +14,646 @@
 # limitations under the License.
 # ------------------------------------------------------------
 
-name: Functional tests
-on:
-  schedule:
-    # Run every 4 hours on weekdays.
-    - cron: "30 0,4,8,12,16,20 * * 1-5"
-    # Run every 12 hours on weekends.
-    - cron: "30 0,12 * * 0,6"
-  # Dispatch on external events
-  repository_dispatch:
-    types: [de-functional-test]
-  workflow_run:
-    workflows: ['Approve Functional Tests']
-    types:
-      - completed
-
-env:
-  # Go version
-  GOVER: '^1.21'
-  GOPROXY: https://proxy.golang.org
-
-  # gotestsum version - see: https://github.com/gotestyourself/gotestsum
-  GOTESTSUM_VER: 1.10.0
-
-  # Helm version
-  HELM_VER: 'v3.12.0'
-  # KinD cluster version
-  KIND_VER: 'v0.20.0'
-  # Dapr version
-  DAPR_VER: '1.12.0'
-  DAPR_DASHBOARD_VER: '0.14.0'
-  # Kubectl version
-  KUBECTL_VER: 'v1.25.0'
-  # Azure Keyvault CSI driver chart version
-  AZURE_KEYVAULT_CSI_DRIVER_VER: '1.4.2'
-  # Azure workload identity webhook chart version
-  AZURE_WORKLOAD_IDENTITY_WEBHOOK_VER: '1.1.0'
-  # Container registry for storing container images
-  CONTAINER_REGISTRY: ghcr.io/radius-project/dev
-  # Container registry for storing Bicep recipe artifacts
-  BICEP_RECIPE_REGISTRY: ghcr.io/radius-project/dev
-  # The radius functional test timeout
-  FUNCTIONALTEST_TIMEOUT: 60m
-  # The Azure Location to store test resources
-  AZURE_LOCATION: westus3
-  # The base directory for storing test logs
-  RADIUS_CONTAINER_LOG_BASE: dist/container_logs
-  # The Radius helm chart location.
-  RADIUS_CHART_LOCATION: deploy/Chart/
-  # The region for AWS resources
-  AWS_REGION: 'us-west-2'
-  # The AWS account ID
-  AWS_ACCOUNT_ID: '${{ secrets.FUNCTEST_AWS_ACCOUNT_ID }}'
-  # The current GitHub action link
-  ACTION_LINK: '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}'
-  # Server where terraform test modules are deployed
-  TF_RECIPE_MODULE_SERVER_URL: "http://tf-module-server.radius-test-tf-module-server.svc.cluster.local"
-
-jobs:
-  build:
-    name: Build Radius for test
-    runs-on: ubuntu-latest
-    if: github.event_name == 'repository_dispatch' || (github.event_name == 'schedule' && github.repository == 'radius-project/radius') || github.event_name == 'workflow_run'
-    env:
-      DE_IMAGE: 'ghcr.io/radius-project/deployment-engine'
-      DE_TAG: 'latest'
-      FUNCTIONAL_TEST_APP_ID: 425843
-    outputs:
-      REL_VERSION: ${{ steps.gen-id.outputs.REL_VERSION }}
-      UNIQUE_ID: ${{ steps.gen-id.outputs.UNIQUE_ID }}
-      PR_NUMBER: ${{ steps.gen-id.outputs.PR_NUMBER }}
-      CHECKOUT_REPO: ${{ steps.gen-id.outputs.CHECKOUT_REPO }}
-      CHECKOUT_REF: ${{ steps.gen-id.outputs.CHECKOUT_REF }}
-      RAD_CLI_ARTIFACT_NAME: ${{ steps.gen-id.outputs.RAD_CLI_ARTIFACT_NAME }}
-      DE_IMAGE: ${{ steps.gen-id.outputs.DE_IMAGE }}
-      DE_TAG: ${{ steps.gen-id.outputs.DE_TAG }}
-    steps:
-      - name: Login as the GitHub App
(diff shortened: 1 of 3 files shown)


rynowak:
Hi @lechnerc77 - thanks for trying to address this. I'm wondering if this is going to have the desired effect.

There are two "check runs" associated with functional tests for each PR:

- `Approve functional test run`
- `Functional test run`

The latter one of these (`Functional test run`) is marked as a required.

This is relevant because after your fix `Approve functional test run` is going to skip, and `Functional test run` will never start. Since `Functional test run` never starts the PR will be stuck in an unmergable state.

We can't allow contributors to directly run `Functional test run` before a code review takes place because it needs access to credentials. 

I am admittedly not an expert on this and so I don't know what the solution is, I just know all of the things we've tried that don't work. GitHub don't have great features for projects that **both** require access to credentials to run tests, and support a wide group of untrusted contributes.

lechnerc77:
@rynowak Thanks for the aditional input. I missed that the functional tests step is a required step. A bit of a more sophisticated solution is needed then - I will work on it and update this PR accordingly so that the current flow is not interrupted but the functional tests are not executed if the only updated file in the PR is the `versions.yaml`.

rynowak:
> @rynowak Thanks for the aditional input. I missed that the functional tests step is a required step. A bit of a more sophisticated solution is needed then - I will work on it and update this PR accordingly so that the current flow is not interrupted but the functional tests are not executed if the only updated file in the PR is the `versions.yaml`.

@lechnerc77 thanks for looking into. Just be forewarned that the solution might be more complicated than is worth pursuing. I'm not sure the 'good first issue' is right for this issue.

We have a use Actions in a pretty complex way to be able to run the functional tests without risk of exposing our credentials to malicious actors. Unfortunately we haven't found a way to simplify this.

lechnerc77:
Rewriting of GH Actions for functional tests based on feedback by @rynowak . The changes reflect the main point that the functional tests workflow is a required workflow (branch protection rule) but also is skipped in case that only the `versions.yaml` file is changed. To fulfill these points the PR introduces the following changes:

## functional-tests-approval.yaml

The workflow contains a second job that: 

- determines the changes that are part of the PR
-evaluates if the only file that changed was the `versions.yaml` file

After the analysis of the changes in the PR the job creates a new dispatch event of type `functional-tests-approved` that contains the information if only the `versions.yaml` was changes in the client payload. 

**Open question**: to create the event a GH token with the scope repository is required. I used the `secrets.GH_RAD_CI_BOT_PAT` but not sure if this is correct and desired.

## functional-test.yaml

I made the following:  

- the GH Action was changed so that it reacts to the new event `functional-tests-approved` and no longer reacts on a successful execution of the ` functional-tests-approval` workflow.
- a new preprocessing job is introduced that checks based on the triggers and the payload of the trigger event if the execution of the checks can be skipped and consolidates the types of triggers. This information is made available as output of the job.
- The `if` conditions of the following jobs and steps are adjusted to work with the output of the `preprocessing` job.

In addition, the following cleanup happened:

- The code block that reacted on pull requests was removed, as this trigger does not exist for the workflow
- The GH Actions using `actions/github-script` have been updated to version 7 to contain the latest node libraries for setting the output and variables
- The code blocks that used the `fs.appendFileSync` have been transferred to `core.exportVariable`

rynowak:
@lechnerc77 - sorry for the delay in getting to this, I was traveling for work. 

@vinayada1 - can you take a look at this?

lechnerc77:
@vinayada1 thanks for your feedback. Your workflows and their interplay are indeed already quite sophisticated (and show some points GitHub might need to address as features or in best practices).

Concerning the passing of data between the workflows: in general, you have two options:

1. You store the intermediate data as described in the reference and chain your workflows based on default events of GitHub.  
2. You use custom events and add data for the receiver in the vent payload.

Both things have their advantages and their tradeoffs. I prefer option 2 when it comes to scenarios like the one, we have here, namely a chaining of workflows with a very limited amount of data that needs to be transferred. 
I would choose option 1 if the data to transfer is a larger/more complex file that needs to be passed between workflows or jobs like build artifacts. I think the documentation [About Workflow Artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#about-workflow-artifacts) contains good examples for this type of data. Using this option, you must make sure that the "right" artifact is fetched from the previous run and that things get cleaned up afterwards.   
Concerning the replay of the event (custom or standard), if something goes wrong: I do not see any big difference that makes one of the options the easier one to maintain. In any case manual calls of the GitHub APIs are needed. 

I do not think that the complexity will be decrease with either option because the chaining of workflows and the propagation of data makes things complex, but you need this setup as far as I have seen.

Long story short: I personally would use custom events with the corresponding event payload in scenarios like the one we have here. I would also pass the PR number as payload and not as an uploaded artifact. But (and this is a big but) I am not maintainer of the repository, so the decision should be based on what you feel most comfortable with as maintainer team. So just let me know in which direction the fix should go.

vinayada1:
Thanks for your inputs and comparing the two options clearly. I was trying to see if we can simplify this further. Let's continue discussion on the approach first and agree upon one before you proceed to make code changes for it. Another way to do this could be:-
1. Run the .github/workflows/functional-tests-approval.yaml with the filter:
    paths-ignore:
       - 'versions.yaml'
     This will cause this workflow to not run when only versions.yaml is changed
 2. However, doing this alone will never satisfy the functional tests run check.
 3. For that, add another workflow (e.g. skip-functional-tests.yaml) which runs only when the versions.yaml alone is changed and add this when the workflow runs successfully, it also executes this code:-
     - uses: LouisBrunner/checks-action@v1.6.1
      if: always()
      with:
        token: ${{ steps.get_installation_token.outputs.token }}
        name: 'Functional Test Run'
        repo: ${{ github.repository }}
        sha: ${{ env.CHECKOUT_REF }}
        status: completed
        conclusion: ${{ job.status }}
        output: |
          {"summary":"Functional Test run not required as only versions.yaml has changed."}
    This should mark the tests run check as complete

     For the second workflow to run, we could add a paths filter:-
     paths:
      - 'versions.yaml' and will also possibly need the code you have to check if that is the only change.

What are your thoughts on this?

lechnerc77:
here some comments on the flow:

ad 1: This works, but as you said, we then need an additional workflow to handle. 

ad 2: Second workflow would need to be implemented like you describe:
 - Check if `versions.yaml` is the only file that changed
 - If yes, we could leverage a combination of “paths” and the “tj-actions/changed-files”. I saw the later but decided not to bring in this dependency. I have no experience with the GitHub API endpoint for calling check actions. This should work, but at least I would need to try it out. I would however use the GitHub API per se and not the action as it seems to be not maintained properly. I anyway wrote a question in that below.

General questions: 

- Do you have any requirements concerning bringing in 3rd part GH Actions into the project? The reason why I am asking is that while some workflows look good, they are not properly maintained. The `LouisBrunner/checks-action` didn't have a new release since mid of last year. As a JavaScript Action I would personally expect more releases due to dependency updates. 
- I think no matter which direction this will be implemented, we need a good documentation about the used workflows, what they do, and why they are implemented this way. Both solutions came with some complexity in a sense that a contributor does not get the relations between the flows at a first glance. What do you think?

Overall, I would say that is a valid solution. I think the advantage is that the single workflows stay simple (or at least do not get more complex), the downside is that the connection between the workflows is not fully transparent for a new contributor. As written above, I think this requirement isn’t an easy one, so the solution will be complex, the question is which is the tradeoff that you want to choose.

vinayada1:
@lechnerc77 - Thanks for the detailed analysis.
1. I have discussed this internally as well and would prefer to keep the individual workflows simple. So adding a second workflow seems to be a better option. Like you suggested, we should also add documentation to explain what we are doing and how these workflows pass data and interact with one another.
2. You are absolutely right that whatever we do here, it's still complex and I believe these should be common requirements for many projects. I will file a request with github team to see hopefully they can simplify this for us.
3. As regards to bringing in github actions, we are okay using 3rd party actions if it simplifies the workflow for us.

Let me know if this clarifies things for you to move ahead.

lechnerc77:
@vinayada1 Thanks for the answers and the clarification with the team. I know what to do now and I will adjust this PR accordingly. Might take a bit, so the updated PR will probably be available next week

lechnerc77:
@vinayada1 I changed the setup as discussed. However, I cannot really test it in a simplified setup and I do not have experience with the `check run` API of GitHub. So basically I do not know if this *new* setup solves the requirement. 

If not I would close the PR tbh, as reproducing the setup outside of the repo is (at least from my perspective) not straightforward for an external contributor.

vinayada1:
Thanks for making these changes. I agree that it is not very easy to test these changes on your own. As you suggested, I will close this PR for now and use this code to test with my private repo. I will cc you on the new PR that I create.

lechnerc77:
@vinayada1 Makes perfect sense. Very interested to see how the solution looks like in the end!

PR #6991, opened 2024-01-04
Add GitHub action to check PR for formatted title

sylvainsf:
# Description

Checks incoming PRs for nicely formatted title.

## Type of change

## Auto-generated summary

copilot:all


.github/actions/check-pr-format/action.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          const title = context.payload.pull_request.title;
+          const regex = /\[(Add|Remove|Update|Fix)\]\s*.+/;
+          const regexNoBrackets = /(Add|Remove|Update|Fix)\s*.+/i;
+          const regexNoSpace = /\[(Add|Remove|Update|Fix)\].+/i;
+          const regexLowerCase = /\[(add|remove|update|fix)\]\s*.+/i;
+          const issue_number = context.payload.pull_request.number;
+          const owner = context.repo.owner;
+          const repo = context.repo.repo;
+          let body = '';
```
ytimocin, line 23: This can be defined within the first if. We don't use it outside of the if cases below.

.github/actions/check-pr-format/action.yml
```diff
@@ -0,0 +1,38 @@
...
+
+          if (!regex.test(title)) {
+            if (regexNoBrackets.test(title)) {
+              body = 'Your PR title is missing square brackets. Please update it to: [Add | Remove | Update | Fix] <Title>. For example: [Add] New feature';
```
shalabhms, line 27: square bracket is not needed. Just the verb.

.github/actions/check-pr-format/action.yml
sylvainsf: I'm not sure what you mean, if it's defined in the first if then it's not available in the other cases. Maybe I'm misunderstanding your comment?

ytimocin: Apologies for not being very clear. I couldn't see body being used outside of the if case. It can be moved inside the first if I think. Like this:
```
if (!regex.test(title)) {
            let body = '';
            if (regexNoSpace.test(title)) {
              body = 'Your PR title is missing a space after the keyword. Please update it to: [Add | Remove | Update | Fix] <Title>. For example: Remove Old feature';
            } else if (regexLowerCase.test(title)) {
              body = 'The keyword in your PR title is not capitalized. Please update it to: [Add | Remove | Update | Fix] <Title>. For example: Update Existing feature';
            } else {
              body = 'Your PR does not have a [Add | Remove | Update | Fix] <Title>. Please add one to help streamline our changelog. For example: Fix Bug in feature';
            }
            github.issues.createComment({ issue_number, owner, repo, body });
}
```

sylvainsf: That's true, though I have a habit of declaring all variables outside of conditionals as it reduces cognitive load to declare everything you're going to work with ahead of the logic.

shalabhms (dismissed):

.github/actions/check-pr-format/action.yml
```diff
@@ -0,0 +1,35 @@
...
+
+          if (!regex.test(title)) {
+            if (regexNoSpace.test(title)) {
+              body = 'Your PR title is missing a space after the keyword. Please update it to: [Add | Remove | Update | Fix] <Title>. For example: Remove Old feature';
+            } else if (regexLowerCase.test(title)) {
+              body = 'The keyword in your PR title is not capitalized. Please update it to: [Add | Remove | Update | Fix] <Title>. For example: Update Existing feature';
+            } else {
+              body = 'Your PR does not have a [Add | Remove | Update | Fix] <Title>. Please add one to help streamline our changelog. For example: Fix Bug in feature';
+            }
+            github.issues.createComment({ issue_number, owner, repo, body });
```
AaronCrawfis, line 32: In addition to adding a comment can we also fail this action? If the action still succeeds we probably can't count on a comment enforcing this behavior.

sylvainsf: I am of the philosophy to warn first and move to blocking if warning doesn't yield results. Several engineers also expressed concern at blocks causing friction, we can always add blocking in the near future if the comments alone don't work.

AaronCrawfis:
Can we also update the contribution documentation to note that PR titles should follow a specific format now? If we want to enforce consistent titles we should make sure we document it.

.github/actions/check-pr-format/action.yml
AaronCrawfis: If we're not going to block on the title, this is a recommendation and not a requirement. If we want consistency in the PR titles, this should be a requirement. If it's a requirement, what's the ask for reviewers? Ignore the title and approve anyway? If it's anything else, where the reviewer asks the contributor to update the title, then this needs to be a requirement.

kachawla: I agree with making it a requirement (can be in a quick follow up PR if we want to test this first). I'd also prefer if we expand the keywords to allow verb<ing> (adding/removing) which is very commonly used by many developers. 

Unrelated to this PR - can this be used to enforce on commit messages as well in addition to PR titles?

kachawla (approved):

sylvainsf:
@AaronCrawfis Yes, I'll add that when I add this bot to the documents repo.

AaronCrawfis:
> Yes, I'll add that when I add this bot to the documents repo.

We should update this repo as well here: https://github.com/radius-project/radius/tree/main/docs/contributing/contributing-pull-requests

PR #6995, opened 2024-01-05
Fix release race condition

### LINKED ISSUE #6874: Race condition in release process
Today, if you follow the release process outlined [here](https://github.com/radius-project/radius/tree/main/docs/contributing/contributing-releases), there will be an error in the "Publish GitHub release" step. This is because the workflow tries to get the release notes from the release branch, which they are not present in when they are merged to main (merges to main happen first). 

To fix this, we should:
1. Remove the trigger to publish the GitHub release from `main` if a release branch already exists.
2. Add a trigger to publish the GitHub release from the release branch.

This way, when the `versions.yaml` update is merged into the release branch (along with the release notes), the GitHub release publishing step will succeed.

[AB#10582](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10582)

willdavsmith:
# Description

This PR fixes a race condition in the release process. If we merge the release notes into main before the release branch, then the "Publish GitHub release" step will fail. This PR should fix that by:
1. adding a trigger to start the release (create the tags) on pushes to `versions.yaml` on `release/*` branches.
2. checking if the release branch exists, and if it does and the workflow is running on a push to main, then stop the workflow. This defers the tag creation to the release branch (which happens after we cherry pick and merge the release notes into the release branch).

## Type of change

Fixes: #6874


.github/workflows/release.yaml
```diff
@@ -163,7 +165,30 @@ jobs:
...
+            if (data.name == '${{ steps.get-version.outputs.release-branch-name }}') {
+              return 'true'
+            } else {
+              return 'false'
+            }
+      - name: Stop workflow if release branch exists and the event is a push to main
+        if: steps.release-branch-exists == 'true' && github.ref == 'refs/heads/main' && github.event_name == 'push'
+        run: |
+          echo "Release branch ${{ steps.get-version.outputs.release-branch-name }} already exists. Stop the workflow."
+          echo "STOPPED=true" >> $GITHUB_ENV
```
lakshmimsft, line 189: nit: consider renaming variable to something more descriptive IS_WORKFLOW_STOPPED / WORKFLOW_STOPPED

.github/workflows/release.yaml
```diff
@@ -172,18 +197,18 @@ jobs:
...
         run: |
           ./radius/.github/scripts/release-create-tag-and-branch.sh bicep ${{ steps.get-version.outputs.release-version }} ${{ steps.get-version.outputs.release-branch-name }}
-        if: success()
       - name: Release radius-project/deployment-engine version ${{ steps.get-version.outputs.release-version }}
+        if: success() && env.STOPPED == 'false'
         run: |
           ./radius/.github/scripts/release-create-tag-and-branch.sh deployment-engine ${{ steps.get-version.outputs.release-version }} ${{ steps.get-version.outputs.release-branch-name }}
-        if: success()
       - name: Release radius-project/recipes version ${{ steps.get-version.outputs.release-version }}
+        if: success() && env.STOPPED == 'false'
```
lakshmimsft, line 212: Question: The only way we can test this is in the upcoming release?is that right?

willdavsmith: correct, unfortunately

willdavsmith: the good news is that if it fails during the release, we can just do this stuff manually (as we do today)

lakshmimsft (dismissed):

lakshmimsft:
needs rebase, functional test runs before merge

kachawla (dismissed):

.github/scripts/cleanup-cluster.sh
```diff
@@ -18,13 +18,18 @@
...
+echo "cleaning up cluster"
+
+# Delete all test resources in queuemessages.
+echo "delete all resources in queuemessages.ucp.dev"
 kubectl delete queuemessages.ucp.dev -n radius-system --all
 
+# Testing deletion of deployment.apps.
+
 # Delete all test resources in resources without proxy resource.
+echo "delete all resources in resources.ucp.dev"
```
ytimocin, line 30: This PR needs a rebase :)

willdavsmith: whoops, fixed :)

.github/workflows/release.yaml
```diff
@@ -163,7 +165,30 @@ jobs:
...
+            if (data.name == '${{ steps.get-version.outputs.release-branch-name }}') {
+              return 'true'
+            } else {
+              return 'false'
+            }
+      - name: Stop workflow if release branch exists and the event is a push to main
+        if: steps.release-branch-exists == 'true' && github.ref == 'refs/heads/main' && github.event_name == 'push'
+        run: |
+          echo "Release branch ${{ steps.get-version.outputs.release-branch-name }} already exists. Stop the workflow."
+          echo "IS_WORKFLOW_STOPPED=true" >> $GITHUB_ENV
```
ytimocin, line 189: Instead of **IS_WORKFLOW_STOPPED**, should we have something more meaningful in the context of having another release branch already?

willdavsmith: what kind of thing are you thinking?

ytimocin: I mean (and again please note that it is not that important) but it could be something like **ACTIVE_RELEASE_BRANCH**. `Workflow stopped` can mean something else. This is my opinion and would like to hear others' ideas too.

willdavsmith: looks like I can remove the env variable all together :) I updated the PR.

.github/workflows/release.yaml
```diff
@@ -163,7 +164,27 @@ jobs:
         id: get-version
         run: |
           ./radius/.github/scripts/release-get-version.sh ${{ steps.get-supported-versions.outputs.result }} radius
+      - name: Check if release branch exists
+        id: release-branch-exists
```
ytimocin, line 168: Could line179 `context.ref == 'refs/heads/main' && context.eventName == 'push'` be moved here?

ytimocin (approved):

.github/workflows/release.yaml
willdavsmith: I don't think so, because we need `steps.release-branch-exists.outputs.result` to be `'false'` for the rest of the workflow run to make sure the other steps are skipped

Issue #7007, opened 2024-01-10
`rad install kubernetes --reinstall` behavior documentation - discussion kick off


ytimocin:
## Engineering Improvement
This issue will be more like an explanation of how `rad install kubernetes --reinstall` works as of now. From here on, we can start a discussion if we would like to change this behavior.

Summary:
1. `rad install kubernetes --reinstall` command recreates **applications-rp**, **bicep-de**, **controller**, and **ucp** pods even though there is no change in the Helm charts. At the same time, the secrets associated with these pods are not recreated. Please see the first attached screenshot.
2. `rad install kubernetes --reinstall` command doesn't recreate **contour** pods. Probably because there isn't any change to the Helm charts of them.
3. If you delete **sh.helm.release.v1.contour.*** and/or **sh.helm.release.v1.radius.*** secrets (which probably act as a checkpoint for the deployments) and then run `rad install kubernetes --reinstall`, the pods do not get recreated but the deleted secrets get recreated.
4. If you delete all the secrets in the **radius-system** namespace and then run `rad install kubernetes --reinstall`, the pods do not get recreated but all secrets get recreated.

### Area for Improvement (required)
We can discuss which one of these scenarios are expected or not expected. We can also discuss what should happen instead of what is stated above.

### Observed behavior (required)
In the first scenario above, because the secrets are not updated, I see failures in the webhook and reconciler functional tests. I believe that the secret also needs to be created like the pods so that we don't see any failures. This is the reason why running functional tests as one time doesn't cause errors. But as soon as we introduce `rad install kubernetes --reinstall` into the picture like we do in the long-haul test cluster, we start to see some errors.

Another behavior I observed is that, recreation of the secrets doesn't affect the Pod and in return I don't see x509 errors. But once the pods get recreated and at the same time secrets don't get recreated, like a link gets broken between the secret object and the pod object, I start to see x509 errors both in controller pods and ucp pods.

I will add my findings as I do more tests.

### Desired behavior (required)
1. `rad install kubernetes --reinstall` should recreate all pods and all secrets (and probably all necessary resources). Or if there is no change to the Helm charts, like contour does, it shouldn't recreate anything at all.

### Proposed Fix (optional)
Here is a comment written a long time ago: https://github.com/radius-project/radius/blob/main/pkg/cli/helm/radiusclient.go#L97. Not sure if that might be related. We should debug further and see why pods are being recreated but related secrets are not.

### rad Version (required)
[...]

PR #7008, opened 2024-01-10
Updating the Radius installation step of the long-haul workflow

ytimocin:
# Description

* Added **Report Failure** step to the workflow.
* Removed the secret deletion workaround step.

Details can be found here: #7007 

## Type of change

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -86,6 +86,9 @@ env:
   # Radius test environment name
   RADIUS_TEST_ENVIRONMENT_NAME: 'kind-radius'
 
+  # The current GitHub action link
+  ACTION_LINK: '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}'
+
 jobs:
   build:
     name: Build Radius for test
@@ -388,10 +391,6 @@ jobs:
           rad env create ${{ env.RADIUS_TEST_ENVIRONMENT_NAME }} --namespace default
           rad env switch ${{ env.RADIUS_TEST_ENVIRONMENT_NAME }}
 
-          # Temporary workaround to fix the x509 certificate error in the controller.
-          # https://github.com/radius-project/radius/issues/6989
-          kubectl delete secrets controller-cert -n radius-system --ignore-not-found
-
           echo "*** Configuring Azure provider ***"
           rad env update ${{ env.RADIUS_TEST_ENVIRONMENT_NAME }} --azure-subscription-id ${{ secrets.INTEGRATION_TEST_SUBSCRIPTION_ID }} \
             --azure-resource-group ${{ env.AZURE_TEST_RESOURCE_GROUP }}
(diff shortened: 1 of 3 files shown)


.github/workflows/long-running-azure.yaml
```diff
@@ -349,6 +352,14 @@ jobs:
...
         env:
           RESOURCE_GROUP: ${{ env.AZURE_TEST_RESOURCE_GROUP }}
+      - name: Check if tests are already running
+        run: |
+          if kubectl get configmap long-running-test-lock -n radius-system --ignore-not-found; then
+            echo "Tests are already running. Exiting..."
+            exit 1
+          fi
+      - name: Set test run flag
+        run: kubectl create configmap long-running-test-lock -n radius-system
```
youngbupark, line 362: What if the clean up step in previous workflow run is failed? the current run will be ignored.

.github/workflows/long-running-azure.yaml
```diff
@@ -368,6 +379,12 @@ jobs:
           export PATH=$GITHUB_WORKSPACE/bin:$PATH
           which rad || { echo "cannot find rad"; exit 1; }
 
+          echo "*** Uninstalling existing Radius installation ***"
+          rad uninstall kubernetes
+
+          echo "*** Deleting radius-system namespace ***"
+          kubectl delete namespace radius-system --ignore-not-found
```
youngbupark, line 386: I do not think this is right approach. it is still workaround. upgrade scenario should work here.

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -32,7 +33,7 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
-    caBundle: {{ b64enc $ca.Cert }}
+    caBundle: {{ $validatingWebhookCaBundle }}
```
ytimocin, line 36: The missing piece was that we were always using a new caBundle here. If the secret stays the same and pod gets recreated, we shouldn't create a new caBundle for this. That creates all the failures of second time test runs.

Also, ucp does it this way: https://github.com/radius-project/radius/blob/main/deploy/Chart/templates/ucp/apiservice.yaml#L42.

cc/ @youngbupark @rynowak @lakshmimsft

youngbupark: Can you please open new PR for helm update? this is the actual fix for upgrading scenario. So it is better to create new pr for this.

deploy/Chart/templates/controller/validating-webhook-configuration.yaml
```diff
@@ -13,7 +13,6 @@ metadata:
   labels:
     app.kubernetes.io/name: controller
     app.kubernetes.io/part-of: radius
-type: kubernetes.io/tls
```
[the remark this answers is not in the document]
ytimocin: Created this issue: https://github.com/radius-project/radius/issues/7035.

lakshmimsft (approved):

lakshmimsft:
as discussed, secrets continue to be 'opaque' as they were earlier.

PR #7010, opened 2024-01-10
A first step to installing Radius on multi tenent clusters

### LINKED ISSUE #6691: Run Radius services and applications in single namespace

Today we require cluster-level permissions to run Radius, specifically because we install cluster-wide CRDs. This can cause issues for users who want to try Radius but only have access to a single namespace.

Quotes from users hitting this issue:

> Avoidance of cluster scope requirements is key to early adoption in large enterprises. Once there is traction we can make a case for breaking the “rules” but if we go in day one saying we need cluster scope for anything we are in for so many meetings that we forget the original goal. I mentioned earlier that we had a similar problem with Dapr. A workaround that would let us remain completely within a namespace would be very helpful for adoption.

> We have a multi-tenant cluster where every team gets a namespace. Within our namespace we can create projects which are namespaces with some additional annotations (name, display name, description etc). As a developer I can create projects  and manage pods within our namespace but I have to get approval to create any new namespaces and I can't change/add anything in the cluster scope.

Opening this feature to track the overall scenario of running Radius in a single namespace. The primary focus in on dev/test environments when learning and trying Radius, but it could potentially include running multiple Radius installations on the same cluster.

## Acceptance criteria

A user who only has access to a single namespace should be able to install Radius and deploy a Radius application. They shouldn't need any cluster-level permissions in order to initialize Radius or deploy a Radius application.

## Additional context

See https://discord.com/channels/1113519723347456110/1171852494016614491 for more information (you may need to first join https://aka.ms/Radius/Discord).

cpossinger:
# Description

This pull request:

- Allows the storage and queue configuration to be passed in helm update using a yaml file in the storage-config directory.
- Dynamic resource provider URLs based on release namespace
- Adds contour-values yaml file to deploy contour helm chart
- Updates documentation

## Type of change

Fixes: #6691

### DIFF
--- a/deploy/Chart/README.md
+++ b/deploy/Chart/README.md
@@ -7,12 +7,19 @@ The Radius helm chart deploys the Radius services on a Kubernetes cluster using 
 - Kubernetes cluster with RBAC enabled
 - Helm 3
 
-### Installing the Chart
+### Installing the Radius Chart
 
 To install the chart with the release name `radius`:
 
 ```console
 helm upgrade --wait --install radius deploy/Chart -n radius-system
+
+### Installing the Contour Chart
+
+To install the chart with the release name `contour`:
+
+helm repo add bitnami https://charts.bitnami.com/bitnami
+helm install contour bitnami/contour --namespace demo -f deploy/contour-values.yaml
 ```
 
 ### Verify the installation
@@ -25,10 +32,15 @@ kubectl get pods -n radius-system
 
 ### Uninstalling the Chart
 
+To uninstall/delete the `contour` deployment:
+
+helm uninstall contour -n radius-system
+
+
 To uninstall/delete the `radius` deployment:
 
 ```console
-helm delete radius
+helm uninstall radius -n radius-system
 ```
 
 The command removes all the Kubernetes components associated with the chart and deletes the release.
--- a/deploy/Chart/templates/controller/configmaps.yaml
+++ b/deploy/Chart/templates/controller/configmaps.yaml
@@ -12,21 +12,13 @@ data:
       host: "0.0.0.0"
       port: 9443
 
-    storageProvider:
-      provider: "apiserver"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
-
+    storageProvider: {{ toYaml .Values.global.storageProvider | nindent 6 }}
+      
     secretProvider:
       provider: kubernetes
 
-    queueProvider:
-      provider: "apiserver"
+    queueProvider: {{ toYaml .Values.global.queueProvider | nindent 6 }}
       name: "ucp"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
 
     profilerProvider:
       enabled: true
--- a/deploy/Chart/templates/rp/configmaps.yaml
+++ b/deploy/Chart/templates/rp/configmaps.yaml
@@ -13,17 +13,9 @@ data:
     environment:
       name: self-hosted
       roleLocation: "global"
-    storageProvider:
-      provider: "apiserver"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
-    queueProvider:
-      provider: "apiserver"
+    storageProvider: {{ toYaml .Values.global.storageProvider | nindent 6 }}
+    queueProvider: {{ toYaml .Values.global.queueProvider | nindent 6 }} 
       name: "radius"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
     metricsProvider:
       prometheus:
         enabled: true
--- a/deploy/Chart/values.yaml
+++ b/deploy/Chart/values.yaml
@@ -10,8 +10,6 @@ global:
     # Dotnet runtime and Go use this environment variable to load the root CA.
     sslCertDirEnvVar: "SSL_CERT_DIR"
 
-  
-
   prometheus:
     enabled: true
     path: "/metrics"
--- a/deploy/Chart/values.yaml
+++ b/deploy/Chart/values.yaml
@@ -10,6 +10,8 @@ global:
     # Dotnet runtime and Go use this environment variable to load the root CA.
     sslCertDirEnvVar: "SSL_CERT_DIR"
 
+  
+
   prometheus:
     enabled: true
     path: "/metrics"
--- a/deploy/Chart/storage-config/apiserver.yaml
+++ b/deploy/Chart/storage-config/apiserver.yaml
@@ -1,0 +1,13 @@
+global:
+  storageProvider:
+    provider: "apiserver"
+    apiserver:
+      context: ""
+      namespace: "radius-system"
+  queueProvider:
+      provider: "apiserver"
+      apiserver:
+        context: ""
+        namespace: "radius-system" 
+
+  
--- a/deploy/Chart/storage-config/inmemory.yaml
+++ b/deploy/Chart/storage-config/inmemory.yaml
@@ -1,0 +1,8 @@
+global:
+  storageProvider:
+    provider: "etcd"
+    etcd:
+      inmemory: true
+  queueProvider:
+    provider: inmemory
+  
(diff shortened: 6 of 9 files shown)


deploy/Chart/values.yaml
```diff
@@ -10,6 +10,8 @@ global:
     # Dotnet runtime and Go use this environment variable to load the root CA.
     sslCertDirEnvVar: "SSL_CERT_DIR"
 
+  
+
```
ytimocin, line 14: nit: extra lines can be removed.

deploy/Chart/README.md
```diff
@@ -7,12 +7,19 @@ The Radius helm chart deploys the Radius services on a Kubernetes cluster using
...
 
 ```console
 helm upgrade --wait --install radius deploy/Chart -n radius-system
+
+### Installing the Contour Chart
+
+To install the chart with the release name `contour`:
+
+helm repo add bitnami https://charts.bitnami.com/bitnami
+helm install contour bitnami/contour --namespace demo -f deploy/contour-values.yaml
```
ytimocin, line 22: Just for my own understanding, could you explain why we need this? `rad install kubernetes` already installs this. Are there other scenarios where we would like to install **contour** separately? I guess on multi-tenent clusters, we have to use manual installations instead of `rad install kubernetes`.

deploy/Chart/templates/controller/configmaps.yaml
```diff
@@ -12,21 +12,13 @@ data:
       host: "0.0.0.0"
       port: 9443
 
-    storageProvider:
-      provider: "apiserver"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
-
+    storageProvider: {{ toYaml .Values.global.storageProvider | nindent 6 }}
```
ytimocin, line 15: Is the reason why we start with **storageProvider** and **queueProvider** that they are prerequisites for a Radius installation? We also have **metricsProvider** in the config but metrics service is not required for Radius to stand up. Is that the reason why we started with **storageProvider** and **queueProvider**? Could you give some more information on that?

deploy/Chart/README.md
cpossinger: My organization uses a multi-tenant cluster where I only have access to a single namespace and I have no cluster scoped permissions. `rad install kubernetes` has no way to change the namespace from 'radius-system' or opt out of CRDs that need cluster-scoped permissions to define. This would not be for production but for people who want to try radius out or get buy in for approval.

deploy/Chart/templates/controller/configmaps.yaml
cpossinger: storageProvider and queueProvider both use CRDs which requires cluster-scoped permissions. This allows users to opt out of those definitions if they want.

youngbupark:
@cpossinger Thanks for your contribution. I have some recommendations here.

* [ ] Make contour as a subchart of radius chart. Please refer to [this prometheus chart](https://github.com/prometheus-community/helm-charts/blob/a5a6939767203a2eac89a6ba6056534225179247/charts/prometheus/Chart.yaml#L27C1-L43C46) how it adds its dependent as subchart. 

For example, we can update the existing radius Chart.yaml like below.
```yaml
apiVersion: v2
description: Radius provides a declarative way to manage your infrastructure and applications.
name: radius
# version is the chart version, which will be replaced when this chart is packaged.
version: '0.42.42-dev'
# appVersion is the version of Radius, which will be replaced when this chart is packaged.
appVersion: 'edge'
home: https://radapp.io
sources:
  - https://github.com/radius-project/radius/tree/main/deploy/Chart
maintainers:
  - name: Radius Authors
    url: https://github.com/radius-project/radius
icon: https://radapp.io/images/logo.png
keywords:
  - radius
  - iac
  - bicep
dependencies:
  - name: contour
    version: "15.0.*"
    repository: https://charts.bitnami.com/bitnami/contour
    condition: contour.enabled
```

* [ ] Move contour configuration values to radius chart values as subchart configuration. Please see [this prometheus example](https://github.com/prometheus-community/helm-charts/blob/a5a6939767203a2eac89a6ba6056534225179247/charts/prometheus/values.yaml#L1207C1-L1219C19)

For example: in values.yaml

```
// Existing radius values

contour:
   enabled: false // let's keep it false to not install contour by default
   // contour specific change   
```

Now, user can install contour while installing radius with following command.
```bash
helm upgrade --wait --install radius deploy/Chart -n radius-system --set contour.enabled=true
```

* [ ] For the first iteration, do not allow user to configure the entire `queueProvider` and `storageProvider`  object. To achieve your goal, you need to set only hardcoded namespace property to `{{ .Release.Namespace }}`. If you want to make the entire provider values configurable, please create the separate pr after completing this PR.  Please keep the PR short. 

* [ ] Once this PR is merged, let's have new PR to migrate manual contour install code to use new radius subchart. With this, user can decide whether to install contour as part of `rad install` or not.

For example, 
```bash
rad install kubernetes --set contour.enabled=false
```

cc/ @rynowak

deploy/Chart/templates/controller/configmaps.yaml
youngbupark: In this iteration, let's not make all provider values configurable. To achieve your goal, it is enough to change only namespace property like below. Let's keep it simple. If you want to make entire object configurable, Please create separate PR later once this pr is merged.

```yaml
      provider: "apiserver"
      apiserver:
        context: ""
        namespace: {{ .Release.Namespace }}
```

deploy/Chart/README.md
rynowak: It's good that you're updating this. Also the docs for chart that are on our website are here: https://github.com/radius-project/docs/blob/v0.29/docs/content/guides/operations/kubernetes/kubernetes-install/index.md

rynowak: I think we're also open to fixing any bugs that come up when contour is not installed. We want to make it an optional component.

youngbupark:
@cpossinger Can you update this PR? Please let me know if you do not have a time. We will prioritize this on our backlog.

deploy/Chart/templates/rp/configmaps.yaml
```diff
@@ -13,17 +13,9 @@ data:
...
       name: self-hosted
       roleLocation: "global"
-    storageProvider:
-      provider: "apiserver"
-      apiserver:
-        context: ""
-        namespace: "radius-system"
-    queueProvider:
-      provider: "apiserver"
+    storageProvider: {{ toYaml .Values.global.storageProvider | nindent 6 }}
```
gpltaylor, line 16: Same comments as above. Here we only need to change the namespace, not the whole block.

gpltaylor: However, I do like how DRY this is :)

cpossinger:
> @cpossinger Can you update this PR? Please let me know if you do not have a time. We will prioritize this on our backlog.

Hey sorry for not updating this sooner.

I'm focused on getting a POC approved within my IT organization so I'm going to close this for now. 

Radius will fundamentally change how we do things at Delta and possibly even shift our org structure to accommodate a self service IT model which requires less resources and a better developer experience.

That being said if you want to target usage from large not-tech organizations like Delta it would be worth thinking about how to create a POC version without cluster level permissions so people can try it out and generate buy in easier. Once people see Radius in action they realize the value but it's more difficult to communicate that value conceptionally.

Once I can get a POC approved I'll start working on getting Radius to be fully functional (if it isn't already) on a ROSA cluster.
Really appreciate the feedback on this PR. 

Thanks,
Camden

rynowak:
> That being said if you want to target usage from large not-tech organizations like Delta it would be worth thinking about how to create a POC version without cluster level permissions so people can try it out and generate buy in easier. 

Would it be easier if you could run Radius outside the cluster? I was thinking something like the following:

- Radius runs locally (or in the cloud) in Docker (or any container infra)
- You point Radius at a Kubernetes cluster (hey you, deploy applications over there)
- Everything else works the same :+1:

PR #7011, opened 2024-01-10
Add Deployment Webhook

lakshmimsft:
# Description

Adding Deployment webhook to pause new radius enabled deployments.
Deployment is unpaused after necessary env variables are created.

## Type of change

- This pull request adds or changes features of Radius and has an approved issue ( #6431 ).

Fixes: Part of #6431

### DIFF
--- a/deploy/Chart/templates/controller/webhook.yaml
+++ b/deploy/Chart/templates/controller/webhook.yaml
@@ -69,6 +69,7 @@ webhooks:
     - v1
     operations:
     - CREATE
+    - UPDATE
     resources:
     - deployments
   sideEffects: None
--- a/deploy/Chart/templates/controller/webhook.yaml
+++ b/deploy/Chart/templates/controller/webhook.yaml
@@ -45,3 +45,30 @@ webhooks:
     resources:
     - recipes
   sideEffects: None
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: deployment-webhook.apps.io
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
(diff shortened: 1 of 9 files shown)


deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -50,4 +50,31 @@ webhooks:
...
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: deployment-webhook.apps.io
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    caBundle: {{ b64enc $ca.Cert }}
```
ytimocin, line 63: We should use the existing caBundle here and not recreate each time. Like here: https://github.com/radius-project/radius/pull/7008/files#diff-69a5c48489019be60a68141f4c76f07622f44ea7a34694aebfeaa088491eab04R36.

lakshmimsft: thank you for catching that. i will update it per related PR.

test/functional/kubernetes/kubernetes_test.go
```diff
@@ -129,6 +130,25 @@ func Test_TutorialApplication_KubernetesManifests(t *testing.T) {
 
 		_, err = client.Get(ctx, deployment.Name, nil)
 		require.NoError(t, err)
+
+		// Get the list of pods
+		labelSelector := labels.Set{"app": "demo"}.AsSelector().String()
```
lakshmimsft, line 135: Checking if crashloopbackoff has occured (since there could be other reasons for pods to be restarted). 
@rynowak , pls review if this is correct.

lakshmimsft:
Closing for now. Will pick it up later, make it work for Update scenario.

PR #7013, opened 2024-01-10
Add check for Terraform submodules

### LINKED ISSUE #7005: Support for Terraform Submodules with Context Variables
It seems like Terraform registries don't work with submodules when using context parameters. I have a TF module uploaded [here](https://registry.terraform.io/modules/sk593/redis/kubernetes/latest) based off this Github repo [here](https://github.com/sk593/terraform-kubernetes-redis/tree/main). Versions 1.0.0 - 3.0.0 have the recipe uploaded as a submodule and these fail deployment when using context parameters. Versions 4.0.0 - 6.0.0 were uploaded with the recipe at the root folder and these deploy successfully.

Recipe deployment with the submodule uses a slightly different module source as the template path (note the `//` in the template path):
```
recipes: {
      'Applications.Core/extenders': {
        default: {
          templateKind: 'terraform'
          templatePath: 'sk593/redis/kubernetes//modules/kubernetes-redis'
          templateVersion: '>= 1.0.0, < 3.0.0'
        }
      }
    }
```

This fails with the following error:
```
Error: {
  "code": "DeploymentFailed",
  "message": "At least one resource deployment operation failed. Please see the details for the specific operation that failed.",
  "target": "/planes/radius/local/resourceGroups/default/providers/Microsoft.Resources/deployments/rad-deploy-3f98b05a-259c-4aa5-bbac-feaef33b0cf9",
  "details": [
    {
      "code": "OK",
      "message": "",
      "target": "/planes/radius/local/resourceGroups/default/providers/Applications.Core/applications/myapp"
    },
    {
      "code": "ResourceDeploymentFailure",
      "message": "Failed",
      "target": "/planes/radius/local/resourceGroups/default/providers/Applications.Core/extenders/corerp-resources-terraform-redis",
      "details": [
        {
          "code": "RecipeDeploymentFailed",
          "message": "terraform apply failure: exit status 1\n\nError: Missing required argument\n\n  on main.tf.json line 16, in module.default:\n  16:     }\n\nThe argument \"context\" is required, but no definition was found.\n"
        }
      ]
    },
    {
      "code": "OK",
      "message": "",
      "target": "/planes/radius/local/resourceGroups/default/providers/Applications.Core/environments/corerp-resources-terraform-redis-env"
    }
  ]
}

TraceId:  bba43582c0fde7f18a6b21d316c98548
```

Ideally submodules should deploy successfully when using context parameters. 

### Proposed Fix (optional)

It seems like this issue occurs because we try to [load the module](https://github.com/radius-project/radius/blob/fcd096e43bf73e47a434708d43d28d56ebb1b1ba/pkg/recipes/terraform/module.go#L62C54-L62C54) that's at the root directory. But we need to access the submodule to find if there's a context variable. Terraform has a hard convention where they denote submodules using a `//` in the source. See here: https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories

Since this is their convention, we could parse the template paths for the double slash and then deploy the recipe that's in the nested submodule instead of only looking at the root. 

## Additional context

[AB#10855](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10855)

sk593:
# Description

This checks for submodules when deploying TF recipes. If the source is a submodule, we'll use that path when loading the module after the download

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius #7005

Fixes: #7005


pkg/recipes/terraform/module.go
```diff
@@ -53,12 +50,18 @@ type moduleInspectResult struct {
...
+func inspectModule(workingDir, localModuleName string, templatePath string) (*moduleInspectResult, error) {
+	moduleRootDir := ".terraform/modules"
 	result := &moduleInspectResult{ContextVarExists: false, RequiredProviders: []string{}, ResultOutputExists: false, Parameters: map[string]any{}}
 
 	// Modules are downloaded in a subdirectory in the working directory.
 	// Name of the module specified in the configuration is used as subdirectory name.
 	// https://developer.hashicorp.com/terraform/tutorials/modules/module-use#understand-how-modules-work
+	//
+	// If the template path is for a submodule, we'll add the submodule path to the module root directory.
+	if strings.Contains(localModuleName, "//") {
```
sk593, line 62: Terraform has a hard convention for the double slash: https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories

We can use this special character to check if the module source is a submodule

sk593: Updated this to find the last index of `//` but the logic for using those characters as a delineator is the same

vishwahiremat:
can we add a functional test for this scenario?

pkg/recipes/terraform/module.go
```diff
@@ -20,17 +20,14 @@ import (
...
 	"github.com/hashicorp/terraform-config-inspect/tfconfig"
 	"github.com/hashicorp/terraform-exec/tfexec"
 	"github.com/radius-project/radius/pkg/recipes"
 	"github.com/radius-project/radius/pkg/recipes/recipecontext"
 )
 
-const (
-	moduleRootDir = ".terraform/modules"
-)
-
```
vishwahiremat, line 33: why was this removed?

sk593: `moduleRootDir` will need to change depending on whether there's a submodule or not, so this value was moved into the `inspectModules` function

pkg/recipes/terraform/module.go
```diff
@@ -53,13 +50,21 @@ type moduleInspectResult struct {
...
 
 	// Modules are downloaded in a subdirectory in the working directory.
 	// Name of the module specified in the configuration is used as subdirectory name.
 	// https://developer.hashicorp.com/terraform/tutorials/modules/module-use#understand-how-modules-work
-	mod, diags := tfconfig.LoadModule(filepath.Join(workingDir, moduleRootDir, localModuleName))
+	//
+	// If the template path is for a submodule, we'll add the submodule path to the module root directory.
+	subModDex := strings.LastIndex(templatePath, "//")
+	// Make sure that we don't mistake the double slashes in a '://' scheme as a submodule indicator
+	if subModDex != -1 && templatePath[subModDex-1:subModDex+len("//")] != "://" {
```
lakshmimsft, line 64: I don't have full context but this code may give error if subModDex ==0. Is that a possible case? Do we need to update the condition?

sk593: Ah, good catch. There shouldn't be any Terraform sources that can start with `//` but this should still handle that case. Will update this check

vishwahiremat: I wonder if this works for non https module source kind.
E.g: github.com/project/module

vishwahiremat: I was thinking as @lakshmimsft said, that would hit `index out of range` at `templatePath[subModDex-1:subModDex+len("//")]`  if it was not a submodule

pkg/recipes/terraform/execute.go
```diff
@@ -281,7 +281,7 @@ func downloadAndInspect(ctx context.Context, tf *tfexec.Terraform, options Optio
 	// Load the downloaded module to retrieve providers and variables required by the module.
 	// This is needed to add the appropriate providers config and populate the value of recipe context variable.
 	logger.Info(fmt.Sprintf("Inspecting the downloaded Terraform module: %s", options.EnvRecipe.TemplatePath))
-	loadedModule, err := inspectModule(tf.WorkingDir(), options.EnvRecipe.Name)
+	loadedModule, err := inspectModule(tf.WorkingDir(), options.EnvRecipe.Name, options.EnvRecipe.TemplatePath)
```
ytimocin, line 284: Just thinking out loud; would it make more sense to have the whole Recipe object as the input to the **inspectModule** function? For future proofing the function... Instead of increasing the number of parameters, we can have a single parameter that would contain all the necessary.

pkg/recipes/terraform/module.go
```diff
@@ -53,13 +50,21 @@ type moduleInspectResult struct {
...
+	moduleRootDir := filepath.Join(".terraform/modules", localModuleName)
 	result := &moduleInspectResult{ContextVarExists: false, RequiredProviders: []string{}, ResultOutputExists: false, Parameters: map[string]any{}}
 
 	// Modules are downloaded in a subdirectory in the working directory.
 	// Name of the module specified in the configuration is used as subdirectory name.
 	// https://developer.hashicorp.com/terraform/tutorials/modules/module-use#understand-how-modules-work
-	mod, diags := tfconfig.LoadModule(filepath.Join(workingDir, moduleRootDir, localModuleName))
+	//
+	// If the template path is for a submodule, we'll add the submodule path to the module root directory.
+	subModDex := strings.LastIndex(templatePath, "//")
```
ytimocin, line 62: I believe that line 62-66 could be extracted to a function for more readability. It could also be easier to just unit test that function for different use cases. Something like **getModulePath**? You know the context much better...

ytimocin: We can also add a good documentation to that function.

sk593: Ended up using an existing TF function so these lines won't be needed

pkg/recipes/terraform/execute.go
sk593: Updated to use the recipe

pkg/recipes/terraform/module.go
sk593: Ended up using an existing TF function so this check is no longer needed

go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/google/uuid v1.3.0
 	github.com/gosuri/uilive v0.0.4
+	github.com/hashicorp/go-retryablehttp v0.7.5
```
ytimocin, line 48: We can run `go mod tidy` just in case.

ytimocin (dismissed):
Looks very clean! Thanks @sk593!

kachawla (dismissed):
Thanks for making this change @sk593. Are you planning to add/update functional test in a separate PR?

sk593:
> Thanks for making this change @sk593. Are you planning to add/update functional test in a separate PR?

I'm planning to address this in this PR, but I'm still working through the best way to upload the submodule folder. Since we compress our recipes and call them over HTTPS, I don't think it will work exactly the same as it does now

go.mod
sk593: I'll rerun again before merging to make sure this is up to date

lakshmimsft (dismissed):

kachawla (commented):
Looks like the functional test is failing..

test/functional/shared/resources/testdata/corerp-resources-terraform-redis.bicep
```diff
@@ -21,7 +21,7 @@ resource env 'Applications.Core/environments@2023-10-01-preview' = {
       'Applications.Core/extenders': {
         default: {
           templateKind: 'terraform'
-          templatePath: '${moduleServer}/kubernetes-redis.zip'
+          templatePath: '${moduleServer}/kubernetes-redis.zip//kubernetes-redis/modules'
```
kachawla, line 24: Looking at the docs https://docs.python.org/3/library/shutil.html#shutil-archiving-example-with-basedir, I think unpack returns directories inside root directory. So we either change the path here to `//modules` or specify base directory while creating the zip here: https://github.com/radius-project/radius/blob/main/.github/scripts/publish-test-terraform-recipes.py#L63 - root dir would change to tmp_dir and base directory would be recipe_dir

go.mod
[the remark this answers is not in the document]
sk593: I think this was automatically added when running `go mod tidy`. It doesn't seem to be a dependency for the hashicorp/go-getter package that was added though

test/functional/shared/resources/testdata/corerp-resources-terraform-redis.bicep
sk593: Updated to `//modules` and that seems to have fixed the path issue. Thanks for finding this!

kachawla (approved):

lakshmimsft (approved):

PR #7022, opened 2024-01-11
Updating the caBundle for the controller webhook

### LINKED ISSUE #6989: "x509: certificate signed by unknown authority" error during webhook call
I ran into this issue when I was trying to fix the long-haul test cluster. Running `make test-functional-kubernetes` gave me the following error:

`
Internal error occurred: failed calling webhook "recipe-webhook.radapp.io": failed to call webhook: Post "https://controller.radius-system.svc:443/validate-radapp-io-v1alpha3-recipe?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "controller-ca")
`

Flakiness during the kubernetes (`make test-functional-kubernetes`) functional tests.

No certificate error.

### Workaround (optional)
Deleting the secret for the controller-cert worked.

### rad Version (required)
0.28.0    v0.28.0   0.28.0    c7217b6f661a790053c1b65ff7972a0c5be33908

### Operating system (required)
macOS, **Apple Chip**, Sonoma 14.2.1

ytimocin:
# Description
Updating the caBundle for the controller webhook

## Type of change
Fixes: #6989

### DIFF
--- a/deploy/Chart/templates/_helpers.tpl
+++ b/deploy/Chart/templates/_helpers.tpl
@@ -12,3 +12,32 @@
 {{- end -}}
 {{- print $version }}
 {{- end -}}
+
+{{/*
+Reuses the value from an existing secret, otherwise sets its value to a default value.
+
+Usage:
+{{ include "secrets.lookup" (dict "secret" "secret-name" "namespace" "ns-name" "key" "key-name" "defaultValue" "default-secret") }}
+
+Params:
+  - secret - String - Required - Name of the 'Secret' resource where the password is stored.
+  - namespace - String - Required - Namespace of the 'Secret' resource where the password is stored.
+  - key - String - Required - Name of the key in the secret.
+  - defaultValue - String - Required - Default value to use if the secret does not exist.
+
+References:
+  - https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_secrets.tpl
+*/}}
+{{- define "secrets.lookup" -}}
+{{- $value := "" -}}
+{{- $namespace := .namespace | toString -}}
+{{- $secretData := (lookup "v1" "Secret" $namespace .secret).data -}}
+{{- if and $secretData (hasKey $secretData .key) -}}
+  {{- $value = index $secretData .key -}}
+{{- else if .defaultValue -}}
+  {{- $value = .defaultValue | toString | b64enc -}}
+{{- end -}}
+{{- if $value -}}
+{{- printf "%s" $value -}}
+{{- end -}}
+{{- end -}}
--- a/deploy/Chart/templates/controller/validating-webhook-configuration.yaml
+++ b/deploy/Chart/templates/controller/validating-webhook-configuration.yaml
@@ -1,5 +1,3 @@
-{{- $existingSecret := lookup "v1" "Secret" .Release.Namespace "controller-cert"}}
-{{- $existingWebhook := lookup "admissionregistration.k8s.io/v1" "ValidatingWebhookConfiguration" .Release.Namespace "recipe-webhook.radapp.io"}}
 {{- $ca := genCA "controller-ca" 3650 }}
 {{- $cn := printf "controller" }}
 {{- $altName1 := printf "controller.%s" .Release.Namespace }}
@@ -15,14 +13,11 @@ metadata:
   labels:
     app.kubernetes.io/name: controller
     app.kubernetes.io/part-of: radius
+type: kubernetes.io/tls
 data:
-  {{ if $existingSecret }}tls.crt: {{ index $existingSecret.data "tls.crt" }}
-  {{ else }}tls.crt: {{ b64enc $cert.Cert }}
-  {{ end }}
-
-  {{ if $existingSecret }}tls.key: {{ index $existingSecret.data "tls.key" }}
-  {{ else }}tls.key: {{ b64enc $cert.Key }}
-  {{ end }}
+  tls.crt: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "tls.crt" "defaultValue" $cert.Cert) }}
+  tls.key: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "tls.key" "defaultValue" $cert.Key) }}
+  ca.crt: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "ca.crt" "defaultValue" $ca.Cert) }}
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
@@ -32,7 +27,7 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
-    caBundle: {{ b64enc $ca.Cert }}
+    caBundle: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "ca.crt" "defaultValue" $ca.Cert) }}
     service:
       name: controller
       namespace: {{ .Release.Namespace }}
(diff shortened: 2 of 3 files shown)


deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -32,7 +33,7 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
-    caBundle: {{ b64enc $ca.Cert }}
+    caBundle: {{ $validatingWebhookCaBundle }}
```
youngbupark, line 36: @ytimocin Can you please make the code consistent with how UCP does? UCP way is more intuitive.

ytimocin: In UCP's case, we don't need to select a **caBundle** from an array of webhook configurations: https://github.com/radius-project/radius/blob/main/deploy/Chart/templates/ucp/apiservice.yaml#L42.

The reason I did it this way is that we have a new webhook coming soon and that will also be added to the array of configs. We will need to map the right webhook configuration to the right caBundle.

The only thing that we will need to do for the next webhook is to call the function with the right webhook name.

Let me know if that makes sense @youngbupark.

youngbupark: For mutatingAdmissionController, we will use existing `controller` service and share the same endpoint. Then I do not think we need additional cert for mutating controller.

lakshmimsft (dismissed):

kachawla (dismissed):

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -32,7 +32,7 @@ webhooks:
 - admissionReviewVersions:
   - v1
   clientConfig:
-    caBundle: {{ b64enc $ca.Cert }}
+    caBundle: {{ if $existingValidatingWebhook }}{{ index $existingValidatingWebhook "webhooks" 0 "clientConfig" "caBundle" }}{{ else }}{{ b64enc $ca.Cert }}{{ end }}
```
youngbupark, line 35: @ytimocin What if we can fetch the cert from secret instead of fetching from webhook resource? Secret resource is source of truth, so it is better to fetch from secret resource. bitnami hosting chart is using this way. Please see the below example.

https://github.com/bitnami/charts/blob/882140cfc5318420fdf563b77201f7ee5f67ca8f/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/validating-webhook-configuration.yaml#L51

If you think it is better approach, please update this way in webhook controller and ucp.

youngbupark (changes requested):
controller-cert secret should be created only if existingSecret is null. Otherwise, it shouldn't create it. Please add `if` statement around `controller-cert` manifest.

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -16,13 +16,9 @@ metadata:
...
-  {{ if $existingSecret }}tls.crt: {{ index $existingSecret.data "tls.crt" }}
-  {{ else }}tls.crt: {{ b64enc $cert.Cert }}
-  {{ end }}
-
-  {{ if $existingSecret }}tls.key: {{ index $existingSecret.data "tls.key" }}
-  {{ else }}tls.key: {{ b64enc $cert.Key }}
-  {{ end }}
+  tls.crt: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "tls.crt" "defaultValue" $cert.Cert) }}
+  tls.key: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "tls.key" "defaultValue" $cert.Key) }}
+  ca.crt: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "ca.crt" "defaultValue" $ca.Cert) }}
```
youngbupark, line 21: Why do we need this ? this is dupe with tls.crt

deploy/Chart/templates/controller/webhook.yaml
```diff
@@ -16,13 +16,9 @@ metadata:
...
     app.kubernetes.io/part-of: radius
 data:
-  {{ if $existingSecret }}tls.crt: {{ index $existingSecret.data "tls.crt" }}
-  {{ else }}tls.crt: {{ b64enc $cert.Cert }}
-  {{ end }}
-
-  {{ if $existingSecret }}tls.key: {{ index $existingSecret.data "tls.key" }}
-  {{ else }}tls.key: {{ b64enc $cert.Key }}
-  {{ end }}
+  tls.crt: {{ include "secrets.lookup" (dict "secret" "controller-cert" "namespace" .Release.Namespace "key" "tls.crt" "defaultValue" $cert.Cert) }}
```
youngbupark, line 19: We do not need to create Secret if secret exists. So we can simplify it like below. 

```
{{- if not $existingSecret }} // ---> UPDATE THIS 
apiVersion: v1
kind: Secret
metadata:
  name: controller-cert
  namespace: {{ .Release.Namespace }}
  labels:
    app.kubernetes.io/name: controller
data:
  tls.crt: {{ b64enc $cert.Cert }}
  tls.key: {{ b64enc $cert.Key }}
{{- end }}
```

youngbupark (changes requested):

deploy/Chart/templates/controller/webhook.yaml
ytimocin: `secrets.lookup` function checks if there is already a Secret and, if there is, it uses the existing one. Like [here](https://github.com/bitnami/charts/blob/882140cfc5318420fdf563b77201f7ee5f67ca8f/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/validating-webhook-configuration.yaml#L51).

`secrets.lookup` gets the secret name and the namespace that secret should be in, checks if it exists, if it does it returns the existing secret. If not, it returns the default value provided. In this case $cert.Cert, $cert.Key, and $ca.Cert.

deploy/Chart/templates/controller/webhook.yaml
ytimocin: `tls.crt` uses `$cert.Cert` while `ca.crt` uses `$ca.Cert`. They are different.

ytimocin: `ca.crt` is also added here: https://github.com/bitnami/charts/blob/882140cfc5318420fdf563b77201f7ee5f67ca8f/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/validating-webhook-configuration.yaml#L29. That is what we use for the **caBundle**.

We need a way to access existing `ca.crt` to use it in the **caBundle**.

deploy/Chart/templates/controller/webhook.yaml
ytimocin: If we do it this (`{{- if not $existingSecret }}`) way, then Secret gets deleted when we do `--reinstall`. In the `--reinstall` case, the **$existingSecret** is not nil, that part is missing from the template. Let me know if that makes sense. Happy to sync offline.

And I did another test, when we do `--reinstall` Secret doesn't get recreated with this way: https://github.com/radius-project/radius/pull/7022/files#diff-dacf6ba48c5b451f0a3a55b9978b57991a50fb27fc2da1c4785e09e52adc4b09R18-R20.

deploy/Chart/templates/_helpers.tpl
```diff
@@ -12,3 +12,32 @@
 {{- end -}}
 {{- print $version }}
 {{- end -}}
+
+{{/*
+Reuses the value from an existing secret, otherwise sets its value to a default value.
+
+Usage:
+{{ include "secrets.lookup" (dict "secret" "secret-name" "namespace" "ns-name" "key" "key-name" "defaultValue" "default-secret") }}
```
lakshmimsft, line 20: nit: perhaps ...."defaultValue" "default-secretvalue") or ...."defaultValue" "default-value") is more accurate

lakshmimsft (approved):

youngbupark (approved):
LGTM. Please ensure that we run the proper test for clean install and upgrade scenarios.

PR #7032, opened 2024-01-16
Report functional test pass only when all subtests finish

### LINKED ISSUE #6939: Functional test checks don't work correctly
Functional tests

The "check" added by our functional tests is marked passing as soon as **one** of the functional test jobs completes successfully.

Our functional tests use a matrix for parallelism. However we use a single 'check' to mark the tests as passing, and we update **the single check** for all of the jobs that are launched by the matrix.

Functional tests check should only be marked passing when **all** of the jobs complete successfully.

### Proposed Fix (optional)

Move the interactions with the checkout out of the matrix.

willdavsmith:
# Description

* Change functional test logic to only report complete when all subtests (shared, samples, msgrp, etc.) are complete.

## Type of change

Fixes: #6939

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -75,7 +75,7 @@ jobs:
   build:
     name: Build Radius for test
     runs-on: ubuntu-latest
-    if: github.event_name == 'repository_dispatch' || (github.event_name == 'schedule' && github.repository == 'radius-project/radius') || github.event_name == 'workflow_run'
+    if: github.event_name == 'repository_dispatch' || (github.event_name == 'schedule' && github.repository == 'willdavsmith/radius') || github.event_name == 'workflow_run'
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -614,7 +640,7 @@ jobs:
           repo: ${{ github.repository }}
           sha: ${{ env.CHECKOUT_REF }}
           status: completed
-          conclusion: ${{ job.status }}
+          conclusion: ${{ needs.tests.outputs.test_status }}
```
rynowak, line 643: Maybe a dumb question about this about this ... does this work?

Since `tests` is a matrix are we going to end up with the same issue because each entry in the matrix will set `test_status`?

Would using `needs.tests.status` do the same thing? https://github.com/orgs/community/discussions/26822#discussioncomment-3253537

PR #7037, opened 2024-01-17
Monitor release progress from single workflow

### LINKED ISSUE #6924: Kick off samples verification, samples release, docs release from `versions.yaml` change
[AB#10698](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10698)

willdavsmith:
# Description

This PR adds some logic to update the release process to:
1. simplify the troubleshooting process for a release - all successes and failures should be reported in one place
2. remove manual steps of triggering the release verification workflow, the samples verification workflow, releasing docs, and releasing samples

## Type of change

Fixes: #6924


.github/workflows/release.yaml
```diff
@@ -187,3 +187,84 @@ jobs:
         run: |
           ./radius/.github/scripts/release-create-tag-and-branch.sh recipes ${{ steps.get-version.outputs.release-version }} ${{ steps.get-version.outputs.release-branch-name }}
         if: success()
+
+      - name: Wait for the radius-project/radius Build and Test workflow to complete
```
youngbupark, line 191: Instead of creating our own script, I would consider using the following action:

1. https://github.com/marketplace/actions/wait-on-check
2. https://github.com/marketplace/actions/await-remote-run

youngbupark: wait-on-check is promising for this scenario.

willdavsmith: wait-on-check doesn't support different repos, so I forked and added these changes: https://github.com/lewagon/wait-on-check-action/compare/master...willdavsmith:wait-on-check-action:master

PR #7079, opened 2024-01-26
Support routes.*.destination of gateway for App Graph

### LINKED ISSUE #7038: [Application graph] data doesnt reflect correctly in graph for some data
deploy below bicep and check for correctness
```
import radius as radius

@description('Specifies the environment for resources.')
param environment string

@description('The ID of your Radius Application. Automatically injected by the rad CLI.')
param application string

// The backend container that is connected to the Dapr state store
resource backend 'Applications.Core/containers@2023-10-01-preview' = {
  name: 'backend'
  properties: {
    application: application
    container: {
      // This image is where the app's backend code lives
      image: 'ghcr.io/radius-project/samples/dapr-backend:latest'
      ports: {
        orders: {
          containerPort: 3000
        }
      }
    }
    connections: {
      orders: {
        source: stateStore.id
      }
      subscribers: {
        source: messageQueue.id
      }
      gateway: {
        source: gateway.id
      }
    }
    extensions: [
      {
        kind: 'daprSidecar'
        appId: 'backend'
        appPort: 3000
      }
    ]
  }
}

resource messageQueue 'Applications.Messaging/rabbitMQQueues@2023-10-01-preview' = {
  name: 'messageQueue'
  properties: {
    environment: environment
    application: application
  }
}

resource uiCache 'Applications.Datastores/redisCaches@2023-10-01-preview' = {
  name: 'uiCache'
  properties: {
    environment: environment
    application: application
  }
}

// The Dapr state store that is connected to the backend container
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    // Provision Redis Dapr state store automatically via the default Radius Recipe
    environment: environment
    application: application
  }
}

// The frontend container that serves the application UI
resource frontend 'Applications.Core/containers@2023-10-01-preview' = {
  name: 'frontend'
  properties: {
    application: application
    container: {
      // This image is where the app's frontend code lives
      image: 'ghcr.io/radius-project/samples/dapr-frontend:latest'
      env: {
        // An environment variable to tell the frontend container where to find the backend
        CONNECTION_BACKEND_APPID: 'backend'
        // An environment variable to override the default port that .NET Core listens on
        ASPNETCORE_URLS: 'http://*:8080'
      }
      // The frontend container exposes port 8080, which is used to serve the UI
      ports: {
        ui: {
          containerPort: 8080
        }
      }
    }
    connections: {
      cookies: {
        source: uiCache.id
      }
      gateway: {
        source: gateway.id
      }
    }
    // The extension to configure Dapr on the container, which is used to invoke the backend
    extensions: [
      {
        kind: 'daprSidecar'
        appId: 'frontend'
      }
    ]
  }
}

resource gateway 'Applications.Core/gateways@2023-10-01-preview' = {
  name: 'gateway'
  properties: {
    application: application
    // The gateway routes traffic from the frontend to the backend
    routes: [
      {
        path: '/frontend'
        destination: 'http://frontend:8080'
      }
      {
        path: '/backend'
        destination: 'http://backend:3000'
      }
    ]
  }
}

resource emailer 'Applications.Core/containers@2023-10-01-preview' = {
  name: 'emailer'
  properties: {
    application: application
    container: {
      image: 'ghcr.io/radius-project/samples/dapr-backend:latest'
    }
    connections: {
      subscribers: {
        source: messageQueue.id
      }
    }
  }
}
```

[AB#10898](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10898)

youngbupark:
# Description

This is to support `routes.*.destination` property for App graph. It enables to show the connections between Gateway and the other resources in App Graph

## Type of change

Fixes: #7038


pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -459,16 +464,10 @@ func outputResourcesFromAPIData(resource generated.GenericResource) []*corerpv20
 	return entries
 }
 
-// connectionsFromAPIData resolves the outbound connections for a resource from the generic resource representation.
-// For example if the resource is an 'Applications.Core/container' then this function can find it's connections
-// to other resources like databases. Conversely if the resource is a database then this function
-// will not find any connections (because they are inbound). Inbound connections are processed later.
-func connectionsFromAPIData(resource generated.GenericResource, allResources []generated.GenericResource) []*corerpv20231001preview.ApplicationGraphConnection {
+func resolveConnections(resource generated.GenericResource, jsonRefPath string, converter func(any) (string, corerpv20231001preview.Direction, error)) []*corerpv20231001preview.ApplicationGraphConnection {
```
youngbupark, line 467: I refactored the code to use the same code for resolving connections, routes, and provides instead of repeating the same code.

pkg/corerp/frontend/controller/applications/testdata/graph-app-gw-out.json
```diff
@@ -0,0 +1,53 @@
+[
```
youngbupark, line 1: @nithyatsu @rynowak Please review this appgraph output carefully. To me, the result makes sense.

pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -356,8 +361,8 @@ func computeGraph(applicationName string, applicationResources []generated.Gener
 		connectionsIn := connectionsByDestination[id]
 
 		entryConnections := make([]*corerpv20231001preview.ApplicationGraphConnection, len(connectionsIn))
-		for i, conn := range connectionsIn {
```
youngbupark, line 359: This is the code bug. `&conn` refered to the same data so that entryConnections array held the same pointer values.

ytimocin: Great catch!

pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -547,61 +551,61 @@ func findSourceResource(source string, allResources []generated.GenericResource)
...
+		return sourceID, corerpv20231001preview.DirectionOutbound, nil
 	}
+}
 
-	raw, _, err := p.Get(&resource)
-	if err != nil {
-		// Not found, this is fine.
-		return []*corerpv20231001preview.ApplicationGraphConnection{}
+func routesPathConverter(resources []generated.GenericResource) func(any) (string, corerpv20231001preview.Direction, error) {
+	return func(item any) (string, corerpv20231001preview.Direction, error) {
```
ytimocin, line 574: Can't we have one function for all three and just pass in the resource type?

ytimocin (dismissed):

pkg/corerp/frontend/controller/applications/graph_util.go
youngbupark: > Can't we have one function for all three and just pass in the resource type?

That way makes more complicated and each one has different logic for scanning resource. it is not worth making.

pkg/corerp/frontend/controller/applications/graph_util.go
rynowak: Ugh. 

The good news is that they are fixing this: https://github.com/golang/go/issues/60078

I wasn't able to find the link, but we can use the Go compiler to detect these cases.

pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -547,61 +556,62 @@ func findSourceResource(source string, allResources []generated.GenericResource)
...
-
-	p, err := jsonpointer.New(portsPath)
-	if err != nil {
-		// This should never fail since we're hard-coding the path.
-		panic("parsing JSON pointer should not fail: " + err.Error())
+// connectionsResolver resolves the outbound connections for a resource from the generic resource representation.
+// For example if the resource is an 'Applications.Core/container' then this function can find it's connections
+// to other resources like databases. Conversely if the resource is a database then this function
+// will not find any connections (because they are inbound). Inbound connections are processed later.
+func connectionsResolver(resources []generated.GenericResource) resolver {
```
rynowak, line 563: Not relevant for this review, but long term I'd like to make this behavior consistent across APIs by using Capabilities + Well-Known API shapes, or Well-Known Child resources.

Consider how we'd make this extensible 👍 When users can define their own types, they should be able to support all of the features of the built-in ones.

rynowak: Basically anyplace we write code that handles a specific resource type like `Applications.Core/gateways`, we might need to reconsider in the future.

pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -248,8 +252,12 @@ func computeGraph(applicationName string, applicationResources []generated.Gener
 			applicationGraphResource.ProvisioningState = &state
 		}
 
-		connections := connectionsFromAPIData(resource, resources)          // Outbound connections based on 'connections'
-		connections = append(connections, providesFromAPIData(resource)...) // Inbound connections based on 'provides'
+		// Resolve Outbound connections based on 'connections'.
+		connections := resolveConnections(resource, connectionsPath, connectionsResolver(resources))
```
rynowak, line 256: Should we be checking the resource type here? It looks like all of these "resolvers" apply to all resource types. So if a resource has a property `.properties.connections` it works like `Applications.Core/containers`, has the same schema, etc.

rynowak: This is similar to the design I was talking about here: https://github.com/radius-project/radius/pull/7079/files#r1467935133 but without being driven by Capabilities.

pkg/corerp/frontend/controller/applications/testdata/graph-app-gw-out.json
rynowak: Looks right to me 👍

pkg/corerp/frontend/controller/applications/graph_util.go
youngbupark: I intentionally didn't try to refactor it further to keep the change short. Instead of strongly typed one, I would try to using jsonpointer to look up the `connections.destination`. We can support multiple types. Also for custom resource, we can ask contributors to use the specific connection properties if they want to use appgraph. If you like this idea, I can make the change now.

rynowak: I don't think we should change it now - we should discuss during the design of user-defined-types. I just wanted to mention the idea.

rynowak (approved):

pkg/corerp/frontend/controller/applications/graph_util.go
rynowak: I'm good with this for now 👍 we will revisit when we do extensibility.

PR #7081, opened 2024-01-26
Fix rad app delete does not clean up all resources when the application deployment originally failed

### LINKED ISSUE #7052: Deleting application does not delete resources in some cases
### Steps to reproduce

ℹ️ This issue is the continuity of [discord forum](https://discord.com/channels/1113519723347456110/1196857104305234001) with @vishwahiremat 

I have created an environment, application and container in 3 different files. I then have applied them one by one.
Finally I decided I didn't want application and container anymore so I run rad application delete myapp. The application seems removed but not the container which was part of it.

Here are some of the files:
```bicep
import radius as radius

resource environment 'Applications.Core/environments@2023-10-01-preview' = {
  name: 'demo5env'
  properties: {
    compute: {
      kind: 'kubernetes'
      namespace: 'demo5env'
    }
  }
}
```
```bicep
import radius as radius

resource myapp 'Applications.Core/applications@2023-10-01-preview' = {
  name: 'myapp'
  properties: {
    environment: '/planes/radius/local/resourcegroups/demo5group/providers/Applications.Core/environments/demo5env'
  }
}
```
```bicep
import radius as radius

import kubernetes as kubernetes {
  namespace: 'default-application'
  context: '...'
  kubeConfig: '...'
}

resource backend_container 'Applications.Core/containers@2023-10-01-preview' = {
  name: 'backend'
  properties: {
    application: '/planes/radius/local/resourcegroups/demo5group/providers/Applications.Core/applications/myapp'
    container: {
      ...
    }
  }
}
```

### Observed behavior

The application is deleted while the k8s.deployment is not.

### Desired behavior

We should never have orphan resources.

### Workaround

To me it seems the problems appears when we deploy the application and the resource with sepparate commands.

But @vishwahiremat thinks it's not the case. For him the issue is linked to the failing container deployment.

I'm surprised as, if I understand correctly, if the application and the container are deployed together, the container is still fialing but deleting the application succeed at deleting the service.

@vishwahiremat, do you mean the "link" is not done as the "deployment of the resource" is failing when deployed separately ? 

### rad Version

```
❯ rad version
edge      16011a5   0.29.0    16011a5da8e7e531da9d59527954fc5336aac044
```

### Operating system

WSL2

### Additional context

_No response_

[AB#10954](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10954)

nithyatsu:
# Description

We do not track the outputResources in the event of deplyment failure, but might have created some, potentially with errors. 
If the deployment of outResource errors, we should cleanup these so that app can be deleted gracefully using rad app delete command. 

## Type of change

Fixes: #7052

### DIFF
--- a/pkg/corerp/backend/deployment/deploymentprocessor.go
+++ b/pkg/corerp/backend/deployment/deploymentprocessor.go
@@ -294,7 +294,10 @@ func (dp *deploymentProcessor) Deploy(ctx context.Context, id resources.ID, rend
 				ID:      outputResource.ID,
 			}
 			deployedOutputResources = append(deployedOutputResources, outputResource)
-			dp.Delete(ctx, id, deployedOutputResources)
+			deleteErr := dp.Delete(ctx, id, deployedOutputResources)
+			if deleteErr != nil {
+				logger.Info("Failed to cleanup deployed output resources for resource %q. Error: %s", id.String(), deleteErr.Error())
+			}
 
 			return rpv1.DeploymentOutput{}, err
 		}
--- a/pkg/corerp/backend/deployment/deploymentprocessor.go
+++ b/pkg/corerp/backend/deployment/deploymentprocessor.go
@@ -287,6 +287,15 @@ func (dp *deploymentProcessor) Deploy(ctx context.Context, id resources.ID, rend
 
 		err := dp.deployOutputResource(ctx, id, rendererOutput, computedValues, &handlers.PutOptions{Resource: &outputResource, DependencyProperties: deployedOutputResourceProperties})
 		if err != nil {
+			// add the last deployed resource  which was deployed with error to deployed list
+			// cleanup all deployed output resources in reverse order since app deployment has failed
+			outputResource := rpv1.OutputResource{
+				LocalID: outputResource.LocalID,
+				ID:      outputResource.ID,
+			}
+			deployedOutputResources = append(deployedOutputResources, outputResource)
+			dp.Delete(ctx, id, deployedOutputResources)
+
 			return rpv1.DeploymentOutput{}, err
 		}
 


pkg/corerp/backend/deployment/deploymentprocessor.go
```diff
@@ -287,6 +287,18 @@ func (dp *deploymentProcessor) Deploy(ctx context.Context, id resources.ID, rend
...
+			outputResource := rpv1.OutputResource{
+				LocalID: outputResource.LocalID,
+				ID:      outputResource.ID,
+			}
+			deployedOutputResources = append(deployedOutputResources, outputResource)
+			deleteErr := dp.Delete(ctx, id, deployedOutputResources)
+			if deleteErr != nil {
+				logger.Info("Failed to cleanup deployed output resources for resource %q. Error: %s", id.String(), deleteErr.Error())
+			}
+
```
rynowak, line 301: WDYT would happen in the following scenario?

- User deploys an application successfully. They are using a Recipe to create a Redis cache, and a cloud-hosted Redis cache is created successfully.
- User updates the application. The recipe runs again, but this time it fails - let's say it was on outage on the cloud-provider's part.
- Then what ???

pkg/corerp/backend/deployment/deploymentprocessor.go
```diff
@@ -287,6 +287,18 @@ func (dp *deploymentProcessor) Deploy(ctx context.Context, id resources.ID, rend
 
 		err := dp.deployOutputResource(ctx, id, rendererOutput, computedValues, &handlers.PutOptions{Resource: &outputResource, DependencyProperties: deployedOutputResourceProperties})
 		if err != nil {
+			// add the last deployed resource  which was deployed with error to deployed list
+			// cleanup all deployed output resources in reverse order since app deployment has failed
+			outputResource := rpv1.OutputResource{
+				LocalID: outputResource.LocalID,
+				ID:      outputResource.ID,
+			}
+			deployedOutputResources = append(deployedOutputResources, outputResource)
```
ytimocin, line 296: I can't understand why we need to add the outputResource to a map? The map will only have one item and then the function will exit based on line 302. Am I missing something?

pkg/corerp/backend/deployment/deploymentprocessor.go
nithyatsu: I think it would unnecessarily delete the cache too, with potential data loss . :(
I am trying to fix a tricky scenario where rad app delete can't clean up resources upon deployment error since we dont track o/p resources at all in that case.. I am not sure it would be an improvement if I deleted just the errored outputresource in that case?

rynowak: > I think it would unnecessarily delete the cache too, with potential data loss . 

Yup, that's right 😆

youngbupark: Instead of deleting resource and returning error, what if we can add `status` in deploymentOutput to track the failure state? For updating scenario (failed scenario), we can upsert to the outputresources to keep old resource references for deletion. WDYT?

PR #7095, opened 2024-01-31
Fix Delete application confirmation message points to the workspace but says environment #7089

### LINKED ISSUE #7089: Delete application confirmation message points to the workspace but says environment
### Steps to reproduce

1. Create an application or use an existing one
2. Run `rad app delete my-app`
3. See the confirmation message: https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/app/delete/delete.go#L36
4. Confirmation message says environment but actually prints the workspace name: https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/app/delete/delete.go#L146

### Observed behavior

Delete application confirmation message points to the workspace but says environment:

image

### Desired behavior

Delete application confirmation message should be updated.

### Workaround

_No response_

### rad Version

RELEASE   VERSION                  BICEP     COMMIT
edge      v0.27.0-rc1-94-g096e202  0.29.0    096e2024cd4b09e3a2eeba67a472b82dc4368c33

### Operating system

macOS Apple, Sonoma 14.3

### Additional context

_No response_

[AB#11020](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11020)

jhandel:
Signed-off-by: Josh <josh@liveoak.ws>

# Description

Parse out the environment name from the Workspace.Environment property using the resources library.

## Type of change

This pull request fixes a bug in Radius and has an approved issue

Fixes: #7089

### DIFF
--- a/pkg/cli/cmd/app/delete/delete.go
+++ b/pkg/cli/cmd/app/delete/delete.go
@@ -29,11 +29,12 @@ import (
 	"github.com/radius-project/radius/pkg/cli/output"
 	"github.com/radius-project/radius/pkg/cli/prompt"
 	"github.com/radius-project/radius/pkg/cli/workspaces"
+	"github.com/radius-project/radius/pkg/ucp/resources"
 	"github.com/spf13/cobra"
 )
 
 const (
-	deleteConfirmation = "Are you sure you want to delete application '%v' from workspace '%v'?"
+	deleteConfirmation = "Are you sure you want to delete application '%v' from environment '%v'?"
 	bicepWarning       = "'%v' is a Bicep filename or path and not the name of a Radius Application. Specify the name of a valid application and try again"
 )
 
@@ -82,6 +83,7 @@ type Runner struct {
 	Output            output.Interface
 
 	ApplicationName string
+	EnvironmentName string
 	Scope           string
 	Confirm         bool
 	Workspace       *workspaces.Workspace
@@ -121,6 +123,16 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Lopkup the environment name for use in the confirmation prompt
+	if workspace.Environment != "" {
+		id, err := resources.ParseResource(workspace.Environment)
+		if err != nil {
+			return err
+		}
+
+		r.EnvironmentName = id.Name()
+	}
+
 	// Throw error if user specifies a Bicep filename or path instead of an application name
 	if strings.HasSuffix(r.ApplicationName, ".bicep") {
 		return clierrors.Message(bicepWarning, r.ApplicationName)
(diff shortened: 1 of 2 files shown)


AaronCrawfis:
Thanks for the contribution @jhandel! For #7089, it's indeed a bug where it says "environment" when the value printed is the workspace name. Although, instead of updating the label I think it might be better to update the value to the environment name. We're trying to make workspaces an advanced concept that only need to be referenced/used if you need to opt-into the complexity of managing multiple environments/installations of Radius from a single machine. For the 99% case we want applications/environments to be the main resources and concepts users interact with.

I think a great solution to the bug in #7089 would be to update https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/app/delete/delete.go#L146 to use the environment name instead of the workspace name, and keep the string as-is.

Thanks again for taking a crack at this! We do bi-weekly triage and you caught the issue before we could take a look at possible fixes and post a reply. Very much appreciate the pro-active fixes!

jhandel:
> Thanks for the contribution @jhandel! For #7089, it's indeed a bug where it says "environment" when the value printed is the workspace name. Although, instead of updating the label I think it might be better to update the value to the environment name. We're trying to make workspaces an advanced concept that only need to be referenced/used if you need to opt-into the complexity of managing multiple environments/installations of Radius from a single machine. For the 99% case we want applications/environments to be the main resources and concepts users interact with.
> 
> I think a great solution to the bug in #7089 would be to update https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/app/delete/delete.go#L146 to use the environment name instead of the workspace name, and keep the string as-is.
> 
> Thanks again for taking a crack at this! We do bi-weekly triage and you caught the issue before we could take a look at possible fixes and post a reply. Very much appreciate the pro-active fixes!

Thanks for the feedback.. I have parsed out the environment name from the Workspace.Environment property using the resource parsing lib. 

Though this brings up the question on if we should be adding env as an optional attribute to the delete command like resource group is.  Just food for thought.

pkg/cli/cmd/app/delete/delete.go
```diff
@@ -121,6 +123,16 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Lopkup the environment name for use in the confirmation prompt
```
shalabhms, line 126:
```suggestion
	// Lookup the environment name for use in the confirmation prompt
```

shalabhms:
@jhandel - thanks for making the contribution . Could you please update the description which is pointing to your earlier commit changes?

jhandel:
> @jhandel - thanks for making the contribution . Could you please update the description which is pointing to your earlier commit changes?

done....  and thank you for the review of the PR.

shalabhms (approved):

ytimocin (approved):
Thanks for your contribution @jhandel !

PR #7108, opened 2024-02-01
Use unique resource names in test Terraform Recipe

kachawla:
# Description

Trying a fix for https://github.com/radius-project/radius/issues/7060. Resource name uniqueness is tied to resource group right now, potentially causing concurrency issues.

## Type of change

Attempting fix for https://github.com/radius-project/radius/issues/7060

### DIFF
--- a/test/functional/shared/resources/testdata/recipes/test-terraform-recipes/azure-storage/main.tf
+++ b/test/functional/shared/resources/testdata/recipes/test-terraform-recipes/azure-storage/main.tf
@@ -7,8 +7,12 @@ terraform {
   }
 }
 
+resource "random_id" "unique_name" {
+  byte_length = 8
+}
+
 resource "azurerm_storage_account" "test_storage_account" {
-  name = var.name
+  name = "acct${random_id.unique_name.hex}"
   resource_group_name = var.resource_group_name
   location = var.location
   account_tier = "Standard"
@@ -16,12 +20,12 @@ resource "azurerm_storage_account" "test_storage_account" {
 }
 
 resource "azurerm_storage_container" "test_container" {
-  name = "test-container"
+  name = "ctr${random_id.unique_name.hex}"
   storage_account_name = azurerm_storage_account.test_storage_account.name
 }
 
 resource "azurerm_storage_blob" "test_blob" {
-  name = "test-blob"
+  name = "blob${random_id.unique_name.hex}"
   storage_account_name = azurerm_storage_account.test_storage_account.name
   storage_container_name = azurerm_storage_container.test_container.name
   type = "Block"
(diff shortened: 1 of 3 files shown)


test/functional/shared/resources/testdata/recipes/test-terraform-recipes/azure-storage/main.tf
```diff
@@ -7,21 +7,25 @@ terraform {
...
 }
 
 resource "azurerm_storage_container" "test_container" {
-  name = "test-container"
+  name = "container-${random_id.unique_name.id}"
   storage_account_name = azurerm_storage_account.test_storage_account.name
 }
 
 resource "azurerm_storage_blob" "test_blob" {
-  name = "test-blob"
```
ytimocin, line 24: Good catch. I think it might be related to this or `test-container` above. I don't think it is due to another long-running test deleting it but a functional test that happens to be running at the same time might be deleting one of these.

ytimocin (dismissed):
Are you going to add anything else to this PR @kachawla? So far LGTM. We can test it out.

ytimocin:
Just saw this error: `"terraform destroy failure: exit status 1\n\nError: name (\"account-a_dH0Fo4N8E\") can only consist of lowercase letters and numbers, and must be between 3 and 24 characters long\n\n  with module.default.azurerm_storage_account.test_storage_account,\n  on .terraform/modules/default/main.tf line 15, in resource \"azurerm_storage_account\" \"test_storage_account\":\n  15:   name = \"account-${random_id.unique_name.id}\"\n\n"`

kachawla:
> Are you going to add anything else to this PR @kachawla? So far LGTM. We can test it out.

No, just wanted to check that it runs successfully with the updated names, but looks like need to make it compliant with naming restrictions, will look into it and update.

test/functional/shared/resources/testdata/recipes/test-terraform-recipes/azure-storage/main.tf
kachawla: Interesting. Are we sharing resource groups between functional tests and long running tests?

The failure is due to `test-container` but I just decided to add uniqueness to all resource names.

kachawla:
> > Are you going to add anything else to this PR @kachawla? So far LGTM. We can test it out.
> 
> No, just wanted to check that it runs successfully with the updated names, but looks like we need to make it compliant with naming restrictions, will look into it and update.

Updated

lakshmimsft (approved):

test/functional/shared/resources/testdata/recipes/test-terraform-recipes/azure-storage/main.tf
ytimocin: > Interesting. Are we sharing resource groups between functional tests and long running tests?

No, we are not actually. They all belong to a resource group, right? Can't seem to find out why this is caused for now.

PR #7125, opened 2024-02-02
Add connection support for ACI containers

willdavsmith:
* Add connection support for ACI containers


pkg/corerp/renderers/aci/renderer.go
```diff
@@ -322,5 +391,62 @@ func (r Renderer) Render(ctx context.Context, dm v1.DataModelInterface, options
 	return renderers.RendererOutput{
 		Resources:      orResources,
 		RadiusResource: dm,
+		ComputedValues: map[string]rpv1.ComputedValueReference{
+			// TODO: Add the computed values here
```
willdavsmith, line 395: Here is where we can add connection information

youngbupark (approved):

PR #7154, opened 2024-02-09
Update Namespace.ValidateNamespace to add workspace to arguments

### LINKED ISSUE #5698: Environment validation in `rad env create` is wrong

1. Create two kubernetes clusters and install Radius on one of them (Cluster A)
2. Create radius workspace pointing to Cluster A
3. Create namespace `test-ns` on Cluster A (`kubectl create ns test-ns`)
4. Set Cluster B as your default cluster in Kubeconfig
5. Run `rad env create kubernetes test-env --namespace test-ns`

The `test-ns` namespace will be created on Cluster B by the `rad` CLI.

The `test-ns` namespace should be created on Cluster A. 

### Workaround (optional)

Use kubectl 😆 

### rad Version (required)

v0.21

## Additional context

The bug is [here](https://github.com/project-radius/radius/blob/4d80a343b4c19e0374439a2e067137ea620a022e/pkg/cli/cmd/env/namespace/namespace.go#LL34C1-L34C1). This code is using the current Kubernetes context rather than the context set in the workspace.

Outside of installation, our code should never use the default Kubernetes context. It should always come from the workspace. 

[AB#8248](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/8248)

jhandel:
Signed-off-by: Josh <josh@liveoak.ws>

# Description

in order for the non current context to be used by ValidateNamespaces so that it can create namespaces in the workspaces kubernetes context we need to pass in the workspace and then use workspace.KuberneteContext() to look up the correct context.

## Type of change

This pull request fixes a bug in Radius and has an approved issue (issue link required).

Fixes: #5698

### DIFF
--- a/pkg/cli/cmd/env/create/create.go
+++ b/pkg/cli/cmd/env/create/create.go
@@ -141,7 +141,7 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	err = r.NamespaceInterface.ValidateNamespace(cmd.Context(), r.Namespace)
+	err = r.NamespaceInterface.ValidateNamespace(cmd.Context(), r.Namespace, *r.Workspace)
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/cmd/env/create/create_test.go
+++ b/pkg/cli/cmd/env/create/create_test.go
@@ -271,13 +271,13 @@ func createMocksWithInvalidResourceGroup(namespaceClient *namespace.MockInterfac
 
 func createValidateNamespaceSuccess(namespaceClient *namespace.MockInterface) {
 	namespaceClient.EXPECT().
-		ValidateNamespace(gomock.Any(), "testingenv").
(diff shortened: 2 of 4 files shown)


ytimocin:
Thanks for your contribution @jhandel! You should run `make generate`. Please see: https://github.com/radius-project/radius/actions/runs/7847801524/job/21417564330?pr=7154.

jhandel:
> Thanks for your contribution @jhandel! You should run `make generate`. Please see: https://github.com/radius-project/radius/actions/runs/7847801524/job/21417564330?pr=7154.

sorry about that.. I thought I read somewhere that the generated code wasn't checked in so it could always be re-generated so I kind of ignored that assuming the build would run the generator.. 

Anyways, generated the updated code and checked..

ytimocin (approved):

ytimocin:
Thanks for your contribution @jhandel !

PR #7157, opened 2024-02-09
Do not try to load recipes for a simulated environment

### LINKED ISSUE #7156: Do not try to load recipes for simulated environments
### Steps to reproduce

1. There are no recipes in the simulated environment
2. rad app deploy <bicep file> to this env
3. Deployment fails

### Observed behavior

Currently, for a simulated env, if a recipe if not found in the env, the deployment fails.

### Desired behavior

This failure is unnecessary for simulated environments as there are no output resources rendered i.e. the result with or without loading the recipe is the same. Therefore, skip the loading of recipes in the case of simulated environments

### Workaround

_No response_

### rad Version

0.30.0

### Operating system

macOS

### Additional context

_No response_

[AB#11143](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11143)

vinayada1:
# Description

Currently, for a simulated env, if a recipe if not found in the env, the deployment fails. This is unnecessary for simulated environments as there are no output resources rendered i.e. the result with or without loading the recipe is the same. Therefore, skip the loading of recipes in the case of simulated environments

## Type of change

Fixes: #7156

### DIFF
--- a/pkg/recipes/driver/bicep.go
+++ b/pkg/recipes/driver/bicep.go
@@ -126,11 +126,6 @@ func (d *bicepDriver) Execute(ctx context.Context, opts ExecuteOptions) (*recipe
 		logger.Info("using Azure provider", "deploymentID", deploymentID, "scope", providerConfig.Az.Value.Scope)
 	}
 
-	if opts.Configuration.Simulated {
-		logger.Info("simulated environment enabled, skipping deployment")
-		return nil, nil
-	}
-
 	poller, err := d.DeploymentClient.CreateOrUpdate(
 		ctx,
 		clients.Deployment{
--- a/pkg/recipes/engine/engine.go
+++ b/pkg/recipes/engine/engine.go
@@ -28,6 +28,7 @@ import (
 	recipedriver "github.com/radius-project/radius/pkg/recipes/driver"
 	"github.com/radius-project/radius/pkg/recipes/util"
 	rpv1 "github.com/radius-project/radius/pkg/rp/v1"
+	"github.com/radius-project/radius/pkg/ucp/ucplog"
 )
 
 // NewEngine creates a new Engine to deploy recipe.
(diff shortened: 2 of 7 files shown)


pkg/recipes/engine/engine.go
```diff
@@ -119,14 +128,18 @@ func (e *engine) Delete(ctx context.Context, opts DeleteOptions) error {
...
 // Any changes to the core logic of the Delete function should be made here.
 func (e *engine) deleteCore(ctx context.Context, recipe recipes.ResourceMetadata, outputResources []rpv1.OutputResource) (*recipes.EnvironmentDefinition, error) {
-	definition, driver, err := e.getDriver(ctx, recipe)
+	configuration, err := e.options.ConfigurationLoader.LoadConfiguration(ctx, recipe)
 	if err != nil {
 		return nil, err
 	}
 
-	configuration, err := e.options.ConfigurationLoader.LoadConfiguration(ctx, recipe)
+	if configuration.Simulated {
```
kachawla, line 136: A log here will be helpful as well for debugging, similar to what you have above in execute.

pkg/recipes/driver/bicep.go
```diff
@@ -126,11 +126,6 @@ func (d *bicepDriver) Execute(ctx context.Context, opts ExecuteOptions) (*recipe
 		logger.Info("using Azure provider", "deploymentID", deploymentID, "scope", providerConfig.Az.Value.Scope)
 	}
 
-	if opts.Configuration.Simulated {
```
kachawla, line 129: We should delete this unit test https://github.com/radius-project/radius/blob/27c1633311f5128e9ac44b143e63761f85f6000a/pkg/recipes/driver/bicep_test.go#L381? Doesn't look like it was doing what we expected it to, since it isn't failing now. :) Same for terraform driver.

pkg/recipes/engine/engine.go
```diff
@@ -71,14 +72,22 @@ func (e *engine) Execute(ctx context.Context, opts ExecuteOptions) (*recipes.Rec
...
+
+	configuration, err := e.options.ConfigurationLoader.LoadConfiguration(ctx, recipe)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, recipes.NewRecipeError(recipes.RecipeConfigurationFailure, err.Error(), util.RecipeSetupError, recipes.GetErrorDetails(err))
 	}
 
-	configuration, err := e.options.ConfigurationLoader.LoadConfiguration(ctx, recipe)
+	// No need to try executing the recipe if it's a simulated environment.
+	if configuration.Simulated {
```
kachawla, line 83: Let's add unit test for this..

kachawla (commented):
Looks great! Couple of quick comments.

Do we have any functional tests for simulated environments scenarios or plan to add any?

pkg/recipes/driver/bicep.go
vinayada1: These tests are actually failing which is good :) But makes sense to remove them now

kachawla (commented):
Thanks for addressing the feedback.

* Is there a blocker that prevented us from adding unit tests for: https://github.com/radius-project/radius/blob/ed54654b6b46c3684b9d6264b7ebc2c4f7a2c4dd/pkg/recipes/configloader/environment.go#L114
(Not directly related to this PR, but related to simulated env test coverage and the code this PR is consuming)

* Do we have any functional tests for simulated environments scenarios or plan to add any?

vinayada1:
> Thanks for addressing the feedback.
> 
> * Is there a blocker that prevented us from adding unit tests for: https://github.com/radius-project/radius/blob/ed54654b6b46c3684b9d6264b7ebc2c4f7a2c4dd/pkg/recipes/configloader/environment.go#L114
                   Thanks for pointing out. Added one now.
>   
>   (Not directly related to this PR, but related to simulated env test coverage and the code this PR is consuming)
> * Do we have any functional tests for simulated environments scenarios or plan to add any?
                   test/functional/datastoresrp/resources/simulated_environment_test.go has some tests. There are UTs too.

kachawla (approved):

PR #7167, opened 2024-02-13
Adding support for terraform private module source for git

### LINKED ISSUE #6911: Support for Private Terraform Modules

Radius can work with publicly-hosted Terraform modules across a few different module sources. We don't support working with privately-hosted Terraform modules, and there's no way to configure authentication.

This is important because organizations write their own Terraform modules and store them in privately-accessible sources. In order for us to support serious use, we need to enable private sources.

The challenge is that each module source has its own authentication mechanism with its own style of configuration. For example the `git` module source relies on the environment hosting Terraform to provide the credentials. 

**For right now the set of required sources is TBD.**

## Acceptance criteria

Other than the "glue" to wire up credentials, the Terraform Recipe driver should behave the same way for a private module. Given the correct configuration of credentials, the Terraform driver should behave the same for a public vs private module.

- Radius API allows users to register credentials for some set of Terraform module sources.
- TBD set of Terraform module sources are supported.
- Documentation updated.

## Additional context

It may be helpful to pair this with the work to isolate the Terraform driver in its own container. Some of the source configuration makes changes to the execution environment (for example logging into git). 

[AB#10652](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10652)

vishwahiremat:
# Description

- Added a RecipeConfig property to the environment
  - Typesspec changes
  - Conversion and unit tests
  - Datamodel changes
- Adding changes to the config to append the template path with credential informaiton
   - changes to query list secret to get secret information
   - creating credential appended template path

Design Doc: https://github.com/radius-project/design-notes/pull/37

## Type of change

Fixes: https://github.com/radius-project/radius/issues/6911


pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -63,6 +64,24 @@ func (src *EnvironmentResource) ConvertTo() (v1.DataModelInterface, error) {
 	}
 	converted.Properties.Compute = *envCompute
 
+	recipeConfig := src.Properties.RecipeConfig
+	if recipeConfig != nil {
+		if recipeConfig.Terraform != nil {
+			if recipeConfig.Terraform.Authentication != nil {
```
sk593, line 70: Can these conditions be collapsed? Or do we expect that there may be future logic that's unique to each config level?

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -151,6 +170,27 @@ func (dst *EnvironmentResource) ConvertFrom(src v1.DataModelInterface) error {
 		dst.Properties.Recipes = recipes
 	}
 
+	if !reflect.DeepEqual(env.Properties.RecipeConfig, datamodel.RecipeConfigProperties{}) {
+		recipeConfig := &RecipeConfigProperties{}
+		if !reflect.DeepEqual(env.Properties.RecipeConfig.Terraform, datamodel.TerraformConfigProperties{}) {
```
sk593, line 175: Same as above, since this logic is specific to Terraform, maybe we can move them to a helper function

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -63,6 +64,24 @@ func (src *EnvironmentResource) ConvertTo() (v1.DataModelInterface, error) {
 	}
 	converted.Properties.Compute = *envCompute
 
+	recipeConfig := src.Properties.RecipeConfig
+	if recipeConfig != nil {
+		if recipeConfig.Terraform != nil {
```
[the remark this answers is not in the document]
ytimocin: I agree with @sk593. We can have a function that takes care of the recipeConfig conversion and that function can make a call to another function which takes care of terraform related stuff conversion. With this approach we could keep things simple. It would be more readable, easier to debug and write unit tests.

pkg/corerp/api/v20231001preview/environment_conversion.go
ytimocin: We can apply the same thing as I mentioned above but this time in reverse.

**This function -> RecipeConfig conversion function -> Terraform related conversion function**

In future, it would be very easy to add new types of provider conversions.

pkg/corerp/api/v20231001preview/environment_conversion_test.go
```diff
@@ -117,6 +117,19 @@ func TestConvertVersionedToDataModel(t *testing.T) {
 							Scope: "/planes/aws/aws/accounts/140313373712/regions/us-west-2",
 						},
 					},
+					RecipeConfig: datamodel.RecipeConfigProperties{
```
ytimocin, line 120: This needs a few different layers and we could run into nil pointer exception easily. We should test it thoroughly.

pkg/corerp/api/v20231001preview/testdata/environmentresource.json
```diff
@@ -16,6 +16,19 @@
...
     },
+    "recipeConfig": {
+      "terraform": {
+        "authentication": {
+          "git": {
+            "pat": {
+              "dev.azure.com":{
+                "secretStore":"/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/github"
+              }
+            }
```
ytimocin, line 27: What if this part doesn't exist (by mistake of the operator), would this cause a nil pointer exception?

pkg/corerp/datamodel/environment.go
```diff
@@ -38,11 +38,42 @@ func (e *Environment) ResourceTypeName() string {
...
 type EnvironmentProperties struct {
-	Compute    rpv1.EnvironmentCompute                           `json:"compute,omitempty"`
-	Recipes    map[string]map[string]EnvironmentRecipeProperties `json:"recipes,omitempty"`
-	Providers  Providers                                         `json:"providers,omitempty"`
-	Extensions []Extension                                       `json:"extensions,omitempty"`
-	Simulated  bool                                              `json:"simulated,omitempty"`
+	Compute      rpv1.EnvironmentCompute                           `json:"compute,omitempty"`
+	Recipes      map[string]map[string]EnvironmentRecipeProperties `json:"recipes,omitempty"`
+	Providers    Providers                                         `json:"providers,omitempty"`
+	RecipeConfig RecipeConfigProperties                            `json:"recipeConfig,omitempty"`
```
ytimocin, line 44: I am a fan of adding new properties to the end. Would the order matter here?

pkg/corerp/datamodel/environment.go
```diff
@@ -38,11 +38,42 @@ func (e *Environment) ResourceTypeName() string {
...
+	Extensions   []Extension                                       `json:"extensions,omitempty"`
+	Simulated    bool                                              `json:"simulated,omitempty"`
+}
+
+// RecipeConfigProperties - Specifies recipe configurations needed for the recipes.
+type RecipeConfigProperties struct {
+	// Specifies the terraform config properties
```
ytimocin, line 51: We should start with the name of the property. All docs should follow that structure.

pkg/recipes/configloader/environment.go
```diff
@@ -101,14 +101,15 @@ func getConfiguration(environment *v20231001preview.EnvironmentResource, applica
...
-		}
+	env, err := environment.ConvertTo()
+	if err != nil {
+		return nil, err
+	}
+	if environment.Properties.Providers != nil {
+		config.Providers = env.(*datamodel.Environment).Properties.Providers
+	}
+	if environment.Properties.RecipeConfig != nil {
+		config.RecipeConfig = env.(*datamodel.Environment).Properties.RecipeConfig
```
ytimocin, line 112: I see that we do this twice: `env.(*datamodel.Environment)`. Should we do it once before the if cases and use that within them?

pkg/recipes/terraform/config/config.go
```diff
@@ -38,10 +44,12 @@ const (
 
 // New creates TerraformConfig with the given module name and its inputs (module source, version, parameters)
 // Parameters are populated from environment recipe and resource recipe metadata.
-func New(moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRecipe *recipes.ResourceMetadata) *TerraformConfig {
+func New(ctx context.Context, moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRecipe *recipes.ResourceMetadata, envConfig *recipes.Configuration, armOptions *arm.ClientOptions) *TerraformConfig {
+	// Getting the module source path with credentials information if its a private source.
+	path, _ := getModuleSource(ctx, envConfig, envRecipe.TemplatePath, armOptions)
```
ytimocin, line 49: Are we ignoring the error here?

pkg/recipes/terraform/config/config.go
```diff
@@ -52,6 +60,84 @@ func New(moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRe
...
+		token, ok := secrets.Data["pat"]
+		if ok {
+			pat = token.Value
+			path += *pat
+		}
+		path += fmt.Sprintf("@%s", strings.TrimPrefix(url.String(), "https://"))
+		return path, err
+	}
+
+	return templatePath, err
```
ytimocin, line 119: You could do the other case first: `if secretStore != "" { return templatePath, err }`. Then the rest doesn't need to be within an if case. I find this more readable: **first exit cases then the actual logic**.

pkg/recipes/terraform/config/config_test.go
```diff
@@ -533,9 +534,54 @@ func Test_Save_InvalidWorkingDir(t *testing.T) {
...
 
-	tfconfig := New(testRecipeName, &envRecipe, &resourceRecipe)
+	tfconfig := New(context.Background(), testRecipeName, &envRecipe, &resourceRecipe, nil, nil)
 
 	err := tfconfig.Save(testcontext.New(t), testDir)
 	require.Error(t, err)
 	require.Equal(t, fmt.Sprintf("error creating file: open %s/main.tf.json: no such file or directory", testDir), err.Error())
 }
+
+func Test_getSecretStoreID(t *testing.T) {
```
ytimocin, line 544: We should add unit tests for all the new functions.

pkg/recipes/terraform/execute.go
```diff
@@ -268,6 +271,8 @@ func downloadAndInspect(ctx context.Context, tf *tfexec.Terraform, options Optio
 	// Download the Terraform module to the working directory.
 	logger.Info(fmt.Sprintf("Downloading Terraform module: %s", options.EnvRecipe.TemplatePath))
 	downloadStartTime := time.Now()
+
+	//path := options.EnvRecipe.TemplatePath
```
ytimocin, line 275: Can be removed.

pkg/recipes/terraform/config/config.go
```diff
@@ -52,6 +60,84 @@ func New(moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRe
...
+		}
+	} else {
+		source = templatePath
+	}
+	return source, err
+}
+
+// getGitSource creates a terraform module source in a generic git format with credential information
+// e.g: git::https://<username>:<pat>@<example.com>/<repo_path>
+func getGitSource(ctx context.Context, envConfig *recipes.Configuration, templatePath string, armOptions *arm.ClientOptions) (string, error) {
```
sk593, line 81: Are we only supporting private registries that are accessed over HTTPS? Terraform supports other Git URLS (just looking at this: https://developer.hashicorp.com/terraform/language/modules/sources#github) so I'm wondering what the expected behavior would be if a user passed one of these formats

pkg/recipes/terraform/config/config.go
```diff
@@ -52,6 +60,84 @@ func New(moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRe
...
+		}
+		path += fmt.Sprintf("@%s", strings.TrimPrefix(url.String(), "https://"))
+		return path, err
+	}
+
+	return templatePath, err
+}
+
+func getGitURL(templatePath string) (*url.URL, error) {
+	paths := strings.Split(templatePath, "git::")
```
sk593, line 123: Similar question as above, what happens if a user tries to use Git over a different protocol (github.com/hashicorp/example or git@github.com:hashicorp/example.git or over SSH)? We should add some sort of error message if an incorrect Git module source format is used

pkg/recipes/terraform/config/config.go
```diff
@@ -52,6 +60,84 @@ func New(moduleName string, envRecipe *recipes.EnvironmentDefinition, resourceRe
 	}
 }
 
+// getModuleSource is called to get the module source path with credential information
+// if template path represent a private source else returns the provided template path
+func getModuleSource(ctx context.Context, envConfig *recipes.Configuration, templatePath string, armOptions *arm.ClientOptions) (string, error) {
+	var source string
+	var err error
+	if strings.HasPrefix(templatePath, "git::") {
```
sk593, line 68: Are we only supporting private registries that are accessed over HTTPS? Terraform supports other Git URLS (just looking at this: https://developer.hashicorp.com/terraform/language/modules/sources#github) so I'm wondering what the expected behavior would be if a user passed one of these formats

vishwahiremat: thats correct, today we don't support private registries that are accessed over ssh. And for private git repository we expect user to provide the source in generic git format so all git platforms are supported.

pkg/corerp/datamodel/environment.go
vishwahiremat: not really but i added before recipes because recipeConfig is usually configured before recipes.

pkg/corerp/api/v20231001preview/environment_conversion.go
vishwahiremat: created a separate func

pkg/corerp/api/v20231001preview/environment_conversion.go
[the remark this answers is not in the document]
vishwahiremat: done

pkg/corerp/api/v20231001preview/environment_conversion.go
vishwahiremat: In any case i think we still need to check nil at every level to avoid null pointer. Let me know if you think otherwise

pkg/corerp/api/v20231001preview/testdata/environmentresource.json
vishwahiremat: I dont think it will nil pointer if pat property doesnt exist

pkg/corerp/datamodel/environment.go
vishwahiremat: done

pkg/recipes/configloader/environment.go
vishwahiremat: done

pkg/recipes/terraform/config/config.go
vishwahiremat: good catch! i handled it my latest pr

pkg/recipes/terraform/config/config_test.go
vishwahiremat: I dont if we can add unit for `getGitSource` func but i added for the other ones

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
```diff
@@ -2745,6 +2745,10 @@
             "type": "object"
           }
         },
+        "recipeConfig": {
```
kachawla, line 2748: Do you know why recipe related properties are getting added to Application properties? I understand it's not directly related to this PR, but weird to see it here.

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
```diff
@@ -4966,6 +5011,16 @@
         }
       }
     },
+    "Secret": {
+      "type": "object",
+      "description": "Specifies the secret details of type personal access token for each different git platforms",
+      "properties": {
+        "secretStore": {
```
kachawla, line 5018: From the design I thought this was expected to be `secret` and not `secretStore` https://github.com/radius-project/design-notes/blob/3644b754152edc97e641f537b20cf3d87a386c43/recipe/2024-01-support-private-terraform-repository.md?plain=1#L106. Did we change that or am I'm interpreting this generated schema incorrectly?

vishwahiremat: I think it should be `ApplicationProperties` instead of `EnvironmentProperties`https://github.com/radius-project/radius/blob/0a1a8680dace317f8bd25bdde2b503f895766d2d/typespec/Applications.Core/applications.tsp#L145
Maybe a typo while creating this file. Will create a fix for this in different PR.
Good Catch!!

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
vishwahiremat: I think it should be `ApplicationProperties` instead of `EnvironmentProperties`https://github.com/radius-project/radius/blob/0a1a8680dace317f8bd25bdde2b503f895766d2d/typespec/Applications.Core/applications.tsp#L145
Maybe a typo while creating this file. Will create a fix for this in different PR.
Good Catch!!

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -180,6 +183,59 @@ func (dst *EnvironmentResource) ConvertFrom(src v1.DataModelInterface) error {
...
+				recipeConfig.Terraform.Authentication = &AuthConfig{}
+				if !reflect.DeepEqual(config.Terraform.Authentication.Git, datamodel.GitAuthConfig{}) {
+					recipeConfig.Terraform.Authentication.Git = &GitAuthConfig{}
+					if config.Terraform.Authentication.Git.PAT != nil {
+						recipeConfig.Terraform.Authentication.Git.Pat = map[string]*Secret{}
+					}
+				}
+			}
+		}
+		for k, v := range config.Terraform.Authentication.Git.PAT {
```
ytimocin, line 229: Would this throw a nil pointer exception if config doesn't have Terraform map? Or Terraform map doesn't have Authentication and so on. We can add unit tests for those cases.

ytimocin: I think this for loop should move to the **last if** in above. See: **if config.Terraform.Authentication.Git.PAT != nil**.

pkg/corerp/api/v20231001preview/environment_conversion_test.go
ytimocin: If we remove the PAT map from this test, we can see if we run into nil pointer exception above in the original conversion logic.

pkg/recipes/configloader/environment.go
```diff
@@ -101,14 +101,16 @@ func getConfiguration(environment *v20231001preview.EnvironmentResource, applica
...
 
-	providers := environment.Properties.Providers
-	if providers != nil {
-		if providers.Aws != nil {
-			config.Providers.AWS.Scope = to.String(providers.Aws.Scope)
-		}
-		if providers.Azure != nil {
-			config.Providers.Azure.Scope = to.String(providers.Azure.Scope)
-		}
+	env, err := environment.ConvertTo()
```
ytimocin, line 104: Is **ConvertTo** a new function? It doesn't make a lot of sense to me. We want to convert environment to what specifically? Not in the scope of this PR, but we should check this function.

pkg/recipes/configloader/environment_test.go
```diff
@@ -177,6 +177,53 @@ func TestGetConfiguration(t *testing.T) {
...
+								Git: &model.GitAuthConfig{
+									Pat: map[string]*model.Secret{
+										"dev.azure.com": &model.Secret{
+											SecretStore: to.Ptr("secretStoreID"),
+										},
+									},
+								},
+							},
+						},
+					},
```
ytimocin: We should have a few cases that would cause errors. Like what happens if we remove the PAT?

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
vishwahiremat: i changed it to secret.

pkg/recipes/configloader/environment.go
vishwahiremat: i am using the existing convertTo() function from environment_conversions

kachawla (dismissed):

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -180,6 +183,59 @@ func (dst *EnvironmentResource) ConvertFrom(src v1.DataModelInterface) error {
...
+		}
+		return recipeConfig
+	}
+	return datamodel.RecipeConfigProperties{}
+}
+
+func fromRecipeConfigDatamodel(config datamodel.RecipeConfigProperties) *RecipeConfigProperties {
+	if !reflect.DeepEqual(config, datamodel.RecipeConfigProperties{}) {
+		recipeConfig := &RecipeConfigProperties{}
+		if !reflect.DeepEqual(config.Terraform, datamodel.TerraformConfigProperties{}) {
```
kachawla, line 217: Do we need to use reflection? Can this be `if config.Terraform != (datamodel.TerraformConfigProperties{})` instead for example?

pkg/corerp/api/v20231001preview/testdata/environmentresource.json
kachawla: Can we add a unit test for this?

pkg/recipes/configloader/environment.go
kachawla, line 104: Let's add a comment here that it converts it to internal data model since it is not as obvious from the function name.

pkg/recipes/configloader/environment_test.go
```diff
@@ -177,6 +177,53 @@ func TestGetConfiguration(t *testing.T) {
 			},
 			errString: ErrUnsupportedComputeKind.Error(),
 		},
+		{
+			name: "recipe config with env resource",
```
kachawla, line 181: Do we need a new test for this or would it be sufficient add recipe config to any of the exiting tests https://github.com/radius-project/radius/blob/e717d111d1380b733d3e1fdd1dd10c0adc16b504/pkg/recipes/configloader/environment_test.go#L53

pkg/recipes/terraform/config/config.go
```diff
@@ -38,18 +44,102 @@ const (
...
+
+	secretStoreID, err := resources.ParseResource(secretStore)
+	if err != nil {
+		return "", err
+	}
+	client, err := v20231001preview.NewSecretStoresClient(secretStoreID.RootScope(), &aztoken.AnonymousCredential{}, armOptions)
+	if err != nil {
+		return "", err
 	}
+	secrets, err := client.ListSecrets(ctx, secretStoreID.Name(), map[string]any{}, nil)
```
kachawla, line 101: I'm not sure if this is the right place for this. I'd lean towards keeping interactions with CoreRP bounded to driver, and everything within Terraform executor and downstream is responsible for interacting with Terraform. @rynowak do you have any strong preference?

pkg/corerp/api/v20231001preview/environment_conversion.go
vishwahiremat: that kept throwing error we cannot do that, so I had to use reflect.
I dont know if i am missing something, i spent lot of time to get it working the way shown in example or the way we do it for providers but i kept getting the same error

pkg/corerp/datamodel/environment.go
```diff
@@ -38,11 +38,42 @@ func (e *Environment) ResourceTypeName() string {
...
+	Compute      rpv1.EnvironmentCompute                           `json:"compute,omitempty"`
+	Recipes      map[string]map[string]EnvironmentRecipeProperties `json:"recipes,omitempty"`
+	Providers    Providers                                         `json:"providers,omitempty"`
+	RecipeConfig RecipeConfigProperties                            `json:"recipeConfig,omitempty"`
+	Extensions   []Extension                                       `json:"extensions,omitempty"`
+	Simulated    bool                                              `json:"simulated,omitempty"`
+}
+
+// RecipeConfigProperties - Specifies recipe configurations needed for the recipes.
+type RecipeConfigProperties struct {
```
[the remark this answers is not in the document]
vishwahiremat: i agree, i was thinking if we can move it to `pkg/recipes/types.go`?

pkg/corerp/api/v20231001preview/environment_conversion.go
```diff
@@ -180,6 +183,59 @@ func (dst *EnvironmentResource) ConvertFrom(src v1.DataModelInterface) error {
...
+						}
+						recipeConfig.Terraform.Authentication.Git.PAT = p
+					}
+
+				}
+			}
+		}
+		return recipeConfig
+	}
+	return datamodel.RecipeConfigProperties{}
```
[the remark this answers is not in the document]
vishwahiremat:
```
	if config == nil || config.Terraform == nil || config.Terraform.Authentication == nil || config.Terraform.Authentication.Git == nil {
		return datamodel.RecipeConfigProperties{}
	}
```
I think it will be an issue when we have more properties in config e.g bicep config or when we add different modes of Auth ?

pkg/recipes/configloader/environment_test.go
vishwahiremat: makes sense! removed it

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
   @doc("Specifies Recipes linked to the Environment.")
   recipes?: Record<Record<RecipeProperties>>;
 
+  @doc("Specifies configurations needed for the recipes registered to the environment.")
```
AaronCrawfis, line 68:
```suggestion
  @doc("Configuration for Recipes. Defines how each type of Recipe should be configured and run.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  recipeConfig?: RecipeConfigProperties;
+
   @doc("The environment extension.")
   @extension("x-ms-identifiers", [])
   extensions?: Array<Extension>;
 }
 
+@doc("Specifies recipe configurations needed for the recipes.")
```
AaronCrawfis, line 76:
```suggestion
@doc("Configuration for Recipes. Defines how each type of Recipe should be configured and run.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+model RecipeConfigProperties {
+  @doc("Specifies the terraform config properties")
```
AaronCrawfis, line 78: Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  terraform?: TerraformConfigProperties;
+}
+
+@doc("Specifies the terraform config properties")
```
AaronCrawfis, line 82:
```suggestion
@doc("Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+model TerraformConfigProperties{
+  @doc("Specifies authentication information needed to access private terraform module sources. Currently supported module sources: Git.")
```
AaronCrawfis, line 84:
```suggestion
  @doc("Authentication information used to access private Terraform module sources. Supported module sources: Git.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  authentication?: AuthConfig;
+}
+
+@doc("Specifies authentication information needed to access private terraform module repositories.")
```
AaronCrawfis, line 88:
```suggestion
@doc("Authentication information used to access private Terraform module sources. Supported module sources: Git.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+model AuthConfig{
+  @doc("Specifies authentication information needed to access private terraform modules from Git repository sources.")
```
AaronCrawfis, line 90:
```suggestion
  @doc("Authentication information used to access private Terraform modules from Git repository sources.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  git?: GitAuthConfig;
+}
+
+@doc("Specifies authentication information needed to access private terraform modules from Git repository sources.")
```
AaronCrawfis, line 94:
```suggestion
@doc("Authentication information used to access private Terraform modules from Git repository sources.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+model GitAuthConfig{
+  @doc("Specifies the secret details of type personal access token for each different git platforms")
```
AaronCrawfis, line 96:
```suggestion
  @doc("Personal Access Token (PAT) configuration used to authenticate to Git platforms.")  
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  git?: GitAuthConfig;
+}
+
+@doc("Specifies authentication information needed to access private terraform modules from Git repository sources.")
+model GitAuthConfig{
+  @doc("Specifies the secret details of type personal access token for each different git platforms")  
+  pat?: Record<SecretConfig>;
+}
+
+@doc("Specifies the secret details of type personal access token for each different git platforms")
```
AaronCrawfis, line 100:
```suggestion
@doc("Personal Access Token (PAT) configuration used to authenticate to Git platforms.")
```

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+model SecretConfig {
+  @doc("The resource id for the Applications.Core/SecretStore resource containing credentials.")
```
AaronCrawfis, line 102: Could you add some information about what the secret name is required to be in the secret store? Do we expect that there needs to be a secret with a specific name?

AaronCrawfis (changes requested):

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
kachawla: @vishwahiremat let's create a tracking issue so that this gets picked up during next sprint planning.

pkg/corerp/datamodel/environment.go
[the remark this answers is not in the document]
kachawla: sure, I thought it would result into an import cycle, but if that's not the case then moving to /recipe/types sound good

pkg/corerp/api/v20231001preview/environment_conversion.go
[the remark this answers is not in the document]
kachawla: I understand your point, but how do we solve that with the currently implemented reverse checks?

pkg/corerp/datamodel/environment.go
lakshmimsft, line 44: @vishwahiremat Can we update RecipeConfig to be a pointer to RecipeConfigProperties? it can be a potentially  large object when envConfig is passed, pointer will be better.

pkg/corerp/api/v20231001preview/environment_conversion.go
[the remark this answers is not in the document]
kachawla: Discussed offline, we are going to keep it as is to be able to support more properties in recipe config in the future.

pkg/recipes/terraform/config/config.go
kachawla: We discussed this in the design review, this will be moved to the config loader layer.

pkg/corerp/api/v20231001preview/environment_conversion.go
kachawla: I agree with creating a Terraform related conversion function. @vishwahiremat I know you have updated it to recipe config specific conversion, have you considered adding Terraform specific function to make it easy to extend the recipe config function in the future?

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
vishwahiremat: I have created a PR for this here :https://github.com/radius-project/radius/pull/7182

kachawla: Just noticed https://github.com/radius-project/radius/pull/7182/files. Thanks for following up on this quickly.

pkg/corerp/datamodel/environment.go
vishwahiremat: discussed it offline! resolving the comment

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+  recipeConfig?: RecipeConfigProperties;
+
   @doc("The environment extension.")
   @extension("x-ms-identifiers", [])
   extensions?: Array<Extension>;
 }
 
+@doc("Configuration for Recipes. Defines how each type of Recipe should be configured and run.")
+model RecipeConfigProperties {
+  @doc("Specifies the terraform config properties")
```
AaronCrawfis, line 78:
```suggestion
  @doc("Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.")
```

pkg/recipes/terraform/execute.go
```diff
@@ -61,6 +62,8 @@ type executor struct {
 
 	// k8sClientSet is the Kubernetes client.
 	k8sClientSet kubernetes.Interface
+
+	armOptions *arm.ClientOptions
```
kachawla, line 66: I don't this is used anymore in this terraform package?

typespec/Applications.Core/environments.tsp
```diff
@@ -65,11 +65,44 @@ model EnvironmentProperties {
...
+
+@doc("Authentication information used to access private Terraform modules from Git repository sources.")
+model GitAuthConfig{
+  @doc("Personal Access Token (PAT) configuration used to authenticate to Git platforms.")  
+  pat?: Record<SecretConfig>;
+}
+
+@doc("Personal Access Token (PAT) configuration used to authenticate to Git platforms.")
+model SecretConfig {
+  @doc("The resource id for the Applications.Core/SecretStore resource containing credentials. Secret names required: 'pat', and 'username' is optional")
```
AaronCrawfis, line 102:
```suggestion
  @doc("The ID of an Applications.Core/SecretStore resource containing the Git platform personal access token (PAT). The secret store must have a secret named 'pat', containing the PAT value. A secret named 'username' is optional, containing the username associated with the pat. By default no username is specified.")
```

pkg/recipes/controllerconfig/config.go
```diff
@@ -88,6 +88,7 @@ func New(options hostoptions.HostOptions) (*RecipeControllerConfig, error) {
 
 	cfg.ConfigLoader = configloader.NewEnvironmentLoader(clientOptions)
 	cfg.Engine = engine.NewEngine(engine.Options{
+		SecretsLoader:       configloader.NewSecretStoreLoader(clientOptions),
 		ConfigurationLoader: cfg.ConfigLoader,
 		Drivers: map[string]driver.Driver{
 			recipes.TemplateKindBicep: driver.NewBicepDriver(
```
kachawla, line 94: nit

```suggestion
	cfg.Engine = engine.NewEngine(engine.Options{
		SecretsLoader:       configloader.NewSecretStoreLoader(clientOptions),
		ConfigurationLoader: cfg.ConfigLoader,
		Drivers: map[string]driver.Driver{
			recipes.TemplateKindBicep: driver.NewBicepDriver(
```
```suggestion
	cfg.Engine = engine.NewEngine(engine.Options{
		ConfigurationLoader: cfg.ConfigLoader,
		SecretsLoader:       configloader.NewSecretStoreLoader(clientOptions),
		Drivers: map[string]driver.Driver{
			recipes.TemplateKindBicep: driver.NewBicepDriver(
```

kachawla (commented):
Still reviewing.. few comments on what I have so far.

pkg/recipes/configloader/environment_test.go
```diff
@@ -63,6 +63,19 @@ func TestGetConfiguration(t *testing.T) {
 							Scope: to.Ptr(azureScope),
 						},
 					},
+					RecipeConfig: &model.RecipeConfigProperties{
+						Terraform: &model.TerraformConfigProperties{
+							Authentication: &model.AuthConfig{
+								Git: &model.GitAuthConfig{
+									Pat: map[string]*model.SecretConfig{
+										"dev.azure.com": &model.SecretConfig{
+											Secret: to.Ptr("secretStoreID"),
```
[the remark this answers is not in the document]
vishwahiremat: updated it

pkg/recipes/terraform/execute.go
vishwahiremat: removed it

pkg/recipes/driver/gitconfig.go
```diff
@@ -0,0 +1,96 @@
...
+
+	path += fmt.Sprintf("@%s", url.Hostname())
+
+	// git config key will be in the format of url.<git url with secret details>.insteadOf
+	// and value returned will the original git url domain, e.g github.com
+	return fmt.Sprintf("url.%s.insteadOf", path), url.Hostname(), nil
+}
+
+// Add the git credentials information to .gitconfig by running
+// git config --global url<template_path_domain_with_credentails>.insteadOf <template_path_domain>
```
kachawla, line 60: A suggestion on adding more details to the comment

```suggestion
// Updates the global Git configuration with credentials for a recipe template path and prefixes the path with environment, application, and resource name to make the entry unique to each recipe execution operation.
//
// Retrieves the git credentials from the provided secrets object
// and adds them to the Git config by running
// git config --global url<template_path_domain_with_credentails>.insteadOf <template_path_domain>.
```

pkg/recipes/controllerconfig/config.go
```diff
@@ -102,7 +103,7 @@ func New(options hostoptions.HostOptions) (*RecipeControllerConfig, error) {
 			recipes.TemplateKindTerraform: driver.NewTerraformDriver(options.UCPConnection, provider.NewSecretProvider(options.Config.SecretProvider),
 				driver.TerraformOptions{
 					Path: options.Config.Terraform.Path,
-				}, cfg.K8sClients.ClientSet),
```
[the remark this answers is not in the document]
vishwahiremat: good catch! not needed anymore. removed it

pkg/recipes/configloader/secrets.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
+	"github.com/radius-project/radius/pkg/ucp/resources"
+)
+
+// NewSecretStoreLoader creates a new SecretsLoader instance with the given ARM Client Options.
+func NewSecretStoreLoader(armOptions *arm.ClientOptions) SecretsLoader {
+	return SecretsLoader{ArmClientOptions: armOptions}
+}
+
+// SecretsLoader struct provides functionality to get secret information from secret stores.
```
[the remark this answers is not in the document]
vishwahiremat: updated it

pkg/recipes/types_test.go
```diff
@@ -80,3 +81,227 @@ func TestRecipeOutput_PrepareRecipeResponse(t *testing.T) {
...
+				require.Equal(t, tt.expEnv, env)
+				require.Equal(t, tt.expResource, res)
+			} else {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "not a valid resource id")
+			}
+		})
+	}
+}
+func Test_GetGitURL(t *testing.T) {
```
kachawla, line 173:
```suggestion
}

func Test_GetGitURL(t *testing.T) {
```

pkg/recipes/types_test.go
```diff
@@ -80,3 +81,227 @@ func TestRecipeOutput_PrepareRecipeResponse(t *testing.T) {
...
+				require.NoError(t, err)
+				require.Equal(t, tt.expectedURL, url.String())
+			} else {
+				require.Error(t, err)
+			}
+		})
+	}
+
+}
+func Test_GetSecretStoreID(t *testing.T) {
```
kachawla, line 205:
```suggestion
}

func Test_GetSecretStoreID(t *testing.T) {
```

kachawla (commented):
Looks good overall. A few minor comments.

typespec/Applications.Core/environments.tsp
kachawla, line 1: Should the example be updated as well https://github.com/radius-project/radius/blob/main/typespec/Applications.Core/examples/2023-10-01-preview/Environments_CreateOrUpdate.json?

AaronCrawfis (dismissed):
Type descriptions look good to me!

typespec/Applications.Core/environments.tsp
vishwahiremat: updated with examples

pkg/corerp/api/v20231001preview/environment_conversion.go
kachawla: Vishwa and I discussed offline, we will abstract it out into Terraform specific function once we add support for other types.

kachawla (dismissed):

kachawla (approved):

PR #7186, opened 2024-02-21
Add dashboard to Radius installation and rad run

### LINKED ISSUE #6951: Simple dashboard: app graph visualization

As a developer or IT operator, when deploying an application using Radius, I would like Radius to also deploy a single-container simple dashboard that displays a visualization of my application graph data so that I may view a summary of my application's architecture. Additionally, I would like the option to opt-out of this dashboard deployment.

## Acceptance criteria

Step 1. Deploy an application and its resources, along with connections and routes using Radius.
Step 2. After running `rad run` to start the application, receive a network address for the application graph dashboard after the application resources and dashboard have completed deployment.
Step 3. Navigate to the application graph dashboard and view application graph visualization.

willdavsmith:
# Description

* Add dashboard install to `rad install kubernetes` and `rad init`
* Add dashboard portforwarding to `rad run`

## Type of change

Fixes: https://github.com/radius-project/radius/issues/6951


deploy/Chart/templates/dashboard/deployment.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if .Values.dashboard.enabled }}
```
willdavsmith, line 1: Is there a better way to do this? (ignore the whole subfolder)

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: dashboard-role
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: dashboard-role
+    app.kubernetes.io/part-of: radius
+rules:
+  - apiGroups: ['', 'api.ucp.dev']
+    resources: ['*']
+    verbs: ['*']
```
AaronCrawfis, line 13: Do we want to limit the dashboard to just GET for now, since the dashboard shouldn't do any write operations? Will help with a least-privilege security model.

deploy/Chart/values.yaml
```diff
@@ -71,3 +71,14 @@ rp:
     deleteRetryDelaySeconds: 60
   terraform:
     path: "/terraform"
+
+dashboard:
+  enabled: true
```
AaronCrawfis, line 76: Do we want this to be enabled by default all the time, even when installed by Helm? The dashboard container adds CPU/memory usage to clusters and when running in a production environment operators may not want to add the container and the security footprint that it also represents.

Would it make sense to have `rad init` enable the dashboard (usually used creating a dev environment), but keep it disabled by default when you do `helm upgrade` (usually used when deploying a production Radius)

deploy/Chart/templates/dashboard/rbac.yaml
willdavsmith: sure, will update

deploy/Chart/values.yaml
willdavsmith: deferring to @willtsai

deploy/Chart/templates/dashboard/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: dashboard
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app.kubernetes.io/name: dashboard
+    app.kubernetes.io/part-of: radius
+spec:
+  ports:
+    - name: http
+      port: 80
+      targetPort: http
```
youngbupark, line 14: This must be port number.

```suggestion
      targetPort: 7007
```

I recommend to define variables for port and targetPort in values.yaml and use them instead of hardcoding port

deploy/Chart/templates/dashboard/deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+        app.kubernetes.io/part-of: radius
+    spec:
+      serviceAccountName: dashboard-account
+      containers:
+      - name: dashboard
+        image: "{{ .Values.dashboard.image }}:{{ .Values.dashboard.tag | default $appversion }}"
+        imagePullPolicy: Always
+        ports:
+        - name: http
+          containerPort: 7007
```
youngbupark, line 31: Define variables for port in values.yaml. e.g. dashboard.contaienrPort=7007

deploy/Chart/templates/dashboard/serviceaccount.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.dashboard.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dashboard-account
```
youngbupark, line 5: Please make it consistent with the others by removing suffix.

```suggestion
  name: dashboard
```

deploy/Chart/templates/dashboard/serviceaccount.yaml
```diff
@@ -0,0 +1,10 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: dashboard-account
```
youngbupark, line 8:
```suggestion
    app.kubernetes.io/name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.dashboard.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: dashboard-role
```
youngbupark, line 5:
```suggestion
  name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: dashboard-role
```
youngbupark, line 8:
```suggestion
    app.kubernetes.io/name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/part-of: radius
+rules:
+  - apiGroups: ['', 'api.ucp.dev']
+    resources: ['*']
+    verbs: ['*']
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: dashboard-role-binding
```
youngbupark, line 18:
```suggestion
  name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: dashboard-role-binding
```
youngbupark, line 21:
```suggestion
    app.kubernetes.io/name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/part-of: radius
+subjects:
+- kind: ServiceAccount
+  name: dashboard-account
```
youngbupark, line 25:
```suggestion
  name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: dashboard-role
```
youngbupark, line 30:
```suggestion
  name: dashboard
```

deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: dashboard-role
+    app.kubernetes.io/part-of: radius
+rules:
+  - apiGroups: ['', 'api.ucp.dev']
+    resources: ['*']
+    verbs: ['*']
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
youngbupark, line 15: I am wondering if dashboard requires cluster-scope permission. Ideally, it shouldn't. cc/ @rynowak

youngbupark (commented):
Please make the naming consistent with the existing chart.

deploy/Chart/values.yaml
willtsai: I think there's some value in keeping it consistent with an opt-out model so that users don't have to figure out when it's opt-in and when it would be opt-out. thoughts? @AaronCrawfis 

hypothetical question: would it be possible to have an opt-in mechanism for `helm upgrade`? @willdavsmith

AaronCrawfis: > I think there's some value in keeping it consistent with an opt-out model so that users don't have to figure out when it's opt-in and when it would be opt-out.

I think there are different users though. An operator wanting a production-ready environment may not want the dashboard, at least at this stage of the project. We've done good work on limiting the size of the dashboard image in terms of disc-footprint, CPU, and memory, but it's still a large image. A developer building a new application will definitely want the dashboard to learn and experiment on. But in the end it's definitely an optional component.

For comparison Dapr made a similar call when they removed the dashboard from their chart: https://github.com/dapr/dapr/issues/4724

deploy/Chart/templates/dashboard/rbac.yaml
willdavsmith: looks like it does as of now. maybe this could be changed in the future.

deploy/Chart/values.yaml
willdavsmith: interesting! I didn't know dapr made this change. how about this - we'll keep the initial work to have dashboard "opt-out" for this release (so folks can try it out), and I will propose the 2 options (dashboard installs with Radius, dashboard installs with `rad dashboard init`) in a design review in the next 2 weeks. Whatever we go forward with, 0.32 will have full support for `rad dashboard` commands. does this sound alright? @willtsai @AaronCrawfis

willtsai: @willdavsmith this sounds like a fine idea to me, any objections @AaronCrawfis ?

AaronCrawfis: Sure, sounds good to me

pkg/cli/cmd/run/run.go
```diff
@@ -159,28 +162,65 @@ func (r *Runner) Run(ctx context.Context) error {
...
-			Out:             os.Stdout,
+			LabelSelector: applicationSelector,
+			Namespace:     namespace,
+			KubeContext:   kubeContext,
+			StatusChan:    applicationStatusChan,
+			Out:           os.Stdout,
+		})
+	})
+
+	// 4. Port-forward dashboard
```
AaronCrawfis, line 212: What happens if the user has opted out of the dashboard in their installation? Does this silently continue without throwing any errors or warnings?

willdavsmith: just added a commit to handle this :) now we will throw a log message with a docs link:

image

deploy/Chart/templates/dashboard/deployment.yaml
youngbupark: We can use subchart, but it is hard to version it. So I would use this way for now.

pkg/cli/cmd/run/run.go
```diff
@@ -159,28 +159,56 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	group.Go(func() error {
+		return r.Portforward.Run(ctx, portforward.Options{
+			LabelSelector: dashboardSelector,
+			Namespace:     "radius-system",
```
youngbupark, line 204: does it make sense to hardcode this namespace?

pkg/cli/cmd/run/run.go
```diff
@@ -159,28 +159,56 @@ func (r *Runner) Run(ctx context.Context) error {
...
-	status := make(chan portforward.StatusMessage)
+	// 1. Display port-forward messages for application
+	applicationStatusChan := make(chan portforward.StatusMessage)
+	group.Go(func() error {
+		r.displayPortforwardMessages(applicationStatusChan)
+		return nil
+	})
+
+	// 2. Display port-forward messages for dashboard
+	dashboardStatusChan := make(chan portforward.StatusMessage)
```
youngbupark, line 183: what happen if user doesn't install dashboard?

pkg/cli/cmd/run/run.go
```diff
@@ -159,28 +159,56 @@ func (r *Runner) Run(ctx context.Context) error {
...
+			Namespace:     namespace,
+			KubeContext:   kubeContext,
+			StatusChan:    applicationStatusChan,
+			Out:           os.Stdout,
+		})
+	})
+
+	// 4. Port-forward dashboard
+	group.Go(func() error {
+		return r.Portforward.Run(ctx, portforward.Options{
```
youngbupark, line 202: Do we need to enable port-forwarding if user doesn't install dashboard?

deploy/Chart/templates/dashboard/rbac.yaml
rynowak: For calling into UCP it does require cluster-scoped. Kubernetes APIs look like:

**Cluster-scoped:**

```
/apis/<group>/<version>/....
```

**Namespace-scoped:**

```
/apis/<group>/<version>/namespaces/....
```

The UCP apis look like `/apis/api.ucp.dev/v1alpha3/<resourceID>`.

pkg/cli/kubernetes/portforward/application_watcher.go
```diff
@@ -60,16 +58,21 @@ func NewApplicationWatcher(options Options) *applicationWatcher {
 func (aw *applicationWatcher) Run(ctx context.Context) error {
 	defer close(aw.done)
 
-	// We use the `radapp.io/application` label to include pods that are part of an application.
-	// This can include the user's Radius containers as well as any Kubernetes resources that are labeled
-	// as part of the application (eg: something created with a recipe).
-	req, err := labels.NewRequirement(kubernetes.LabelRadiusApplication, selection.Equals, []string{aw.Options.ApplicationName})
```
rynowak, line 66: Is this refactor because the "selector" is calculated outside this code?

pkg/cli/kubernetes/portforward/application_watcher.go
```diff
@@ -60,16 +58,21 @@ func NewApplicationWatcher(options Options) *applicationWatcher {
...
-		return err
+		return fmt.Errorf("failed to list deployments: %w", err)
 	}
 
-	deployments := aw.Options.Client.AppsV1().Deployments(aw.Options.Namespace)
-	listOptions := metav1.ListOptions{LabelSelector: labels.NewSelector().Add(*req).String()}
+	if len(labelledDeployments.Items) == 0 {
+		dashboardLabels := CreateLabelsForDashboard()
+		if aw.Options.LabelSelector.Matches(dashboardLabels) {
+			fmt.Println("Radius Dashboard not found, please see https://docs.radapp.io/guides/tooling/dashboard for more information")
```
rynowak, line 73: Is there a better option here than using `fmt.Println`? I try to avoid using `fmt.Println` anywhere in CLI code because it's inflexible. 

I think it would also be better to augment the `Options` to have a "warn if no deployments found" feature. This hardcodes the dashboard functionality in the middle of the code.

pkg/cli/kubernetes/portforward/util.go
```diff
@@ -39,16 +37,11 @@ const (
 // This is useful because we frequently run a port-forward right after completion of a Radius
 // deployment. We want to make sure we're port-forwarding to fresh replicas, not the ones
 // that are being scaled-down.
-func findStaleReplicaSets(ctx context.Context, client k8sclient.Interface, namespace, applicationName, desiredRevision string) (map[string]bool, error) {
```
rynowak, line 42: 👍

test/functional/shared/cli/cli_test.go
```diff
@@ -397,13 +397,26 @@ func Test_Run_Portforward(t *testing.T) {
...
+			t.Log("found matching output", line)
+
+			// Found the portforward local port.
+			port, err := strconv.Atoi(string(dashboardMatches[1]))
+			require.NoErrorf(t, err, "port is not an integer")
+			t.Logf("found local port %d", port)
+			require.Equal(t, 7007, port, "dashboard port should be 7007")
+		}
+
+		matches := appRegex.FindSubmatch([]byte(line))
```
rynowak, line 419: Nice 👍

deploy/Chart/values.yaml
rynowak: > I will propose the 2 options (dashboard installs with Radius, dashboard installs with rad dashboard init) in a design review in the next 2 weeks

Can I make a suggestion? What if we filed an issue and let it sit in the backlog until someone asks for it? This doesn't feel like it's absolutely required.

pkg/cli/cmd/run/run.go
willdavsmith: The portforward attempt will end and report to the user that the dashboard is not installed: https://github.com/radius-project/radius/pull/7186#discussion_r1500994822

pkg/cli/cmd/run/run.go
willdavsmith: technically no, we could do a k8s get on the dashboard to see if it is installed before here to see if we should try to portforward or not. Which do you think is better? the current approach or this one?

pkg/cli/kubernetes/portforward/application_watcher.go
willdavsmith: exactly

pkg/cli/kubernetes/portforward/application_watcher.go
willdavsmith: yeah I don't really like putting this here either - maybe we could move the dashboard check up a bit and remove this custom functionality from application_watcher. I will try it

willdavsmith: Newest commit moves the logic to run instead of applicationWatcher.

pkg/cli/cmd/run/run.go
willdavsmith: I updated this logic, it won't try to port-forward now if the user doesn't install the dashboard.

pkg/cli/cmd/run/run.go
willdavsmith: not really, fixed

pkg/cli/cmd/run/run.go
```diff
@@ -243,3 +256,23 @@ func (r *Runner) displayPortforwardMessages(status <-chan portforward.StatusMess
 		fmt.Printf("%s %s [port-forward] %s from localhost:%d -> ::%d\n", regular.Sprint(message.ReplicaName), bold.Sprint(message.ContainerName), message.Kind, message.LocalPort, message.RemotePort)
 	}
 }
+
+func shouldPortforwardDashboard(ctx context.Context, kubeContext string, dashboardLabelSelector labels.Selector) bool {
```
ytimocin, line 260: I think this function checks if the Dashboard deployment exists or not, right? I believe that using a function name that represents that would be more readable and cleaner.

pkg/cli/kubernetes/portforward/labels_test.go
```diff
@@ -50,27 +50,3 @@ func Test_CreateLabelSelectorForDashboard(t *testing.T) {
...
-	// Create labels for the dashboard
-	labels := CreateLabelsForDashboard()
-	require.NotNil(t, labels)
-	require.True(t, labels.Has("app.kubernetes.io/name"))
-	require.True(t, labels.Has("app.kubernetes.io/part-of"))
-	require.Equal(t, "dashboard", labels.Get("app.kubernetes.io/name"))
-	require.Equal(t, "radius", labels.Get("app.kubernetes.io/part-of"))
-}
-
-func Test_CreateLabelsForApplication(t *testing.T) {
```
ytimocin: Should these tests be added somewhere else?

pkg/cli/cmd/run/run.go
willdavsmith: fixed

pkg/cli/cmd/run/run.go
```diff
@@ -215,3 +258,33 @@ func (r *Runner) displayPortforwardMessages(status <-chan portforward.StatusMess
...
+	// If a kubernetes client is provided, use it. Otherwise, create a new client.
+	// This helps with testing.
+	if kubernetesClient != nil {
+		client = kubernetesClient
+	} else {
+		client, _, err = kubernetes.NewClientset(kubeContext)
+		if err != nil {
+			return false
+		}
+	}
```
youngbupark, line 276: No need to declare client variable here. 

```suggestion
	if kubernetesClient == nil {
		kubernetesClient, _, err = kubernetes.NewClientset(kubeContext)
		if err != nil {
			return false
		}
	}
```

pkg/cli/cmd/run/run.go
```diff
@@ -215,3 +258,29 @@ func (r *Runner) displayPortforwardMessages(status <-chan portforward.StatusMess
...
 	}
 }
+
+// dashboardDeploymentExists checks if a dashboard deployment exists in the given Kubernetes context.
+func dashboardDeploymentExists(ctx context.Context, kubernetesClient k8sclient.Interface, kubeContext string, dashboardLabelSelector labels.Selector) bool {
+	// If a kubernetes client is provided, use it. Otherwise, create a new client.
+	// This helps with testing.
+	if kubernetesClient == nil {
+		var err error
+		kubernetesClient, _, err = kubernetes.NewClientset(kubeContext)
```
youngbupark, line 268: Is this used for testing? If kubeContext doesn't exist in local machine, you will get the failure. Do not use this code. instead, we need to pass fake client to this. Can you pass fake client to dashboardDeploymentExists in the test without having this if ?

pkg/cli/cmd/run/run.go
```diff
@@ -215,3 +267,20 @@ func (r *Runner) displayPortforwardMessages(status <-chan portforward.StatusMess
 		fmt.Printf("%s %s [port-forward] %s from localhost:%d -> ::%d\n", regular.Sprint(message.ReplicaName), bold.Sprint(message.ContainerName), message.Kind, message.LocalPort, message.RemotePort)
 	}
 }
+
+// dashboardDeploymentExists checks if a dashboard deployment exists in the given Kubernetes context.
+func dashboardDeploymentExists(ctx context.Context, kubernetesClient k8sclient.Interface, kubeContext string, dashboardLabelSelector labels.Selector) bool {
```
youngbupark, line 272: We do not need to pass kubeContext now.
```suggestion
func dashboardDeploymentExists(ctx context.Context, kubernetesClient k8sclient.Interface, dashboardLabelSelector labels.Selector) bool {
```

youngbupark (approved):

Issue #6539, opened 2023-10-20
Allow any Terraform provider to be configured and used in a Recipe


AaronCrawfis:
## Overview of feature request

Today we support Terraform templates for Recipes. For the Azure, Kubernetes, and AWS providers we automatically setup the provider configuration with the identity of where to deploy and we pass in where to deploy via the `context` parameter.

Additionally, any additional provider can be used in a Recipe, however there is no way to securely configure the provider with credentials. For example, there is no way to use the GCP Terraform Provider in a Recipe without hard-coding the credentials in the Recipe module.

This feature is for allowing credentials to be securely configured and passed into the Recipe so any Terraform Provider can be used.

## Acceptance criteria

Once this feature is implemented I should be able to use the GCP, Oracle Cloud, or any other Terraform provider and pass in the credentials. I should be able to follow documentation that walks me through the steps I should take to accomplish this, with snippets showing an example.

There should be unit tests confirming that credentials are passed in correctly.

## Additional context

There are a couple ways to implement this, as each Terraform provider handles auth in various ways (_managed identities, service principals, IAM roles, etc._). We should take a look at how most providers handle credentials and see what makes sense to implement to cast the widest nest. Could be parameters that are passed in with the credentials, or could be a Terraform RC file pass through.

## Community ask

Please comment below with any Terraform modules that you would like to use.

[AB#9883](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/9883)

willdavsmith:
Here's one reported on Discord: https://registry.terraform.io/providers/ovh/ovh/latest/docs

ockhamlabs:
@AaronCrawfis @shalabhms is this on your roadmap ? If so any ETA?

AaronCrawfis:
> @AaronCrawfis @shalabhms is this on your roadmap ? If so any ETA?

Hi @ockhamlabs, this is high up on our roadmap for Recipes. We're working on a published Roadmap link, hope to have that available soon. We'll make sure this Issue is added to the roadmap project and stack ranked soon.

PR #7189, opened 2024-02-22
Add logic to build configuration for multiple Terraform providers support

lakshmimsft:
# Description

Adding logic to build configuration for multiple Terraform provider support. This is constructed from a combination of environment level recipe configuration located under RecipeConfig/Terraform/Providers section and the provider configurations registered with UCP. The environment level recipe configuration for providers takes precedence over UCP provider configurations.
[link to PR for design doc](https://github.com/radius-project/design-notes/pull/39)

The design document describes a SecretReference type and ability to fetch data from secrets and populate Provider Configuration. This will be implemented in subsequent PRs.

## Type of change
- This pull request adds or changes features of Radius and has an approved issue #6539 

Fixes: Part of #6539

### DIFF
--- a/pkg/recipes/terraform/config/config.go
+++ b/pkg/recipes/terraform/config/config.go
@@ -113,8 +113,8 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
 // by Radius to generate custom provider configurations. Save() must be called to save
 // the generated providers config. requiredProviders contains a list of provider names
 // that are required for the module.
-func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders []string, supportedProviders map[string]providers.Provider, envConfig *recipes.Configuration) error {
-	providerConfigs, err := getProviderConfigs(ctx, requiredProviders, supportedProviders, envConfig)
+func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders []string, ucpSupportedProviders map[string]providers.Provider, envConfig *recipes.Configuration) error {
+	providerConfigs, err := getProviderConfigs(ctx, requiredProviders, ucpSupportedProviders, envConfig)
 	if err != nil {
 		return err
 	}
@@ -166,12 +166,22 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
 }
 
 // getProviderConfigs generates the Terraform provider configurations for the required providers.
-func getProviderConfigs(ctx context.Context, requiredProviders []string, supportedProviders map[string]providers.Provider, envConfig *recipes.Configuration) (map[string]any, error) {
+func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpSupportedProviders map[string]providers.Provider, envConfig *recipes.Configuration) (map[string]any, error) {
 	providerConfigs := make(map[string]any)
+
+	// Get recipe provider configurations from the environment configuration
+	providerConfigs = getRecipeProviderConfigs(ctx, envConfig)
+
+	// Build provider configurations for required providers excluding the ones already present in providerConfigs
 	for _, provider := range requiredProviders {
-		builder, ok := supportedProviders[provider]
+		if _, ok := providerConfigs[provider]; ok {
+			// If provider is in providerConfigs, skip this iteration
+			continue
+		}
+
+		builder, ok := ucpSupportedProviders[provider]
 		if !ok {
-			// No-op: For any other provider, Radius doesn't generate any custom configuration.
+			// No-op: For any other provider under required_providers, Radius doesn't generate any custom configuration.
 			continue
 		}
 
@@ -179,12 +189,29 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
 		if err != nil {
 			return nil, err
 		}
+
 		if len(config) > 0 {
 			providerConfigs[provider] = config
 		}
 	}
 
 	return providerConfigs, nil
+}
+
+// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers.
+func getRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string]any {
+	providerConfigs := make(map[string]any)
+
+	// If the provider is not configured, or has empty configuration, skip this iteration
+	if envConfig.RecipeConfig.Terraform.Providers != nil {
+		for provider, config := range envConfig.RecipeConfig.Terraform.Providers {
+			if len(config) > 0 {
+				providerConfigs[provider] = config
+			}
+		}
+	}
+
+	return providerConfigs
 }
 
 // AddTerraformBackend adds backend configurations to store Terraform state file for the deployment.
--- a/pkg/recipes/terraform/config/config.go
+++ b/pkg/recipes/terraform/config/config.go
@@ -169,7 +169,7 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
 // providers and the provider configurations registered with UCP. The environment level recipe configuration for providers takes precedence over UCP provider configurations.
 func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpConfiguredProviders map[string]providers.Provider, envConfig *recipes.Configuration) (map[string]any, error) {
 	// Get recipe provider configurations from the environment configuration
-	providerConfigs := getRecipeProviderConfigs(ctx, envConfig)
+	providerConfigs := providers.GetRecipeProviderConfigs(ctx, envConfig)
 
 	// Build provider configurations for required providers excluding the ones already present in providerConfigs
 	for _, provider := range requiredProviders {
@@ -195,32 +195,6 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpConf
 	}
 
 	return providerConfigs, nil
-}
-
-// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers
-// specified under the RecipeConfig/Terraform/Providers section under environment configuration.
-func getRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string]any {
-	providerConfigs := make(map[string]any)
-
-	// If the provider is not configured, or has empty configuration, skip this iteration
-	if envConfig != nil && envConfig.RecipeConfig.Terraform.Providers != nil {
-		for provider, config := range envConfig.RecipeConfig.Terraform.Providers {
-			if len(config) > 0 {
-				configList := make([]any, 0)
-
-				for _, configDetails := range config {
-					//for _, additionalProperty := range configDetails.AdditionalProperties {
-					//	configList = append(configList, additionalProperty)
-					//}
-					configList = append(configList, configDetails.AdditionalProperties)
-				}
-
-				providerConfigs[provider] = configList
-			}
-		}
-	}
-
-	return providerConfigs
 }
 
 // AddTerraformBackend adds backend configurations to store Terraform state file for the deployment.
--- a/pkg/recipes/terraform/config/config.go
+++ b/pkg/recipes/terraform/config/config.go
@@ -113,8 +113,8 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
 // by Radius to generate custom provider configurations. Save() must be called to save
 // the generated providers config. requiredProviders contains a list of provider names
 // that are required for the module.
-func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders []string, ucpSupportedProviders map[string]providers.Provider, envConfig *recipes.Configuration) error {
-	providerConfigs, err := getProviderConfigs(ctx, requiredProviders, ucpSupportedProviders, envConfig)
+func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders []string, ucpConfiguredProviders map[string]providers.Provider, envConfig *recipes.Configuration) error {
+	providerConfigs, err := getProviderConfigs(ctx, requiredProviders, ucpConfiguredProviders, envConfig)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/recipes/terraform/config/config.go
```diff
@@ -150,13 +167,24 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
...
 
 	return providerConfigs, nil
 }
 
+// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers.
+func getRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string][]any {
+	// If the provider is not configured, return an empty config
+	if envConfig.RecipeConfig.Terraform.Providers.ProviderConfigs == nil {
+		// Handle the case where one or more objects in the chain are nil...
+		return nil
```
kachawla, line 182: This should return an empty map instead. In current this if condition isn't really doing anything, this return statement would result into same value as L185.

pkg/recipes/terraform/config/config.go
```diff
@@ -150,13 +167,24 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
...
 		}
 	}
 
 	return providerConfigs, nil
 }
 
+// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers.
+func getRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string][]any {
+	// If the provider is not configured, return an empty config
+	if envConfig.RecipeConfig.Terraform.Providers.ProviderConfigs == nil {
```
kachawla, line 180: What does it do if envConfig.RecipeConfig is nil?

pkg/recipes/terraform/config/config.go
```diff
@@ -150,13 +167,24 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
...
 		if len(config) > 0 {
-			providerConfigs[provider] = config
+			providerConfigs[provider] = []any{config}
 		}
 	}
 
 	return providerConfigs, nil
 }
 
+// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers.
```
kachawla, line 177: Can you elaborate this comment to include context on where these configurations are coming from?

pkg/recipes/terraform/config/config.go
```diff
@@ -136,12 +136,29 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
+	recipeProviderConfigs := getRecipeProviderConfigs(ctx, envConfig)
+
+	// Add provider config from recipeProviderConfigs to providerConfigs
+	for provider, config := range recipeProviderConfigs {
+		if len(config) > 0 {
+			providerConfigs[provider] = config
+		}
+	}
+
+	// Build provider configurations for required providers excluding the ones already present in providerConfigs
```
kachawla, line 152: Could you remind me the user experience we decided on again - unlike the current requirement, a user doesn't *need* to specify providers used in the module in the required providers anymore?

The reason I'm asking is because this code will add all the providers configured on the environment to the terraform config even if they are not used by the module. We should validate that this doesn't cause unexpected behavior.

pkg/recipes/terraform/config/config.go
```diff
@@ -136,12 +136,29 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
 	for _, provider := range requiredProviders {
-		builder, ok := supportedProviders[provider]
+		if _, ok := providerConfigs[provider]; ok {
+			// If provider is in providerConfigs, skip this iteration
```
kachawla, line 155:
```suggestion
			// Environment level recipe configuration for providers will take precedence over UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
```

pkg/recipes/terraform/config/config.go
```diff
@@ -150,13 +167,24 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
 			return nil, err
 		}
 		if len(config) > 0 {
-			providerConfigs[provider] = config
+			providerConfigs[provider] = []any{config}
```
kachawla, line 170: What's causing this to change?

pkg/recipes/terraform/config/providers/types.go
```diff
@@ -34,10 +34,10 @@ type Provider interface {
 	BuildConfig(ctx context.Context, envConfig *recipes.Configuration) (map[string]any, error)
 }
 
-// GetSupportedTerraformProviders returns a map of Terraform provider names to provider config builder.
-// Providers represent Terraform providers for which Radius generates custom provider configurations.
+// GetUCPSupportedTerraformProviders returns a map of Terraform provider names with configuration details stored in UCP, to provider config builder.
+// These providers represent Terraform providers for which Radius generates custom provider configurations based on credentials stored with UCP.
```
kachawla, line 38: It also adds non credential provider configuration (subscription id for azure, for example)

```suggestion
// These providers represent Terraform providers for which Radius generates custom provider configurations based on credentials stored with UCP and providers configured on the Radius environment.
```

pkg/recipes/terraform/config/config.go
```diff
@@ -150,13 +167,24 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, support
...
 		}
 	}
 
 	return providerConfigs, nil
 }
 
+// getRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers.
+func getRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string][]any {
```
kachawla, line 178: Since this is specific to providers, what do you think of moving this under terraform/config/providers package in [types.go](https://github.com/radius-project/radius/blob/12e1e49ba11fe387ad27630b0cd2d154c20f2700/pkg/recipes/terraform/config/providers/types.go)?

kachawla (commented):
Were you able to test these changes end to end locally?

Also, can you rebase with main since some of the code in this pr is duplicate of https://github.com/radius-project/radius/commit/c0b50a58b65621cf8d82087ed4a682577d342f43

pkg/recipes/terraform/config/config.go
lakshmimsft: code updated after rebase.

pkg/corerp/datamodel/environment.go
```diff
@@ -38,11 +38,48 @@ func (e *Environment) ResourceTypeName() string {
...
+	PAT map[string]SecretConfig `json:"pat,omitempty"`
+}
+
+// SecretConfig - Specifies the secret details of type personal access token for each different git platforms
+type SecretConfig struct {
+	// Secret represent the resource id for the secret store containing credentials
+	Secret string `json:"secret,omitempty"`
+}
+
+// TerraformProviders - Providers to be referenced by terraform recipes
```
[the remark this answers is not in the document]
lakshmimsft: pls ignore pkg/corerp/datamodel files. Current code rebased with main and copied changes from Provider datamodel PR which will be rebased again.

pkg/recipes/terraform/config/config_test.go
```diff
@@ -305,17 +305,19 @@ func Test_AddProviders(t *testing.T) {
...
-		expectedProviders  []map[string]any
-		expectedConfigFile string
-		Err                error
+		desc                          string
+		envConfig                     recipes.Configuration
+		requiredProviders             []string
+		expectedUCPSupportedProviders []map[string]any
+		expectedConfigFile            string
+		Err                           error
+		times                         int
```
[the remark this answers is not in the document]
lakshmimsft: updated to numFuncCall. This is to make sure existing tests using GoMock run as before and new tests run as expected. can discuss further.

pkg/recipes/terraform/config/config.go
lakshmimsft: code updated after rebase and tests added for nil checks

pkg/recipes/terraform/config/config_test.go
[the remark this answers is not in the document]
kachawla: > updated to numMockFuncCall. This is to make sure existing tests using GoMock run as before and new tests run as expected. can discuss further.

Hmm this is confusing. We don't use this pattern anywhere else, can you share what's different about these tests here that we need to use this pattern?

pkg/recipes/terraform/config/config.go
lakshmimsft: Currently we are adding to main.tf.json all providers configured in the envConfig/RecipeConfig/Terraform/Providers section and adding any additional configurations from the required_provider section the user may have configured in recipes. Let's discuss if this is not expected.

kachawla (commented):
PR title and description are out of date. Please update them. Also, it is good practice to write the description in a way that anyone can understand what the PR is about, think about someone looking at the commit in the future to understand what this change is, why it was etc. Same feedback for https://github.com/radius-project/radius/pull/7192

pkg/recipes/terraform/config/config_test.go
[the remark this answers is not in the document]
lakshmimsft: discussed and removed it.

pkg/recipes/terraform/config/config.go
lakshmimsft: moved getRecipeProviderConfigs() to terraform/config/providers/types.go

kachawla (commented):
This PR doesn't implement handling secrets, can you please call that out in the description?

Also, the typespec PR contains secrets, should that be split out into a separate PR given that backend isn't implemented?

pkg/recipes/terraform/config/providers/types.go
```diff
@@ -34,13 +34,37 @@ type Provider interface {
...
 	return map[string]Provider{
 		AWSProviderName:        NewAWSProvider(ucpConn, secretProvider),
 		AzureProviderName:      NewAzureProvider(ucpConn, secretProvider),
 		KubernetesProviderName: &kubernetesProvider{},
 	}
 }
+
+// GetRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers
+// specified under the RecipeConfig/Terraform/Providers section under environment configuration.
+func GetRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string]any {
```
kachawla, line 50: Can you add unit tests for this function

pkg/recipes/terraform/config/providers/types.go
```diff
@@ -34,13 +34,37 @@ type Provider interface {
...
+	// If the provider is not configured, or has empty configuration, skip this iteration
+	if envConfig != nil && envConfig.RecipeConfig.Terraform.Providers != nil {
+		for provider, config := range envConfig.RecipeConfig.Terraform.Providers {
+			if len(config) > 0 {
+				configList := make([]any, 0)
+
+				// Retrieve configuration details from 'AdditionalProperties' property and add to the list.
+				for _, configDetails := range config {
+					configList = append(configList, configDetails.AdditionalProperties)
+				}
```
kachawla, line 62: Should check if `AdditionalProperties` is non-empty/non-nil?

```suggestion
				for _, configDetails := range config {
    if configDetails.AdditionalProperties != nil && len(configDetails.AdditionalProperties) > 0 {
					configList = append(configList, configDetails.AdditionalProperties)
					}
				}
```

pkg/recipes/terraform/config/providers/types.go
```diff
@@ -34,13 +34,37 @@ type Provider interface {
...
+// GetRecipeProviderConfigs returns the Terraform provider configurations for Terraform providers
+// specified under the RecipeConfig/Terraform/Providers section under environment configuration.
+func GetRecipeProviderConfigs(ctx context.Context, envConfig *recipes.Configuration) map[string]any {
+	providerConfigs := make(map[string]any)
+
+	// If the provider is not configured, or has empty configuration, skip this iteration
+	if envConfig != nil && envConfig.RecipeConfig.Terraform.Providers != nil {
+		for provider, config := range envConfig.RecipeConfig.Terraform.Providers {
+			if len(config) > 0 {
+				configList := make([]any, 0)
```
kachawla, line 57:
```suggestion
				configList := make([]map[string]any, 0)
```

pkg/recipes/terraform/config/providers/types.go
```diff
@@ -34,13 +34,37 @@ type Provider interface {
 	BuildConfig(ctx context.Context, envConfig *recipes.Configuration) (map[string]any, error)
 }
 
-// GetSupportedTerraformProviders returns a map of Terraform provider names to provider config builder.
-// Providers represent Terraform providers for which Radius generates custom provider configurations.
-// For example, the Azure subscription id is added to Azure provider config using Radius Environment's Azure provider scope.
-func GetSupportedTerraformProviders(ucpConn sdk.Connection, secretProvider *ucp_provider.SecretProvider) map[string]Provider {
+// GetUCPConfiguredTerraformProviders returns a map of Terraform provider names with configuration details stored in UCP, to provider config builder.
```
kachawla, line 37: UCP part is covered in the next sentence, this update I think is unclear..

```suggestion
// GetUCPConfiguredTerraformProviders returns a map of Terraform provider names to provider config builder.
```

kachawla (commented):
Looks good overall. Couple of comments

pkg/recipes/terraform/config/providers/types.go
lakshmimsft: @kachawla this is allowed here to accomodate 'default provider configuration' where no details are input. Tests are added for this case both in types_test.go and here: https://github.com/radius-project/radius/pull/7189/files#diff-c5c89506f1ba2fe9bd8ecf4eedc856e93714c251c295c32a15d875f5924df3bdR515-R530
a example resulting main.tf.json is here: [pkg/recipes/terraform/config/testdata/providers-envrecipedefaultconfig.tf.json ](https://github.com/radius-project/radius/pull/7189/files#diff-f607148044721443b79901397b6794542895307f824510496a86ad52cd5e1b61R12-R18)
I did a manual test to see provider configuration load successfully.

kachawla: Sounds good, thank you. During manual testing, in addition to checking that the config was loaded successfully, could you also confirm that terraform apply runs successfully for this scenario?

pkg/recipes/terraform/config/providers/types.go
lakshmimsft: updated

pkg/recipes/terraform/config/providers/types.go
kachawla: Also, can you please expand on "default provider configuration"? I'm not sure what we mean by this. Feel free to point me to the right section in the design if it is captured in the doc already.

lakshmimsft: yes, confirming, terraform apply runs successfully for this scenario.

kachawla: Why do we want to populate empty config though?

pkg/recipes/terraform/config/testdata/providers-envrecipedefaultconfig.tf.json
```diff
@@ -0,0 +1,30 @@
...
+      "kubernetes": {
+        "config_path": "/home/radius/.kube/config",
+        "namespace": "radius-system",
+        "secret_suffix": "test-secret-suffix"
+      }
+    }
+  },
+  "provider": {
+    "azurerm": [
+      null,
```
ytimocin, line 13: Why do we have **null** here? Would/Why would a Radius User ever add **null** to the Config?

ytimocin (dismissed):
Asked a question but other than that LGTM

pkg/recipes/terraform/config/providers/types.go
kachawla: There is no concept of default empty configuration afaik, let's add this nil check unless there is a terraform documentation that mentions otherwise.

kachawla (changes requested):

pkg/recipes/terraform/config/testdata/providers-envrecipedefaultconfig.tf.json
lakshmimsft: this is covered in discussion in above comment and is now removed.

pkg/recipes/terraform/config/providers/types.go
lakshmimsft: added nil check.

kachawla: From the doc - "Terraform assumes an empty default configuration for any provider that is not explicitly configured." - this is already handled by Terraform. By adding empty configuration here, we are overriding any non-empty config that a user might have in their module. Unless there is a strong use case for passing empty config, I wouldn't recommend doing this.

kachawla (approved):

PR #7266, opened 2024-03-05
Changes to load secrets as part of environment configuration

vishwahiremat:
# Description

- Remove Secret Loader
- Adding changes to load secrets as part of LoadConfiguration
- Refactored the code accordingly

## Type of change

Fixes: #issue_number

### DIFF
--- a/pkg/corerp/datamodel/recipe_types.go
+++ b/pkg/corerp/datamodel/recipe_types.go
@@ -52,7 +52,8 @@ type SecretConfig struct {
 	// The ID of an Applications.Core/SecretStore resource containing the Git platform personal access token (PAT). The secret
 	// store must have a secret named 'pat', containing the PAT value. A secret named
 	// 'username' is optional, containing the username associated with the pat. By default no username is specified.
-	Secret string `json:"secret,omitempty"`
+	Secret     string `json:"secret,omitempty"`
+	SecretData map[string]string
 }
 
 // EnvironmentVariables represents the environment variables to be set for the recipe execution.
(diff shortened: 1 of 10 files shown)


pkg/corerp/datamodel/recipe_types.go
```diff
@@ -52,7 +52,8 @@ type SecretConfig struct {
 	// The ID of an Applications.Core/SecretStore resource containing the Git platform personal access token (PAT). The secret
 	// store must have a secret named 'pat', containing the PAT value. A secret named
 	// 'username' is optional, containing the username associated with the pat. By default no username is specified.
-	Secret string `json:"secret,omitempty"`
+	Secret     string `json:"secret,omitempty"`
+	SecretData map[string]string
```
rynowak, line 56: Why is the datamodel changing as part of this change? The data model is the internal representation of our API.

ytimocin: Is Secret the Key and the SecretData the Value?

pkg/recipes/configloader/environment.go
```diff
@@ -114,6 +114,17 @@ func getConfiguration(environment *v20231001preview.EnvironmentResource, applica
 
 	if environment.Properties.RecipeConfig != nil {
 		config.RecipeConfig = envDatamodel.Properties.RecipeConfig
+		secrets := config.RecipeConfig.Terraform.Authentication.Git.PAT
```
ytimocin, line 117: This is specific to Terraform, right? Where do we check if this recipe is a Terraform recipe?

kachawla, line 117: Ideally all the recipe type specific operations should be contained within driver, so configloader is not really the right place for this.

PR #7294, opened 2024-03-07
Add workflow_dispatch trigger to functional tests

willdavsmith:
# Description

* Add workflow_dispatch trigger to functional test workflow to allow for manual runs
* Re-worded some steps for clarity

## Type of change

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -21,6 +21,8 @@ on:
     - cron: "30 0,4,8,12,16,20 * * 1-5"
     # Run every 12 hours on weekends.
     - cron: "30 0,12 * * 0,6"
+  # Allow manual dispatch
+  workflow_dispatch:
   # Dispatch on external events
   repository_dispatch:
     types: [de-functional-test]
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -352,7 +354,6 @@ jobs:
         with:
           repository: radius-project/samples
           ref: refs/heads/edge
-          token: ${{ secrets.GH_RAD_CI_BOT_PAT }}
```
willdavsmith, line 355: this was unnecessary because radius-project/samples is public.

.github/workflows/functional-test.yaml
```diff
@@ -414,7 +415,7 @@ jobs:
           # AZURE_OIDC_ISSUER
           eval "export $(echo "${{ secrets.FUNCTEST_AZURE_OIDC_JSON }}" | jq -r 'to_entries | map("\(.key)=\(.value)") | @sh')"
 
-          AUTHKEY=$(echo -n "${{ github.actor }}:${{ secrets.GH_RAD_CI_BOT_PAT }}" | base64)
+          AUTHKEY=$(echo -n "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}" | base64)
```
willdavsmith, line 418: if github.actor is not rad-ci-bot (as is the case with workflow_dispatch), we should use GITHUB_TOKEN instead.

.github/workflows/functional-test.yaml
```diff
@@ -21,6 +21,8 @@ on:
     - cron: "30 0,4,8,12,16,20 * * 1-5"
     # Run every 12 hours on weekends.
     - cron: "30 0,12 * * 0,6"
+  # Allow manual dispatch
+  workflow_dispatch:
```
rynowak, line 25: Do we need to also accept parameters here? Like the branch name?

Where were you thinking we'd configure that?

See example of how we configure this in other cases here: https://github.com/radius-project/radius/pull/7294/files#diff-c79f364a9293abaaa8595776b74674e24bec6287834e63ab8aa7aec6a42f0dbcR101

PR #7296, opened 2024-03-09
Log errors for failed requests

### LINKED ISSUE #7297: Log Errors for Failed Requests
### Problem

We currently lack error logging upon failure for many operations which hinders our ability to debug failures. This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.

Here is an example of where error was not logged: https://github.com/radius-project/radius/issues/7264

### Proposed Solution
To address this, the proposed change is to move error logging to the setFailed method. By doing so, we ensure that error logging is not reliant on every individual operation to log errors.

[AB#11425](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11425)

kachawla:
# Description

We currently lack error logging upon failure for many operations which hinders our ability to debug failures. This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.

To address this, the proposed change is to move error logging to the setFailed method. By doing so, we ensure that error logging is not reliant on every individual operation to log errors.

## Type of change

Fixes: https://github.com/radius-project/radius/issues/7297

### DIFF
--- a/pkg/armrpc/asyncoperation/controller/result.go
+++ b/pkg/armrpc/asyncoperation/controller/result.go
@@ -17,7 +17,11 @@ limitations under the License.
 package controller
 
 import (
+	"context"
+	"fmt"
+
 	v1 "github.com/radius-project/radius/pkg/armrpc/api/v1"
+	"github.com/radius-project/radius/pkg/ucp/ucplog"
 )
 
 // Result is the response of async operation controller.
@@ -40,20 +44,24 @@ func NewCanceledResult(message string) Result {
 }
 
 // NewFailedResult creates a new Result object with the given error details and sets the failed flag to true.
-func NewFailedResult(err v1.ErrorDetails) Result {
+func NewFailedResult(ctx context.Context, err error, errorDetails v1.ErrorDetails) Result {
 	r := Result{}
-	r.SetFailed(err, false)
+	r.SetFailed(ctx, err, errorDetails, false)
+
 	return r
 }
 
 // SetFailed sets the error response with Failed status.
-func (r *Result) SetFailed(err v1.ErrorDetails, requeue bool) {
+func (r *Result) SetFailed(ctx context.Context, err error, errorDetails v1.ErrorDetails, requeue bool) {
+	logger := ucplog.FromContextOrDiscard(ctx)
+	logger.Error(err, fmt.Sprintf("Operation Failed: %s", errorDetails.Message))
+
 	if r == nil {
 		r = &Result{}
 	}
 	r.Requeue = requeue
 	r.SetProvisioningState(v1.ProvisioningStateFailed)
-	r.Error = &err
+	r.Error = &errorDetails
 }
 
 // SetCanceled sets the Result's Requeue field to false, sets the ProvisioningState to Canceled and sets the Error field
--- a/pkg/armrpc/asyncoperation/worker/worker.go
+++ b/pkg/armrpc/asyncoperation/worker/worker.go
@@ -179,8 +179,7 @@ func (w *AsyncRequestProcessWorker) Start(ctx context.Context) error {
 
 			if msgreq.DequeueCount > w.options.MaxOperationRetryCount {
 				errMsg := fmt.Sprintf("exceeded max retry count to process async operation message: %d", msgreq.DequeueCount)
-				opLogger.Error(nil, errMsg)
-				failed := ctrl.NewFailedResult(v1.ErrorDetails{
+				failed := ctrl.NewFailedResult(ctx, nil, v1.ErrorDetails{
 					Code:    v1.CodeInternal,
 					Message: errMsg,
 				})
@@ -266,8 +265,7 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 		if !errors.Is(asyncReqCtx.Err(), context.Canceled) {
 			if err != nil {
 				armErr := extractError(err)
-				result.SetFailed(armErr, false)
-				logger.Error(err, "Operation Failed")
+				result.SetFailed(ctx, err, armErr, false)
 			}
 
 			w.completeOperation(ctx, message, result, asyncCtrl.StorageClient())
--- a/pkg/armrpc/asyncoperation/controller/result_test.go
+++ b/pkg/armrpc/asyncoperation/controller/result_test.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"errors"
+	"testing"
+
+	v1 "github.com/radius-project/radius/pkg/armrpc/api/v1"
+	"github.com/radius-project/radius/test/testcontext"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewFailedResult(t *testing.T) {
+	ctx := testcontext.New(t)
+	err := errors.New("test error")
+	errorDetails := v1.ErrorDetails{
+		Message: "test error message",
+	}
+
+	result := NewFailedResult(ctx, err, errorDetails)
+	require.Equal(t, errorDetails, *result.Error)
+	require.Equal(t, false, result.Requeue)
+	require.Equal(t, v1.ProvisioningStateFailed, *result.state)
+}
+
+func TestSetFailed(t *testing.T) {
+	err := v1.ErrorDetails{
+		Message: "test error message",
+	}
+
+	result := &Result{}
+	result.SetFailed(testcontext.New(t), errors.New("test error"), err, true)
+	require.Equal(t, err, *result.Error)
+	require.Equal(t, true, result.Requeue)
+}
--- a/pkg/corerp/backend/controller/createorupdateresource.go
+++ b/pkg/corerp/backend/controller/createorupdateresource.go
@@ -108,7 +108,7 @@ func (c *CreateOrUpdateResource) Run(ctx context.Context, request *ctrl.Request)
 
 	deploymentDataModel, ok := dataModel.(rpv1.DeploymentDataModel)
 	if !ok {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Message: "deployment data model conversion error"}), err
+		return ctrl.NewFailedResult(ctx, nil, v1.ErrorDetails{Message: "deployment data model conversion error"}), nil
 	}
 
 	oldOutputResources := deploymentDataModel.OutputResources()
@@ -136,5 +136,5 @@ func (c *CreateOrUpdateResource) Run(ctx context.Context, request *ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
-	return ctrl.Result{}, err
+	return ctrl.Result{}, nil
 }
--- a/pkg/corerp/backend/controller/deleteresource.go
+++ b/pkg/corerp/backend/controller/deleteresource.go
@@ -43,7 +43,7 @@ func NewDeleteResource(opts ctrl.Options) (ctrl.Controller, error) {
 func (c *DeleteResource) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {
 	obj, err := c.StorageClient().Get(ctx, request.ResourceID)
 	if err != nil {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Message: err.Error()}), err
+		return ctrl.NewFailedResult(ctx, err, v1.ErrorDetails{Message: err.Error()}), err
 	}
 
 	// This code is general and we might be processing an async job for a resource or a scope, so using the general Parse function.
@@ -63,7 +63,7 @@ func (c *DeleteResource) Run(ctx context.Context, request *ctrl.Request) (ctrl.R
 
 	deploymentDataModel, ok := dataModel.(rpv1.DeploymentDataModel)
 	if !ok {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Message: "deployment data model conversion error"}), nil
+		return ctrl.NewFailedResult(ctx, nil, v1.ErrorDetails{Message: "deployment data model conversion error"}), nil
 	}
 
 	err = c.DeploymentProcessor().Delete(ctx, id, deploymentDataModel.OutputResources())
--- a/pkg/portableresources/backend/controller/createorupdateresource.go
+++ b/pkg/portableresources/backend/controller/createorupdateresource.go
@@ -19,7 +19,6 @@ package controller
 import (
 	"context"
 	"errors"
-	"fmt"
 
 	ctrl "github.com/radius-project/radius/pkg/armrpc/asyncoperation/controller"
 	"github.com/radius-project/radius/pkg/portableresources/datamodel"
@@ -97,7 +96,6 @@ func (c *CreateOrUpdateResource[P, T]) Run(ctx context.Context, req *ctrl.Reques
 	recipeOutput, err := c.executeRecipeIfNeeded(ctx, data, previousOutputResources, config.Simulated)
 	if err != nil {
 		if recipeError, ok := err.(*recipes.RecipeError); ok {
-			logger.Error(err, fmt.Sprintf("failed to execute recipe. Encountered error while processing %s ", recipeError.ErrorDetails.Target))
 			// Set the deployment status to the recipe error code.
 			recipeDataModel.Recipe().DeploymentStatus = util.RecipeDeploymentStatus(recipeError.DeploymentStatus)
 			update := &store.Object{
@@ -111,7 +109,7 @@ func (c *CreateOrUpdateResource[P, T]) Run(ctx context.Context, req *ctrl.Reques
 			if err != nil {
 				return ctrl.Result{}, err
 			}
-			return ctrl.NewFailedResult(recipeError.ErrorDetails), nil
+			return ctrl.NewFailedResult(ctx, err, recipeError.ErrorDetails), nil
 		}
 		return ctrl.Result{}, err
 	}
--- a/pkg/portableresources/backend/controller/deleteresource.go
+++ b/pkg/portableresources/backend/controller/deleteresource.go
@@ -60,7 +60,7 @@ func NewDeleteResource[P interface {
 func (c *DeleteResource[P, T]) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {
 	obj, err := c.StorageClient().Get(ctx, request.ResourceID)
 	if err != nil {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Message: err.Error()}), err
+		return ctrl.NewFailedResult(ctx, err, v1.ErrorDetails{Message: err.Error()}), err
 	}
 
 	// This code is general and we might be processing an async job for a resource or a scope, so using the general Parse function.
@@ -95,7 +95,7 @@ func (c *DeleteResource[P, T]) Run(ctx context.Context, request *ctrl.Request) (
 		})
 		if err != nil {
 			if recipeError, ok := err.(*recipes.RecipeError); ok {
-				return ctrl.NewFailedResult(recipeError.ErrorDetails), nil
+				return ctrl.NewFailedResult(ctx, err, recipeError.ErrorDetails), nil
 			}
 			return ctrl.Result{}, err
 		}
--- a/pkg/ucp/backend/controller/resourcegroups/trackedresourceprocess.go
+++ b/pkg/ucp/backend/controller/resourcegroups/trackedresourceprocess.go
@@ -57,7 +57,7 @@ func NewTrackedResourceProcessController(opts ctrl.Options) (ctrl.Controller, er
 func (c *TrackedResourceProcessController) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {
 	resource, err := store.GetResource[datamodel.GenericResource](ctx, c.StorageClient(), request.ResourceID)
 	if errors.Is(err, &store.ErrNotFound{}) {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Code: v1.CodeNotFound, Message: fmt.Sprintf("resource %q not found", request.ResourceID), Target: request.ResourceID}), nil
+		return ctrl.NewFailedResult(ctx, err, v1.ErrorDetails{Code: v1.CodeNotFound, Message: fmt.Sprintf("resource %q not found", request.ResourceID), Target: request.ResourceID}), nil
 	} else if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -69,9 +69,9 @@ func (c *TrackedResourceProcessController) Run(ctx context.Context, request *ctr
 
 	downstreamURL, err := resourcegroups.ValidateDownstream(ctx, c.StorageClient(), originalID)
 	if errors.Is(err, &resourcegroups.NotFoundError{}) {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Code: v1.CodeNotFound, Message: err.Error(), Target: request.ResourceID}), nil
+		return ctrl.NewFailedResult(ctx, err, v1.ErrorDetails{Code: v1.CodeNotFound, Message: err.Error(), Target: request.ResourceID}), nil
 	} else if errors.Is(err, &resourcegroups.InvalidError{}) {
-		return ctrl.NewFailedResult(v1.ErrorDetails{Code: v1.CodeInvalid, Message: err.Error(), Target: request.ResourceID}), nil
+		return ctrl.NewFailedResult(ctx, err, v1.ErrorDetails{Code: v1.CodeInvalid, Message: err.Error(), Target: request.ResourceID}), nil
 	} else if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to validate downstream: %w", err)
 	}
@@ -82,7 +82,7 @@ func (c *TrackedResourceProcessController) Run(ctx context.Context, request *ctr
 	if errors.Is(err, &trackedresource.InProgressErr{}) {
 		// The resource is still being processed, so we can sleep for a while.
 		result := ctrl.Result{}
-		result.SetFailed(v1.ErrorDetails{Code: v1.CodeConflict, Message: err.Error(), Target: request.ResourceID}, true)
+		result.SetFailed(ctx, err, v1.ErrorDetails{Code: v1.CodeConflict, Message: err.Error(), Target: request.ResourceID}, true)
 
 		return result, nil
 	} else if err != nil {
--- a/pkg/ucp/backend/controller/resourcegroups/trackedresourceprocess_test.go
+++ b/pkg/ucp/backend/controller/resourcegroups/trackedresourceprocess_test.go
@@ -84,6 +84,7 @@ func Test_Run(t *testing.T) {
 
 	t.Run("retry", func(t *testing.T) {
 		pc, updater, storageClient := setup(t)
+		ctx := testcontext.New(t)
 
 		storageClient.EXPECT().
 			Get(gomock.Any(), trackingID.String(), gomock.Any()).
@@ -101,33 +102,35 @@ func Test_Run(t *testing.T) {
 		updater.Result = &trackedresource.InProgressErr{}
 
 		expected := controller.Result{}
-		expected.SetFailed(v1.ErrorDetails{Code: v1.CodeConflict, Message: updater.Result.Error(), Target: trackingID.String()}, true)
+		expected.SetFailed(ctx, nil, v1.ErrorDetails{Code: v1.CodeConflict, Message: updater.Result.Error(), Target: trackingID.String()}, true)
 
-		result, err := pc.Run(testcontext.New(t), &controller.Request{ResourceID: trackingID.String()})
+		result, err := pc.Run(ctx, &controller.Request{ResourceID: trackingID.String()})
 		require.Equal(t, expected, result)
 		require.NoError(t, err)
 	})
 
 	t.Run("Failure (resource not found)", func(t *testing.T) {
 		pc, _, storageClient := setup(t)
+		ctx := testcontext.New(t)
 
 		storageClient.EXPECT().
 			Get(gomock.Any(), trackingID.String(), gomock.Any()).
 			Return(nil, &store.ErrNotFound{}).Times(1)
 
-		expected := controller.NewFailedResult(v1.ErrorDetails{
+		expected := controller.NewFailedResult(ctx, nil, v1.ErrorDetails{
 			Code:    v1.CodeNotFound,
 			Message: fmt.Sprintf("resource %q not found", trackingID.String()),
 			Target:  trackingID.String(),
 		})
 
-		result, err := pc.Run(testcontext.New(t), &controller.Request{ResourceID: trackingID.String()})
+		result, err := pc.Run(ctx, &controller.Request{ResourceID: trackingID.String()})
 		require.Equal(t, expected, result)
 		require.NoError(t, err)
 	})
 
 	t.Run("Failure (validate downstream: not found)", func(t *testing.T) {
 		pc, _, storageClient := setup(t)
+		ctx := testcontext.New(t)
 
 		storageClient.EXPECT().
 			Get(gomock.Any(), trackingID.String(), gomock.Any()).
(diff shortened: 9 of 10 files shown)


rynowak:
> This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.

Can you explain this in more detail, especially the part about errors being propagated vs not propagated? 

Are there cases we return a failure from an async operation and it's **not** logged?

/cc @youngbupark

kachawla:
> > This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.
> 
> Can you explain this in more detail, especially the part about errors being propagated vs not propagated?
> 
> Are there cases we return a failure from an async operation and it's **not** logged?

Yes. Looks like it is expected that we always log an error explicitly if an async operation fails, which is the reason why log statements like [these](https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L182) are spread across our code base. Current approach is unreliable because it clearly has left us with logging gaps, and this is not the first time it has prevented us from debugging, past example [here](https://github.com/radius-project/radius/pull/6276). I'm sure @nithyatsu and @youngbupark can had more historical context on this choice of implementation.

rynowak:
I'm trying to figure out whether we've got the wrong design or whether this is "just" a bug 🐛 FWIW, I wrote the guidance on logging - so you can rest the blame on me if this is the wrong idea.  

In the case we're talking about, I'd call [this region](https://github.com/radius-project/radius/pull/7296/files#diff-5d9b9964242d77606f4868e3fe049978ea1c3f1903afcfeb60c76e19d9785fc5L253-L275) of the code the place "where the error is handled". 

Is the problem that there are two ways to represent a failure?

- A go `error`.
- A `controller.Result` with a failure status?

---

I'm pretty nervous about relying on "constructors" for any kind of important logic. Go is more like C than it is Java, there's no requirement to use a "constructor" and lots of alternative ways to initialize a struct.

kachawla:
> I'm trying to figure out whether we've got the wrong design or whether this is "just" a bug 🐛 FWIW, I wrote the guidance on logging - so you can rest the blame on me if this is the wrong idea. I'd be happy to spend some time discussing and coming up with better guidance. I've found that error handling in Go is usually the most challenging thing for people to get right.
> 
> In the case we're talking about, I'd call [this region](https://github.com/radius-project/radius/pull/7296/files#diff-5d9b9964242d77606f4868e3fe049978ea1c3f1903afcfeb60c76e19d9785fc5L253-L275) of the code the place "where the error is handled".

Does the region of code you linked handle [this scenario]( https://github.com/radius-project/radius/blob/84d4ca36ac6df46fb10489a3d666ff798d2a1904/pkg/corerp/backend/controller/deleteresource.go#L66) as well where the error returned from the controller is nil with a failed result?

> Is the problem that there are two ways to represent a failure?

The problem is that the errors aren't getting logged. There is no way to get the stack trace for debugging. If the root cause identified here is not correct, then this needs further investigation on why we are dropping logs for certain requests.

pkg/armrpc/asyncoperation/worker/worker.go
```diff
@@ -179,8 +179,7 @@ func (w *AsyncRequestProcessWorker) Start(ctx context.Context) error {
 
 			if msgreq.DequeueCount > w.options.MaxOperationRetryCount {
 				errMsg := fmt.Sprintf("exceeded max retry count to process async operation message: %d", msgreq.DequeueCount)
-				opLogger.Error(nil, errMsg)
```
nithyatsu, line 182: when we have an async op, the traceID should have been propagated, and this line should have printed the log along with traceID, was it not happening? trying to understand the issue.

kachawla:
> > I'm trying to figure out whether we've got the wrong design or whether this is "just" a bug 🐛 FWIW, I wrote the guidance on logging - so you can rest the blame on me if this is the wrong idea. I'd be happy to spend some time discussing and coming up with better guidance. I've found that error handling in Go is usually the most challenging thing for people to get right.
> > In the case we're talking about, I'd call [this region](https://github.com/radius-project/radius/pull/7296/files#diff-5d9b9964242d77606f4868e3fe049978ea1c3f1903afcfeb60c76e19d9785fc5L253-L275) of the code the place "where the error is handled".
> 
> Does the region of code you linked handle [this scenario](https://github.com/radius-project/radius/blob/84d4ca36ac6df46fb10489a3d666ff798d2a1904/pkg/corerp/backend/controller/deleteresource.go#L66) as well where the error returned from the controller is nil with a failed result?
> 
> > Is the problem that there are two ways to represent a failure?
> 
> The problem is that the errors aren't getting logged. There is no way to get the stack trace for debugging. If the root cause identified here is not correct, then this needs further investigation on why we are dropping logs for certain requests.

This log is also odd https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L260. Why is it always hardcoding the `result.Error` to nil? @youngbupark do you have context on what was the motivation here or is it a bug? If it is a bug then I think updating it will be a better approach then what I'm doing in this PR.

youngbupark:
> > > This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.
> > 
> > 
> > Can you explain this in more detail, especially the part about errors being propagated vs not propagated?
> 
> Async errors aren't returned. I had linked the documentation https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45. It should be updated if it isn't providing clear understanding.
> 
> > Are there cases we return a failure from an async operation and it's **not** logged?
> 
> Yes. Current implementation expects separate explicit error logging if an async operation fails, which is the reason why log statements like [these](https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L182) are spread across our code base. Current approach is unreliable because it clearly has left us with logging gaps, and this is not the first time it has prevented us from debugging, past example [here](https://github.com/radius-project/radius/pull/6276). I'm sure @nithyatsu and @youngbupark can add more historical context on this choice of implementation.

Logging error for that case is important to understand server-side issue because retry failure could be caused by server side  problem. That's why we not only return the error to client but also record the error on server-side log.

youngbupark:
> > > I'm trying to figure out whether we've got the wrong design or whether this is "just" a bug 🐛 FWIW, I wrote the guidance on logging - so you can rest the blame on me if this is the wrong idea. I'd be happy to spend some time discussing and coming up with better guidance. I've found that error handling in Go is usually the most challenging thing for people to get right.
> > > In the case we're talking about, I'd call [this region](https://github.com/radius-project/radius/pull/7296/files#diff-5d9b9964242d77606f4868e3fe049978ea1c3f1903afcfeb60c76e19d9785fc5L253-L275) of the code the place "where the error is handled".
> > 
> > 
> > Does the region of code you linked handle [this scenario](https://github.com/radius-project/radius/blob/84d4ca36ac6df46fb10489a3d666ff798d2a1904/pkg/corerp/backend/controller/deleteresource.go#L66) as well where the error returned from the controller is nil with a failed result?
> > > Is the problem that there are two ways to represent a failure?
> > 
> > 
> > The problem is that the errors aren't getting logged. There is no way to get the stack trace for debugging. If the root cause identified here is not correct, then this needs further investigation on why we are dropping logs for certain requests.
> 
> This log is also odd https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L260. Why is it always hardcoding `result.Error` to nil? @youngbupark do you have context on what was the motivation here or is it a bug? If it is a bug then I think updating it will be a better approach over what I'm doing in this PR.

It is intended. 

https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260

```golang
logger.Info("Operation returned", "success", result.Error == nil, "code", code, "provisioningState", result.ProvisioningState(), "err", err)
```

If the operation is succeeded, result.Error must be nil.  When logger processes this, it will be represented as JSON format log like below. The arguments from `"success"` until `err` are treated as a log dimension or attribution. Having such attributions such as success/code/provisioningState help indexing for search backend.

```
{"severity":"info","timestamp":"2024-03-12T20:43:42.002Z","name":"radius.radiusasyncworker","caller":"worker/worker.go:260","message":"Operation returned","serviceName":"radius","version":"edge","hostName":"applications-rp-bf77896d9-rfnk7","resourceId":"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/containers/backendcontainerdns","operationId":"64d729f1-2936-48eb-aabe-9a5fcaee2626","operationType":"APPLICATIONS.CORE/CONTAINERS|PUT","dequeueCount":1,"traceId":"9684bed9d6f01c4db253ab973274dfd1","spanId":"a15c41cd88fea0e1","success":true,"code":"","provisioningState":"Succeeded","err":null}
```

pkg/armrpc/asyncoperation/worker/worker.go
```diff
@@ -266,8 +265,7 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 		if !errors.Is(asyncReqCtx.Err(), context.Canceled) {
 			if err != nil {
 				armErr := extractError(err)
-				result.SetFailed(armErr, false)
-				logger.Error(err, "Operation Failed")
```
youngbupark, line 270: This is required for server side debugging. If we remove this, it is hard to debug only with server log. e.g. when we debug the problem with server log, it is hard to know which step has failed because there is no indication. Operation Failed will be the indicator at which step has been failed.

pkg/armrpc/asyncoperation/controller/result.go
```diff
@@ -40,20 +44,24 @@ func NewCanceledResult(message string) Result {
 }
 
 // NewFailedResult creates a new Result object with the given error details and sets the failed flag to true.
-func NewFailedResult(err v1.ErrorDetails) Result {
+func NewFailedResult(ctx context.Context, err error, errorDetails v1.ErrorDetails) Result {
```
youngbupark, line 47: What's the benefit to have err arg here? The initial idea is that we want to have a detail error description specified by Controller instead of just golang error. So we can deliver the detail error to clients.

pkg/armrpc/asyncoperation/controller/result.go
```diff
@@ -40,20 +44,24 @@ func NewCanceledResult(message string) Result {
...
 	r := Result{}
-	r.SetFailed(err, false)
+	r.SetFailed(ctx, err, errorDetails, false)
+
 	return r
 }
 
 // SetFailed sets the error response with Failed status.
-func (r *Result) SetFailed(err v1.ErrorDetails, requeue bool) {
+func (r *Result) SetFailed(ctx context.Context, err error, errorDetails v1.ErrorDetails, requeue bool) {
```
youngbupark, line 55: Same here.

pkg/armrpc/asyncoperation/worker/worker.go
kachawla: This is not being removed in this PR, but being abstracted out. I have explained it in the PR description.

kachawla: This is not being removed in this PR, but being abstracted out as I also had explained in the PR description.

pkg/armrpc/asyncoperation/controller/result.go
kachawla: You can see how it is consumed by this function to understand the reasoning behind adding it in parameters here.

pkg/armrpc/asyncoperation/controller/result.go
kachawla: This is being used in on L57.

kachawla:
> > > > This gap exists because errors that aren't propagated back up (async operations) [aren't logged automatically](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45). The current state of the code is inconsistent, with some operations logging errors while others do not.
> > > 
> > > 
> > > Can you explain this in more detail, especially the part about errors being propagated vs not propagated?
> > 
> > 
> > Async errors aren't returned. I had linked the documentation https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/logging.md?plain=1#L45. It should be updated if it isn't providing clear understanding.
> > > Are there cases we return a failure from an async operation and it's **not** logged?
> > 
> > 
> > Yes. Current implementation expects separate explicit error logging if an async operation fails, which is the reason why log statements like [these](https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L182) are spread across our code base. Current approach is unreliable because it clearly has left us with logging gaps, and this is not the first time it has prevented us from debugging, past example [here](https://github.com/radius-project/radius/pull/6276). I'm sure @nithyatsu and @youngbupark can add more historical context on this choice of implementation.
> 
> Logging error for that case is important to understand server-side issue because retry failure could be caused by server side problem. That's why we not only return the error to client but also record the error on server-side log.

Well that's the problem I'm fixing in this PR. Errors aren't being logged today preventing us from being able to effectively debug issues. We have seen many instances of this, here is an example for your reference https://github.com/radius-project/radius/issues/7264

kachawla:
> > > > I'm trying to figure out whether we've got the wrong design or whether this is "just" a bug 🐛 FWIW, I wrote the guidance on logging - so you can rest the blame on me if this is the wrong idea. I'd be happy to spend some time discussing and coming up with better guidance. I've found that error handling in Go is usually the most challenging thing for people to get right.
> > > > In the case we're talking about, I'd call [this region](https://github.com/radius-project/radius/pull/7296/files#diff-5d9b9964242d77606f4868e3fe049978ea1c3f1903afcfeb60c76e19d9785fc5L253-L275) of the code the place "where the error is handled".
> > > 
> > > 
> > > Does the region of code you linked handle [this scenario](https://github.com/radius-project/radius/blob/84d4ca36ac6df46fb10489a3d666ff798d2a1904/pkg/corerp/backend/controller/deleteresource.go#L66) as well where the error returned from the controller is nil with a failed result?
> > > > Is the problem that there are two ways to represent a failure?
> > > 
> > > 
> > > The problem is that the errors aren't getting logged. There is no way to get the stack trace for debugging. If the root cause identified here is not correct, then this needs further investigation on why we are dropping logs for certain requests.
> > 
> > 
> > This log is also odd https://github.com/radius-project/radius/blob/main/pkg/armrpc/asyncoperation/worker/worker.go#L260. Why is it always hardcoding `result.Error` to nil? @youngbupark do you have context on what was the motivation here or is it a bug? If it is a bug then I think updating it will be a better approach over what I'm doing in this PR.
> 
> It is intended.
> 
> https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260
> 
> ```go
> logger.Info("Operation returned", "success", result.Error == nil, "code", code, "provisioningState", result.ProvisioningState(), "err", err)
> ```
> 
> If the operation is succeeded, result.Error must be nil. When logger processes this, it will be represented as JSON format log like below. The arguments from `"success"` until `err` are treated as a log dimension or attribution. Having such attributions such as success/code/provisioningState help indexing for search backend.
> 
> ```
> {"severity":"info","timestamp":"2024-03-12T20:43:42.002Z","name":"radius.radiusasyncworker","caller":"worker/worker.go:260","message":"Operation returned","serviceName":"radius","version":"edge","hostName":"applications-rp-bf77896d9-rfnk7","resourceId":"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/containers/backendcontainerdns","operationId":"64d729f1-2936-48eb-aabe-9a5fcaee2626","operationType":"APPLICATIONS.CORE/CONTAINERS|PUT","dequeueCount":1,"traceId":"9684bed9d6f01c4db253ab973274dfd1","spanId":"a15c41cd88fea0e1","success":true,"code":"","provisioningState":"Succeeded","err":null}
> ```

The problem with this code that I'm highlighting is that this log isn't exercised only for success scenarios, the current implementation updates the value for `code` [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L257) in case of an error, and then adds it [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) with the impression that the request has succeeded.

pkg/armrpc/asyncoperation/controller/result.go
youngbupark: I do not think printing out error in this helper is the right approach. If we add this approach, then we need to add the similar code to cancelled case as well. We might not want to have the same logging code here and there. I think we need to first figure it out the root cause about why error is not logged in main worker loop before making this change.

JFYI, all these helpers in this package are to create `ctrl.Result` with ARM compatible error like we use `rest.Response` in frontend controller. In the design of controllers, we want to have the consistent controller interfaces between frontend and backend controllers. In short, when we want to return the detail error response, we can return `rest.Response` for frontend and `ctrl.Result` for backend. If the error is internal error or hard to be categorized, controller can return `nil, err`. (err is go error), which is not recommended. Let's chat offline if you need more info here. 

* Frontend controller interface

https://github.com/radius-project/radius/blob/c77088f794f18e7478b10b34ee3041f600b613d9/pkg/armrpc/frontend/controller/controller.go#L103-L106

* Backend controller interface

https://github.com/radius-project/radius/blob/c77088f794f18e7478b10b34ee3041f600b613d9/pkg/armrpc/asyncoperation/controller/controller.go#L50

pkg/armrpc/asyncoperation/controller/result.go
```diff
@@ -40,20 +44,24 @@ func NewCanceledResult(message string) Result {
...
+	logger := ucplog.FromContextOrDiscard(ctx)
+	logger.Error(err, fmt.Sprintf("Operation Failed: %s", errorDetails.Message))
```
youngbupark, line 57: As I mentioned above, if we want to log out for the failure, cancellation is also failure. we need to put the same code there. We might not want to have the same code here and there. So we have logged the failure or status in main worker loop. I think we first need to figure it out why the current logging is not working as expected.

youngbupark (changes requested):

pkg/armrpc/asyncoperation/controller/result.go
kachawla: > I think we first need to figure it out why the current logging is not working as expected.

Some details on why it is not working is here: https://github.com/radius-project/radius/issues/7297#issuecomment-1989167067. Current design expects each controller to individually log an error, which isn't reliable as it requires specific knowledge and clearly hasn't worked well with misses in multiple places. The approach in this PR is one way to fix it. How would you recommend fixing this gap? If the root cause is something else according to your understanding, then please add details on that.

pkg/armrpc/asyncoperation/controller/result.go
kachawla: > I do not think printing out error in this helper is the right approach. If we add this approach, then we need to add the similar code to cancelled case as well. We might not want to have the same logging code here and there. 

I do think this should be added to cancelled requests, the same problem exists there. I intentionally left it out of the first iteration of this PR to agree on the approach and then extend it to cancelled.

> I think we need to first figure it out the root cause about why error is not logged in main worker loop before making this change.

I did call out the why in other comments. Here is one problem I called out https://github.com/radius-project/radius/pull/7296#issuecomment-2002620464, another here: https://github.com/radius-project/radius/pull/7296#discussion_r1529260580. Can you provide some context on why the original design was implemented expecting each controller to log the error?

> all these helpers in this package are to create ctrl.Result with ARM compatible error like we use rest.Response in frontend controller. In the design of controllers, we want to have the consistent controller interfaces between frontend and backend controllers. In short, when we want to return the detail error response, we can return rest.Response for frontend and ctrl.Result for backend. If the error is internal error or hard to be categorized, controller can return nil, err. (err is go error), which is not recommended.

I understand that. This PR is not changing controller interfaces or the result, it is purely abstracting out logging into where the result is set so that it is guaranteed to be logged. If NewFailedResult/SetFailed is called, there should be a corresponding error that is causing it to get to this state. If the same goal can be achieved by moving the logging to async worker, that's fine by me, as long as we have a shared component/piece of code that guarantees logging, and isn't leaving it to human memory.

pkg/armrpc/asyncoperation/controller/result.go
youngbupark: > > I think we first need to figure it out why the current logging is not working as expected.
> 
> Some details on why it is not working is here: [#7297 (comment)](https://github.com/radius-project/radius/issues/7297#issuecomment-1989167067). Current design expects each controller to individually log an error, which isn't reliable as it requires specific knowledge and clearly hasn't worked well with misses in multiple places.

To be clear, we have logged error individually in the main worker loop, not in controller. When controller creates Result, it needs to be expected to return Result and error. then the following code must log out the error. This the centralized logging part for all controllers.

https://github.com/radius-project/radius/blob/c77088f794f18e7478b10b34ee3041f600b613d9/pkg/armrpc/asyncoperation/worker/worker.go#L253-L260

> The approach in this PR is one way to fix it. How would you recommend fixing this gap? If the root cause is something else according to your understanding, then please add details on that.

I think the PR is the workaround. We first need to understand the root cause and then have the right fix here. in this comments - https://github.com/radius-project/radius/issues/7297#issuecomment-1989167067, we might not be able to repro the same issue.

I do not know what happens. but it might have numerous reasons. For example, RP could crash before logging. Another hypothesis is that when workflow step collects the logs, the log might be missing. Functional tests save the RP logs by using kubectl logs, which is not reliable.

youngbupark:
> The problem with this code that I'm highlighting is that this log isn't exercised only for success scenarios, the current implementation updates the value for `code` [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L257) in case of an error, and then adds it [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) with the impression that the request has succeeded.

Then I would change this code only to use logger.Error for error case and use logger.Info for normal scenario.

pkg/armrpc/asyncoperation/controller/result.go
kachawla: @youngbupark Let me ask this to build a better understanding of the current state -

1. Can you show where the code is that logs the error caught [here](https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/deleteresource.go#L98)?

2. Is this log statement unnecessary https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/createorupdateresource.go#L100?

kachawla:
> > The problem with this code that I'm highlighting is that this log isn't exercised only for success scenarios, the current implementation updates the value for `code` [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L257) in case of an error, and then adds it [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) with the impression that the request has succeeded.
> 
> Then I would change this code only to use logger.Error for error case and use logger.Info for normal scenario.

I'm not following this. Are you agreeing that there is a problem with the code I linked above?

youngbupark:
> > > The problem with this code that I'm highlighting is that this log isn't exercised only for success scenarios, the current implementation updates the value for `code` [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L257) in case of an error, and then adds it [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) with the impression that the request has succeeded.
> > 
> > 
> > Then I would change this code only to use logger.Error for error case and use logger.Info for normal scenario.
> 
> I'm not following this. Are you agreeing that there is a problem with the code I linked above?

I do not have the strong opinion to use logger.Error and logger.Info separately or not [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260). If you think using the different logger helper method is right for error case, it might be good to add logger.Error to [this code](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) instead of changing the other code including NewFailedResult helper.

BTW, I dug it out the logs on the issue, please see my comment - https://github.com/radius-project/radius/issues/7297#issuecomment-2005043589

kachawla:
> > > > The problem with this code that I'm highlighting is that this log isn't exercised only for success scenarios, the current implementation updates the value for `code` [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L257) in case of an error, and then adds it [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) with the impression that the request has succeeded.
> > > 
> > > 
> > > Then I would change this code only to use logger.Error for error case and use logger.Info for normal scenario.
> > 
> > 
> > I'm not following this. Are you agreeing that there is a problem with the code I linked above?
> 
> I do not have the strong opinion to use logger.Error and logger.Info separately or not [here](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260). If you think using the different logger helper method is right for error case, it might be good to add logger.Error to [this code](https://github.com/radius-project/radius/blob/02b7b205919dbb5d1cfbb3f7b53a10c90a44f4fe/pkg/armrpc/asyncoperation/worker/worker.go#L260) instead of changing the other codes including NewFailedResult helper.
> 
> BTW, I dug it out the original issue, please see my comment - [#7297 (comment)](https://github.com/radius-project/radius/issues/7297#issuecomment-2005043589)

Ok I will look into fixing it in async worker. Will sync up with you offline on these questions I have for my clarity https://github.com/radius-project/radius/pull/7296#discussion_r1529301772

pkg/armrpc/asyncoperation/controller/result.go
youngbupark: If the error was coming from the following code in the original issue, 

https://github.com/radius-project/radius/blob/c77088f794f18e7478b10b34ee3041f600b613d9/pkg/portableresources/processors/resourceclient.go#L279
 
All errors related to bicep should be caught in the following code first and logged before reaching worker's error handling code.

https://github.com/radius-project/radius/blob/c77088f794f18e7478b10b34ee3041f600b613d9/pkg/recipes/driver/bicep.go#L212-L226

But I could not find this error logs in the log files. It means that it

kachawla: So the test that failed [doesn't use a recipe](https://github.com/radius-project/radius/blob/main/test/functional-portable/daprrp/noncloud/resources/testdata/daprrp-resources-statestore-manual.bicep#L57) and won't go inside the driver code, which is why the log above isn't present. Looking through the logs and all the resources that the failed test creates, it failed during deletion of the container resource and the error in the test failure is likely coming from [here](https://github.com/radius-project/radius/blob/main/pkg/corerp/handlers/kubernetes.go#L169). I'm doing further investigation to see why the error wasn't logged.

The trace id that I see in the logs for deletion is different from the trace id in the functional test failure logs - so that's another area which needs further digging.

`
{"severity":"info","timestamp":"2024-03-04T21:30:59.433Z","name":"radius.radiusasyncworker","caller":"deployment/deploymentprocessor.go:341","message":"Deleting output resource: LocalID: KubernetesRole, resource type: \"Provider: kubernetes, Type: rbac.authorization.k8s.io/Role\"\n","serviceName":"radius","version":"edge","hostName":"applications-rp-7b774d7494-k6dv2","resourceId":"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/containers/dapr-sts-manual-ctnr","operationId":"489b1b1a-36dc-4219-bbd5-48ef3d23e90d","operationType":"APPLICATIONS.CORE/CONTAINERS|DELETE","dequeueCount":1,"traceId":"6cde20fc5dcbe244c35799b5ff94fea6","spanId":"35daee63ef2a07aa"}
`

youngbupark: I do not think that it happened while deleting secret as part of container deletion. Please see the following log:
image

It went further after deleting secret resource.

pkg/armrpc/asyncoperation/controller/result.go
youngbupark: > @youngbupark Let me ask this to build a better understanding of the current state -
> 
> 1. Can you share link to where the code is that logs the error caught [here](https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/deleteresource.go#L98)?

This is a backend controller. asyncCtrl.Run calls `func (c *DeleteResource[P, T]) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {`.

https://github.com/radius-project/radius/blob/2a681be2a36280a81177e5a1221f01accd33c89d/pkg/armrpc/asyncoperation/worker/worker.go#L253

For additional code flow, I recommend to debug the code with the debugger.

> 2. Is this log statement unnecessary https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/createorupdateresource.go#L100?

This is a server-side code. It may or may not. This is the error logging in controller. If controller author thinks it is useful with the better message and print out the error earlier, they can add it.

kachawla: > > @youngbupark Let me ask this to build a better understanding of the current state -
> > 
> > 1. Can you share link to where the code is that logs the error caught [here](https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/deleteresource.go#L98)?
> 
> This is a backend controller. asyncCtrl.Run calls `func (c *DeleteResource[P, T]) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {`.
> 
> https://github.com/radius-project/radius/blob/2a681be2a36280a81177e5a1221f01accd33c89d/pkg/armrpc/asyncoperation/worker/worker.go#L253
> 
> For additional code flow, I recommend to debug the code with the debugger.

I'm aware of where it is called from. I'm asking where is the error logged for the scenario I linked above.

> > 2. Is this log statement unnecessary https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/createorupdateresource.go#L100?
> 
> This is a server-side code. It may or may not. This is the error logging in controller. If controller author thinks it is useful with the better message and print out the error earlier, they can add it.

My question is - is this redundant?

youngbupark: > > > @youngbupark Let me ask this to build a better understanding of the current state -
> > > 
> > > 1. Can you share link to where the code is that logs the error caught [here](https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/deleteresource.go#L98)?
> > 
> > 
> > This is a backend controller. asyncCtrl.Run calls `func (c *DeleteResource[P, T]) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {`.
> > https://github.com/radius-project/radius/blob/2a681be2a36280a81177e5a1221f01accd33c89d/pkg/armrpc/asyncoperation/worker/worker.go#L253
> > 
> > For additional code flow, I recommend to debug the code with the debugger.
> 
> I'm aware of where it is called from. I'm asking where is the error logged for the scenario I linked above.

There are two places:

https://github.com/radius-project/radius/blob/2a681be2a36280a81177e5a1221f01accd33c89d/pkg/armrpc/asyncoperation/worker/worker.go#L260

https://github.com/radius-project/radius/blob/2a681be2a36280a81177e5a1221f01accd33c89d/pkg/armrpc/asyncoperation/worker/worker.go#L270

And I strongly recommend to debug the code flow to understand the details.

> 
> > > 2. Is this log statement unnecessary https://github.com/radius-project/radius/blob/main/pkg/portableresources/backend/controller/createorupdateresource.go#L100?
> > 
> > 
> > This is a server-side code. It may or may not. This is the error logging in controller. If controller author thinks it is useful with the better message and print out the error earlier, they can add it.
> 
> My question is - is this redundant?

As I said, "It may or may not". In general, If the log is useful for debugging the server, then it should not be redundant. From what I read, it is more readable than just printing out pure go error. I would reach out to yetkin who contribute to this code.

youngbupark, line 57: Another downside to add logging here is that whenever backend controller calls this method, it will log out. All these helpers is to construct and set the failure reason, not print out the log. adding logging here doesn't seem right..

kachawla: If we are logging the same error twice, then that is redundant imo. It was [added originally](https://github.com/radius-project/radius/pull/6276) because the error wasn't getting logged anywhere else (same problem that this PR was trying to solve).

PR #7298, opened 2024-03-10
Add cli validation for missing Bicep parameters

rynowak:
# Description

This change adds validation for parameters that required by a Bicep template but not supplied by the CLI. There are some nasty cases today that can occur when the application parameter is defined, but not supplied. Short circuiting this behavior on the CLI improves the user-experience.

## Type of change

Fixes: #issue_number

### DIFF
--- a/pkg/cli/bicep/envinject.go
+++ b/pkg/cli/bicep/envinject.go
@@ -35,12 +35,13 @@ func InjectApplicationParam(deploymentTemplate map[string]any, parameters map[st
 }
 
 func injectParam(deploymentTemplate map[string]any, parameters map[string]map[string]any, parameter string, value string) error {
-	if deploymentTemplate["parameters"] == nil {
-		return nil
+	formalParams, err := ExtractParameters(deploymentTemplate)
+	if err != nil {
+		return err
 	}
 
-	innerParameters := deploymentTemplate["parameters"].(map[string]any)
-	if innerParameters[parameter] == nil {
+	if _, ok := formalParams[parameter]; !ok {
+		// If we got here, it means 'parameter' is not a formal parameter of the template.
 		return nil
 	}
 
--- a/pkg/cli/bicep/parameters_test.go
+++ b/pkg/cli/bicep/parameters_test.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bicep
+
+import (
+	"encoding/json"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_ExtractParameters(t *testing.T) {
+	b, err := os.ReadFile("testdata/test-extractparameters.json")
+	require.NoError(t, err)
+
+	template := map[string]any{}
+	err = json.Unmarshal(b, &template)
+	require.NoError(t, err)
+
+	params, err := ExtractParameters(template)
+	require.NoError(t, err)
+
+	expected := map[string]any{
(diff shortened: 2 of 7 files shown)


rynowak:
Example output:

```txt
❯ go run cmd/rad/main.go deploy /Users/ryan/testing/app.bicep
Building /Users/ryan/testing/app.bicep...
/Users/ryan/testing/app.bicep(5,7) : Warning no-unused-params: Parameter "location" is declared but never used. [https://aka.ms/bicep/linter/no-unused-params]
The template "/Users/ryan/testing/app.bicep" could not be deployed because of the following errors:

  - The template requires an application. Use --application to specify the application name.
  - The template requires a parameter "location". Use --parameters location=<value> to specify the value.
```

pkg/cli/bicep/envinject.go
```diff
@@ -35,12 +35,13 @@ func InjectApplicationParam(deploymentTemplate map[string]any, parameters map[st
...
-		return nil
+	formalParams, err := ExtractParameters(deploymentTemplate)
+	if err != nil {
+		return err
 	}
 
-	innerParameters := deploymentTemplate["parameters"].(map[string]any)
-	if innerParameters[parameter] == nil {
+	if _, ok := formalParams[parameter]; !ok {
+		// If we got here, it means 'parameter' is not a formal parameter of the template.
```
rynowak, line 44: I need to make an update here because parameters are actually case-sensitive. We didn't handle this case correctly for `application` and `environment`. Now that we're validating this more strictly it's causing failures in tests.

vinayada1 (dismissed):

pkg/cli/bicep/parameters_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	template := map[string]any{}
+	err = json.Unmarshal(b, &template)
+	require.NoError(t, err)
+
+	params, err := ExtractParameters(template)
+	require.NoError(t, err)
+
+	expected := map[string]any{
+		"application": map[string]any{
```
ytimocin, line 39: Should we try to see if it fails with a capital letter (Application)?

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -303,3 +320,73 @@ func (r *Runner) Run(ctx context.Context) error {
...
+func (r *Runner) injectAutomaticParameters(template map[string]any) error {
+	if r.Providers.Radius.EnvironmentID != "" {
+		err := bicep.InjectEnvironmentParam(template, r.Parameters, r.Providers.Radius.EnvironmentID)
+		if err != nil {
+			return err
+		}
+	}
+
+	if r.Providers.Radius.ApplicationID != "" {
+		err := bicep.InjectApplicationParam(template, r.Parameters, r.Providers.Radius.EnvironmentID)
```
ytimocin, line 333:
```suggestion
		err := bicep.InjectApplicationParam(template, r.Parameters, r.Providers.Radius.ApplicationID)
```

ytimocin: How can this be covered in unit tests?

ytimocin (changes requested):

pkg/cli/cmd/deploy/deploy.go
rynowak: Good catch, adding a test for this.

pkg/cli/bicep/parameters_test.go
rynowak: This code path doesn't compare parameter names, it just reads them.

rynowak:
@ytimocin thanks! updated

ytimocin (approved):

pkg/cli/bicep/envinject.go
```diff
@@ -35,12 +35,13 @@ func InjectApplicationParam(deploymentTemplate map[string]any, parameters map[st
 }
 
 func injectParam(deploymentTemplate map[string]any, parameters map[string]map[string]any, parameter string, value string) error {
-	if deploymentTemplate["parameters"] == nil {
-		return nil
+	formalParams, err := ExtractParameters(deploymentTemplate)
```
kachawla, line 38: Could be just me - what do we mean by "formal" params?

kachawla (approved):

pkg/cli/bicep/envinject.go
rynowak: You need a term to tell the difference between parameters that were declared, and the values associated with those parameters.

- declaration: `param foo string`
- value `--parameters foo=bar`

Usually the declaration is called formal parameters.

PR #7303, opened 2024-03-11
Only set Functional Test Run as passing if all subtests pass

### LINKED ISSUE #6939: Functional test checks don't work correctly

willdavsmith:
# Description

* Fixing bug with test reporting
* Fixing some issues in functional test workflow

Note: Since we don't have a way to test this in the main repo, I tested the different paths (failed tests, cancelled tests, passing tests) manually on my fork here: https://github.com/willdavsmith/radius/pull/7

## Type of change

Fixes: #6939

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -309,7 +309,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest]
-        name: [ucp,kubernetes,shared,msgrp,daprrp,samples,cli]
+        name: [ucp,kubernetes,msgrp,daprrp,samples]
         include:
           # datastorerp functional tests need the larger VM.
           - os: ubuntu-latest-m
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -644,22 +644,22 @@ jobs:
         id: get_test_status
         run: |
           # from: https://github.com/orgs/community/discussions/26526#discussioncomment-3252209
-          JOB_STATUS=$(curl -X GET -s -u "admin:${{ steps.get_installation_token.outputs.token }}" https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs | jq ".jobs[] | {job_status: .conclusion, matrix: .name}")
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -414,7 +413,7 @@ jobs:
           # AZURE_OIDC_ISSUER
           eval "export $(echo "${{ secrets.FUNCTEST_AZURE_OIDC_JSON }}" | jq -r 'to_entries | map("\(.key)=\(.value)") | @sh')"
 
-          AUTHKEY=$(echo -n "${{ github.actor }}:${{ secrets.GH_RAD_CI_BOT_PAT }}" | base64)
+          AUTHKEY=$(echo -n "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}" | base64)
```
rynowak, line 416: Does this work? What's this token being used for? I would expect the github token to no longer be valid once the test run is over.

.github/workflows/functional-test.yaml
```diff
@@ -628,6 +627,40 @@ jobs:
...
+          TEST_STATUS="success"
+          for job_status in $(echo "$ALL_JOBS_STATUS" | jq -r '.[]'); do
+            echo "Job Status: $job_status"
+            if [[ "$job_status" == "failure" ]]; then
+              echo "Found failed test. Setting test status to failure"
+              TEST_STATUS="failure"
+              break
+            elif [[ "$job_status" == "cancelled" ]]; then
+              echo "Found cancelled test. Setting test status to cancelled"
+              TEST_STATUS="cancelled"
```
rynowak, line 659: Should we also have a break here?

.github/workflows/functional-test.yaml
willdavsmith: I can remove this - only had it there for fork testing.

.github/workflows/functional-test.yaml
willdavsmith: I added a break for the failure but not the cancelled because I think it's better to report "tests failed" instead of "tests cancelled" if both occur. If all tests pass but some are cancelled, we will still report the status as cancelled because we only set as "success" if we don't go in this conditional.

rynowak: ok

rynowak (approved):

youngbupark (approved):

PR #7306, opened 2024-03-12
Creating an optional interface to get secret id references for driver with secrets.

vishwahiremat:
# Description

Refactoring module secret workflow.
- Adding FindSecretIds function to get module secret references for driver with secrets.

## Type of change

- Creating an optional interface to get secret id references for driver with secrets. 

Fixes: #issue_number


pkg/recipes/driver/types.go
```diff
@@ -42,6 +42,13 @@ type Driver interface {
 	GetRecipeMetadata(ctx context.Context, opts BaseOptions) (map[string]any, error)
 }
 
+type DriverWithSecrets interface {
+	Driver
+
+	// FindSecretIDs gets the secret ID references from environment definition.
+	FindSecretIDs(ctx context.Context, config recipes.Configuration, definition recipes.EnvironmentDefinition) (string, error)
```
rynowak, line 49: Should this return `([]string, error)`?

What would happen in the future if a driver needed to read multiple secrets?

pkg/recipes/engine/engine.go
```diff
@@ -92,19 +92,19 @@ func (e *engine) executeCore(ctx context.Context, recipe recipes.ResourceMetadat
...
+		secretStore, err := driverWithSecrets.FindSecretIDs(ctx, *configuration, *definition)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to fetch secrets from the secret store resource id %s for Terraform recipe %s deployment: %w", secretStore, definition.TemplatePath, err)
+			return nil, nil, err
+		}
+		// Retrieves the secret values from the secret store ID provided.
+		if secretStore != "" {
+			secrets, err = e.options.SecretsLoader.LoadSecrets(ctx, secretStore)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to fetch secrets from the secret store resource id %s for Terraform recipe %s deployment: %w", secretStore, definition.TemplatePath, err)
```
rynowak, line 106: Does this need its own error code?

pkg/recipes/driver/terraform.go
```diff
@@ -249,6 +249,11 @@ func (d *terraformDriver) GetRecipeMetadata(ctx context.Context, opts BaseOption
 	return recipeData, nil
 }
 
+func (d *terraformDriver) FindSecretIDs(ctx context.Context, envConfig recipes.Configuration, definition recipes.EnvironmentDefinition) (string, error) {
```
youngbupark, line 252: nit: Please add comment.

pkg/recipes/driver/types.go
```diff
@@ -42,6 +42,13 @@ type Driver interface {
 	GetRecipeMetadata(ctx context.Context, opts BaseOptions) (map[string]any, error)
 }
 
+type DriverWithSecrets interface {
```
youngbupark, line 45: nit: Please add comment

pkg/recipes/driver/terraform.go
```diff
@@ -249,6 +249,11 @@ func (d *terraformDriver) GetRecipeMetadata(ctx context.Context, opts BaseOption
...
+
+	return recipes.GetSecretStoreID(envConfig, definition.TemplatePath)
```
kachawla, line 254: Why does it call into another function in recipes pkg? Any reason why we aren't moving the implementation here?

pkg/recipes/driver/types.go
```diff
@@ -42,6 +42,13 @@ type Driver interface {
...
+	Driver
+
+	// FindSecretIDs gets the secret ID references from environment definition.
```
kachawla, line 48: Could you add more details in this comment? What is the secret ID for example? Where does it read it from in the environment definition - recipe config. Why is it needed - to access private modules (for now, with providers change it would extend to secrets passed into the providers config as well)

pkg/recipes/driver/types.go
vishwahiremat: currently we are keeping it return module secret. But in the future when we have provider secrets etc, return type will be map because we may need to specify secret kind then ex: module secret or provider secret. It will be discussed in detail during the design for provider secrets.

pkg/recipes/driver/terraform.go
vishwahiremat: currently we cannot, because we are using this function in config.go to get the template path prefix, we can move the implementation here when we have containerization.

kachawla: Ah right. Can you please add that in the comment and call it out in the ?

pkg/recipes/engine/engine.go
vishwahiremat: I agree, we could add a new error code for load secrets failures. 
thoughts on naming it as   `LOAD_SECRETS_FAILED` ?

rynowak: ARM-like error codes PascalCase so it would be `LoadSecretsFailed`

pkg/recipes/driver/types.go
```diff
@@ -42,6 +42,16 @@ type Driver interface {
...
 }
 
+// DriverWithSecrets is an optional interface and used when the driver needs to load secrets for recipe deployment.
+type DriverWithSecrets interface {
+	// Driver is an interface to implement recipe deployment and recipe resources deletion.
+	Driver
+
+	// FindSecretIDs gets the secret ID references associated with git private terraform repository source.
+	// In the future it will be extended to get secret references for provider secrets.
+	FindSecretIDs(ctx context.Context, config recipes.Configuration, definition recipes.EnvironmentDefinition) (string, error)
```
kachawla, line 52: Can we generate mocks for this and leverage that for unit testing in engine? I'm not sure why go generate was originally added to the [bicep driver](https://github.com/radius-project/radius/blob/9de3a3e02eec675bc408aa88d57f41ac87baa7b9/pkg/recipes/driver/bicep.go#L48) instead of the types file.

vishwahiremat: added mocks and unit tests

pkg/recipes/driver/terraform.go
```diff
@@ -249,6 +249,14 @@ func (d *terraformDriver) GetRecipeMetadata(ctx context.Context, opts BaseOption
 	return recipeData, nil
 }
 
+// FindSecretIDs is used to get the secret reference associated with git private terraform repository source.
```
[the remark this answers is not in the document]
vishwahiremat: updated it

pkg/recipes/driver/types.go
```diff
@@ -42,6 +44,18 @@ type Driver interface {
...
 }
 
+// DriverWithSecrets is an optional interface and used when the driver needs to load secrets for recipe deployment.
+//
+//go:generate mockgen -destination=./mock_driver_with_secrets.go -package=driver -self_package github.com/radius-project/radius/pkg/recipes/driver github.com/radius-project/radius/pkg/recipes/driver DriverWithSecrets
+type DriverWithSecrets interface {
+	// Driver is an interface to implement recipe deployment and recipe resources deletion.
+	Driver
+
+	// FindSecretIDs gets the secret ID references associated with git private terraform repository source.
```
kachawla, line 54:
```suggestion
	// FindSecretIDs gets the secret store resource ID references associated with git private terraform repository source.
```

kachawla (dismissed):

kachawla (approved):

kachawla (commented):
@vishwahiremat can this be merged?

lakshmimsft (approved):

PR #7337, opened 2024-03-15
Use workload identity in workflow

### LINKED ISSUE #7309: Switch GitHub Actions over to using Azure managed identities
Today we use a service principal secret in order to authenticate to Azure for functional tests. We should move this over to managed identities using OIDC federation. This prevents credentials from needing to be rotated.

Link here: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure

[AB#11483](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11483)

youngbupark:
# Description

This is to use workload identity instead of secret auth for service principal.

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -306,12 +306,14 @@ jobs:
           mkdir -p ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}
           helm package ${{ env.HELM_CHARTS_DIR }} --version ${{ env.CHART_VERSION }} --app-version ${{ env.REL_VERSION }} --destination ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}
       # TODO: Delete this step once we use GHCR as the helm chart repo.
-      - name: Login to Azure
-        uses: azure/login@v2
-        with:
-          client-id: ${{ secrets.AZURE_SP_TESTS_APPID }}
-          tenant-id: ${{ secrets.AZURE_SP_TESTS_TENANTID }}
-          subscription-id: ${{ secrets.AZURE_SUBSCRIPTIONID_TESTS }}
+      - name: Setup Azure CLI
+        run: curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+      - name: az CLI login
+        run: |
+          az login --service-principal \
+            --username ${{ secrets.AZURE_SP_TESTS_APPID }} \
+            --password ${{ secrets.AZURE_SP_TESTS_PASSWORD }} \
+            --tenant ${{ secrets.AZURE_SP_TESTS_TENANTID }}
       # TODO: Delete this step once we use GHCR as the helm chart repo.
       - name: Push helm chart to ACR
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -380,7 +386,9 @@ jobs:
       - name: Login to Azure
         uses: azure/login@v1
         with:
-          creds: '{"clientId":"${{ secrets.INTEGRATION_TEST_SP_APP_ID }}","clientSecret":"${{ secrets.INTEGRATION_TEST_SP_PASSWORD }}","subscriptionId":"${{ secrets.INTEGRATION_TEST_SUBSCRIPTION_ID }}","tenantId":"${{ secrets.INTEGRATION_TEST_TENANT_ID }}"}'
+          client-id: ${{ secrets.INTEGRATION_TEST_SP_APP_ID }}
```
AaronCrawfis, line 389: While we're in here making changes can we switch this over to secrets.AZURE_SP_TESTS_APPID?

It's the same service principal behind the scenes, but we've been using separate secrets for it. A couple years ago we switched to an org-scoped secret but this workflow wasn't updated to use it.

AaronCrawfis: image

.github/workflows/functional-test.yaml
```diff
@@ -380,7 +386,9 @@ jobs:
       - name: Login to Azure
         uses: azure/login@v1
```
AaronCrawfis, line 387: V2 was also released last month which adds Node 20 support and fixes some error messages.

.github/workflows/functional-test.yaml
youngbupark: Updated. Do you have new subscription id for test ?

AaronCrawfis: Ah yeah `AZURE_SUBSCRIPTIONID_TESTS` should be the correct value:
image

Although, would a variable be better than a secret since it's not a sensitive value?

ytimocin (dismissed):

.github/workflows/functional-test.yaml
youngbupark: > Ah yeah `AZURE_SUBSCRIPTIONID_TESTS` should be the correct value: image
> 
> Although, would a variable be better than a secret since it's not a sensitive value?

secret is better than variable. if you can create new secret, please let me know.

kachawla: If it is being added as a secret, then can you make sure to document the value for AZURE_SUBSCRIPTIONID_TESTS in our wiki?

youngbupark: > If it is being added as a secret, then can you make sure to document the value for AZURE_SUBSCRIPTIONID_TESTS in our wiki?

looks like Aaron added it already.

.github/workflows/build.yaml
```diff
@@ -306,6 +306,7 @@ jobs:
           mkdir -p ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}
           helm package ${{ env.HELM_CHARTS_DIR }} --version ${{ env.CHART_VERSION }} --app-version ${{ env.REL_VERSION }} --destination ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}
       # TODO: Delete this step once we use GHCR as the helm chart repo.
+      # Cannot use Workload Identity because azure federated identity doesn't accept wildcard tag version.
```
youngbupark, line 309: I cannot remove this code because we still use ACR for helm chart. I think we need to move helm chart to ghcr as soon as possible.  cc/ @willdavsmith  @rynowak  @AaronCrawfis

AaronCrawfis (approved):

kachawla (commented):
Thanks for doing this @youngbupark! Could you also update [this wiki](https://dev.azure.com/azure-octo/Project%20Radius/_wiki/wikis/Project-Radius.wiki/246/App-registrations) to reflect the current state?

PR #7357, opened 2024-03-20
Migrate long-running test to new cluster to enable monitoring

youngbupark:
# Description

This is to migrate long-running test to new cluster:

1. Update bicep template for test cluster due to the deprecated resource
2. Enable manual test run
3. Move to new cluster.
4. Create Issue only for schedule trigger

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/scripts/cleanup-cluster.sh
+++ b/.github/scripts/cleanup-cluster.sh
@@ -22,7 +22,7 @@ echo "cleaning up cluster"
 
 # Delete all test resources in queuemessages.
 echo "delete all resources in queuemessages.ucp.dev"
-kubectl delete queuemessages.ucp.dev -n radius-system --all
+kubectl delete queuemessages.ucp.dev -n radius-system --all || true
 
 # Testing deletion of deployment.apps.
 
--- a/.github/scripts/cleanup-cluster.sh
+++ b/.github/scripts/cleanup-cluster.sh
@@ -28,7 +28,7 @@ kubectl delete queuemessages.ucp.dev -n radius-system --all || true
 
 # Delete all test resources in resources without proxy resource.
 echo "delete all resources in resources.ucp.dev"
-resources=$(kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name")
+resources=$(kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" || true)
 for r in $resources; do
     if [[ $r == scope.local.* || $r == scope.aws.* || -z "$r" ]]; then
         echo "skip deletion: $r"
@@ -40,7 +40,7 @@ done
 
 # Delete all test namespaces.
 echo "delete all test namespaces"
-namespaces=$(kubectl get namespace |
+namespaces=$(kubectl get namespace || true |
     grep -E '^kubernetes-interop-tutorial.*|^corerp.*|^default-.*|^radiusfunctionaltestbucket.*|^radius-test.*|^kubernetes-cli.*|^dpsb-.*|^dsrp-.*|^azstorage-workload.*|^dapr-serviceinvocation|^daprrp-rs-.*|^mynamespace.*|^demo.*|^tutorial-demo.*|^ms.+' |
     awk '{print $1}')
 for ns in $namespaces; do
--- a/.github/scripts/cleanup-cluster.sh
+++ b/.github/scripts/cleanup-cluster.sh
@@ -40,7 +40,7 @@ done
 
 # Delete all test namespaces.
 echo "delete all test namespaces"
-namespaces=$(kubectl get namespace || true |
+namespaces=$(kubectl get namespace |
     grep -E '^kubernetes-interop-tutorial.*|^corerp.*|^default-.*|^radiusfunctionaltestbucket.*|^radius-test.*|^kubernetes-cli.*|^dpsb-.*|^dsrp-.*|^azstorage-workload.*|^dapr-serviceinvocation|^daprrp-rs-.*|^mynamespace.*|^demo.*|^tutorial-demo.*|^ms.+' |
     awk '{print $1}')
 for ns in $namespaces; do
--- a/.github/scripts/cleanup-cluster.sh
+++ b/.github/scripts/cleanup-cluster.sh
@@ -21,22 +21,26 @@ set -e
 echo "cleaning up cluster"
 
 # Delete all test resources in queuemessages.
-echo "delete all resources in queuemessages.ucp.dev"
-kubectl delete queuemessages.ucp.dev -n radius-system --all || true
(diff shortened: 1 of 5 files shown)


.github/scripts/cleanup-cluster.sh
```diff
@@ -22,13 +22,13 @@ echo "cleaning up cluster"
 
 # Delete all test resources in queuemessages.
 echo "delete all resources in queuemessages.ucp.dev"
-kubectl delete queuemessages.ucp.dev -n radius-system --all
+kubectl delete queuemessages.ucp.dev -n radius-system --all || true
```
kachawla, line 25: Is there any potential downside in ignoring failures here?

.github/workflows/long-running-azure.yaml
```diff
@@ -44,6 +44,7 @@ permissions:
   packages: write # Required for uploading the package
 
 on:
+  workflow_dispatch:
```
kachawla, line 47: Could you add a comment on when is it ok to manually trigger this workflow? Anything to keep in mind?

.github/workflows/long-running-azure.yaml
```diff
@@ -82,9 +83,9 @@ env:
   VALID_RADIUS_BUILD_WINDOW: 86400
 
   # The AKS cluster name
-  AKS_CLUSTER_NAME: "radiuse2e00-aks"
+  AKS_CLUSTER_NAME: "radiuse2e000-aks"
```
kachawla, line 86: Where do we get this name from?

.github/scripts/cleanup-cluster.sh
youngbupark: This is not the final code. I am still working on it.

.github/workflows/long-running-azure.yaml
youngbupark: This is new cluster name. We can define whatever we want. I deployed this cluster using the bicep template in /test/infra/azure. I will document it in on-call wiki page once this change is completed.

kachawla: I see. Could you also include this information in the comment [here](https://github.com/radius-project/radius/blob/24de5f5d0f9537c66685a78f825ac60bc0760e0d/.github/workflows/long-running-azure.yaml#L85) to make it easier for anyone looking at it to understand.

kachawla: Also, could we give it a more explicit name, something like longhaultest? What is the significance of `2e000`?

youngbupark: I will create new one, `rad-longrunningtest-00`. `00` is the sequence number of cluster for the case when we want to migrate.

.github/scripts/cleanup-cluster.sh
kachawla: Ok, I'll stop looking at it until it is ready then.

.github/workflows/long-running-azure.yaml
youngbupark: added the comment

.github/workflows/long-running-azure.yaml
youngbupark: There is the resource name limitation for grafana dashboard and loganalytics. We can use only alphabet + numbers but we cannot exceed 13 lengths. So I created new cluster with `radlrtest00`.

.github/scripts/cleanup-cluster.sh
youngbupark: please review it again.

.github/workflows/long-running-azure.yaml
kachawla: sounds good, thanks!

.github/scripts/cleanup-cluster.sh
```diff
@@ -21,22 +21,26 @@ set -e
...
 
 # Delete all test resources in queuemessages.
-echo "delete all resources in queuemessages.ucp.dev"
-kubectl delete queuemessages.ucp.dev -n radius-system --all
+if kubectl get crd queuemessages.ucp.dev > /dev/null 2>&1; then
+    echo "delete all resources in queuemessages.ucp.dev"
+    kubectl delete queuemessages.ucp.dev -n radius-system --all
+fi
 
 # Testing deletion of deployment.apps.
```
kachawla, line 29: Is this comment valid or was it added for some testing? I know it is unrelated to the changes in the PR but I just came across it so thought to ask

kachawla (approved):

.github/workflows/long-running-azure.yaml
```diff
@@ -82,9 +84,9 @@ env:
   VALID_RADIUS_BUILD_WINDOW: 86400
 
   # The AKS cluster name
-  AKS_CLUSTER_NAME: "radiuse2e00-aks"
+  AKS_CLUSTER_NAME: "radlrtest00-aks"
```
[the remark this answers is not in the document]
youngbupark: > Would these names be better as Actions Variables instead of hard-coded in the workflow file? Would allow us to quickly update the values without needing to check in a change.

Sure. Please add new variable for cluster name. I can update it.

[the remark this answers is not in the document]
AaronCrawfis: Since these would probably be best as repo-scoped variables, you should be able to edit them here: https://github.com/radius-project/radius/settings/variables/actions. Lemme know if you hit any access issues.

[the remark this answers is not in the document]
youngbupark: > Since these would probably be best as repo-scoped variables, you should be able to edit them here: https://github.com/radius-project/radius/settings/variables/actions. Lemme know if you hit any access issues.

I do not have the permission.

.github/scripts/cleanup-cluster.sh
ytimocin: I added this and forgot to remove.

.github/workflows/long-running-azure.yaml
```diff
@@ -34,7 +34,7 @@
 # the previous build is still valid. If valid, the workflow skips the build steps
 # and uses the cached 'rad cli' for testing.
 #
-# Grafana dashboard URL: https://radiuse2e00-dashboard-audycmffgberbghy.wus3.grafana.azure.com/
+# Grafana dashboard URL:  https://radlrtest00-dashboard-e4ffc0cwggchdhba.wus3.grafana.azure.com
```
ytimocin, line 37: nit: there are 2 spaces here

ytimocin (approved):

.github/workflows/long-running-azure.yaml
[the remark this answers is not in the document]
youngbupark: I merged it now. I will create new pr to use new var.

PR #7364, opened 2024-03-21
Add required_providers block to main.tf.config, functional test for Terraform Provider

### LINKED ISSUE #7311: Radius currently does not read Terraform `required_providers` block which causes issues for providers not directly from Hashicorp
### Steps to reproduce

1) Create a `main.tf` that contains for your TF Recipe:

```
terraform {
  required_providers {
    azuredevops = {
      source  = "microsoft/azuredevops"
      version = ">=0.1.0"
    }
  }
}

resource "azuredevops_project" "project" {
  name        = "Test Project"
  description = "Test Project Description"
}
```

2) Create a Radius Environment in a `env.bicep` and replace with your:

`templatePath` is a param that needs to be changed and pointed to your github repo

Please also fill out the `providers` params in the `recipeConfig` with the appropriate values

```
import radius as radius
 
 
resource env 'Applications.Core/environments@2023-10-01-preview' = {
  name: 'azd-resources-environment-azdql-recipe-env'
  location: 'global'
  properties: {
    compute: {
      kind: 'kubernetes'
      resourceId: 'self'
      namespace: 'azd-resources-environment-azdql-recipe-ns'
    }
    recipeConfig: {
      terraform:{
        providers:{
          azuredevops:[{
            org_service_url : 'https://dev.azure.com/xxxx'
            client_id     : 'xxx'
  	    tenant_id     : 'xxx'
            client_secret : 'xxx'
          }]
        }
      }	
    }
    recipes: {
      'Applications.Core/extenders': {
        defaultazd: {
          templateKind: 'terraform'
          templatePath:'git::https://git......
        }
      }
    }
  }
}
 
resource app 'Applications.Core/applications@2023-10-01-preview' = {
  name: 'azd-resources-azdql-recipe-app'
  location: 'global'
  properties: {
    environment: env.id
    extensions: [
      {
        kind: 'kubernetesNamespace'
        namespace: 'azd-resources-azdql-default-recipe-app'
      }
    ]
  }
}
 
resource azdapp 'Applications.Core/extenders@2023-10-01-preview' = {
  name: 'azd-resources-terraform-azdapp1'
  properties: {
    application: app.id
    environment: env.id
    recipe: {
      name: 'defaultazd'
    }
  }
}
```

3) Deploy the bicep file

### Observed behavior

During deployment the Recipe fails with the output similar to:

```
"message": "terraform init failure: exit status 1\n\nError: Failed to query available provider packages\n\nCould not retrieve the list of available versions for provider\nhashicorp/azuredevops: provider registry registry.terraform.io does not have\na provider named registry.terraform.io/hashicorp/azuredevops\n\nDid you intend to use microsoft/azuredevops? If so, you must specify that\nsource address in each module which requires that provider. To see which\nmodules are currently depending on hashicorp/azuredevops, run the following\ncommand:\n    terraform providers\n\n"
```

### Desired behavior

Deployment is successful

### Workaround

_No response_

### rad Version

0.31.0    v0.31.0   0.31.0    3fa840730295681733357e7c94e6813d04cfae6d

### Operating system

_No response_

### Additional context

- https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/guides/authenticating_service_principal_using_a_client_secret

[AB#11488](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11488)

lakshmimsft:
# Description

PR includes updates to Terraform Providers functionality including:
1. Added required_providers block to main.tf.config file, (reading data from the loaded recipe module).
2. Updated Save to include a json encoder to not escape HTML characters so data can be saved into config file as authored.
3. Added a 'providers' block to module section if providers include 'alias', in main.tf.config in UpdateModuleWithProviderAliases()
4.  Updated unit tests and added functional test.

## Type of change

- This pull request fixes a bug in Radius and has an approved issue (#7311 ).

Fixes: #7311


pkg/recipes/terraform/config/config.go
```diff
@@ -190,23 +237,26 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpConf
...
+func (cfg *TerraformConfig) AddTerraformInfrastructure(resourceRecipe *recipes.ResourceMetadata, backend backends.Backend, requiredProviders map[string]*RequiredProviderInfo) (map[string]any, error) {
 	backendConfig, err := backend.BuildBackend(resourceRecipe)
 	if err != nil {
 		return nil, err
 	}
+
 	cfg.Terraform = &TerraformDefinition{
-		Backend: backendConfig,
+		Backend:           backendConfig,
+		RequiredProviders: requiredProviders,
```
kachawla, line 259: What's the motivation of adding this with backend instead of where providers are added?

pkg/recipes/terraform/config/config.go
```diff
@@ -190,23 +237,26 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpConf
...
 	return providerConfigs, nil
 }
 
-// AddTerraformBackend adds backend configurations to store Terraform state file for the deployment.
+// AddTerraformInfrastructure adds backend configurations to store Terraform state file for the deployment.
+// It also sets the required providers for the Terraform configuration.
 // Save() must be called to save the generated backend config.
 // Currently, the supported backend for Terraform Recipes is Kubernetes secret. https://developer.hashicorp.com/terraform/language/settings/backends/kubernetes
-func (cfg *TerraformConfig) AddTerraformBackend(resourceRecipe *recipes.ResourceMetadata, backend backends.Backend) (map[string]any, error) {
+func (cfg *TerraformConfig) AddTerraformInfrastructure(resourceRecipe *recipes.ResourceMetadata, backend backends.Backend, requiredProviders map[string]*RequiredProviderInfo) (map[string]any, error) {
```
kachawla, line 251: Not sure if `Infrastructure` is the right term to use here

pkg/recipes/terraform/config/config.go
```diff
@@ -167,19 +209,24 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
 	// Get recipe provider configurations from the environment configuration
 	providerConfigs := providers.GetRecipeProviderConfigs(ctx, envConfig)
 
 	// Build provider configurations for required providers excluding the ones already present in providerConfigs
-	for _, provider := range requiredProviders {
-		if _, ok := providerConfigs[provider]; ok {
-			// Environment level recipe configuration for providers will take precedence over
-			// UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
+	for providerName := range requiredProviders {
+		if _, ok := ucpConfiguredProviders[providerName]; ok { // requiredProviders can contain providers not configured with UCP
```
kachawla, line 218: Why is this additional check needed?

lakshmimsft: Existing code was working with scenarios where required_providers matched/were subset of ucpConfiguredProviders. We are now expanding to include scenarios where required_providers can contain providers not configured with UCP (eg. cyrilgdn/postgresql). 
The loop is for provider configuration at the envConfig level to supercede ucpconfiguration for providers when they exist in required_providers.
pls let me know if more comments are needed.

pkg/recipes/terraform/config/config.go
lakshmimsft: Terraform expects the required_providers configuration to be included with backend. This is needed in order to ensure processing of providers which it does not recognize having a default configuration.
It is also beneficial for providers which Terraform may use default configuration for, so we use the user-intended required_provider configuration instead of default versions.

pkg/recipes/terraform/config/config.go
lakshmimsft: yes, open to suggestions, AddTerraformBackendAndReqProviders is another option.

pkg/recipes/terraform/config/config.go
```diff
@@ -96,14 +97,20 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
...
-	jsonData, err := json.MarshalIndent(cfg, "", "  ")
-	if err != nil {
+	// Create a buffer to write the JSON to
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "  ")
+
+	// Encode the Terraform config to JSON
+	if err := enc.Encode(cfg); err != nil {
```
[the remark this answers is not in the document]
lakshmimsft: updated

pkg/recipes/terraform/config/config.go
```diff
@@ -96,14 +97,20 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
...
 		return fmt.Errorf("error marshalling JSON: %w", err)
 	}
+	jsonData := strings.TrimSuffix(buf.String(), "\n")
```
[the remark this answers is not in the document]
lakshmimsft: The newline is coming from json encoding. We typically do not have trailing new lines in our test files, across in the codebase. They will need to be updated for this if we were to remove this. I'm keeping this in for now.

pkg/recipes/terraform/config/config.go
```diff
@@ -96,24 +97,32 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
...
 	// https://developer.hashicorp.com/terraform/language/syntax/json
 
-	// Convert the Terraform config to JSON
-	jsonData, err := json.MarshalIndent(cfg, "", "  ")
-	if err != nil {
+	// Create a buffer to write the JSON to
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "  ")
```
ytimocin, line 104: Why are there two spaces here? Should we add a comment here so that the next person to work on this part could know better?

pkg/recipes/terraform/config/config.go
```diff
@@ -127,6 +136,41 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
 }
 
+// UpdateModuleWithProviderAliases updates the module provider configuration in the Terraform config
+// by adding aliases to the provider configurations.
+func (cfg *TerraformConfig) UpdateModuleWithProviderAliases(ctx context.Context) error {
+	if cfg == nil {
+		return fmt.Errorf("terraform configuration is not initialized")
+	}
+
+	providerConfigs := cfg.Provider
```
ytimocin, line 146: Can this be empty?

ytimocin: **providerConfigs** is plural but **cfg.Provider** is singular. Should we update one of them so that we have consistency?

pkg/recipes/terraform/config/config.go
```diff
@@ -127,6 +136,41 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+
+	// For each provider in the providerConfigs, if provider has a property "alias",
+	// add entry to the module provider configuration
+	for providerName, providerConfigList := range providerConfigs {
+		providerConfigDetails, ok := providerConfigList.([]map[string]any)
+		if !ok {
+			continue
+		}
+		for _, providerConfig := range providerConfigDetails {
+			if alias, ok := providerConfig["alias"]; ok {
```
ytimocin, line 157: Is **alias** only going to be a string? Above in line 152 we are casting the **providerConfigList** to map[string]any.

pkg/recipes/terraform/config/config.go
```diff
@@ -167,19 +211,24 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
-			// UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
+	for providerName := range requiredProviders {
+		if _, ok := ucpConfiguredProviders[providerName]; ok { // requiredProviders can contain providers not configured with UCP
+			if _, ok := providerConfigs[providerName]; ok {
+				// Environment level recipe configuration for providers will take precedence over
+				// UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
+				continue
+			}
+		} else {
+			// If the provider under required_providers is not configured with UCP, skip this iteration.
```
ytimocin, line 227: I think a good documentation here is necessary. I see that we `continue` if both `ucpConfiguredProviders` and `providerConfigs` contain the **providerName**. Or if `ucpConfiguredProviders` doesn't contain... It is a little bit complicated here.

pkg/recipes/terraform/module.go
```diff
@@ -73,9 +75,32 @@ func inspectModule(workingDir string, recipe *recipes.EnvironmentDefinition) (*m
...
-	// Extract the list of required providers.
-	for providerName := range mod.RequiredProviders {
-		result.RequiredProviders = append(result.RequiredProviders, providerName)
+	// Extract the details of required providers.
+	for k, v := range mod.RequiredProviders {
+		requiredprovider := &config.RequiredProviderInfo{}
+
+		if v.Source != "" {
+			requiredprovider.Source = v.Source
+		}
```
ytimocin, line 84: Here we can just do `requiredprovider.Source = v.Source` because Source is already an empty string based on the definition above in line 80: `requiredprovider := &config.RequiredProviderInfo{}`.

pkg/recipes/terraform/config/config.go
lakshmimsft: updated comments

pkg/recipes/terraform/config/config.go
lakshmimsft: added comment

pkg/recipes/terraform/config/testdata/providers-emptyazureconfig.tf.json
```diff
@@ -6,6 +6,12 @@
         "namespace": "radius-system",
         "secret_suffix": "test-secret-suffix"
       }
+    },
+    "required_providers": {
+      "azurerm": {
+        "source": "hashicorp/azurerm",
+        "version": "~> 2.0"
+      }
```
ytimocin, line 14: Should we have a test with multiple required providers?

pkg/recipes/terraform/config/types.go
```diff
@@ -61,4 +61,14 @@ type TerraformDefinition struct {
 	// Backend defines where Terraform stores its state.
 	// https://developer.hashicorp.com/terraform/language/state
 	Backend map[string]interface{} `json:"backend"`
+
+	// RequiredProviders is the list of required Terraform providers.
+	RequiredProviders map[string]*RequiredProviderInfo `json:"required_providers,omitempty"`
```
ytimocin, line 66: Is there a reason why `json:"required_providers,omitempty"` is not camel case?

pkg/recipes/terraform/config/types.go
```diff
@@ -61,4 +61,14 @@ type TerraformDefinition struct {
...
+}
+
+// RequiredProviderInfo represents details for a provider listed under the required_providers block in a Terraform module.
+type RequiredProviderInfo struct {
+	Source               string   `json:"source,omitempty"`                // The source of the provider.
+	Version              string   `json:"version,omitempty"`               // The version of the provider.
+	ConfigurationAliases []string `json:"configuration_aliases,omitempty"` // The configuration aliases for the provider.
```
ytimocin, line 73: Is there a reason why `json:"configuration_aliases,omitempty"` is not camel case?

ytimocin (dismissed):
I'd also get @kachawla's approval. Also asked a few more questions. LGTM for now.

pkg/recipes/terraform/config/testdata/providers-emptyazureconfig.tf.json
lakshmimsft: Two tests with multiple require_providers exist: 
testdata/providers-overrideucpproviderconfig.tf.json
testdata/providers-valid.tf.json

pkg/recipes/terraform/config/types.go
lakshmimsft: this is because in the tf.json file,  the serialized json output is expected to have the field name as 'required_providers'

pkg/recipes/terraform/config/types.go
lakshmimsft: same answer as above. In the tf.json file, the serialized json output is expected to have the field name as 'configuration_aliases'

pkg/recipes/terraform/config/config.go
```diff
@@ -96,24 +97,32 @@ func (cfg *TerraformConfig) Save(ctx context.Context, workingDir string) error {
...
-	if err != nil {
+	// Create a buffer to write the JSON to
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "  ") // Indent with 2 spaces to make the JSON file human-readable and consistent with codebase.
+
+	// Encode the Terraform config to JSON. JSON encoding is being used to ensure that special characters
+	// in the original text are preserved when writing to the file.
+	if err := enc.Encode(cfg); err != nil {
```
kachawla, line 108: Can you please elaborate the comment with specific example on what breaks if we don't attach the buffer to the encoder first? This will help anyone reading the code easily get the context on which fields require this extra step.

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
kachawla: @lakshmimsft got it. Can you please add a comment on the reasoning behind this?

> They will need to be updated for this if we were to remove this.

Why do the test files need to be updated?

pkg/recipes/terraform/config/config.go
kachawla: I understand required_providers need to be added to the config. My question is the code path where you are adding this. We shouldn't be overloading the backend config generation with providers, all the providers stuff should be processed in a single place.

pkg/recipes/terraform/config/config.go
```diff
@@ -127,6 +136,42 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+
+	for providerName, providerConfigList := range providerConfigs {
+		providerConfigDetails, ok := providerConfigList.([]map[string]any)
+		if !ok {
+			continue
+		}
+
+		// For each provider in the providerConfigs, if provider has a property "alias",
+		// add entry to the module provider configuration
+		for _, providerConfig := range providerConfigDetails {
```
kachawla, line 157: Instead of iterating over this again, can we process this in `AddProviders`?

kachawla (changes requested):

pkg/recipes/terraform/config/config.go
kachawla: @lakshmimsft I'm still unclear on why this code path needs to be updated, maybe I'm missing something. What error are you seeing if this code is kept as is without the updates being made in this PR?

kachawla: @lakshmimsft I'm still unclear on why this code path needs to be updated, maybe I'm missing something. What error are you seeing if this code is kept as is without the updates being made in this PR?

pkg/recipes/terraform/config/types.go
kachawla: @lakshmimsft this question can come up for others reading this code as well, adding link to the relevant doc in the comment will help here.

pkg/recipes/terraform/config/types.go
kachawla, line 73: Interesting, `configuration_aliases` property is needed at the root module level as well?

pkg/recipes/terraform/module.go
```diff
@@ -38,7 +40,7 @@ type moduleInspectResult struct {
 	ContextVarExists bool
 
 	// RequiredProviders is a list of names of required providers for the module.
```
kachawla, line 42: Comment doesn't match the updated data type

pkg/recipes/terraform/module.go
```diff
@@ -73,9 +75,29 @@ func inspectModule(workingDir string, recipe *recipes.EnvironmentDefinition) (*m
...
+		if len(v.VersionConstraints) > 0 {
+			requiredprovider.Version = strings.Join(v.VersionConstraints, ",")
+		}
+
+		if len(v.ConfigurationAliases) > 0 {
+			var aliases []string
+			for _, alias := range v.ConfigurationAliases {
+				// Concatenate Name and Alias from ProviderRef.
+				if alias.Name != "" && alias.Alias != "" {
+					aliases = append(aliases, alias.Name+"."+alias.Alias)
```
kachawla, line 92: Could you add a comment with link to the doc referencing why this is needed?

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
lakshmimsft: discussed and also updated comment.

pkg/recipes/terraform/config/config.go
lakshmimsft: Removed updates to this function as discussed. All work of adding provider configuration is now in AddProviders() function

pkg/recipes/terraform/config/config.go
lakshmimsft: reverted changes to this function as discussed.

pkg/recipes/terraform/config/config.go
```diff
@@ -127,6 +136,42 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+	}
+
+	providerConfigs := cfg.Provider
+	moduleAliasConfig := map[string]string{}
+
+	for providerName, providerConfigList := range providerConfigs {
+		providerConfigDetails, ok := providerConfigList.([]map[string]any)
+		if !ok {
+			continue
+		}
```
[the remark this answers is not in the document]
lakshmimsft: Ln 177. cfg.Provider is of type map[string]any
Ln 182. We're proceeding only if providerConfigList (type any) can be resolved to type []map[string]any. 
It's a check for unexpected data condition.

pkg/recipes/terraform/config/config.go
lakshmimsft: we do not iterate through the details of provider configuration in AddProviders()/getProviderConfigs() functions. Keeping this function in but made updates to call it from AddProviders so we have one function which adds all provider details to Terraform configuration.

pkg/recipes/terraform/config/config.go
```diff
@@ -127,6 +136,42 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+
+		// For each provider in the providerConfigs, if provider has a property "alias",
+		// add entry to the module provider configuration
+		for _, providerConfig := range providerConfigDetails {
+			if alias, ok := providerConfig["alias"]; ok {
+				moduleAliasConfig[providerName+"."+fmt.Sprintf("%v", alias)] = providerName + "." + fmt.Sprintf("%v", alias)
```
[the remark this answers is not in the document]
lakshmimsft: Updated comments Ln 186-Ln 193 for more details.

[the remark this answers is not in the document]
lakshmimsft: note: I'm updating this function to make sure I get the modules' config aliases are in provider config and passing in values for those values. Pls hold off on review.

pkg/recipes/terraform/config/config.go
lakshmimsft: updated comments with a detailed example. Pls check.

pkg/recipes/terraform/config/types.go
lakshmimsft: updated in comments

pkg/recipes/terraform/module.go
lakshmimsft: updated comment with link.

pkg/recipes/terraform/module.go
lakshmimsft: updated

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
lakshmimsft: Updated. Pls review. Thank you!

pkg/recipes/terraform/config/config.go
kachawla: Thanks

pkg/recipes/terraform/config/config.go
```diff
@@ -124,6 +153,72 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+	if cfg.Terraform == nil {
+		cfg.Terraform = &TerraformDefinition{}
+	}
+	cfg.Terraform.RequiredProviders = requiredProviders
+
+	return nil
+}
+
+// updateModuleWithProviderAliases updates the module provider configuration in the Terraform config
+// by adding aliases to the provider configurations.
```
kachawla, line 173: A link will help providing more context of example

```suggestion
// updateModuleWithProviderAliases updates the module provider configuration in the Terraform config
// by adding aliases to the provider configurations.
// https://developer.hashicorp.com/terraform/language/syntax/json#module-blocks
```

pkg/recipes/terraform/config/config.go
```diff
@@ -124,6 +153,72 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+func (cfg *TerraformConfig) updateModuleWithProviderAliases(requiredProviders map[string]*RequiredProviderInfo) error {
+	if cfg == nil {
+		return fmt.Errorf("terraform configuration is not initialized")
+	}
+	providerConfigs := cfg.Provider
```
kachawla, line 178: Do we need to initialize this additional variable? `providerConfigs` seems to be used only once on L181

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
kachawla: @lakshmimsft The reason I asked was because maybe there is an expected scenario which is why the implementation is ignoring it. If it is checking for unexpected scenario then shouldn't we be surfacing it through an internal error for instead of a no-op?

pkg/recipes/terraform/config/config.go
```diff
@@ -167,19 +262,25 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
+	// Required providers that are not configured with UCP will be skipped.
+	for providerName := range requiredProviders {
+		if _, ok := ucpConfiguredProviders[providerName]; ok { // Check if the provider is configured with UCP
+			if _, ok := providerConfigs[providerName]; ok {
+				// Environment level recipe configuration for providers will take precedence over
+				// UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
+				continue
+			}
+		} else {
+			// If the provider under required_providers is not configured with UCP, skip this iteration.
```
kachawla, line 279: I'm still confused why this code needs to be updated. The scenario you mentioned in https://github.com/radius-project/radius/pull/7364#discussion_r1544817535 should be handled by existing code. Do we have a unit test for that scenario that I can look at to understand? Also, curious to hear what error are you seeing if this code is kept in it current state?

pkg/recipes/terraform/config/config.go
```diff
@@ -167,19 +262,25 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
...
 
-	// Build provider configurations for required providers excluding the ones already present in providerConfigs
-	for _, provider := range requiredProviders {
-		if _, ok := providerConfigs[provider]; ok {
-			// Environment level recipe configuration for providers will take precedence over
-			// UCP provider configuration (currently these include azurerm, aws, kubernetes providers)
+	// Build provider configurations for required providers excluding the ones already present in providerConfigs (environment level configuration).
+	// Required providers that are not configured with UCP will be skipped.
+	for providerName := range requiredProviders {
+		if _, ok := ucpConfiguredProviders[providerName]; ok { // Check if the provider is configured with UCP
```
kachawla, line 272: L283 is checking this

pkg/recipes/terraform/config/config_test.go
```diff
@@ -555,22 +582,22 @@ func Test_AddProviders(t *testing.T) {
...
-			expectedConfigFile:             "testdata/providers-empty.tf.json",
+			expectedConfigFile:             "testdata/providers-emptyproviders.tf.json",
 		},
 		{
 			desc:                           "envConfig not populated",
 			expectedUCPConfiguredProviders: nil,
 			Err:                            nil,
 			requiredProviders:              nil,
-			expectedConfigFile:             "testdata/providers-empty.tf.json",
+			expectedConfigFile:             "testdata/providers-emptyproviders.tf.json",
```
kachawla, line 600: `providers` is already in the prefix

pkg/recipes/terraform/config/testdata/providers-emptyproviders.tf.json
```diff
@@ -8,19 +8,6 @@
       }
     }
   },
-  "provider": {
```
kachawla, line 11: Why is this not a valid test scenario anymore?

pkg/recipes/terraform/config/config_test.go
```diff
@@ -666,6 +695,213 @@ func Test_AddOutputs(t *testing.T) {
 	}
 }
 
+func Test_updateModuleWithProviderAliases(t *testing.T) {
```
kachawla, line 698: I think we are making the same assumption in test cases that alias names in configuration_alias field have to match the alias names in providers configs, which doesn't seem to be true from official docs.

pkg/recipes/terraform/config/config_test.go
```diff
@@ -581,8 +608,10 @@ func Test_AddProviders(t *testing.T) {
 
 			tfconfig, err := New(ctx, testRecipeName, &envRecipe, &resourceRecipe, &tc.envConfig)
 			require.NoError(t, err)
-			for _, p := range tc.expectedUCPConfiguredProviders {
-				mProvider.EXPECT().BuildConfig(ctx, &tc.envConfig).Times(1).Return(p, nil)
+			if !tc.overrideUCPProviderConfig {
```
kachawla, line 611: It's really hard to understand that setting `overrideUCPProviderConfig` translates into building configs for providers configured on UCP. :) Can you share some details on how this flag should be used, or maybe make it more explicit?

pkg/recipes/terraform/config/config.go
```diff
@@ -124,6 +153,72 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+			continue
+		}
+
+		// For each provider in the providerConfigs, if provider has a property "alias",
+		// add entry to the module provider configuration, per documentation:
+		// Provider configurations (those with the alias argument set) are never inherited automatically by modules,
+		// and so must always be passed explicitly using the providers map.
+		// https://developer.hashicorp.com/terraform/language/modules/develop/providers#legacy-shared-modules-with-provider-configurations
+
+		// Note: We're building configuration from user input, we're currently mapping the provider.alias names in
```
kachawla, line 193: > we're currently mapping the provider.alias names in

To make sure I understand - what do we mean by **currently** mapping? Is there a plan to follow up with another approach?

pkg/recipes/terraform/config/config.go
```diff
@@ -124,6 +153,72 @@ func (cfg *TerraformConfig) AddProviders(ctx context.Context, requiredProviders
...
+
+		for _, providerConfig := range providerConfigDetails {
+			if alias, ok := providerConfig["alias"]; ok {
+				aliasProviderConfig := providerName + "." + fmt.Sprintf("%v", alias)
+
+				// Check if the alias is in the required providers' configuration aliases. If there is a match, add the alias to the module provider configuration.
+				if requiredProviders[providerName] != nil && len(requiredProviders[providerName].ConfigurationAliases) > 0 {
+					for _, alias := range requiredProviders[providerName].ConfigurationAliases {
+						if alias == aliasProviderConfig {
+							moduleAliasConfig[alias] = alias
```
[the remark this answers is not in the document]
kachawla: Where was this requirement/assumption decided on? Design/product spec discussion, somewhere else? If not, then I'd suggest not adding assumptions at the implementation stage.

pkg/recipes/terraform/config/config.go
lakshmimsft: whoops! i missed addressing this earlier.. digging in here now.

pkg/recipes/terraform/config/config_test.go
lakshmimsft: This check is to ensure we're building configs for UCP provider only when we're not overriding it. I added a comment.

pkg/recipes/terraform/config/testdata/providers-emptyproviders.tf.json
lakshmimsft: this is a git diff issue i think. it's recognizing this as a rename from a unrelated file.
	image

pkg/recipes/terraform/config/config.go
lakshmimsft: reverting changes and running tests..

pkg/recipes/terraform/config/config_test.go
kachawla: Can we call it useUCPProviderConfig then to make it more intuitive?

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
lakshmimsft: still thinking on this..

pkg/recipes/terraform/config/config.go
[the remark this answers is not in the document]
lakshmimsft: updating this thread. Design discussion took place where this was discussed. Design doc updated [here](https://github.com/radius-project/design-notes/pull/39) (pls see 'Alias' section). We're going ahead with the 1:1 mapping at this time.

pkg/recipes/terraform/config/testdata/providers-emptyproviders.tf.json
kachawla: It's because [this](https://github.com/radius-project/radius/blob/main/pkg/recipes/terraform/config/testdata/providers-overridereqproviders.tf.json) file was deleted. Why are we deleting it?

pkg/recipes/terraform/config/config_test.go
kachawla: Since we are going ahead with only supporting the scenarios with 1-1 mapping, could you add a unit test for scenario where config_alias has a name that doesn't exist in the provider config?

kachawla (commented):
Can't seem to find the new functional test in `recipe_terraform_test.go` anymore, maybe it got reverted somehow?

pkg/recipes/terraform/config/config_test.go
lakshmimsft: This test exists: ln 775 : "Test with subset of required_provider aliases in provider config". I can add another one where it's not a subset.

kachawla: My suggestion was not for subset - it was for use case 2 discussed yeterday.

"scenario where config_alias has a name that doesn't exist in the provider config"

pkg/recipes/terraform/config/testdata/providers-emptyproviders.tf.json
lakshmimsft: this file is not deleted. It was renamed to providers-overrideucpproviderconfig.tf.json [link](https://github.com/radius-project/radius/blob/b947644e9f5890534b367355f741875eda36c7f5/pkg/recipes/terraform/config/testdata/providers-overrideucpproviderconfig.tf.json)

pkg/recipes/terraform/config/config.go
```diff
@@ -189,8 +284,9 @@ func getProviderConfigs(ctx context.Context, requiredProviders []string, ucpConf
 		if err != nil {
 			return nil, err
 		}
+
 		if len(config) > 0 {
-			providerConfigs[provider] = config
+			providerConfigs[provider] = []map[string]any{config}
```
kachawla, line 289: What is causing this to change? Return type is `map[string]any`

kachawla (commented):
One question about recent update in config.go: https://github.com/radius-project/radius/pull/7364#discussion_r1607404612. Looks good apart from that.

pkg/recipes/terraform/config/config.go
lakshmimsft: We're making sure the structure of provider config is the same when UCP provider config is built or we get it from env-recipe config. the UCP provider config was earlier generating it in map[string]any format

kachawla: We chatted offline and decided to the config provider map value strongly typed to avoid assumptions and better readability/maintainability of this code path.

lakshmimsft: updated in latest commit + rebased. Pls take a look.

pkg/recipes/terraform/config/config.go
```diff
@@ -261,7 +256,10 @@ func newModuleConfig(moduleSource string, moduleVersion string, params ...Recipe
 
 // getProviderConfigs generates the Terraform provider configurations. This is built from a combination of environment level recipe configuration for
 // providers and the provider configurations registered with UCP. The environment level recipe configuration for providers takes precedence over UCP provider configurations.
-func getProviderConfigs(ctx context.Context, requiredProviders map[string]*RequiredProviderInfo, ucpConfiguredProviders map[string]providers.Provider, envConfig *recipes.Configuration) (map[string]any, error) {
+// The function returns a map where the keys are provider names and the values are slices of maps.
+// Each map in the slice represents a specific configuration for the corresponding provider.
+// This structure allows for multiple configurations per provider.
```
kachawla, line 261: This is a helpful comment, thanks for updating.

kachawla (dismissed):

kachawla (approved):

PR #7370, opened 2024-03-22
Update logging for failed async operations

kachawla:
# Description

Controllers currently have two ways of bubbling up errors to the async worker:
1. By returning the error, which is then added to the result by the async worker.
2. By directly returning a non-empty result with error message populated and provisioning state set to "failed".

We have a few gaps in the logging today which results into false data or no data, this PR aims to fix these gaps -
* Error is not logged if it is returned as a part of the result (scenario 2 above).
* Operation is logged as successful if a direct error is returned (scenario 1 above).
* Provisioning state is set to succeeded even if the controller returns an error (scenario 1 above).

## Type of change

Part of https://github.com/radius-project/radius/issues/7369


youngbupark:
> * Error is not logged if it is returned as a part of the result (scenario 2 above).

This is not true. err is always logged. L270 has been logged the error unless it is cancelled one. So operation returned might log wrong info, but it still needs to log error correctly. 

https://github.com/radius-project/radius/blob/22d69efce8d61eb03e723edeeabc56d95c3d6c94/pkg/armrpc/asyncoperation/worker/worker.go#L262-L274

I would combine with L267-271 and then refactor it like below.

```golang
                 result, err := asyncCtrl.Run(asyncReqCtx, asyncReq)
                 if err != nil {
                                 // Just in case, controller returns result and error both. (it should not happen)
				logger.Error(err, "returned controller error")
				armErr := extractError(err)
                                 if result.Error == nil {
				      result.SetFailed(armErr, false)
                                 }
		}

		logger.Info("Operation result", "success", result.Error == nil, "code", result.Error.Code, "provisioningState", result.ProvisioningState(), "err", err)

		// There are two cases when asyncReqCtx is canceled.
		// 1. When the operation is timed out, w.completeOperation will be called in L186
		// 2. When parent context is canceled or done, we need to requeue the operation to reprocess the request.
		// Such cases should not call w.completeOperation.
		if !errors.Is(asyncReqCtx.Err(), context.Canceled) {
			w.completeOperation(ctx, message, result, asyncCtrl.StorageClient())
		}
```

kachawla:
> * Error is not logged if it is returned as a part of the result (scenario 2 above).

> This is not true. err is always logged. L270 has been logged the error unless it is cancelled one

I debugged it locally and verified before writing what I wrote. Have you tried this scenario?

youngbupark:
> > * Error is not logged if it is returned as a part of the result (scenario 2 above).
> 
> > This is not true. err is always logged. L270 has been logged the error unless it is cancelled one
> 
> I debugged it locally and verified before writing what I wrote. Have you tried this scenario?

If L270 doesn't log the error, then we need to fix L270 block as well. let me debug this scenario.

kachawla:
> I would combine with L267-271 and then refactor it like below.

I spent quite a bit of time on this code before sending this PR and intentionally didn't combine the two because I believe we don't want to update the provisioning state if the operation was cancelled, which is why the error handling is currently inside the if block on L266. I'm guessing you would have the most context on the existing code, so you can confirm what the rationale was behind this. It would have been easier if more comments were added on this code path when it was written.

ytimocin (dismissed):

kachawla:
> > > > * Error is not logged if it is returned as a part of the result (scenario 2 above).
> > > 
> > > 
> > > > This is not true. err is always logged. L270 has been logged the error unless it is cancelled one
> > > 
> > > 
> > > I debugged it locally and verified before writing what I wrote. Have you tried this scenario?
> > 
> > 
> > If L270 doesn't log the error, then we need to fix L270 block as well. let me debug this scenario.
> 
> Apologies, I was thinking that `Error` is `err` returned from controller. If `Error` is `result.Error`, then yes, it is right that we don't log its context correctly. looks like Ryan added this part in his tracked resource work.
> 
> > > I would combine with L267-271 and then refactor it like below.
> > 
> > 
> > I spent quite a bit of time on this code before sending this PR and intentionally didn't combine the two because I believe we don't want to update the provisioning state if the operation was cancelled, which is why the error handling is currently inside the if block on L266. I'm guessing you would have the most context on the existing code, so you can confirm what the rationale was behind this. It would have been easier if more comments were added on this code path when it was written.
> 
> The current update is to extract the error from `result.Error` and `err`. Instead of doing this, converting `err` to `result.Error` is simpler earlier.
> 
> As you can see, the context is marked as cancelled for the following scenario. I recall I added the test code for the following scenario as well.
> 
> https://github.com/radius-project/radius/blob/22d69efce8d61eb03e723edeeabc56d95c3d6c94/pkg/armrpc/asyncoperation/worker/worker.go#L262-L265
> 
> In most of the cases, it happens by opCancel() when the timeout happens. As you can see the following code, we update the resource metadata after opCancel(). This happens before L267-271 in the go routine is processed. The main reason we check the cancellation is not to overwrite provisioningState by the following code. So if we do not call `w.completeOperation(ctx, message, result, asyncCtrl.StorageClient())`, it is okay to convert `err` to `result.Error`. I think this way allows us to simplify this logic. We can also move L260 logging after L274. Then L255-L258 are not required.
> 
> https://github.com/radius-project/radius/blob/22d69efce8d61eb03e723edeeabc56d95c3d6c94/pkg/armrpc/asyncoperation/worker/worker.go#L295-L300

Updated. Moved error logging after error handling so that it is consistent for all error scenarios. https://github.com/radius-project/radius/blob/8754c50c827e912f38bf824d6ae3de709b1ecef3/pkg/armrpc/asyncoperation/worker/worker.go#L262-L265

pkg/armrpc/asyncoperation/worker/worker.go
```diff
@@ -251,27 +251,28 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 
 		logger.Info("Start processing operation.")
 		result, err := asyncCtrl.Run(asyncReqCtx, asyncReq)
+		// Update the result if an error is returned from the controller.
+		// Check that the result is empty to ensure we don't override it, it shouldn't happen.
+		// Controller should always either return non-empty error or non-empty result, but not both.
```
rynowak, line 256: I think I have the same comment I did on the previous PR. I'm reflecting back on previous discussions we've had so please tell me if my memory is faulty.

This isn't a comment on the code, but rather the semantics of error handling, and whether our design decisions have aged well. 

We distinguish between the following case:

- Async action completed successfully.
- Async action failed with a known (expected) error. This isn't a bug in Radius.
- Async action failed with an unknown error. This is possibly a bug in Radius.

This matches the error handling in the front-end. 

- Async action completed successfully. -> `HTTP 2XX`
- Async action failed with a known (expected) error. -> `HTTP 4XX`
- Async action failed with an unknown error. -> `HTTP 5XX`

Relevant to this PR - our principle for logging errors is something like:

> error in logs == Radius is misbehaving. 

This is why a _failed_ async result is logged as info. It's an error that's already been handled and classified. 

---

Re: the code This 3-valued logic. 

- Async action completed successfully. -> Empty result + `nil` err
- Async action failed with a known (expected) error. -> Failed result + `nil` err
- Async action failed with an unknown error. -> Empty result + non-`nil` err

If we **wanted** 2-valued logic, we could consolidate two of those cases and simplify all of the code.

rynowak: I'm not trying to belabor the point, if we can let's spend 10 mins on this in the design meeting. Ideally with @youngbupark present.

kachawla: Let's discuss this in the design meeting, but adding my thoughts here to drive the discussion -

1. We don't log handled errors (`Failed result + nil error` scenario) on the server side at all today, this needs to be fixed. The client side errors are out of our control, but it is important to have visibility into them for several reasons - easier debugging, visibility into patterns in client side errors to get insights into areas users are constantly hitting, detecting an outlier, etc.

2. We aren't always handling errors perfectly - [here](https://github.com/radius-project/radius/blob/main/pkg/corerp/backend/controller/createorupdateresource.go#L81) is an example of where this is a known error but controllers returned instead of handling it. So current implementation is brittle. Here is an issue that I had created to track revisiting the implementation: https://github.com/radius-project/radius/issues/7369.

3. "This is why a failed async result is logged as info. It's an error that's already been handled and classified." -- I disagree with this approach to logging. Any error regardless of its type is logged as error. Filtering them separately is done based on the code.

4. [This additional check](https://github.com/radius-project/radius/blob/8754c50c827e912f38bf824d6ae3de709b1ecef3/pkg/armrpc/asyncoperation/worker/worker.go#L255) was added as extra safety net based on the suggestion of @youngbupark [here](https://github.com/radius-project/radius/pull/7370#issuecomment-2018404593), we know this is not expected and the comment calls it out, I do not have strong opinions on keeping it or removing it.


This PR is intended fix #1 above in the most straightforward way possible because it is a pressing issue, has caused enough frustration during debugging that it needs to be fixed asap. The overall design should be revisited improve this code path - I had created an issue here https://github.com/radius-project/radius/issues/7369.

kachawla: Discussed and decided on moving forward with -
* All failures including client side will be logged as errors.
* Skip info log if error is logged since it is redundant. -- updated [here](https://github.com/radius-project/radius/blob/7669a05bba8714f2c4aa3d5f66328b9713cd7203/pkg/armrpc/asyncoperation/worker/worker.go#L273).

pkg/armrpc/asyncoperation/worker/worker.go
```diff
@@ -251,27 +251,29 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
...
 			w.completeOperation(ctx, message, result, asyncCtrl.StorageClient())
 		}
+
+		// Log the error if the operation failed.
+		if result.Error != nil {
+			logger.Error(result.Error, "Operation Failed")
+		} else {
+			logger.Info("Operation returned", "success", result.Error == nil, "provisioningState", result.ProvisioningState())
+		}
+
```
youngbupark, line 276: I checked how logger.Error and logger.Info represent as a log finally.

Here is the example log `logger.Error` emits:

`{"severity":"error","timestamp":"2024-03-17T00:59:28.451Z","name":"radius.radiusasyncworker","caller":"worker/worker.go:270","message":"Operation Failed","serviceName":"radius","version":"edge","hostName":"applications-rp-59dfdc9c54-fjt56","resourceId":"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/containers/invalid-ctnr","operationId":"0cc83dfc-5dff-4441-be3d-d22ed0f727d1","operationType":"APPLICATIONS.CORE/CONTAINERS|PUT","dequeueCount":1,"traceId":"b82c1943e6b0b5aa4a64e20119a3144b","spanId":"9cab3ac40dc3ea1a","error":"code BadRequest: err linked \"/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/default/providers/applications.core/environments/env\" has invalid Applications.Core/applications resource type.","stacktrace":"github.com/radius-project/radius/pkg/armrpc/asyncoperation/worker.(*AsyncRequestProcessWorker).runOperation.func1\n\t/home/runner/work/radius/radius/pkg/armrpc/asyncoperation/worker/worker.go:270"}`

Here is the example log `logger.Info` emits:
`{"severity":"info","timestamp":"2024-03-17T00:59:28.625Z","name":"radius.radiusasyncworker","caller":"worker/worker.go:260","message":"Operation returned","serviceName":"radius","version":"edge","hostName":"applications-rp-59dfdc9c54-fjt56","resourceId":"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/gateways/invalid-gtwy","operationId":"1ff4da3e-7043-4869-a862-311f10ae7ecd","operationType":"APPLICATIONS.CORE/GATEWAYS|PUT","dequeueCount":1,"traceId":"b82c1943e6b0b5aa4a64e20119a3144b","spanId":"535fe18ef22aff07","success":true,"code":"","provisioningState":"Succeeded","err":"code BadRequest: err application ID \"not_an_id\" for the resource \"/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/gateways/invalid-gtwy\" is not a valid id. Error: 'not_an_id' is not a valid resource id"}`

One of the benefits to use logger.Error is adding `stacktrace` attribute. If we want to create query for displaying chart, the query might be complicated. 
It seems like using only `logger.Info` might be easier for query the logs because we can filter only `Operation returned` message with success attribute and error attribute (for the failed operation)
I initially said I did not have the strong opinion, but when I look at the actual log output, using only logger.Info might be easier to make the query for chart.

youngbupark (approved):

pkg/armrpc/asyncoperation/worker/worker.go
kachawla: Discussed offline and updated everything to be logged as info.

PR #7392, opened 2024-03-25
Randomize the id with GITHUB_RUN_NUMBER for scheduled test 

### LINKED ISSUE #7388: Randomize test resource name for scheduled functional test
### Steps to reproduce

Failed tests:
https://github.com/radius-project/radius/issues/7383
#7382
#7380

Multiple long-running-test are failed because they use the same name of resource group. The current unique id generation logic is not enough to randomize the id when multiple test workflow are running concurrently. 

`radtest-aa342ef1df-e2e-all`

### Observed behavior

_No response_

### Desired behavior

Resource group name should not conflict when multiple tests are running.

### Workaround

_No response_

### rad Version

edge

### Operating system

_No response_

### Additional context

_No response_

[AB#11623](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11623)

youngbupark:
# Description

We generate unique id by concatenating GitHub Actions' system variables, hashing the string, and taking 10 characters. This way will reduce the chance to create multiple resource for the same PR. In scheduled test, the most of variables are same, so there is a little chance to face the conflicts. particularly, when resource group is being in progress triggered by the previous action runs like below failures. This change 1) adds GITHUB_RUN_NUMBER, which is updated for each action run, to the raw id before hashing and 2) adds the prefix to each hashed string to avoid the conflicts between different workflows.

https://github.com/radius-project/radius/issues/7383
https://github.com/radius-project/radius/issues/7382
https://github.com/radius-project/radius/issues/7380

## Type of change

Fixes: #7388

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -164,7 +164,7 @@ jobs:
       - name: Generate ID for release
         id: gen-id
         run: |
-          BASE_STR="RAD|${GITHUB_SHA}|${GITHUB_SERVER_URL}|${GITHUB_REPOSITORY}|${GITHUB_RUN_ID}|${GITHUB_RUN_ATTEMPT}"
+          BASE_STR="RADIUS|${GITHUB_SHA}|${GITHUB_SERVER_URL}|${GITHUB_REPOSITORY}|${GITHUB_RUN_ID}|${GITHUB_RUN_ATTEMPT}"
           if [ "$GITHUB_EVENT_NAME" == "schedule" ]; then
             # Add run number to randomize unique id for scheduled runs.
             BASE_STR="${GITHUB_RUN_NUMBER}|${BASE_STR}"
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -164,15 +164,12 @@ jobs:
       - name: Generate ID for release
         id: gen-id
         run: |
-          if [ "$GITHUB_EVENT_NAME" == "schedule"]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -165,7 +165,11 @@ jobs:
         id: gen-id
         run: |
           BASE_STR="RADIUS|${GITHUB_SHA}|${GITHUB_SERVER_URL}|${GITHUB_REPOSITORY}|${GITHUB_RUN_ID}|${GITHUB_RUN_ATTEMPT}"
-          UNIQUE_ID=$(echo $BASE_STR | sha1sum | head -c 10)
+          if [ "$GITHUB_EVENT_NAME" == "schedule" ]; then
+            # Add run number to randomize unique id for scheduled runs.
+            BASE_STR="${GITHUB_RUN_NUMBER}|${BASE_STR}"
```
willdavsmith, line 170: is there any reason we only do this for scheduled runs? is there a downside to adding the run number on all runs?

ytimocin (approved):

willdavsmith (approved):

.github/workflows/functional-test.yaml
youngbupark: If we use GITHUB_RUN_NUMBER for PR event, workflow will create resource groups and resources whenever we trigger functional tests. This results in unnecessary resource creation in our subscription. To avoid this issue and save the cost, we do not use GITHUB_RUN_NUMBER for every event.

youngbupark: Even though we have purge workflow, it cleans up only 3 days+ old resources. So we still need to pay 3 days resoruce usage. so it is better not to use `GITHUB_RUN_NUMBER` for every event.

PR #7415, opened 2024-03-29
Upgrade action versions of workflow to resolve openSSF issues

### LINKED ISSUE #7404: Failing OpenSSF Workflow
OpenSSF Workflow has started failing since last week https://github.com/radius-project/radius/actions/workflows/scorecard.yaml. Needs further investigation on if this is caused by a change we made or something else.

[AB#11657](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11657)

youngbupark:
# Description

* Upgrade action versions of workflow to the latest
* Remove Delete artifact because we do not have storage limit anymore

## Type of change

Fixes: #7404

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -317,10 +317,11 @@ jobs:
           echo ${{ secrets.GITHUB_TOKEN }} | helm registry login -u ${{ github.actor }} --password-stdin ${{ env.OCI_REGISTRY }}
           helm push ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}/radius-${{ env.CHART_VERSION }}.tgz oci://${{ env.OCI_REGISTRY }}/${{ env.OCI_REPOSITORY }}
 
-  publish-release-test:
+  publish-release:
     name: Publish GitHub Release
     needs: ['build-and-push-cli']
     runs-on: ubuntu-latest
+    if: github.repository == 'radius-project/radius' && startsWith(github.ref, 'refs/tags/v')
     env:
       GITHUB_TOKEN: ${{ secrets.GH_RAD_CI_BOT_PAT }}
     steps:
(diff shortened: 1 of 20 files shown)


.github/workflows/build.yaml
```diff
@@ -123,18 +123,11 @@ jobs:
         run: |
           mkdir ${{ env.RELEASE_PATH }}
           cp ./dist/${{ matrix.target_os}}_${{ matrix.target_arch}}/release/rad.exe ${{ env.RELEASE_PATH }}/rad_${{ matrix.target_os}}_${{ matrix.target_arch}}.exe
-      - name: Upload Release binaries
```
kachawla, line 126: Is this being done elsewhere now?

.github/workflows/build.yaml
```diff
@@ -123,18 +123,11 @@ jobs:
...
-          name: rad_cli_release
-          path: ${{ env.RELEASE_PATH }}
       - name: Upload CLI binary
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: rad_cli_${{ matrix.target_os}}_${{ matrix.target_arch}}
-          path: |
-            ./dist/${{ matrix.target_os}}_${{ matrix.target_arch}}/release/rad
-            ./dist/${{ matrix.target_os}}_${{ matrix.target_arch}}/release/rad.exe
```
kachawla, line 137: Is there any way to validate these changes?

.github/workflows/build.yaml
```diff
@@ -333,16 +326,19 @@ jobs:
...
+        uses: actions/checkout@v4
       - name: Parse release version and set environment variables
         run: python ./.github/scripts/get_release_version.py
       - name: Download release artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
-          name: rad_cli_release
+          pattern: rad_cli_*
+          merge-multiple: true
```
kachawla, line 336: What's causing this to be needed?

.github/workflows/build.yaml
```diff
@@ -360,21 +356,3 @@ jobs:
             --title "Radius v${{ env.REL_VERSION }}" \
             --notes-file docs/release-notes/v${{ env.REL_VERSION }}.md \
             --verify-tag
-
-  delete_artifacts:
```
kachawla, line 364: For `we do not have storage limit anymore` - I might have missed this, could you share more details on this?

.github/workflows/build.yaml
youngbupark: We uploaded cli binary twice. So i deleted one and also upload action v4 does not support append operation anymore. in v4, we need to upload separately and then download action needs to download all artifact with the pattern.

.github/workflows/build.yaml
youngbupark: I explained before. This is because of v4 action change.

.github/workflows/build.yaml
youngbupark: Yes I validated it.

.github/workflows/build.yaml
youngbupark: This was because of private repo limitation. For public repo, storage is unlimited space. So we don't need to delete it anymore.

.github/workflows/build.yaml
youngbupark: We uploaded same binary twice with different name. This change will upload it once.

.github/workflows/build.yaml
youngbupark: I validated this by changing the workflow to list files. - https://github.com/radius-project/radius/actions/runs/8486662436/job/23253525031

image

kachawla (approved):

PR #7436, opened 2024-04-03
Adding functional test for private terraform repo support and fixing delete issue with private repo

vishwahiremat:
# Description

- Added functional test for terraform recipe using module stored in private git repository.
- Adding support for delete functionality for private repos.
- Adding required unit tests

## Type of change

Fixes: #issue_number


pkg/recipes/driver/terraform.go
```diff
@@ -135,13 +135,28 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
...
+	}
 	err = d.terraformExecutor.Delete(ctx, terraform.Options{
 		RootDir:        requestDirPath,
 		EnvConfig:      &opts.Configuration,
 		ResourceRecipe: &opts.Recipe,
 		EnvRecipe:      &opts.Definition,
 	})
+
+	// Unset credential information from .gitconfig if module source is of type git.
+	if strings.HasPrefix(opts.Definition.TemplatePath, "git::") && !reflect.DeepEqual(opts.BaseOptions.Secrets, v20231001preview.SecretStoresClientListSecretsResponse{}) {
```
kachawla, line 153: Was this something caught during functional testing? Could you please update the PR description to include things that are being changed here in addition to adding functional test?

pkg/recipes/driver/terraform.go
```diff
@@ -135,13 +135,28 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
 			logger.Info(fmt.Sprintf("Failed to cleanup Terraform execution directory %q. Err: %s", requestDirPath, err.Error()))
 		}
 	}()
-
+	// Add credential information to .gitconfig if module source is of type git.
+	if strings.HasPrefix(opts.Definition.TemplatePath, "git::") && !reflect.DeepEqual(opts.BaseOptions.Secrets, v20231001preview.SecretStoresClientListSecretsResponse{}) {
```
ytimocin, line 139: Should this be a function? Like `IsOfTypeGit` or something like that?

pkg/recipes/driver/terraform.go
ytimocin: I see this is being used a few times. We can make this a function itself.

pkg/recipes/driver/terraform.go
```diff
@@ -135,13 +135,28 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
...
 	err = d.terraformExecutor.Delete(ctx, terraform.Options{
 		RootDir:        requestDirPath,
 		EnvConfig:      &opts.Configuration,
 		ResourceRecipe: &opts.Recipe,
 		EnvRecipe:      &opts.Definition,
 	})
+
+	// Unset credential information from .gitconfig if module source is of type git.
+	if strings.HasPrefix(opts.Definition.TemplatePath, "git::") && !reflect.DeepEqual(opts.BaseOptions.Secrets, v20231001preview.SecretStoresClientListSecretsResponse{}) {
+		unsetError := unsetSecretsFromGitConfig(opts.BaseOptions.Secrets, opts.Definition.TemplatePath)
```
ytimocin, line 154: Is it like first we have to set so that it can be processed and then we have to remove because we don't want user to see the secrets?

ytimocin: I think this can also be written as one if:
```suggestion
		if err := unsetSecretsFromGitConfig(opts.BaseOptions.Secrets, opts.Definition.TemplatePath); err != nil  {
		return err
		}
```

pkg/recipes/engine/engine.go
```diff
@@ -219,3 +212,23 @@ func (e *engine) getDriver(ctx context.Context, recipeMetadata recipes.ResourceM
 	}
 	return definition, driver, nil
 }
+
+func (e *engine) getRecipeConfigSecrets(ctx context.Context, driver recipedriver.Driver, configuration *recipes.Configuration, definition *recipes.EnvironmentDefinition) (v20231001preview.SecretStoresClientListSecretsResponse, error) {
+	secrets := v20231001preview.SecretStoresClientListSecretsResponse{}
+	driverWithSecrets, ok := driver.(recipedriver.DriverWithSecrets)
+	if ok {
```
ytimocin, line 219: Why not check for `!ok` and do the rest after that `!ok` if?

pkg/recipes/driver/terraform.go
vishwahiremat: yes! this was caught during functional testing during clean up phase. Will update the PR description

vishwahiremat: @ytimocin this refactoring will done as part of supporting `rad show` command

pkg/recipes/engine/engine.go
vishwahiremat: we can do that as well. but is there is a reason why checking `!ok` better?

pkg/recipes/driver/terraform.go
kachawla, line 139: Can you please abstract this out in a separate function? That will make it easier to support for more private sources in the future.

Also, let's please update the comment to include context on why this is needed - to access private module sources.

pkg/recipes/driver/terraform.go
```diff
@@ -135,13 +135,28 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
 			logger.Info(fmt.Sprintf("Failed to cleanup Terraform execution directory %q. Err: %s", requestDirPath, err.Error()))
 		}
 	}()
-
+	// Add credential information to .gitconfig if module source is of type git.
+	if strings.HasPrefix(opts.Definition.TemplatePath, "git::") && !reflect.DeepEqual(opts.BaseOptions.Secrets, v20231001preview.SecretStoresClientListSecretsResponse{}) {
+		err := addSecretsToGitConfig(opts.BaseOptions.Secrets, &opts.Recipe, opts.Definition.TemplatePath)
+		if err != nil {
+			return err
```
kachawla, line 142: Should this be wrapped into recipe error? Do we need a new error code for interactions with git config or a generic error for secrets configuration?

kachawla, line 142: We might have discussed this already.. Could you remind me if there's an issue with mocking a dependency to add a unit test here? If not, and we're mainly skipping them since it's covered in gitconfig tests, then let's add them at least for the error scenario, so that a future change doesn't accidentally ignore the error.

pkg/recipes/driver/terraform.go
kachawla: > yes! this was caught during functional testing during clean up phase. Will update the PR description

I see. What issue was it causing? Complete failure to delete or issues with execution of next recipe request?

pkg/recipes/driver/terraform.go
kachawla, line 154: @vishwahiremat would this be needed after the latest changes discussed for nested modules? Wondering if deletion of execution directory would take care of this with the updated approach.

pkg/recipes/engine/engine.go
```diff
@@ -219,3 +212,23 @@ func (e *engine) getDriver(ctx context.Context, recipeMetadata recipes.ResourceM
...
+		secretStore, err := driverWithSecrets.FindSecretIDs(ctx, *configuration, *definition)
+		if err != nil {
```
kachawla, line 221: can we add unit tests for this?

pkg/recipes/engine/engine.go
```diff
@@ -219,3 +212,23 @@ func (e *engine) getDriver(ctx context.Context, recipeMetadata recipes.ResourceM
...
+			return v20231001preview.SecretStoresClientListSecretsResponse{}, err
+		}
+
+		// Retrieves the secret values from the secret store ID provided.
+		if secretStore != "" {
+			secrets, err = e.options.SecretsLoader.LoadSecrets(ctx, secretStore)
+			if err != nil {
```
kachawla, line 228: can we add unit tests for this?

pkg/recipes/driver/terraform.go
vishwahiremat: this will not be needed, but that chenges will go in part of the other PR which add support for nested modules that I plan to go on top of this PR.

pkg/recipes/driver/terraform.go
vishwahiremat: Yes I will add the comment, but can we skip the part of making a separate function ? because as part of supporting nested modules i will get rid of these setting and unsetting part of the code here.

pkg/recipes/engine/engine.go
vishwahiremat: done

pkg/recipes/engine/engine.go
vishwahiremat: Done

pkg/recipes/engine/engine.go
ytimocin: 1. Error handling upfront
2. Improve readability and maintainability
3. Avoid deeper nesting

pkg/recipes/driver/terraform.go
vishwahiremat: complete failure of delete! as secrets were not added to delete path

pkg/recipes/driver/terraform.go
vishwahiremat: did not update this as we are gitting rid of this in https://github.com/radius-project/radius/pull/7519/files

pkg/recipes/engine/engine.go
vishwahiremat: updated it

pkg/recipes/driver/terraform.go
vishwahiremat: as discussed offline, we can ignore this as it will be taken care in suporting nested modules PR

pkg/recipes/driver/terraform.go
vishwahiremat: we are making it a function in a different PR which will go on top of this one, which add supports for nested modules. because some changes are going to be done to the add/unset git config functions.

ytimocin: Could you also please post the link to the PR here?

test/testutil/testutil.go
```diff
@@ -125,6 +125,19 @@ func GetTerraformRecipeModuleServerURL() string {
...
 
+func GetTerraformPrivateModuleSource() string {
+	u := os.Getenv("TF_RECIPE_PRIVATE_GIT_SOURCE")
+	if u == "" {
+		return "privateGitModule=git::https://github.com/radius-project/terraform-private-modules//kubernetes-redis"
+	}
+	return "privateGitModule=" + u
+}
+
+func GetGitPAT() string {
```
ytimocin, line 136: We should always write docs for the functions added.

ytimocin (dismissed):

pkg/recipes/engine/engine_test.go
```diff
@@ -824,3 +973,85 @@ func Test_Engine_Execute_With_Secrets_Success(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, result, recipeResult)
 }
+
+func Test_Engine_Delete_With_Secrets_Success(t *testing.T) {
```
ytimocin, line 977: nit: is this the naming convention we follow?

pkg/recipes/engine/engine_test.go
```diff
@@ -824,3 +973,85 @@ func Test_Engine_Execute_With_Secrets_Success(t *testing.T) {
...
+	configLoader.EXPECT().
+		LoadConfiguration(ctx, recipeMetadata).
+		Times(1).
+		Return(envConfig, nil)
+	driverWithSecrets.EXPECT().
+		FindSecretIDs(ctx, *envConfig, *recipeDefinition).
+		Times(1).
+		Return("/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/azdevopsgit", nil)
+	secretsLoader.EXPECT().
+		LoadSecrets(ctx, gomock.Any()).
```
ytimocin, line 1035: Should we be specific on how many times this will be called? We may be missing some edge cases here if we don't specify number of times this function will be called.

pkg/recipes/engine/engine_test.go
vishwahiremat: the tests in this file are of similar format

ytimocin (approved):

PR #7445, opened 2024-04-04
Remove double imports and some dead code

ytimocin:
# Description
1. Removing all the double imports
2. Removing some dead code

## Type of change

### DIFF
--- a/test/radcli/cli.go
+++ b/test/radcli/cli.go
@@ -128,44 +128,6 @@ func (cli *CLI) Deploy(ctx context.Context, templateFilePath string, environment
 	return cliErr
 }
 
-// Run runs the rad run command. This returns the command output on success or an error.
-func (cli *CLI) Run(ctx context.Context, templateFilePath string, applicationName string, parameters ...string) (string, error) {
-	// Check if the template file path exists
-	if _, err := os.Stat(templateFilePath); err != nil {
-		return "", fmt.Errorf("could not find template file: %s - %w", templateFilePath, err)
-	}
-
-	args := []string{
-		"run",
-		templateFilePath,
-	}
-
-	if applicationName != "" {
-		args = append(args, "--application", applicationName)
-	}
-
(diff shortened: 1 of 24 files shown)


test/radcli/cli.go
```diff
@@ -179,18 +141,6 @@ func (cli *CLI) ApplicationShow(ctx context.Context, applicationName string) (st
 	return cli.RunCommand(ctx, args)
 }
 
-// ApplicationList runs a command to list applications and returns the output as a string. It returns an
-// error if the command fails.
-func (cli *CLI) ApplicationList(ctx context.Context) (string, error) {
```
kachawla, line 184: Are these covered elsewhere in the tests?

ytimocin: These functions are not being referenced anywhere in the code. I am not sure if it is wise to keep these just in case if we need them one day. I believe that if we need, we can add them in future. As of now, it is showing up as `deadcode` when I run `staticcheck`. That is why I believe it is good to remove them.

kachawla (approved):

test/radcli/cli.go
rynowak: > I was curious why don't we just use these functions to add test coverage unless they are being tested through other code paths, but doesn't look like we are testing these commands. I don't think it is in scope of this PR to add tests, but I have created this issue #7453, we can reference this PR if these functions need to be brought back.
> 
> Thanks for doing this cleanup!

Our coverage results don't include functional tests. To do that we'd have to collect coverage from inside the containers where Radius is running.

kachawla: I'm saying test coverage, not coverage results, I know coverage results are for unit tests. What I'm trying to understand is that why don't we have functional tests for these commands. Looks like a bunch of helper functions were added to be able to write functional tests but were never used.

rynowak: Ah sorry. No idea why then :)

PR #7455, opened 2024-04-05
Switch Radius Helm chart pull from ACR to GHCR

### LINKED ISSUE #6801: Migrate chart location from ACR to GHCR
## Summary

When helm retrieves the chart from GHCR, it always uses local docker credential to authenticate ghcr.io even for public repos. So if the credential is expired or invalid, it will return 403 Forbidden error. To mitigate this issue, we can override the authorization header to inject anonmous token (here is the PR #6799), but this way is just workaround solution. So it is better to show users the proper message to fix the credential problem. 

 This is to migrate chart location from ACR to GHCR:
- [ ] Revert this commit to reuse the old code - https://github.com/radius-project/radius/commit/6a4c73049d254abd64e39efe845465d088cc99bb
- [ ] When pull.Run returns 401 or 403, show the message to update ghcr.io credential or logout the ghcr.io

[AB#10416](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10416)

willdavsmith:
# Description

* Switch Radius Helm chart pull from ACR to GHCR
* Add error handling for 403 from GHCR

Example outputs:

Expired credentials:
```sh
❯ go run ./cmd/rad/main.go install kubernetes --reinstall
# command-line-arguments
ld: warning: -bind_at_load is deprecated on macOS
Reinstalling Radius version edge to namespace: radius-system...
Error: failed to load Helm chart, err: recieved 403 unauthorized when downloading helm chart from the registry. you may want to perform a `docker logout ghcr.io` and re-try the command, Helm output: 

TraceId:  dff0bdbe3da3f80fc6e8a84a9b5d38ea

exit status 1
```

## Type of change

Fixes: #6801


pkg/cli/helm/helm.go
```diff
@@ -101,8 +103,48 @@ func helmChartFromContainerRegistry(version string, config *helm.Configuration,
...
+
+	_, err = pull.Run(chartRef)
 	if err != nil {
+		// Error handling for a specific case where credentials are stale.
+		// This happens for ghcr in particular because ghcr does not use
+		// subdomains - the scope of a login is all of ghcr.io.
+		// https://github.com/helm/helm/issues/12584
+		var errUnexpectedStatus containerdErrors.ErrUnexpectedStatus
+		if errors.As(err, &errUnexpectedStatus) {
+			unwrappedErr := UnwrapAll(err)
```
willdavsmith, line 139: is there a better way to do this?

pkg/cli/helm/helm.go
```diff
@@ -20,15 +20,18 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	containerdErrors "github.com/containerd/containerd/remotes/errors"
```
ytimocin, line 29: We shouldn't use capital letters in import aliases.

willdavsmith: fixed

willdavsmith:
a note about the code coverage - the majority of changes in this PR are not tested by our unit tests and have not been so far

pkg/cli/helm/helm.go
```diff
@@ -136,3 +178,13 @@ func runUpgrade(upgradeClient *helm.Upgrade, releaseName string, helmChart *char
 	}
 	return err
 }
+
+func UnwrapAll(err error) error {
```
youngbupark, line 182: nit: missing comment

pkg/cli/helm/helm.go
```diff
@@ -136,3 +178,13 @@ func runUpgrade(upgradeClient *helm.Upgrade, releaseName string, helmChart *char
...
+	for {
+		unwrapped := errors.Unwrap(err)
+		if unwrapped == nil {
+			return err
+		}
+		err = unwrapped
+	}
```
ytimocin, line 189: This is to unwrap until you can't find any more errors, right?

willdavsmith: exactly

willdavsmith: I'm not sure if there's a better way to do this though

rynowak: what's the problem you're trying to solve by doing this?

rynowak: It looks like you're doing an `errors.As` and then ignoring the result ....

willdavsmith: yep this is exactly what I was doing. no idea how I missed that functionality for errors.As()....

pkg/cli/helm/helm.go
```diff
@@ -136,3 +172,17 @@ func runUpgrade(upgradeClient *helm.Upgrade, releaseName string, helmChart *char
...
+// If the error is not found, it returns nil.
+func extractHelmError(err error) error {
+	var errUnexpectedStatus containerderrors.ErrUnexpectedStatus
+	if errors.As(err, &errUnexpectedStatus) {
+		if errUnexpectedStatus.StatusCode == http.StatusForbidden && strings.Contains(errUnexpectedStatus.RequestURL, "ghcr.io") {
+			return fmt.Errorf("recieved 403 unauthorized when downloading helm chart from the registry. you may want to perform a `docker logout ghcr.io` and re-try the command")
+		}
+	}
+
+	return nil
```
rynowak, line 187: We need to also handle the case where the error is NOT an `containerderrors.ErrUnexpectedStatus`. Think about a case where the user is not connected to the internet. 

Every time we special case one type of error, we need to also handle the case where it's some other type of error.

rynowak: Oh, nevermind. I see you did that above.

rynowak: I'd suggest putting this line in here as well for clarity: https://github.com/radius-project/radius/pull/7455/files#diff-35b8f26ef00e63853d6a8303dfdb8755355d7cb6d48ba9449d5a04af62a723cbR142

pkg/cli/helm/helm.go
```diff
@@ -136,3 +172,17 @@ func runUpgrade(upgradeClient *helm.Upgrade, releaseName string, helmChart *char
...
 }
+
+// extractHelmError is a helper function to extract a specific error
+// (403 unauthorized when downloading a helm chart from ghcr.io) from a chain of errors.
+// If the error is not found, it returns nil.
+func extractHelmError(err error) error {
+	var errUnexpectedStatus containerderrors.ErrUnexpectedStatus
+	if errors.As(err, &errUnexpectedStatus) {
+		if errUnexpectedStatus.StatusCode == http.StatusForbidden && strings.Contains(errUnexpectedStatus.RequestURL, "ghcr.io") {
+			return fmt.Errorf("recieved 403 unauthorized when downloading helm chart from the registry. you may want to perform a `docker logout ghcr.io` and re-try the command")
```
rynowak, line 183: I'd suggest using `clierrors.Message` for this.

pkg/cli/helm/helm.go
```diff
@@ -101,8 +103,42 @@ func helmChartFromContainerRegistry(version string, config *helm.Configuration,
...
+		// Error handling for a specific case where credentials are stale.
+		// This happens for ghcr in particular because ghcr does not use
+		// subdomains - the scope of a login is all of ghcr.io.
+		// https://github.com/helm/helm/issues/12584
+		err := extractHelmError(err)
+		if err != nil {
+			return nil, err
+		}
+
 		return nil, fmt.Errorf("error downloading helm chart from the registry for version: %s, release name: %s. Error: %w", version, releaseName, err)
```
rynowak, line 142: I'd suggest using `clierrors.Message` for this. `rad` has special formatting for that type.

Basically it's helpful to know the difference between:

- You made a mistake, and it's not a bug in Radius. (`clierrors.Message`)
- Something weird happened, no idea 🤷 (any other error)

rynowak (approved):
:shipit: - awesome, thanks @willdavsmith

PR #7456, opened 2024-04-05
Updating Azure and AWS packages

ytimocin:
# Description
* Updating the Azure and AWS packages.
* We were using azcore 1.7.0 which didn't have this change.

## Type of change


pkg/sdk/connection.go
```diff
@@ -31,4 +31,8 @@ type Connection interface {
 	// Endpoint returns the endpoint (aka. base URL) of the Radius API. This definitely includes
 	// the URL scheme and authority, and may include path segments.
 	Endpoint() string
+
+	// InsecureAllowCredentialWithHTTP returns true if the connection allows sending credentials
+	// over HTTP. This is disabled by default.
+	InsecureAllowCredentialWithHTTP() bool
```
rynowak, line 37: Can you explain why this is needed? I don't want to clutter up the API with things we're doing.

pkg/sdk/connection_direct.go
```diff
@@ -30,11 +30,14 @@ var _ Connection = (*directConnection)(nil)
...
 	endpoint string
+
+	// insecureAllowCredentialWithHTTP allows sending credentials over HTTP. This is disabled by default.
+	insecureAllowCredentialWithHTTP bool
 }
 
 // NewDirectConnection parses the given endpoint string and returns a direct connection if the endpoint uses the http or
 // https scheme, otherwise it returns an error.
-func NewDirectConnection(endpoint string) (Connection, error) {
+func NewDirectConnection(endpoint string, insecureAllowCredentialWithHTTP bool) (Connection, error) {
```
rynowak, line 40: Do we ever set this to true? I'd really not like for this to be part of the API if we don't have a clear usecase for it.

pkg/sdk/connection.go
ytimocin: We have these tests: https://github.com/radius-project/radius/blob/main/pkg/portableresources/processors/resourceclient_test.go#L295.

And when we update the **azcore** package, these tests start to fail with this error: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_bearer_token.go#L120.

Here is where we create the necessary client for the tests mentioned above: https://github.com/radius-project/radius/blob/main/pkg/portableresources/processors/resourceclient.go#L182. I couldn't find any other way to set this flag to true just for the tests so that tests can pass.

Happy to discuss offline.

pkg/sdk/connection_direct.go
ytimocin: We set this to true for some tests: https://github.com/radius-project/radius/blob/main/pkg/portableresources/processors/resourceclient_test.go#L295.

pkg/sdk/connection.go
rynowak: Per discussion offline, can see if it's possible to remove the `BearerTokenPolicy`? None of the UCP scenarios should be using a bearer token, so we don't need this.

rynowak:
Hey @ytimocin - I think we need to tweak the logic for the failing check (codecov/patch). I'm not sure it's reasonable for it to fail on a trivial change like this. 

Let me know if you need me to bypass the check for you.

pkg/sdk/pipeline.go
```diff
@@ -46,6 +46,13 @@ func NewClientOptions(connection Connection) *arm.ClientOptions {
 				&removeAuthorizationHeaderPolicy{},
 			},
 			Transport: connection.Client(),
+			// When updating azcore to 1.11.1 from 1.7.0, we saw that HTTPS check for Authentication was added.
+			// Link to the check: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_bearer_token.go#L118
+			//
+			// This check was failing for some unit tests because the ARM requests are made over HTTP and the bearer token is being sent in the header.
+			// This is a temporary fix to allow sending the bearer token over HTTP.
```
kachawla, line 53: Do we have a longer term fix in mind? If not, then I'm not sure if the comment about "temporary fix" is conveying the intended message.

rynowak:
Force merging this to bypass the new codecov check. We don't want the check to fail on trival changes like this, it will be reviewed 👍

PR #7478, opened 2024-04-09
Fix dashboard app graph not showing bug

### LINKED ISSUE #68: Dashboard App Graph: Failed to load resource: the server responded with a status of 403 (Forbidden)
### Steps to reproduce

Install Radius v0.32
deploy an application
rad run or rad deploy and open dashboard
visit the APP GRAPH tab on the application in the dashboard and an infinite load will happen

### Observed behavior

Checking the logs shows a Failed to load resource: the server responded with a status of 403 (Forbidden) for a POST request to the `http://localhost:7007/api/kubernetes/proxy/apis/api.ucp.dev/v1alpha3//planes/radius/local/resourcegroups/default/providers/Applications.Core/applications/dashboarddemo/getGraph?api-version=2023-10-01-preview` endpoint

### Desired behavior

App graph shows

### Workaround

k edit clusterrole -n radius-system dashboard, switch get and list to *

### Additional context

_No response_

willdavsmith:
# Description

* Fixing dashboard app graph bug

Note that this matches the deploy.yaml spec in the dashboard: https://github.com/radius-project/dashboard/blob/main/deploy/dashboard.yaml

## Type of change

Fixes: https://github.com/radius-project/dashboard/issues/68

### DIFF
--- a/deploy/Chart/templates/dashboard/rbac.yaml
+++ b/deploy/Chart/templates/dashboard/rbac.yaml
@@ -10,7 +10,7 @@ metadata:
 rules:
   - apiGroups: ['api.ucp.dev']
     resources: ['*']
-    verbs: ['get', 'list']
+    verbs: ['*']
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/deploy/Chart/templates/dashboard/rbac.yaml
+++ b/deploy/Chart/templates/dashboard/rbac.yaml
@@ -10,6 +10,7 @@ metadata:
 rules:
   - apiGroups: ['api.ucp.dev']
     resources: ['*']
+    # dashboard needs get, list, and post privileges for api.ucp.dev
     verbs: ['*']
 ---
 apiVersion: rbac.authorization.k8s.io/v1


deploy/Chart/templates/dashboard/rbac.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
 rules:
   - apiGroups: ['api.ucp.dev']
     resources: ['*']
-    verbs: ['get', 'list']
+    verbs: ['*']
```
ytimocin, line 13: Instead of `*` should we only have the ones **dashboard** needs?

ytimocin (dismissed):
Approved just in case if this is needed for the community call tomorrow.

deploy/Chart/templates/dashboard/rbac.yaml
rynowak: we also need `post` here (that was the bug). I don't have a strong preference between being explicit and using the wildcard. This is just our API and the tradeoffs are small.

kachawla, line 13: Could you add a comment here for context for anyone looking at this code later?

kachawla (dismissed):

deploy/Chart/templates/dashboard/rbac.yaml
willdavsmith: added

ytimocin (approved):

kachawla (approved):

Issue #6946, opened 2023-12-13
Root certificate authority requires escape characters on windows OS


sk593:
### Steps to reproduce (required)

On Windows, follow this docs page on installing Radius with your own root certificate authority certificate: https://docs.radapp.io/guides/operations/kubernetes/install/#use-your-own-root-certificate-authority-certificate

### Observed behavior (required)

When the `rad install` command is run with a rootCA, an error is returned that the file path for the certificate cannot be found 
```
PS C:\WINDOWS\system32> rad install kubernetes --set-file global.rootCA.cert="C:\Users\Desktop\test.crt"
Installing Radius version v0.28.0 to namespace: radius-system...
Error: failed to add Radius values, err: open C:UsersDesktoptest.crt: The system cannot find the file specified., Helm output:
 
TraceId:  a9319a4e612397470986b0b125d574a6`
 
PS C:\WINDOWS\system32> rad install kubernetes --set-file global.rootCA.cert="C:\\Users\\Desktop\\test.crt"
Installing Radius version v0.28.0 to namespace: radius-system...
PS C:\WINDOWS\system32>
```

### Desired behavior (required)

We'd like this flag to succeed in setting the cert on Windows

Some options for fixing this issue: 
1. We handle the escape characters in the file path when the OS is Windows
2. We update our documentation to note that escape characters may be needed when providing the file path on Windows 

### Workaround (optional)

Using escape characters on the file path resolves the issue 
```
PS C:\WINDOWS\system32> rad install kubernetes --set-file global.rootCA.cert="C:\\Users\\Desktop\\test.crt"
Installing Radius version v0.28.0 to namespace: radius-system...
PS C:\WINDOWS\system32>
```

### rad Version (required)

RELEASE   VERSION   BICEP     COMMIT
0.28.0    v0.28.0   0.28.0    c7217b6f661a790053c1b65ff7972a0c5be33908

### Operating system (required)

Windows OS 

## Additional context

[AB#10750](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10750)

gpltaylor:
@rynowak can this be closed now?

rynowak:
Yes, thanks for the reminder. This was fixed by #7479

PR #7479, opened 2024-04-09
PR - Root certificate authority requires escape characters on windows OS

gpltaylor:
# Description
Resolve windows path being escaped by helm pkg

## Type of change

## Additional Description
Helm.runesUntil strips \\ (x92) rune from string when converting Buffer RuneReader to []runes. This issues is outside this package. Alternatives is to import this func and remove this or mutate path before sending to parser. Care also has to be take to support users who are using "c:\\\\dir\\\\file.crt" as a hack. This only works as Helm's runesUntil strips \ by using the next rune!

fixes: Root certificate authority requires escape characters on windows OS #6946


pkg/cli/helm/radiusclient.go
```diff
@@ -141,6 +141,13 @@ func AddRadiusValues(helmChart *chart.Chart, options *RadiusOptions) error {
 	}
 
 	for _, arg := range options.SetFileArgs {
+		// Continue to support: c:\\testdir\\testfile.crt
+		// and convert: c:\testdir\testfile.crt to c:/testdir/testfile.crt
+
+		if !strings.Contains(arg, "\\\\") && strings.Contains(arg, "\\") {
```
rynowak, line 147: Should we check for windows first? `\` has a different meaning in linux paths. It's a pretty rare case but `\` is legal as part of a filename in linux.

rynowak: Even better we could check [os.PathSeparator](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/os/path_unix.go;l=10)

gpltaylor: How about adding an additional check to ` runtime.GOOS`.
This along with the `os.PathSeparator` to stop using magic strings in the search.

There is precedence of this being used within the project

https://github.com/radius-project/radius/blob/74a3f25de96c0cee2b689fe33c6f3499e725c02b/pkg/cli/tools/util.go#L24

gpltaylor: @rynowak clean up the logic... tested on both Windows and Linux.

image

rynowak: My first time seeing [filepath.ToSlash](https://pkg.go.dev/path/filepath#ToSlash).

Fix looks great!

rynowak:
@gpltaylor - thanks! the code change looks good. I've kicked off a functional test run, and I'll take care of merging when it passes.

rynowak:
Note: the codecoverage check that's failing is new and we're stilling adjusting it. Right now it fails on small changes like this one, so I'll need to bypass the check.

rynowak:
Merged! thanks

PR #7485, opened 2024-04-11
Create the issue only when there are consecutive failures

youngbupark:
# Description

To better balance issue detection and noise reduction, I propose the following adjustments:

* Only generate issues after at least two consecutive test failures. This approach reduces noise while ensuring that recurring problems are addressed.
* Reevaluate the alert triggering thresholds. Considering that no system achieves 100% SLO and that our actual SLO is a product of multiple dependent services' SLOs/SLAs, our current trigger for alerts appears overly aggressive. Allowing for a few failures could provide a more realistic measure of system robustness and service reliability.

To understand the details, please see https://github.com/radius-project/radius/issues/7513.

## Type of change

https://github.com/radius-project/radius/issues/7513


.github/workflows/functional-test.yaml
```diff
@@ -694,14 +696,38 @@ jobs:
...
               ...context.repo,
               title: `Scheduled functional test failed - Run ID: ${context.runId}`,
               labels: ['bug', 'test-failure'],
-              body: `## Bug information \n\nThis bug is generated automatically if the scheduled functional test fails. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](${process.env.ACTION_LINK}).`
+              body: `## Bug information \n\nThis bug is generated automatically if the scheduled functional test fails more than ${process.env.ISSUE_CREATE_THRESHOLD}. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](${process.env.ACTION_LINK}).`
```
ytimocin, line 732:
```suggestion
              body: `## Bug information \n\nThis bug is generated automatically if the scheduled functional test fails more than ${process.env.ISSUE_CREATE_THRESHOLD} times. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](${process.env.ACTION_LINK}).`
```

.github/workflows/functional-test.yaml
```diff
@@ -694,14 +696,38 @@ jobs:
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            response = await github.rest.actions.listWorkflowRuns({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'long-running-azure.yaml',
+              event: 'schedule',
+              per_page: 10
+            });
```
ytimocin, line 710: Is this going to check for a specific run? Because this seems like there will always be more than one workflow that failed and it is always going to create an issue?

youngbupark: > Is this going to check for a specific run? Because this seems like there will always be more than one workflow that failed and it is always going to create an issue?

Yes it will check the specific workflow. Please see L707
```
workflow_id: 'long-running-azure.yaml',
```

youngbupark: Ah I have to update it to functional test.

.github/workflows/functional-test.yaml
```diff
@@ -694,14 +696,38 @@ jobs:
...
     steps:
+      - name: Count recently failed tests
+        id: count_failures
+        uses: actions/github-script@v7
+        with:
+          script: |
+            response = await github.rest.actions.listWorkflowRuns({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'functional-test.yaml',
```
youngbupark, line 707: @ytimocin workflow_id will query`functional-test.yaml` workflow.

.github/workflows/long-running-azure.yaml
```diff
@@ -509,14 +512,38 @@ jobs:
...
     steps:
+      - name: Count recently failed tests
+        id: count_failures
+        uses: actions/github-script@v7
+        with:
+          script: |
+            response = await github.rest.actions.listWorkflowRuns({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: 'long-running-azure.yaml',
```
youngbupark, line 523: @ytimocin this will query long-running-azure workflow only.

.github/workflows/functional-test.yaml
youngbupark: No it should always return more than 2. The API returns workflow runs data in chronical order. L713-L719 counts only consecutive failures before the current run. the initial value of failureCount might make you confused. I can update it to 0.

ytimocin (approved):

lakshmimsft (approved):

.github/workflows/functional-test.yaml
```diff
@@ -694,14 +696,38 @@ jobs:
...
               ...context.repo,
               title: `Scheduled functional test failed - Run ID: ${context.runId}`,
               labels: ['bug', 'test-failure'],
-              body: `## Bug information \n\nThis bug is generated automatically if the scheduled functional test fails. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](${process.env.ACTION_LINK}).`
+              body: `## Bug information \n\nThis bug is generated automatically if the scheduled functional test fails more than ${process.env.ISSUE_CREATE_THRESHOLD} times. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](${process.env.ACTION_LINK}).`
```
lakshmimsft, line 732: should this be 'equal to or more than' ?

youngbupark: > should this be '..if the scheduled functional test fails **atleast** ${process.env.ISSUE_CREATE_THRESHOLD} times' ? same comment for long-running-azure.yaml

@lakshmimsft You're right :). Please review this PR - https://github.com/radius-project/radius/pull/7527

PR #7497, opened 2024-04-12
Update TF Exec package version

ytimocin:
# Description
Update TF Exec package version

## Type of change


pkg/recipes/driver/terraform.go
```diff
@@ -152,7 +152,7 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
 // prepareRecipeResponse populates the recipe response from the module output named "result" and the
 // resources deployed by the Terraform module. The outputs and resources are retrieved from the input Terraform JSON state.
 func (d *terraformDriver) prepareRecipeResponse(ctx context.Context, definition recipes.EnvironmentDefinition, tfState *tfjson.State) (*recipes.RecipeOutput, error) {
-	if tfState == nil || (*tfState == tfjson.State{}) {
+	if tfState == nil || reflect.DeepEqual(*tfState, tfjson.State{}) {
```
sk593, line 155: Was this causing issues before using DeepEqual?

sk593 (dismissed):

pkg/recipes/driver/terraform.go
ytimocin: Nope. With the updated TF Exec, it started to throw an error.

vishwahiremat (approved):

pkg/recipes/driver/terraform.go
gpltaylor: World unit tests be useful here?  To demonstrating that this fixes the issue and explain why a Deep Equal is required.
It's unclear to me at least, what problem this solves and why.

rynowak: +1 to the suggestion from @gpltaylor. Or a comment explaining why this check is important.

ytimocin: [This](https://github.com/hashicorp/terraform-json/blob/main/state.go#L41) was added and that is why the equals check started to throw an error.

pkg/recipes/driver/terraform.go
```diff
@@ -152,14 +152,17 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error
...
+	// The reason is that Go does not allow comparison of structs that contain slices.
+	// Please see: https://go.dev/ref/spec#Comparison_operators.
+	if tfState == nil || reflect.DeepEqual(*tfState, tfjson.State{}) {
 		return &recipes.RecipeOutput{}, errors.New("terraform state is empty")
 	}
 
 	recipeResponse := &recipes.RecipeOutput{}
-	moduleOutputs := tfState.Values.Outputs
-	if moduleOutputs != nil {
+	if tfState.Values != nil && tfState.Values.Outputs != nil {
```
ytimocin, line 163: Could you look at this part and let me know if this makes sense @kachawla @vishwahiremat? Can the **Values** ever be nil?

pkg/recipes/driver/terraform.go
```diff
@@ -174,9 +177,13 @@ func (d *terraformDriver) prepareRecipeResponse(ctx context.Context, definition
 		TemplateVersion: definition.TemplateVersion,
 	}
 
-	deployedResources, err := d.getDeployedOutputResources(ctx, tfState.Values.RootModule)
-	if err != nil {
-		return &recipes.RecipeOutput{}, err
+	var deployedResources []string
+	if tfState.Values != nil && tfState.Values.RootModule != nil {
```
ytimocin, line 181: Same question here as above: Could you look at this part and let me know if this makes sense @kachawla @vishwahiremat? Can the **Values** ever be nil?

pkg/recipes/driver/terraform.go
gpltaylor: Well this has me worried about all my code, every :) !
So we are guarding against a default implementation. AKA an unpopulated new state object...
Go does not compare Structs (who knew) so DeepEquals can walk the path for us..

My next question is why does the code allow us to get into this bad state branch. However, I'm not qualified to understand how the code gets to this execution point ATM.

So, this is good code IMO 👍 

I made a Gist to get it clear in my head.
https://gist.github.com/gpltaylor/3fe882a8f276f14b15e9a601cd9f6e8f

ytimocin: We updated `github.com/hashicorp/terraform-json` from **0.15.0** to **0.19.0** in this PR and we noticed that there was an addition to the **tfjson.State** struct: https://github.com/hashicorp/terraform-json/blob/main/state.go#L41.

Because of this addition our code started to complain about this check: `if tfState == nil || (*tfState == tfjson.State{})`. Now that there is an array in the struct, the code wanted us to do a deeper check, which is the changed line above.

@rynowak both nil and empty check is to make sure that the passed-in state is not nil or empty. It makes sense to me to check for both as we do this type of check in a lot of different places in our code. If we would like more information, maybe @kachawla or @vishwahiremat can add their thoughts here.

gpltaylor (approved):
Based on the research I've done this feels like a valid change. As the comments clearly state a simple compare will not work any more and a deep equals is required. This of course is under the condition that an empty state can reach this code branch.

For reference: https://www.youtube.com/watch?v=1rIDxEjNtCI&&ab_channel=gpltaylor

pkg/recipes/driver/terraform.go
vishwahiremat: I am not sure if values can be Nil, but i think its good to have the nil check on Values.

pkg/recipes/driver/terraform.go
vishwahiremat, line 181: nit: we are already doing the nil check for `RootModule` inside d.getDeployedOutputResources()
```suggestion
	if tfState.Values != nil {
```

gpltaylor: It's GoLang, you can never have too many nil checks

rynowak (approved):

pkg/recipes/driver/terraform.go
kachawla: It is possible for extender resource since we don't do enforce any outputs for that, so I think it makes sense to add the nil check. Thanks @ytimocin!

kachawla (commented):
Looks good to me overall. @ytimocin can we add unit tests that execute with nil `tfstate.Values` and `tfState.Values.Outputs`?

PR #7500, opened 2024-04-14
[DO NOT REVIEW] Making some parameters pointers to decrease memory usage

ytimocin:
# Description
This is a side project I am working on. I used **gocritic** to find potential performance improvements for Radius.

Running this command `gocritic check -enable hugeParam ./pkg/...` gave me a lot of parameters that could be turned into pointers.

I will be testing this out. I created an AKS cluster to test the changes: https://benchmark-dashboard-ezd4hsh8hrabekcv.eus.grafana.azure.com/d/Qw-xtsxVz/radius-resource-provider?orgId=1&from=1713040653820&to=1713085544923 (Please let me know if you can see this link).

image
There is a slight drop in memory usage. I need more tests to see if this is actually working.

## Type of change

### DIFF
--- a/pkg/armrpc/api/v1/types_test.go
+++ b/pkg/armrpc/api/v1/types_test.go
@@ -89,8 +89,9 @@ func TestBaseResource_UpdateMetadata(t *testing.T) {
 
 	armCtx := &ARMRequestContext{Location: "global"}
 	var err error
-	armCtx.ResourceID, err = resources.ParseResource(newResourceID)
+	resourceID, err := resources.ParseResource(newResourceID)
 	require.NoError(t, err)
+	armCtx.ResourceID = *resourceID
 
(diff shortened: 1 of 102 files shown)


pkg/armrpc/api/v1/types_test.go
```diff
@@ -89,8 +89,9 @@ func TestBaseResource_UpdateMetadata(t *testing.T) {
 
 	armCtx := &ARMRequestContext{Location: "global"}
 	var err error
-	armCtx.ResourceID, err = resources.ParseResource(newResourceID)
+	resourceID, err := resources.ParseResource(newResourceID)
```
rynowak, line 92: If this is a good change overall, it might make sense to make `resources.ParseResource` return a pointer. Then you'd be encouraged to use it as a pointer since you'd have to do work to do otherwise.

PR #7506, opened 2024-04-15
Update bicep generator to use bicep-types and new contract

sk593:
# Description

Update bicep generator to use bicep-types and new contract

## Type of change

Fixes: #issue_number


rynowak:
Are there functionality changes in this PR or just updates to the new types?

sk593:
> Are there functionality changes in this PR or just updates to the new types?

A little of both. Since the type contract for bicep was recently updated, there has been some change in functionality (mostly with adding functions to resources and getting our repo in sync with the `bicep-types` repo). But the outcome of the changes are the same w.r.t. to the code being used to generate types/index files

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
```diff
@@ -242,113 +242,20 @@ async function findReadmePaths(specsPath: string) {
...
-    }
+  const typeFiles: TypeFile[] = [];
+  for (const typePath of typesPaths) {
+    const content = await readFile(typePath, { encoding: 'utf8' });
+    typeFiles.push({
+      relativePath: path.relative(baseDir, typePath),
+      types: readTypesJson(content),
+    });
   }
+  const indexContent = await buildIndex(typeFiles,  log => logOut(logger, log));
```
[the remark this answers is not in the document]
lakshmimsft: ooo.. we're now importing it from `bicep-types` , okay

build/generate.mk
```diff
@@ -116,7 +116,9 @@ generate-go: generate-mockgen-installed ## Generates go with 'go generate' (Mock
 generate-bicep-types: generate-node-installed ## Generate Bicep extensibility types
 	@echo "$(ARROW) Generating Bicep extensibility types from OpenAPI specs..."
 	@echo "$(ARROW) Build autorest.bicep..."
-	cd hack/bicep-types-radius/src/autorest.bicep; \
+	cd bicep-types/src/bicep-types; \
+	npm ci && npm run build; \
+	cd ../../../hack/bicep-types-radius/src/autorest.bicep; \
```
lakshmimsft, line 121: fmi.. more info on reason for this path change?

hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.md
```diff
@@ -37,20 +26,13 @@
...
 
+## TestType1Properties
+### Properties
+* **basicString**: string: Description for a basic string property.
+* **stringEnum**: 'Bar' | 'Foo': Description for a basic enum property.
+
 ## TrackedResourceTags
 ### Properties
 ### Additional Properties
 * **Additional Properties Type**: string
```
lakshmimsft, line 37: FoosRequest. FoosResponse still exists under :
hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.md and
sk593/radius/hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.json .. should it be deleted here?

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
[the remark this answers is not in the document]
sk593: Yes, we're using the function that the `bicep-types` repo exposes

hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.md
sk593: This might be because of how the `bicep-types` markdown generator is written. Ideally, it should still be generated but I'll double check if this was missed in that repo or if it's an issue in our code

build/generate.mk
sk593: Just some root directory details. Earlier we start with cd-ing into the hack folder as the working directory. But now we need to build the `bicep-types` library first so we're cd-ing into that as the working directory. `cd ../../../hack/bicep-types-radius/src/autorest.bicep;` is how we go from the bicep-types folder to the autorest folder to change the execution directory

hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.md
lakshmimsft: Following up here: We need some clarity/more info on the .md files under the hack/bicep-types-radius/generated/ folder not added to github (they are in .gitignore folder) but then there is a test folder which also includes these generated files which gets into github and there is some discrepancy in there.

sk593: I think the test folder can keep the MD files in case we ever add them back to the generated types folder. From what I remember, the other MD files were added to the .gitignore because there was a discrepancy between the MD files and the docs page and we wanted to keep them the same. @willtsai or the PMs may have more insight on this. This is the PR that removed the MD files from the hack folder: https://github.com/radius-project/radius/pull/6331

hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
```diff
@@ -142,66 +138,74 @@ export function generateTypes(host: AutorestExtensionHost, definition: ProviderD
       }
 
       const { descriptor, bodyType } = output;
+      let resourceTypeFunctions: Record<string, ResourceTypeFunction> = {};
+      
+      let actions = resourceFunctionType[fullyQualifiedType];
```
sk593, line 143: We are now tracking functions per resource instead of overall. So each function will be in the types file associated with that resource's namespace.

hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.md
willtsai: okay, so it looks like Aaron removed the MD files from the hack folder in [PR#6331](https://github.com/radius-project/radius/pull/6331) because these MD documents gets autogenerated in the workflow added by the same PR. Thus, I believe the right call is to continue keeping the MD files in the .gitignore so that the docs site can depend on the autogeneration workflow that keeps the documentation (for [example](https://docs.radapp.io/reference/api/applications.core/applications/)) up to date.

sk593: sounds good, thanks for looking into that. I still think it makes sense to keep the MD file in the test folder since we do generate those (even if they're not uploaded to Github). since the types in that file are just test resources, I don't think it'll be a point for confusion. thoughts @lakshmimsft ?

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
```diff
@@ -242,113 +243,50 @@ async function findReadmePaths(specsPath: string) {
...
-        }
-        resourceTypes.add(resourceType.Name.toLowerCase());
-  
-        resDictionary[resourceType.Name] = {
-          RelativePath: path.relative(baseDir, typeFilePath),
-          Index: types.indexOf(type),
-        };
-        
-        continue;
+function getRadiusRelease(): Promise<string> {
```
sk593, line 264: Thinking about this more and this actually might not work. PRs should get merged in with the "latest" tag but when we test, we'd want to test against a PR-specific tag, not "latest". So this might end up generating the wrong version number. Will look into this more

willdavsmith (approved):
LGTM, but can you quickly summarize how you tested this in a comment?

sk593:
> LGTM, but can you quickly summarize how you tested this in a comment?

A lot of it was manual testing but we have an integration test file that verifies the generation is accurate. The test wasn't updated but the test files were

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
sk593: Leaving this for now. Will address in a follow up PR when we get to uploading to ACR

PR #7523, opened 2024-04-19
Update build workflows to allow manual runs 

sk593:
# Description

This updates the build workflow to allow manual runs for testing feature branch changes

## Type of change

Fixes: #issue_number


.github/workflows/build.yaml
```diff
@@ -16,6 +16,8 @@
 
 name: Build and Test
 on:
+  # Enable manual trigger
+  workflow_dispatch:
```
youngbupark, line 20: You might want to get the input for your feature branch. Please see this - https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

sk593: Where would the inputs be used for this? From my understanding, the manual run would use the workflow version from the feature branch that has the needed changes. Unless I'm missing something

youngbupark: So do you just want to run build action manually from main branch? Or do you want to specify which feature branch you want to build ?

sk593: Yes to the second question. I want to be able to run the workflow from the feature branch. So with the `workflow_dispatch` I can specify which branch to build the workflow off of

youngbupark: image

It fetches the branches of the upstream repo. Is this expected? don't you want to build the branch in the forked repo?

youngbupark (approved):

.github/workflows/build.yaml
youngbupark: I just approved this PR. but you might need additional input if you want to support branches in forked repo.

sk593: Ah no, so the feature branch is in the main repo not in a forked repo. So the branches of the upstream repo is correct

.github/workflows/functional-test.yaml
```diff
@@ -22,6 +22,8 @@ permissions:
   packages: write # Required for uploading the package
 
 on:
+  # Enable manual trigger
+  workflow_dispatch:
```
willdavsmith, line 26: if run via workflow dispatch, the tests might not work. but we should try it.

willdavsmith (approved):

Issue #7521, opened 2024-04-19
Move all container images in test to ghcr.io


youngbupark:
docker.io has the rate limit - https://docs.docker.com/docker-hub/download-rate-limit/

It can cause the following issues in functional test. It is better to not use docker.io image in the test. We need to copy it to ghcr and then consume it in the test. 

<img width="1201" alt="image" src="https://github.com/radius-project/radius/assets/11863108/e821d0d4-7d6b-48db-b900-9395a62c3e36">

[AB#11798](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11798)

PR #7530, opened 2024-04-23
Update tests to pull test images from ghcr instead of docker

### LINKED ISSUE #7512: Scheduled functional test failed - Run ID: 8722019265
## Bug information 

This bug is generated automatically if the scheduled functional test fails. The Radius functional test operates on a schedule of every 4 hours during weekdays and every 12 hours over the weekend. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](https://github.com/radius-project/radius/actions/runs/8722019265).

[AB#11789](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11789)

lakshmimsft:
# Description

We encountered intermittent errors in scheduled functional tests when downloading redis image from docker either due to network issues/rate limiting. eg. https://github.com/radius-project/radius/issues/7512
List of errors: https://github.com/radius-project/radius/issues?q=is%3Aissue+is%3Aopen+docker.io%2Flibrary%2Fredis

This PR will pull redis image from ghcr.io instead of docker and then we will continue to monitor if the update eliminates this error.

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius.

Fixes: #7512 , #7509 , #7504 , #7521

### DIFF
--- a/test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
+++ b/test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
@@ -41,7 +41,7 @@ resource redisContainer 'Applications.Core/containers@2023-10-01-preview' = {
   properties: {
     application: app.id
     container: {
-      image: 'redis:6.2'
+      image: 'ghcr.io/radius-project/dev/redis:6.2'
       ports: {
         redis: {
           containerPort: 6379
--- a/test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
+++ b/test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
@@ -41,7 +41,7 @@ resource redisContainer 'Applications.Core/containers@2023-10-01-preview' = {
   properties: {
     application: app.id
     container: {
-      image: 'ghcr.io/radius-project/dev/redis:6.2'
+      image: 'ghcr.io/radius-project/mirror/redis:6.2'
       ports: {
         redis: {
           containerPort: 6379
--- a/test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-cmd-args.bicep
+++ b/test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-cmd-args.bicep
@@ -26,7 +26,7 @@ resource container 'Applications.Core/containers@2023-10-01-preview' = {
   properties: {
     application: app.id
     container: {
-      image: 'debian'
+      image: 'ghcr.io/radius-project/mirror/debian:latest'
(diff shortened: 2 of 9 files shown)


test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
```diff
@@ -41,7 +41,7 @@ resource redisContainer 'Applications.Core/containers@2023-10-01-preview' = {
   properties: {
     application: app.id
     container: {
-      image: 'redis:6.2'
+      image: 'ghcr.io/radius-project/dev/redis:6.2'
```
youngbupark, line 44: We're purging the images which has `dev/` prefix regularly by [this workflow](https://github.com/radius-project/radius/blob/5c50b81184cd117df90c6fdf5e25f3e904507030/.github/workflows/purge-test-resources.yaml#L38-L45). So I would not recommend to use `dev/`

I would choose the different path name like below.

 ```suggestion
      image: 'ghcr.io/radius-project/mirror/redis:6.2'
```

**Comment by lakshmimsft** [comment]
Since redis image is working with tests, I will update PR for other images later today and after tests pass, will then request permissions to be updated for new images on ghcr to be restricted to maintainer admin access.

**Review by kachawla — APPROVED** [review]

File: test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep [hunk 2]
```diff
@@ -41,7 +41,7 @@ resource redisContainer 'Applications.Core/containers@2023-10-01-preview' = {
   properties: {
     application: app.id
     container: {
-      image: 'redis:6.2'
+      image: 'ghcr.io/radius-project/mirror/redis:6.2'
```
kachawla, line 44: So is the recommendation in this PR to set `mirror` as convention for subdirectory path for all test images?

ytimocin (approved):

test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
lakshmimsft: I inferred it as we're creating a mirror subdirectory for images we're using from docker for test purposes.

kachawla: Makes sense @lakshmimsft. Are there any docs that need to be updated for any future test images we add?

rynowak (commented):
:shipit: - code changes look good! I also like the suggestion of adding this to our contributing docs.

PR #7551, opened 2024-04-26
Update purge test resources (both azure and aws) workflows

### LINKED ISSUE #7550: `AADSTS700024: Client assertion is not within its valid time range` error in Purge Test Resources workflow
### Steps to reproduce

Today I noticed that our functional tests were failing due to number of resource groups we have: https://teams.microsoft.com/l/message/19:ve3Qj3hke37Z4b1EdjJy-K4nzh3nDgRxZJJARwusRgc1@thread.tacv2/1714089575308?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=b33db407-e89b-4fb6-87cc-9b2f9f8ec265&parentMessageId=1714089575308&teamName=Radius%20Core%20Team&channelName=General&createdTime=1714089575308.

I, then, checked the **Purge Test Resources** workflow to see that it had been failing for a while: https://github.com/radius-project/radius/actions/workflows/purge-test-resources.yaml.

### Observed behavior

Test resources were not being purged for a while.

### Desired behavior

1. Test resources being purged
2. Issue gets created if there is a problem with the `Purge test resources` workflow

### Workaround

Manually deleting the resource groups.

### rad Version

0.32

### Operating system

_No response_

### Additional context

_No response_

[AB#12161](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12161)

ytimocin:
# Description
Update purge test resources (both azure and aws) workflows

## Type of change
Fixes: #7550

### DIFF
--- a/.github/workflows/purge-aws-test-resources.yaml
+++ b/.github/workflows/purge-aws-test-resources.yaml
@@ -4,7 +4,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#    
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
@@ -23,8 +23,8 @@ on:
 
 env:
   AWS_REGION: us-west-2
-  AWS_RESOURCE_TYPES: 'AWS::RDS::DBSubnetGroup,AWS::RDS::DBInstance,AWS::S3::Bucket,AWS::Logs::MetricFilter,AWS::Logs::LogGroup'
-  
+  AWS_RESOURCE_TYPES: "AWS::RDS::DBSubnetGroup,AWS::RDS::DBInstance,AWS::S3::Bucket,AWS::Logs::MetricFilter,AWS::Logs::LogGroup"
+
 jobs:
   purge_aws_resources:
     name: Delete old AWS resources created by tests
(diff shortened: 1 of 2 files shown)


.github/workflows/purge-test-resources.yaml
```diff
@@ -93,5 +93,17 @@ jobs:
           cat ${{ env.AZURE_RG_DELETE_LIST_FILE}} | while read line
           do
               echo " * $line" >> $GITHUB_STEP_SUMMARY
-              az group delete --resource-group $line --yes --verbose
+              az group delete --resource-group $line --yes --verbose --no-wait
```
ytimocin, line 96: I think that because we have a lot of resource groups, it takes more than 5 minutes to delete them. Within that 5 minutes, the session loses its access because login is only granted for 5 minutes. Please see: https://github.com/Azure/login/issues/372.

I may be wrong about this. Open to suggestions. cc/ @youngbupark

.github/workflows/purge-aws-test-resources.yaml
```diff
@@ -40,3 +40,15 @@ jobs:
       - name: Delete old AWS resources
         run: |
           ./.github/scripts/delete-aws-resources.sh ${{ env.AWS_RESOURCE_TYPES }}
+      - name: Create issue for failing purge aws test resources run
```
youngbupark, line 43: Don't you think it is too verbose?

.github/workflows/purge-test-resources.yaml
```diff
@@ -93,5 +93,17 @@ jobs:
...
           done
+      - name: Create issue for failing purge test resources run
```
youngbupark, line 98: I do not think we need this.. it is too verbose and noisy.

.github/workflows/purge-test-resources.yaml
sk593: I think this is a similar issue to this one: https://github.com/radius-project/radius/issues/7494. There was a regression in the az cli with the login expiration so they provided some workarounds here: https://github.com/Azure/azure-cli/issues/28737#issuecomment-2051026300

But we already use a service principal and that still seems to be failing for us. I think they plan to release a fix on 4/30 so we may need to revisit this once that version is out

youngbupark: Looks like az login issues 5 mins expiry token. I am ok to add `--no-wait`. The fix will be released in 2.60 CLI.

youngbupark:
@ytimocin I do not think we need to create the issue for the deletion failure. It seems too noisy. other then them, lgtm.

kachawla (approved):

PR #7571, opened 2024-05-07
Adding missing parts to the purge workflows

ytimocin:
# Description
Adding missing parts to the purge workflows:
1. Check out code step was missing in one of the workflows.
2. ACTION_LINK was missing in both of the workflows.
3. Moving the Create Issue step to the workflow level because there are two jobs, and we want to create issues if one of them or both fails.

## Type of change
Fixes: #issue_number

### DIFF
--- a/.github/workflows/purge-aws-test-resources.yaml
+++ b/.github/workflows/purge-aws-test-resources.yaml
@@ -24,6 +24,8 @@ on:
 env:
   AWS_REGION: us-west-2
   AWS_RESOURCE_TYPES: "AWS::RDS::DBSubnetGroup,AWS::RDS::DBInstance,AWS::S3::Bucket,AWS::Logs::MetricFilter,AWS::Logs::LogGroup"
+  # The current GitHub action link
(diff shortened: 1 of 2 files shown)


kachawla (approved):

kachawla (commented):
@ytimocin looks like there are still issues with this workflow: https://github.com/radius-project/radius/actions/runs/9013693181/job/24764966492

ytimocin:
> @ytimocin looks like there are still issues with this workflow: https://github.com/radius-project/radius/actions/runs/9013693181/job/24764966492

Thanks for the heads up @kachawla! I will take a look at this now.

ytimocin:
> @ytimocin looks like there are still issues with this workflow: https://github.com/radius-project/radius/actions/runs/9013693181/job/24764966492

Seems like this workflow has always failed. Here is the PR to fix the issue: https://github.com/radius-project/radius/pull/7581. You can also see the successful runs here: https://github.com/radius-project/radius/actions/workflows/purge-aws-test-resources.yaml.

PR #7578, opened 2024-05-09
Stern - maxLogRequests option update

### LINKED ISSUE #7565: `stern` log collector prevents `rad run` from working
### Steps to reproduce

- Create an application that runs more than 10 containers. In my case, I'm running a few (test) MongoDb containers and a Jaeger Daemonset next to my App.
- Execute `rad run`
- This will fail because of `stern` 
- Notice the error message

### Observed behavior

This is the output that is generated:

```
Starting log stream...

+ backendapi-58d766c8f9-2mbjd › backendapi
+ backendapi-58d766c8f9-2mbjd › daprd
+ frontend-754c68fd46-lw2l8 › frontend
+ jaeger-soasl3vdz6bj4-j995d › jaeger
+ jaeger-soasl3vdz6bj4-jfvkt › jaeger
+ mongo-dkna2yzxpcneq-67645c4fd-8g4gl › mongo
+ mongo-iqc733ktshkvc-78f5795dd9-rsf49 › mongo
+ mongo-iwp4qeiff62fy-6c8c7ff769-db6sg › mongo
+ mongo-p6ua5xnjo33vu-6c64d6d476-6fqw2 › mongo
failed to retry: context canceled
- mongo-dkna2yzxpcneq-67645c4fd-8g4gl › mongo
- mongo-iqc733ktshkvc-78f5795dd9-rsf49 › mongo
- mongo-iwp4qeiff62fy-6c8c7ff769-db6sg › mongo
- jaeger-soasl3vdz6bj4-j995d › jaeger
- mongo-p6ua5xnjo33vu-6c64d6d476-6fqw2 › mongo
Cannot list ReplicaSets with error: Get "https://aks-dns.hcp.northeurope.azmk8s.io:443/apis/apps/v1/namespaces/radius-system/replicasets?labelSelector=app.kubernetes.io%2Fname%3Ddashboard%2Capp.kubernetes.io%2Fpart-of%3Dradius": context canceled
- backendapi-58d766c8f9-2mbjd › backendapi
- frontend-754c68fd46-lw2l8 › frontend
Cannot list ReplicaSets with error: Get "https://aks-dns.hcp.northeurope.azmk8s.io:443/apis/apps/v1/namespaces/azure-radius/replicasets?labelSelector=radapp.io%2Fapplication%3Dradius": context canceled
- jaeger-soasl3vdz6bj4-jfvkt › jaeger
Error: stern reached the maximum number of log requests (10), use --max-log-requests to increase the limit
```

### Desired behavior

I would like to be able to run my application without `stern` preventing me.

### Workaround

I am using Recipes to create MongoDb and Jaeger. 
As a workaround, I can remove the label `radapp.io/application': context.application == null ? '' : context.application.name` from the created Deployment. This way the output won't be collected by `stern`, and the app will run. (But I won't see the logs.)

### rad Version

```
0.33.0    v0.33.0   0.33.0    6ffdc9980420c00f7af35b024dd0e62e598efef2
```

### Operating system

Windows 11 (Radius runs on AKS)

### Additional context

The `stern` tool allows specifying a configuration file, but adding it here `%APPDATA%/.config/stern/config.yaml` did not help.

[AB#12318](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12318)

ytimocin:
# Description

In our code, we set `MaxLogRequests` to 10: https://github.com/radius-project/radius/blob/main/pkg/cli/kubernetes/logstream/stern.go#L84.

In stern repository, I see that if this option is set to -1, based on another parameter it can be 5 or 50: https://github.com/stern/stern/blob/master/cmd/cmd.go#L287.

Here is the place where it checks for the number of requests and the limit: https://github.com/stern/stern/blob/master/stern/stern.go#L194.

We have 2 options:

1. Adding a parameter to let users set `MaxLogRequests`, or
2. Setting it to a higher number like 999 or something like that.

Would like to listen to your feedback on this topic.

cc/ @rynowak @kachawla @loekd

**_This PR is a WIP as of now (10 May 2024) and will be updated after the discussion._**

## Type of change
Fixes: #7565

### DIFF
--- a/pkg/cli/kubernetes/logstream/stern.go
+++ b/pkg/cli/kubernetes/logstream/stern.go
@@ -18,6 +18,7 @@ package logstream
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"text/template"
 	"time"
@@ -48,7 +49,6 @@ type Impl struct {
 // Stream() configures and runs Stern, a library for streaming logs from Kubernetes pods, with custom filters and output formats
 // based on the provided parameters. It returns an error if there is an issue configuring or running Stern.
 func (i *Impl) Stream(ctx context.Context, options Options) error {
-
 	// The functionality of the package is provided almost entirely be github.com/stern/stern.
 	// Under the covers, stern is watching pods based on a set of filters and then piping
 	// all of the matching logstreams to the writer.
@@ -81,7 +81,7 @@ func (i *Impl) Stream(ctx context.Context, options Options) error {
 		ErrOut:     options.Out,
 
 		// Limit the number of concurrent log fetch function unlike Kubernetes client rate limitter.
-		MaxLogRequests: 10,
+		MaxLogRequests: 50,
 	}
 
 	// This is the only Radius-specific customization we make.
@@ -99,7 +99,10 @@ func (i *Impl) Stream(ctx context.Context, options Options) error {
 	// This will block until the context is cancelled.
 	err = stern.Run(ctx, options.KubeClient, &cfg)
 	if err != nil {
-		return err
+		// Not returning the error and just logging is intentional!
+		// We don't want the process to exit if there's an error streaming logs.
+		// We just want to log the error and continue.
+		fmt.Println("Error running stern: ", err)
 	}
 
 	return nil


rynowak:
Can you do some testing around the behavior when the maximum is reached? I'm curious about things like:

- When the maximum is reached, so stern stop? or does it keep watching logs up to the max?
- If we handled the error coming back from stern and printed a message like "you have too many containers, so you might not get all of the logs" would that be possible?

You could test some of these scenarios by setting a lower maximum

ytimocin:
> Can you do some testing around the behavior when the maximum is reached? I'm curious about things like:
> 
> * When the maximum is reached, so stern stop? or does it keep watching logs up to the max?
> * If we handled the error coming back from stern and printed a message like "you have too many containers, so you might not get all of the logs" would that be possible?
> 
> You could test some of these scenarios by setting a lower maximum

I set the `MaxLogRequests` to 1 and deployed a bicep that has 4 containers, here is what happens:

image

I was able to log the error but continue the flow. In this approach, logging will stop because the Run method of **stern** will exit. You can see the Run method here: https://github.com/stern/stern/blob/master/stern/stern.go#L40.

image

I believe here we have two options:
1. Give the user the ability to pass in the **MaxLogRequests**,
2. Set the **MaxLogRequests** to 50 or some high number like that.

I say we go with the second approach because we can't add all the parameters that the libraries Radius is using are using.

Any ideas? cc/ @rynowak

rynowak:
I'm good with setting to 50. I wonder if we could also log the error coming from stern and not fail the command.

pkg/cli/kubernetes/logstream/stern.go
```diff
@@ -99,7 +99,10 @@ func (i *Impl) Stream(ctx context.Context, options Options) error {
 	// This will block until the context is cancelled.
 	err = stern.Run(ctx, options.KubeClient, &cfg)
 	if err != nil {
-		return err
+		// Not returning the error and just logging is intentional!
+		// We don't want the process to exit if there's an error streaming logs.
+		// We just want to log the error and continue.
+		fmt.Println("Error running stern: ", err)
```
rynowak, line 105: Thanks for adding this.

rynowak (approved):
Looks great! :shipit:

pkg/cli/kubernetes/logstream/stern.go
```diff
@@ -81,7 +81,7 @@ func (i *Impl) Stream(ctx context.Context, options Options) error {
 		ErrOut:     options.Out,
 
 		// Limit the number of concurrent log fetch function unlike Kubernetes client rate limitter.
-		MaxLogRequests: 10,
+		MaxLogRequests: 50,
```
kachawla, line 84: Should we update this comment with some context on user experience if an app exceeds this limit, and that this is randomly selected number?

pkg/cli/kubernetes/logstream/stern.go
kachawla, line 105: Do they provide a reliable error code that we can use to filter out and ignore only the limit exceeded error?

kachawla (approved):
Looks good with a couple of questions.

Thanks for investigating and fixing this!

pkg/cli/kubernetes/logstream/stern.go
ytimocin: I will create a follow-up PR to update the comment here. (Apologies for missing this comment.)

pkg/cli/kubernetes/logstream/stern.go
ytimocin: From what I see, they don't. I can check one more time and let you know.

PR #7581, opened 2024-05-10
Updating Purge AWS Test Resources workflow

ytimocin:
# Description
Fixing the Purge AWS Test Resources workflow. There was a reformatting issue that made the runs fail. Fixed that and also added a null check to the tags of the resources.

## Type of change
Fixes: #issue_number

### DIFF
--- a/.github/scripts/delete-aws-resources.sh
+++ b/.github/scripts/delete-aws-resources.sh
@@ -6,7 +6,7 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-#    
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


.github/scripts/delete-aws-resources.sh
```diff
@@ -41,24 +41,21 @@ function delete_old_aws_resources() {
...
-            echo "Deleting resource of type: $resource_type with identifier: $identifier"
-            echo "$identifier\n" >> $DELETED_RESOURCES_FILE
-            aws cloudcontrol delete-resource --type-name "$resource_type" --identifier "$identifier"
-          fi
-        done
+  for resource_type in ${RESOURCE_TYPES//,/ }; do
+    aws cloudcontrol list-resources --type-name "$resource_type" --query "ResourceDescriptions[].Identifier" --output text | tr '\t' '\n' | while read identifier; do
+      aws cloudcontrol get-resource --type-name "$resource_type" --identifier "$identifier" --query "ResourceDescription.Properties" --output text | while read resource; do
+        resource_tags=$(jq -c -r .Tags <<<"$resource")
+        if [[ "$resource_tags" != "null" && "$resource_tags" != "" ]]; then
```
ytimocin, line 48: Added this because of this error: `jq: error (at <stdin>:1): Cannot iterate over null (null)`.

.github/scripts/delete-aws-resources.sh
```diff
@@ -41,24 +41,21 @@ function delete_old_aws_resources() {
...
-        do
-          key=$(jq -r '.Key' <<< "$tag")
-          value=$(jq -r '.Value' <<< "$tag")
-          if [[ "$key" == "$LABEL" && $((CURRENT_TIME - value)) -gt $MAX_AGE]]
-          then
-            echo "Deleting resource of type: $resource_type with identifier: $identifier"
-            echo "$identifier\n" >> $DELETED_RESOURCES_FILE
-            aws cloudcontrol delete-resource --type-name "$resource_type" --identifier "$identifier"
-          fi
-        done
```
ytimocin, line 61: The reformat was off and that is why this workflow had almost always failed: https://github.com/radius-project/radius/actions/workflows/purge-aws-test-resources.yaml.

.github/scripts/delete-aws-resources.sh
```diff
@@ -72,28 +69,33 @@ function delete_old_aws_resources() {
...
+    # Increase the delay exponentially
+    RETRY_DELAY=$((RETRY_DELAY * 2))
+  fi
 done
 
 # Check if the maximum number of retries exceeded
 if [ $RETRY_COUNT -eq $MAX_RETRIES ]; then
-    echo "Maximum number of retries exceeded"
+  echo "Maximum number of retries exceeded"
+  exit 1
```
ytimocin, line 100: Exiting here because if we reach this step the workflow must fail, and we must investigate what happened.

.github/scripts/delete-aws-resources.sh
```diff
@@ -72,28 +69,33 @@ function delete_old_aws_resources() {
...
+  RETRY_COUNT=$((RETRY_COUNT + 1))
+
+  # Check if there are more retries left
+  if [ $RETRY_COUNT -lt $MAX_RETRIES ]; then
+    # Retry after delay
+    sleep $RETRY_DELAY
+    echo "Retrying in $RETRY_DELAY seconds..."
+
+    # Increase the delay exponentially
+    RETRY_DELAY=$((RETRY_DELAY * 2))
```
ytimocin, line 93: Moving the sleep before the retry so that we make sure to wait before each retry not after. This will make sure that the process sleeps after the first one and before the second one. So that we don't run into throttling issues on AWS side.

.github/scripts/delete-aws-resources.sh
```diff
@@ -72,28 +69,33 @@ function delete_old_aws_resources() {
...
+  fi
 done
 
 # Check if the maximum number of retries exceeded
 if [ $RETRY_COUNT -eq $MAX_RETRIES ]; then
```
nithyatsu, line 98: Is this If clause needed since it is implied when the while at line 71 is exitted?

.github/scripts/delete-aws-resources.sh
kachawla: I didn't understand this.. shouldn't we print `Retrying in $RETRY_DELAY seconds...` before executing sleep?

ytimocin: That is right. I will swap sleep and echo commands.

.github/scripts/delete-aws-resources.sh
ytimocin: There is an extra **break** case within the while loop: https://github.com/radius-project/radius/pull/7581/files#diff-5625fc3520f54a6e805114f672e280334dac76957f1f5d54cbf3b661f9efeb12R80. That is the success case and if we remove this if `if [ $RETRY_COUNT -eq $MAX_RETRIES ];` then the **echo** command below will always be called.

rynowak (approved):

PR #7583, opened 2024-05-10
add functional test results to artifacts

### LINKED ISSUE #6727: Include failure details in scheduled test failure reporting

Testing

We have an automated process for reporting functional test failures in our scheduled test runs. This creates an issue like: https://github.com/radius-project/radius/issues/6682 for each scheduled test run that fails. 

This is great, and the functionality that's there already is really valuable. The automation creates and issue and links to the failing workflow. It doesn't include any of the details or logs in the issue itself. 

This can cause problems because a re-run of the tests will overwrite the results for the failing workflow. Additionally workflows time-out and their data will be erased after a period of tiem. We also expect the on-call engineer to copy over the relevant logs and do an RCA in the issue. We could automate the part where the on-call engineer has to hunt through the logs.

The goal I'm proposing is that the issues created by our scheduled test runs are self-contained and can be investigated without needing additional data. 

The automation should include:

- The names of failing tests (if tests ran)
- The console logs of those failing tests
- The artifacts from the test run as a zip (includes the logs of all pods in Kubernetes)

### Proposed Fix (optional)

We have some existing actions/scripts to:

- Output a JSON results file from Go test (this is a Go-specific format and not widely used)
- Convert that JSON results file into JUnit format (supported by a wide array of tools)
- Parse the JUnit results file and add annotations to the test run

We use this for our UNIT TEST failure reporting. Not many people know about it because unit test failures are pretty rare in CI 😉.

We could extend this to build most of what I'm suggesting for our functional tests.

[AB#10297](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10297)

nithyatsu:
# Description
Add functional test results as an artifact so we can download it. 
Currently, rerun erases a functional test run's result and there is no easy way to download/process the results.

Sample artifact that is produced after adding these changes is in the summary of  [Functional tests · radius-project/radius@65bed3d (github.com)](https://github.com/radius-project/radius/actions/runs/9023474283) 

Name of the artifact is functional_test_results

## Type of change

Fixes: #6727

### DIFF
--- a/.github/actions/process-test-results/action.yaml
+++ b/.github/actions/process-test-results/action.yaml
@@ -56,6 +56,7 @@ runs:
       uses: EnricoMi/publish-unit-test-result-action@v2
       if: always()
       with:
+        check_name: ${{ inputs.test_group_name }}
         files: |
           ${{ inputs.result_directory }}/processed/*.xml
     - name: 'Upload ${{ inputs.test_group_name }} Results'
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -324,7 +324,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest]
-        name: [ucp,kubernetes,shared,msgrp,daprrp,samples,cli]
+        name: [shared,cli]
         include:
           # datastorerp functional tests need the larger VM.
           - os: ubuntu-latest-m
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -546,6 +546,8 @@ jobs:
         run: |
           # Ensure rad cli is in path before running tests.
           export PATH=$GITHUB_WORKSPACE/bin:$PATH
+          # Make directory to capture functional test results
+          mkdir -p ./dist/functional_test 
           cd $GITHUB_WORKSPACE
 
           which rad || { echo "cannot find rad"; exit 1; }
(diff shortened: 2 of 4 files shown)


.github/workflows/functional-test.yaml
```diff
@@ -575,6 +577,15 @@ jobs:
           BICEP_RECIPE_REGISTRY: ${{ env.BICEP_RECIPE_REGISTRY }}
           BICEP_RECIPE_TAG_VERSION: ${{ env.BICEP_RECIPE_TAG_VERSION }}
           GH_TOKEN: ${{ steps.get_installation_token.outputs.token }}
+          GOTESTSUM_OPTS: '--junitfile ./dist/functional_test/results.xml'
+      - name: Process Functional Test Results
+        uses: ./.github/actions/process-test-results
+        # In case of failure, upload functional_test_results to artifacts so that they are not erased by subsequent runs.
```
nithyatsu, line 583: If I make the condition always the results captured along with test duration could identify some tests that take long. Would this be useful or is it a better idea to make that change as and when we are troubleshooting?

rynowak: There's basically no cost to adding something to artifacts. 

If you do use `always()` then think about what happens when the files don't exist.

.github/workflows/functional-test.yaml
```diff
@@ -575,6 +577,15 @@ jobs:
...
+        if: failure() && github.repository == 'radius-project/radius'
+        with:
+          test_group_name: 'Functional Tests'
+          artifact_name: 'functional_test_results'
```
rynowak, line 587: These names need to be unique using the matrix name

nithyatsu: fixed

.github/actions/process-test-results/action.yaml
```diff
@@ -56,6 +56,7 @@ runs:
       uses: EnricoMi/publish-unit-test-result-action@v2
       if: always()
       with:
+        check_name: ${{ inputs.test_group_name }}
```
rynowak, line 59: I'll be interested to see what the comment looks like. Are the values of `test_group_name` reasonable? would they make sense if it shows up as a comment on a pr?

rynowak (approved):

ytimocin (commented):
I just triggered this action: https://github.com/radius-project/radius/actions/runs/9110085495. You can test it out before merging.

nithyatsu:
I have a sample run result at [functional_test_results-datastoresrp](https://github.com/radius-project/radius/actions/runs/9103197352/artifacts/1507092349). We have to manually add the workflow_dispatch trigger and remove a few things so that with the manual run we can simulate failure and generate the logs. We can see the name looks meaningful for the  artifact from the test run I linked. But I have not tested the issue creation and comment itself. I will monitor the PR once it is merged to make sure it is meaningful. I am also following up with another PR to enhance the description of issues we create for the failures. @ytimocin @rynowak

Issue #7520, opened 2024-04-18
rad deploy should be able to use environments in different groups to deploy application  


nithyatsu:
```rad deploy``` today does not allow this since it searches for environments in current scope only.

```
nithya@Nithyas-MacBook-Pro bug7052 % rad env list
RESOURCE    TYPE                            GROUP     STATE
a           Applications.Core/environments  default   Succeeded
default     Applications.Core/environments  default   Succeeded
demo        Applications.Core/environments  default   Succeeded
production  Applications.Core/environments  default   Succeeded

nithya@Nithyas-MacBook-Pro bug7052 % rad group switch b

nithya@Nithyas-MacBook-Pro bug7052 % rad env list
RESOURCE  TYPE      GROUP     STATE
nithya@Nithyas-MacBook-Pro bug7052 % rad deploy ctnr.bicep --parameters magpieimage=ghcr.io/radius-project/magpiego:latest --group b --environment production
The environment "production" does not exist in scope "/planes/radius/local/resourceGroups/b". Run `rad env create` first.
```

[AB#11797](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11797)

PR #7594, opened 2024-05-14
Update CLI client code and add tests

rynowak:
# Description

This change is a major update to the CLI's client code for interacting with the Radius API.

The client grew organically over time as we added features. Since it's client code it wasn't well tested because testing it would require a lot of mocking. Unfortunately in some places we *do* have complex logic in this code and unfortunately we also have some bugs.

I'm working on another fix to address #7520, and encountered a lot of limitations with the client code so I decided to fix it.

This change addresses the following:

- Reviewed the API and made updates for consistency.
- Added tests for ALL of the functions on the client.
- Updated each API to accept either a resource ID or a resource name (needed for #7520).

## Type of change


pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
 // ApplicationsManagementClient is used to interface with management features like listing resources by app, show details of a resource.
 type ApplicationsManagementClient interface {
-	ListAllResourcesByType(ctx context.Context, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesOfTypeInApplication(ctx context.Context, applicationName string, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesByApplication(ctx context.Context, applicationName string) ([]generated.GenericResource, error)
-	ListAllResourcesOfTypeInEnvironment(ctx context.Context, environmentName string, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesByEnvironment(ctx context.Context, environmentName string) ([]generated.GenericResource, error)
-	ShowResource(ctx context.Context, resourceType string, resourceName string) (generated.GenericResource, error)
-	DeleteResource(ctx context.Context, resourceType string, resourceName string) (bool, error)
+	// ListResourcesOfType lists all resources of a given type in the configured scope.
```
rynowak, line 140: This is where I did a major pass for consistency of the APIs.

pkg/cli/clients/management.go
```diff
@@ -42,8 +42,13 @@ import (
...
 type UCPApplicationsManagementClient struct {
-	RootScope     string
-	ClientOptions *arm.ClientOptions
+	RootScope                        string
+	ClientOptions                    *arm.ClientOptions
+	genericResourceClientFactory     func(scope string, resourceType string) (genericResourceClient, error)
+	applicationResourceClientFactory func(scope string) (applicationResourceClient, error)
+	environmentResourceClientFactory func(scope string) (environmentResourceClient, error)
+	resourceGroupClientFactory       func() (resourceGroupClient, error)
+	capture                          func(ctx context.Context, capture **http.Response) context.Context
```
rynowak, line 51: Please review this file in detail. Each of these functions have been updated to be more consistent (naming, structure, etc). 

Adding mocks was pretty complex.

pkg/cli/clients/management.go
```diff
@@ -480,170 +443,301 @@ func (amc *UCPApplicationsManagementClient) ListEnvironmentsAll(ctx context.Cont
...
+	return false
+}
+
+func (amc *UCPApplicationsManagementClient) captureResponse(ctx context.Context, capture **http.Response) context.Context {
+	if capture == nil {
+		return runtime.WithCaptureResponse(ctx, nil)
 	}
 
-	return corerpv20231001.RecipeGetMetadataResponse(resp.RecipeGetMetadataResponse), nil
+	return amc.capture(ctx, capture)
```
rynowak, line 742: Mocking the behavior of `runtime.WithCaptureResponse` is ugllllly....

pkg/cli/clients/management_mocks.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/radius-project/radius/pkg/cli/clients_new/generated"
+	corerpv20231001 "github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
+	ucpv20231001 "github.com/radius-project/radius/pkg/ucp/api/v20231001preview"
+)
+
+// We define interfaces so we can mock the interactions with the Radius API. These
```
rynowak, line 28: This is everything that was required to mock the client code 😓

pkg/cli/cmd/app/graph/graph_test.go
```diff
@@ -50,7 +50,7 @@ func Test_Validate(t *testing.T) {
 			},
 			ConfigureMocks: func(mocks radcli.ValidateMocks) {
 				mocks.ApplicationManagementClient.EXPECT().
-					ShowApplication(gomock.Any(), "test-app").
+					GetApplication(gomock.Any(), "test-app").
```
rynowak, line 53: The changes to the client API resulted in changes to most commands, but they are just renames.

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
-	ShowApplication(ctx context.Context, applicationName string) (corerp.ApplicationResource, error)
-	GetGraph(ctx context.Context, applicationName string) (corerp.ApplicationGraphResponse, error)
 
-	// CreateOrUpdateApplication creates or updates an application.
-	CreateOrUpdateApplication(ctx context.Context, applicationName string, resource corerp.ApplicationResource) error
+	// GetApplication retrieves an application by its name (or id).
+	GetApplication(ctx context.Context, applicationNameOrID string) (corerp.ApplicationResource, error)
+
+	// GetApplicationGraph retrieves the application graph of an application by its name (or id).
+	GetApplicationGraph(ctx context.Context, applicationNameOrID string) (corerp.ApplicationGraphResponse, error)
```
nithyatsu, line 168: nice to see we can use ID or name everywhere.

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
 type ApplicationsManagementClient interface {
-	ListAllResourcesByType(ctx context.Context, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesOfTypeInApplication(ctx context.Context, applicationName string, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesByApplication(ctx context.Context, applicationName string) ([]generated.GenericResource, error)
-	ListAllResourcesOfTypeInEnvironment(ctx context.Context, environmentName string, resourceType string) ([]generated.GenericResource, error)
-	ListAllResourcesByEnvironment(ctx context.Context, environmentName string) ([]generated.GenericResource, error)
-	ShowResource(ctx context.Context, resourceType string, resourceName string) (generated.GenericResource, error)
-	DeleteResource(ctx context.Context, resourceType string, resourceName string) (bool, error)
+	// ListResourcesOfType lists all resources of a given type in the configured scope.
+	ListResourcesOfType(ctx context.Context, resourceType string) ([]generated.GenericResource, error)
```
nithyatsu, line 141: Did we decide to remove the ability to have a single call that fetches all resources in an application?

nithyatsu: never mind, I just realized its still there.

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
+
+	// ListApplications lists all applications in the configured scope.
 	ListApplications(ctx context.Context) ([]corerp.ApplicationResource, error)
-	ShowApplication(ctx context.Context, applicationName string) (corerp.ApplicationResource, error)
-	GetGraph(ctx context.Context, applicationName string) (corerp.ApplicationGraphResponse, error)
 
-	// CreateOrUpdateApplication creates or updates an application.
-	CreateOrUpdateApplication(ctx context.Context, applicationName string, resource corerp.ApplicationResource) error
+	// GetApplication retrieves an application by its name (or id).
+	GetApplication(ctx context.Context, applicationNameOrID string) (corerp.ApplicationResource, error)
```
nithyatsu, line 165: I am a bit confused with the "NameOrID" changes to variable name. I dont see addition of logic to detecting whether we passed in a name or ID for application or environment or elsewhere. Am I missing those changes?

pkg/cli/cmd/workspace/create/create.go
```diff
@@ -177,7 +177,7 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-		_, err := client.ShowUCPGroup(cmd.Context(), "local", group)
+		_, err := client.GetResourceGroup(cmd.Context(), "local", group)
```
nithyatsu, line 180: Is it better to call this GetRadiusResourceGroup or add a comment due to confusing terminology with az resource group?

rynowak: I think it's much better to just say `ResouceGroup`. The context makes it clear that is about the Radius feature.

pkg/cli/clients/clients.go
rynowak: You should see that any function that accepts a `nameOrId` is calling `extractScope`. That's where the magic happens.

rynowak:
Looks like there's another bug for me to fix that's blocking functional tests. This is the downside of mock-based testing, you could have bugs in the code that's mocked out.

pkg/cli/clients/management.go
```diff
@@ -64,138 +69,148 @@ var (
...
+			return nil, err
 		}
-		applicationList := nextPage.GenericResourcesList.Value
-		for _, application := range applicationList {
-			results = append(results, *application)
+
+		for _, resource := range page.Value {
+			results = append(results, *resource)
 		}
 	}
```
nithyatsu, line 90: Were we doing this in unnecessarily complicated way before? I am not sure why we changed this to page.Value from page.GenericResourcesList.Value

pkg/cli/clients/management.go
```diff
@@ -480,170 +443,301 @@ func (amc *UCPApplicationsManagementClient) ListEnvironmentsAll(ctx context.Cont
...
-		for _, resourceGroup := range resourceGroupList {
-			resourceGroupResources = append(resourceGroupResources, *resourceGroup)
+		return id.RootScope(), id.Name(), nil
+	}
+
+	// Treat this as a resource name
+	return amc.RootScope, nameOrID, nil
+}
 
+func (amc *UCPApplicationsManagementClient) fullyQualifyID(nameOrID string, resourceType string) (string, error) {
```
nithyatsu, line 684: nice 😊

pkg/cli/clients/clients.go
nithyatsu: hmm, does this make -g flag somewhat redundant? that now would make sense only if the "name" was used, not id right?

pkg/cli/clients/management.go
rynowak: It's the same. `page.Value` and `page.GenericResourcesList.Value` are actually the same field.

What's happening here is that the `GenericResourcesList` is embedded. 


```go
type GenericResourcesListResponse struct {
    GenericResourcesList
}

type GenericResourcesList struct {
    Value []GenericResource
}
```

rynowak: When two things are equivalent, prefer less code 😆

pkg/cli/clients/clients.go
rynowak: Not really. We'd still need to update the rest of the CLI to pass in a resource ID instead of a name. Right now the CLI only works with names and assumes the "default scope". 

So this change only really improves this specific layer.

rynowak:
FYI - I found https://github.com/radius-project/radius/issues/7597 while troubleshooting this code and added a workaround for it.

nithyatsu (dismissed):

pkg/cli/clients/management.go
```diff
@@ -205,136 +220,112 @@ func (amc *UCPApplicationsManagementClient) DeleteResource(ctx context.Context,
...
 	if err != nil {
 		return err
 	}
 
-	_, err = client.CreateOrUpdate(ctx, applicationName, resource, nil)
+	// See: https://github.com/radius-project/radius/issues/7597
+	//
+	// This is a workaround because the server can return invalid system data, which
+	// fails to roundtrip when the client does a "GET -> modify -> PUT".
+	resource.SystemData = nil
```
rynowak, line 326: This is a new bug that was exposed while making this change. Fortunately the workaround for it pretty isolated.

Existing functionality in the client code was already impacted by this bug, but we weren't aware of it. With the changes I made to the environment code, it hit the bug in our functional tests!

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
+	CreateOrUpdateEnvironment(ctx context.Context, environmentNameOrID string, resource *corerp.EnvironmentResource) error
+
+	// DeleteEnvironment deletes an environment and all of its resources by its name (in the configured scope) or resource ID.
+	DeleteEnvironment(ctx context.Context, environmentNameOrID string) (bool, error)
+
+	// ListResourceGroups lists all resource groups in the configured scope.
+	ListResourceGroups(ctx context.Context, planeName string) ([]ucp_v20231001preview.ResourceGroupResource, error)
+
+	// GetResourceGroup retrieves a resource group by its name.
+	GetResourceGroup(ctx context.Context, planeName string, resourceGroupName string) (ucp_v20231001preview.ResourceGroupResource, error)
```
vinayada1, line 201: Are the resourcegroup APIs not with NameOrID for a reason?

rynowak: I thought about this decided against. Right now they take the `planeName` and the `resourceGroupName` which means that they can always build the resource id fully. So taking in a `resourceGroupNameOrId` would make the plane name redundant.

rynowak: Does this sound right? any better ideas?

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
+	ListResourcesInApplication(ctx context.Context, applicationNameOrID string) ([]generated.GenericResource, error)
+
+	// ListResourcesInEnvironment lists all resources in a given environment in the configured scope.
+	ListResourcesInEnvironment(ctx context.Context, environmentNameOrID string) ([]generated.GenericResource, error)
+
+	// GetResource retrieves a resource by its type and name (or id).
+	GetResource(ctx context.Context, resourceType string, resourceNameOrID string) (generated.GenericResource, error)
+
+	// DeleteResource deletes a resource by its type and name (or id).
+	DeleteResource(ctx context.Context, resourceType string, resourceNameOrID string) (bool, error)
```
vinayada1, line 159: There are no tests to test by ID

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
+	ListResourcesOfTypeInApplication(ctx context.Context, applicationNameOrID string, resourceType string) ([]generated.GenericResource, error)
+
+	// ListResourcesOfTypeInEnvironment lists all resources of a given type in a given environment in the configured scope.
+	ListResourcesOfTypeInEnvironment(ctx context.Context, environmentNameOrID string, resourceType string) ([]generated.GenericResource, error)
+
+	// ListResourcesInApplication lists all resources in a given application in the configured scope.
+	ListResourcesInApplication(ctx context.Context, applicationNameOrID string) ([]generated.GenericResource, error)
+
+	// ListResourcesInEnvironment lists all resources in a given environment in the configured scope.
+	ListResourcesInEnvironment(ctx context.Context, environmentNameOrID string) ([]generated.GenericResource, error)
```
vinayada1, line 153: There is no test for this API

pkg/cli/clients/clients.go
```diff
@@ -137,40 +137,74 @@ type LogStream struct {
...
+
+	// GetApplicationGraph retrieves the application graph of an application by its name (or id).
+	GetApplicationGraph(ctx context.Context, applicationNameOrID string) (corerp.ApplicationGraphResponse, error)
+
+	// CreateOrUpdateApplication creates or updates an application by its name (or id).
+	CreateOrUpdateApplication(ctx context.Context, applicationNameOrID string, resource *corerp.ApplicationResource) error
 
 	// CreateApplicationIfNotFound creates an application if it does not exist.
-	CreateApplicationIfNotFound(ctx context.Context, applicationName string, resource corerp.ApplicationResource) error
+	CreateApplicationIfNotFound(ctx context.Context, applicationNameOrID string, resource *corerp.ApplicationResource) error
```
vinayada1, line 174: Do we need this API since we already have a Create or update?

vinayada1: Or does this need to be renamed to CreateDefaultApplication?

rynowak: I'd like to keep this, it's used by a few commands, and it's nice for testability.

I'd rather keep the name, I think it's already clear and I didn't change it in this PR.

pkg/cli/clients/clients.go
rynowak: I'll double check this. I don't plan to add a "name test" + "id test" for every API. The difference between the name case and the ID case are handled in common code.

pkg/cli/clients/management.go
```diff
@@ -363,111 +355,94 @@ func (amc *UCPApplicationsManagementClient) CreateApplicationIfNotFound(ctx cont
...
 	}
 
-	return envResourceList, nil
+	return environments, nil
 }
 
 // ListEnvironmentsAll queries the scope for all environment resources and returns a slice of environment resources or an error if one occurs.
 func (amc *UCPApplicationsManagementClient) ListEnvironmentsAll(ctx context.Context) ([]corerpv20231001.EnvironmentResource, error) {
-	scope, err := resources.ParseScope("/" + amc.RootScope)
+	scope, err := resources.ParseScope(amc.RootScope)
```
ytimocin, line 445: Why did we remove `"/"`? Was this unnecessary?

ytimocin (approved):
Thanks for the change and all the unit tests.

pkg/cli/clients/management.go
rynowak: The generated clients don't like it when the scope has a leading slash. We used to initialize the client without the leading slash (`planes/radius/local/...` instead of `/planes/radius/local/...`)

I moved the code that trims the leading slash to the place where the generated clients are constructed for clarity. It's weird to have the "root scope" in a field without it being in the correct format.

pkg/cli/clients/clients.go
rynowak: I looked again and there's a good mix of tests that use names and tests that use IDs. It feels like a lot of wasted effort to test both cases for each of the 15 functions. They all call into the same code to resolve the full id. 

Please let me know if there's something specific you want to see added.

rynowak:
I'm going to merge this. @vinayada1 - if you have additional feedback for me to address I'll do it in a follow-up 👍

kachawla (approved):
👏

pkg/cli/clients/clients.go
vinayada1: This is fine too but I feel it would be better if all APIs accept similar input formats.

rynowak: What do you think the *right* signature is?

An API like `GetEnvironment(ctx, nameOrId)` works because the client has a notion of "default scope" that can be prepended to the name.

What do you think the behavior of `GetResourceGroup(ctx, nameOrId)` should be?

PR #7620, opened 2024-05-21
Update Terraform recipe download error to include module version

kachawla:
# Description

Update Terraform recipe download error to include module version + minor refactoring (removed unnecessary wrapper function over tf.get).

Here's the error that's returned today - 
`
message": "failed to run terraform get to download the module from source \"kachawla/test/aws\": exit status 1\n\nError: Unresolvable module version constraint\n\n  on main.tf.json line 4, in module:\n   4:     \"default\": {\n\nThere is no available version of module\n\"registry.terraform.io/kachawla/test/aws\" (main.tf.json:4) which matches the\ngiven version constraint. The newest available version is 1.2.0.\n\n"
            }
 `

## Type of change

### DIFF
--- a/pkg/recipes/terraform/module.go
+++ b/pkg/recipes/terraform/module.go
@@ -20,12 +20,16 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	getter "github.com/hashicorp/go-getter"
 	"github.com/hashicorp/terraform-config-inspect/tfconfig"
 	"github.com/hashicorp/terraform-exec/tfexec"
+	"github.com/radius-project/radius/pkg/metrics"
 	"github.com/radius-project/radius/pkg/recipes"
 	"github.com/radius-project/radius/pkg/recipes/recipecontext"
+	"github.com/radius-project/radius/pkg/recipes/util"
+	"github.com/radius-project/radius/pkg/ucp/ucplog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/recipes/terraform/module.go
```diff
@@ -49,6 +53,38 @@ type moduleInspectResult struct {
 	// Any other module information required in the future can be added here.
 }
 
+// downloadAndInspect handles downloading the TF module and retrieving the necessary information
+func downloadAndInspect(ctx context.Context, tf *tfexec.Terraform, options Options) (*moduleInspectResult, error) {
```
kachawla, line 57: This is copied over from https://github.com/radius-project/radius/blob/main/pkg/recipes/terraform/execute.go#L308, with main change being directly calling `tf.Get` instead of another function `downloadModule` calling `tf.Get`.

ytimocin: Why did we need to move this function to a new location?

pkg/recipes/terraform/module.go
```diff
@@ -99,14 +135,3 @@ func inspectModule(workingDir string, recipe *recipes.EnvironmentDefinition) (*m
 
 	return result, nil
 }
-
-// downloadModule downloads the module to the workingDir from the module source specified in the Terraform configuration.
-// It uses Terraform's Get command to download the module using the Terraform executable available at execPath.
-// An error is returned if the module could not be downloaded.
-func downloadModule(ctx context.Context, tf *tfexec.Terraform, templatePath string) error {
-	if err := tf.Get(ctx); err != nil {
-		return fmt.Errorf("failed to run terraform get to download the module from source %q: %w", templatePath, err)
```
ytimocin, line 108: What if we just remove the `fmt.Errorf`? Wouldn't it undo the error wrapping?

pkg/recipes/terraform/module.go
kachawla: I don't know how this ended up being in execute as the code evolved, but same reason why inspect module is in `module.go`. Abstracting out downloading and reading module into a separate file for maintainability. Let me know if you have any concerns that I might have missed.

pkg/recipes/terraform/module.go
kachawla: Not sure if I follow what you mean, can you elaborate a bit?

The rationale for removing this function was (if that's what the question was about), the only thing this function `downloadModule` was doing was one line for calling another function `tf.Get`, so it didn't make sense to keep this additional function around.

ytimocin: I see what you mean now. That makes sense... We also got rid of `fmt.Errorf("failed to run terraform get to download the module from source %q: %w", templatePath, err)` which is the wrapping of the error, right?

ytimocin (approved):

lakshmimsft (approved):

pkg/recipes/terraform/module.go
kachawla: partially.. replaced by this string formatting https://github.com/radius-project/radius/blob/b19dbc20a8df988ec5582dafa9ce3b4df4ac38c8/pkg/recipes/terraform/module.go#L69. We need to add more context since Terraform error doesn't include all the information we need like the module version used, which operation did it fail for etc

PR #7622, opened 2024-05-21
Update purge workflow with manual trigger

sk593:
# Description

The purge test workflow is failing with authentication issues. The documentation says to use GITHUB_TOKEN with package permissions but we need to be able to test this first. This PR will update the manual trigger so we can add permissions in a followup PR

https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#maintaining-access-to-packages-in-github-actions-workflows
https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius #7621 .

Fixes: #issue_number


ytimocin (commented):
If you open this PR from the Radius repo instead of your fork and if this workflow has workflow dispatch, you can test your changes.

ytimocin:
You should also sign your commit.

ytimocin (dismissed):

.github/workflows/purge-test-resources.yaml
```diff
@@ -19,8 +19,11 @@ name: Purge test resources
 permissions:
   id-token: write # Required for requesting the JWT
   contents: read # Required for actions/checkout
+  packages: write # Required for updating packages
```
nithyatsu, line 22: looks good

nithyatsu:
Since you are on-call you might have permissions to push to radius-project directly, instead of to your fork. Then you can run the workflow on that branch to test it out.
image

nithyatsu (dismissed):

ytimocin (dismissed):

rynowak (dismissed):
:shipit: Looks great! Thanks

.github/workflows/purge-test-resources.yaml
```diff
@@ -21,6 +21,8 @@ permissions:
   contents: read # Required for actions/checkout
 
 on:
+  # Enable manual trigger
+  workflow_dispatch:
```
rynowak, line 25: FYI is this is going to create an issue on failure when run manually. If you want to avoid that, you could change the condition [here](https://github.com/radius-project/radius/pull/7622/files#diff-9270fafc31ef4b50d35d1295e9f99f96013853c295c29c92c387492f62cd570dL102) to skip the issue for a manual run.

sk593: Good catch, will update to skip on a manual run

nithyatsu (dismissed):

.github/workflows/purge-test-resources.yaml
```diff
@@ -101,9 +103,9 @@ jobs:
           done
   create_issue_on_failure:
     name: Create issue for failing purge test resources run
-    needs: [purge_gchr_dev, purge_azure_resources]
+    needs: [purge_ghcr_dev, purge_azure_resources]
     runs-on: ubuntu-latest
-    if: always() && (needs.purge_gchr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
+    if: ${{ github.event_name == 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
```
rynowak, line 108: don't we want to check `github.event_name == 'schedule'`?

This logic is like: `is workflow dispatch AND ( purge GHCR failed OR purge Azure failed)`

sk593: This should be `!= workflow_dispatch`. Updated that

rynowak (approved):

.github/workflows/purge-test-resources.yaml
```diff
@@ -101,9 +103,9 @@ jobs:
           done
   create_issue_on_failure:
     name: Create issue for failing purge test resources run
-    needs: [purge_gchr_dev, purge_azure_resources]
+    needs: [purge_ghcr_dev, purge_azure_resources]
     runs-on: ubuntu-latest
-    if: always() && (needs.purge_gchr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
+    if: always() && ${{ github.event_name != 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
```
ytimocin, line 108: 👍

.github/workflows/purge-test-resources.yaml
```diff
@@ -33,7 +35,7 @@ env:
   ACTION_LINK: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
 jobs:
-  purge_gchr_dev:
+  purge_ghcr_dev:
```
ytimocin, line 38: 🕵️

ytimocin (approved):

PR #7629, opened 2024-05-22
Update kubebuilder to the latest working version

sk593:
# Description

We have a dependency on the controller-runtime package that is failing during our build tests. Reverting the package to the last known working version instead of latest. 

Package: https://github.com/kubernetes-sigs/controller-runtime
Error: 
```
/bin/sh: 1: /home/runner/go/bin/setup-envtest: not found
=> Installing Kubebuilder test tools...
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240522175850-2e9781e9fc60
go: downloading github.com/spf13/afero v1.6.0
go: downloading go.uber.org/zap v1.26.0
go: downloading sigs.k8s.io/yaml v1.3.0
go: downloading golang.org/x/text v0.14.0
go: downloading go.uber.org/multierr v1.10.0
=> Instructions:
=> Set environment variable KUBEBUILDER_ASSETS for tests.
=> KUBEBUILDER_ASSETS=""
unable to find a version that was supported for platform linux/amd64
KUBEBUILDER_ASSETS="" CGO_ENABLED=1 gotestsum --junitfile ./dist/unit_test/results.xml -- -v ./pkg/... -race -coverprofile ./dist/unit_test/ut_coverage.out
```

## Type of change

Fixes: #issue_number

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -49,7 +49,7 @@ test: test-get-envtools ## Runs unit tests, excluding kubernetes controller test
 .PHONY: test-get-envtools
 test-get-envtools:
 	@echo "$(ARROW) Installing Kubebuilder test tools..."
-	$(call go-install-tool,$(ENV_SETUP),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
+	$(call go-install-tool,$(ENV_SETUP),sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240521074430-fbb7d370bebc)
 	@echo "$(ARROW) Instructions:"
 	@echo "$(ARROW) Set environment variable KUBEBUILDER_ASSETS for tests."
 	@echo "$(ARROW) KUBEBUILDER_ASSETS=\"$(shell $(ENV_SETUP) use -p path ${K8S_VERSION} --arch amd64)\""


build/test.mk
```diff
@@ -49,7 +49,7 @@ test: test-get-envtools ## Runs unit tests, excluding kubernetes controller test
 .PHONY: test-get-envtools
 test-get-envtools:
 	@echo "$(ARROW) Installing Kubebuilder test tools..."
-	$(call go-install-tool,$(ENV_SETUP),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
+	$(call go-install-tool,$(ENV_SETUP),sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240521074430-fbb7d370bebc)
```
rynowak, line 52: Can you create an issue for the following (if you didn't already):

We should either:

- Revert this in the future and use `latest` when they fix the issue.
- Move this into a place where depdendabot can manage it for it. 

We don't want this to be hardcoded forever.

rynowak (approved):
:shipit:

build/test.mk
sk593: Sure, just opened a bug here: https://github.com/radius-project/radius/issues/7630. I also opened one in the controller-runtime repo that I'll monitor for responses: https://github.com/kubernetes-sigs/controller-runtime/issues/2838

rynowak: Looks like #7630 is tracking the build-break. We should also have a plan to remove the workaround.

kachawla (approved):

build/test.mk
sk593: I considered them parts of the same problem but I updated the issue to explicitly call out removing the workaround

PR #7636, opened 2024-05-27
Update autorest related dependencies

ytimocin:
# Description
Update autorest related dependencies

## Type of change
Fixes: #issue_number


.github/workflows/lint.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
-      GOVER: '1.22.2'
```
rynowak, line 36: Is there a specific formatter you're using for this?

IMO if we're going to spend time on minor formatting changes like this, it should be automated.

.github/workflows/lint.yaml
```diff
@@ -46,31 +46,31 @@ jobs:
...
         with:
-          node-version: '18'
+          node-version: "18"
       - name: Install TypeSpec compiler
         run: |
           pushd typespec
           npm ci
           popd
       - name: Install autorest
-        run: npm install -g autorest@3.6.1
```
rynowak, line 56: Are there ways we could get dependabot to manage this? I know dependabot understands things like `go.mod`, can we teach it to understand versions in other kinds of files?

It would be nice if we could make some improvements that ensure we have **no** hardcoded versions (everything managed by dependabot).

rynowak (approved):
:shipit: - code changes look good. Thanks for doing this @ytimocin. I left some suggestions for future ideas, but non-blocking for this PR.

.github/workflows/lint.yaml
ytimocin: I will have a separate PR later (hopefully in this sprint) and probably a pipeline step to do the formatting for us.

I use prettier on VS Code by the way.

.github/workflows/lint.yaml
ytimocin: I can check this but, as far as I know, **dependabot** doesn't have something like you mentioned above. I will check again, and, if there is, I will create a PR to add that.

rynowak: One idea would be to move this to a `project.json` so Dependabot can manage it for us. Then from the pipeline we'd call `npm install` or `npx`

rynowak: No need to do that in this PR.

PR #7640, opened 2024-05-28
Add Azure Workload Identity support

willdavsmith:
# Description

* Add Azure Workload Identity support to the Azure provider

Associated changes in DE: https://github.com/radius-project/deployment-engine/pull/372

Note: I expect the functional tests to fail because I am updating them in this PR. I will run them manually to confirm that the changes work.

## Type of change

Issue Reference: https://github.com/radius-project/radius/issues/7308


pkg/azure/armauth/auth.go
```diff
@@ -101,8 +104,3 @@ func GetAuthMethod() string {
 		return CliAuth
 	}
 }
-
-// IsServicePrincipalConfigured checks if ServicePrincipalAuth is the authentication method configured.
-func IsServicePrincipalConfigured() (bool, error) {
```
willdavsmith, line 106: this was unused

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/radius-project/radius/pkg/cli/framework"
+	"github.com/radius-project/radius/pkg/cli/output"
+	"github.com/radius-project/radius/pkg/cli/workspaces"
+	"github.com/radius-project/radius/pkg/to"
+	ucp "github.com/radius-project/radius/pkg/ucp/api/v20231001preview"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCommand creates an instance of the command and runner for the `rad credential create azure` command.
```
ytimocin, line 39:
```suggestion
// NewCommand creates an instance of the command and runner for the `rad credential create azure wi` command.
```

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
```diff
@@ -0,0 +1,182 @@
...
+Radius Environments, and Radius portable resources.
+
+Radius will use the provided subscription and resource group as the default target scope for Bicep deployment.
+The provided service principal must have the Contributor or Owner role assigned for the provided resource group
+in order to create or manage resources contained in the group. The resource group should be created before
+calling 'rad credential register azure'.
+` + common.LongDescriptionBlurb,
+		Example: `
+# Register (Add or update) cloud provider credential for Azure with service principal authentication
+rad credential register azure sp --client-id <client id/app id> --client-secret <client secret/password> --tenant-id <tenant id>
```
ytimocin, line 65:
```suggestion
rad credential register azure wi --client-id <client id/app id> --client-secret <client secret/password> --tenant-id <tenant id>
```

pkg/cli/cmd/credential/register/azure/wi/workloadidentity_test.go
```diff
@@ -0,0 +1,233 @@
...
+					Default: "a",
+					Items: map[string]workspaces.Workspace{
+						"a": {
+							Connection: map[string]any{
+								"kind":    workspaces.KindKubernetes,
+								"context": "my-context",
+							},
+							Source: workspaces.SourceUserConfig,
+
+							// Will have provider info added
```
ytimocin, line 116: Is this to be addressed in a different PR?

pkg/cli/cmd/credential/register/register.go
```diff
@@ -32,17 +32,19 @@ func NewCommand(factory framework.Factory) *cobra.Command {
...
-rad credential register azure --client-id <client id> --client-secret <client secret> --tenant-id <tenant id> 	
+rad credential register azure sp --client-id <client id> --client-secret <client secret> --tenant-id <tenant id>
+# Register (Add or update) cloud provider credential for Azure with workload identity authentication
+rad credential register azure wi --client-id <client id> --tenant-id <tenant id>
 # Register (Add or Update) cloud provider credential for AWS with IAM authentication
 rad credential register aws --access-key-id <access-key-id> --secret-access-key <secret-access-key>	
 `,
 	}
 
-	azure, _ := credential_register_azure.NewCommand(factory)
```
ytimocin, line 45: Why was the second output removed?

pkg/cli/cmd/radinit/environment.go
```diff
@@ -96,8 +96,12 @@ func (r *Runner) CreateEnvironment(ctx context.Context) error {
 	}
 
 	if r.Options.CloudProviders.Azure != nil {
-		credential := r.getAzureCredential()
-		err := credentialClient.PutAzure(ctx, credential)
+		credential, err := r.getAzureCredential()
+		if err != nil {
+			return clierrors.MessageWithCause(err, "Failed to configure Azure credentials.")
```
ytimocin, line 101: Should this be `Failed to get Azure credentials`? It looks like it is the same as the error message below (after **PutAzure**).

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,56 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
 		}
 
 		return nil, err
 	}
 
-	if credentials == nil || credentials.ClientID == "" || credentials.TenantID == "" || credentials.ClientSecret == "" {
+	switch credentials.Kind {
+	case ucp_datamodel.AzureServicePrincipalCredentialKind:
+		azureServicePrincipal := credentials.ServicePrincipal
+		if azureServicePrincipal == nil || azureServicePrincipal.ClientID == "" || azureServicePrincipal.TenantID == "" || azureServicePrincipal.ClientSecret == "" {
```
ytimocin, line 133:
```suggestion
		if credentials.ServicePrincipal == nil || credentials.ServicePrincipal.ClientID == "" || credentials.ServicePrincipal.TenantID == "" || credentials.ServicePrincipal.ClientSecret == "" {
```

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,56 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
+			logger.Info("Azure service principal credentials are not registered, skipping credentials configuration.")
+			return nil, nil
+		}
+
+		return credentials, nil
+	case ucp_datamodel.AzureWorkloadIdentityCredentialKind:
+		azureWorkloadIdentity := credentials.WorkloadIdentity
+		if azureWorkloadIdentity == nil || azureWorkloadIdentity.ClientID == "" || azureWorkloadIdentity.TenantID == "" {
```
ytimocin, line 141:
```suggestion
		if credentials.WorkloadIdentity == nil || credentials.WorkloadIdentity.ClientID == "" || credentials.WorkloadIdentity.TenantID == "" {
```

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,56 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
 		configMap[azureSubIDParam] = subscriptionID
 	}
 
-	if credentials != nil && credentials.ClientID != "" && credentials.TenantID != "" && credentials.ClientSecret != "" {
-		configMap[azureClientIDParam] = credentials.ClientID
-		configMap[azureClientSecretParam] = credentials.ClientSecret
-		configMap[azureTenantIDParam] = credentials.TenantID
+	switch credentials.Kind {
+	case ucp_datamodel.AzureServicePrincipalCredentialKind:
+		azureServicePrincipal := credentials.ServicePrincipal
```
ytimocin, line 161: Can be in-lined as mentioned above.

pkg/ucp/credentials/azure.go
```diff
@@ -69,10 +69,17 @@ func (p *AzureCredentialProvider) Fetch(ctx context.Context, planeName, name str
 		case *ucpapi.InternalCredentialStorageProperties:
 			storage = c
 		default:
-			return nil, errors.New("invalid AzureServicePrincipalProperties")
+			return nil, errors.New("invalid InternalCredentialStorageProperties")
```
ytimocin, line 72: Is this error customer facing?

pkg/ucp/credentials/azure.go
```diff
@@ -69,10 +69,17 @@ func (p *AzureCredentialProvider) Fetch(ctx context.Context, planeName, name str
...
+		}
+	case *ucpapi.AzureWorkloadIdentityProperties:
+		switch c := p.Storage.(type) {
+		case *ucpapi.InternalCredentialStorageProperties:
+			storage = c
+		default:
+			return nil, errors.New("invalid InternalCredentialStorageProperties")
```
ytimocin, line 79: Is this error customer facing?

pkg/ucp/credentials/azure.go
```diff
@@ -69,10 +69,17 @@ func (p *AzureCredentialProvider) Fetch(ctx context.Context, planeName, name str
...
 		}
 	default:
-		return nil, errors.New("invalid InternalCredentialStorageProperties")
+		return nil, errors.New("invalid AzureCredentialProperties")
```
ytimocin, line 82: Is this error customer facing?

swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AzureCredential_CreateOrUpdate_WI.json
ytimocin, line 1: Should the name be `AzureCredential_WorkloadIdentity_CreateOrUpdate`? Seems like it could be more readable.

typespec/UCP/azure-credentials.tsp
```diff
@@ -76,9 +78,9 @@ model AzureCredentialProperties {
   provisioningState?: ProvisioningState;
 }
 
-@doc("The properties of Service Principal credential storage")
+@doc("The properties of Azure Service Principal credential storage")
 model AzureServicePrincipalProperties extends AzureCredentialProperties {
-  @doc("Service principal kind")
+  @doc("Workload Identity kind")
   kind: AzureCredentialKind.ServicePrincipal;
```
ytimocin, line 84: Doc and Kind don't match.

kachawla (commented):
On `I expect the functional tests to fail because I am updating them in this PR.` -- is there a way we can break this into two step update? If using service principal/client id instead of workload identity is still supported, then I would expect functional tests to still work..

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +103,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
...
-	// Do not instantiate new client unless the secret is rotated.
-	if c.credential != nil && c.credential.ClientSecret == s.ClientSecret &&
-		c.credential.ClientID == s.ClientID && c.credential.TenantID == s.TenantID {
 		c.refreshExpiry()
 		return nil
-	}
+	case datamodel.AzureWorkloadIdentityCredentialKind:
+		azureWorkloadIdentityCredential := s.WorkloadIdentity
+		if azureWorkloadIdentityCredential.ClientID == "" || azureWorkloadIdentityCredential.TenantID == "" {
+			return errors.New("invalid azure workload identity credential info")
```
kachawla, line 146: nit: I'd suggest providing an explicit error message here

```suggestion
			return errors.New("empty clientID or tenantID provided for Azure workload identity")
```

pkg/azure/armauth/auth.go
```diff
@@ -76,6 +77,8 @@ func NewARMCredential(opt *Options) (azcore.TokenCredential, error) {
 		return azidentity.NewEnvironmentCredential(nil)
 	case ManagedIdentityAuth:
 		return azidentity.NewManagedIdentityCredential(nil)
+	case WorkloadIdentityAuth:
+		return azidentity.NewDefaultAzureCredential(nil)
```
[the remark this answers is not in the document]
willdavsmith: I don't think we actually need this code. It's only called for testing afaik. but in the case we ever need it, it's the right code to have

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +103,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
...
+			c.credential.ServicePrincipal.TenantID == azureServicePrincipalCredential.TenantID {
+			c.refreshExpiry()
+			return nil
+		}
+
+		logger.Info("Retreived Azure Credential - ClientID: " + azureServicePrincipalCredential.ClientID)
+
+		// Rotate credentials by creating new ClientSecretCredential.
+		var opt *azidentity.ClientSecretCredentialOptions
+		if c.options.ClientOptions != nil {
```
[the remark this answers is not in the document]
willdavsmith: I left the code the same as today. so it seems to be okay if it's nil: https://github.com/radius-project/radius/blob/main/pkg/azure/credential/ucpcredentials.go#L126

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +103,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
...
+			opt = &azidentity.ClientSecretCredentialOptions{
+				ClientOptions: *c.options.ClientOptions,
+			}
+		}
+
+		azCred, err := azidentity.NewClientSecretCredential(azureServicePrincipalCredential.TenantID, azureServicePrincipalCredential.ClientID, azureServicePrincipalCredential.ClientSecret, opt)
```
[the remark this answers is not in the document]
willdavsmith: I didn't change this code from before, see above: https://github.com/radius-project/radius/blob/main/pkg/azure/credential/ucpcredentials.go#L126

pkg/cli/cmd/credential/register/azure/wi/workloadidentity_test.go
willdavsmith: I copied the code from the existing `rad credential register azure` code, so it looks like this hasn't been added yet.

pkg/cli/cmd/credential/register/register.go
willdavsmith: since this now calls into `rad credential register azure` (which is a noop command), we don't have a runner. it exists in the subcommands `rad credential register azure sp` and `rad credential register wi` now.

pkg/cli/cmd/radinit/environment.go
willdavsmith: `getAzureCredential` just returns the `AzureCredentialResource` so I think the error message with the error attached it more correct than `Failed to get Azure credentials`.

```go
func (r *Runner) getAzureCredential() (ucp.AzureCredentialResource, error) {
	switch r.Options.CloudProviders.Azure.CredentialKind {
	case azure.AzureCredentialKindServicePrincipal:
		return ucp.AzureCredentialResource{
			Location: to.Ptr(v1.LocationGlobal),
			Type:     to.Ptr(cli_credential.AzureCredential),
			Properties: &ucp.AzureServicePrincipalProperties{
				Storage: &ucp.CredentialStorageProperties{
					Kind: to.Ptr(ucp.CredentialStorageKindInternal),
				},
				TenantID:     &r.Options.CloudProviders.Azure.ServicePrincipal.TenantID,
				ClientID:     &r.Options.CloudProviders.Azure.ServicePrincipal.ClientID,
				ClientSecret: &r.Options.CloudProviders.Azure.ServicePrincipal.ClientSecret,
			},
		}, nil
	case azure.AzureCredentialKindWorkloadIdentity:
		return ucp.AzureCredentialResource{
			Location: to.Ptr(v1.LocationGlobal),
			Type:     to.Ptr(cli_credential.AzureCredential),
			Properties: &ucp.AzureWorkloadIdentityProperties{
				Storage: &ucp.CredentialStorageProperties{
					Kind: to.Ptr(ucp.CredentialStorageKindInternal),
				},
				TenantID: &r.Options.CloudProviders.Azure.WorkloadIdentity.TenantID,
				ClientID: &r.Options.CloudProviders.Azure.WorkloadIdentity.ClientID,
			},
		}, nil
	default:
		return ucp.AzureCredentialResource{}, fmt.Errorf("unsupported Azure credential kind: %s", r.Options.CloudProviders.Azure.CredentialKind)
	}
}
```

pkg/ucp/credentials/azure.go
willdavsmith: maybe, it looks like it can be returned from the credential API call

pkg/ucp/credentials/azure.go
willdavsmith: maybe, it looks like it can be returned from the credential API call

pkg/ucp/credentials/azure.go
willdavsmith: see above

deploy/Chart/templates/ucp/deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
         control-plane: ucp
         app.kubernetes.io/name: ucp
         app.kubernetes.io/part-of: radius
+        {{- if eq .Values.global.azureWorkloadIdentity.enabled true }}
+        azure.workload.identity/use: "true"
```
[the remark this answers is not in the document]
willdavsmith: I'm not certain but I don't think the controller needs to talk to azure. I will ask ryan

willdavsmith:
> On `I expect the functional tests to fail because I am updating them in this PR.` -- is there a way we can break this into two step update? If using service principal/client id instead of workload identity is still supported, then I would expect functional tests to still work..

It's a very minor breaking change that unfortunately breaks the functional tests. `rad credential register azure` has been renamed to `rad credential register azure sp` per the design review, so unfortunately this isn't backwards compatible. We could potentially separate this change out or we could leave it as is and run the tests manually. What do you think is best?

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,62 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
 	credentials, err := azureCredentialsProvider.Fetch(ctx, credentials.AzureCloud, "default")
 	if err != nil {
 		if errors.Is(err, &secret.ErrNotFound{}) {
-			logger.Info("AWS credentials are not registered, skipping credentials configuration.")
+			logger.Info("Azure credentials are not registered, skipping credentials configuration.")
```
willdavsmith, line 123: this was a typo I think

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,62 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
 			return nil, nil
 		}
 
 		return nil, err
 	}
 
-	if credentials == nil || credentials.ClientID == "" || credentials.TenantID == "" || credentials.ClientSecret == "" {
+	switch credentials.Kind {
```
willdavsmith, line 130: @kachawla please review this file. not sure if I did this correctly.

kachawla:
> > On `I expect the functional tests to fail because I am updating them in this PR.` -- is there a way we can break this into two step update? If using service principal/client id instead of workload identity is still supported, then I would expect functional tests to still work..
> 
> It's a very minor breaking change that unfortunately breaks the functional tests. `rad credential register azure` has been renamed to `rad credential register azure sp` per the design review, so unfortunately this isn't backwards compatible. We could potentially separate this change out or we could leave it as is and run the tests manually. What do you think is best?

Ah nvm, expected workflow failure with manually testing sounds right for this scenario. Thanks for the clarification.

pkg/recipes/terraform/config/providers/azure.go
kachawla: sure, didn't get a chance to finish looking at everything, will look into this file today

deploy/Chart/values.yaml
```diff
@@ -32,7 +35,7 @@ controller:
       memory: "300Mi"
 
 de:
-  image: ghcr.io/radius-project/deployment-engine
+  image: ghcr.io/willdavsmith/radius/deployment-engine
```
kachawla, line 38: `ghcr.io/willdavsmith/`

was this for local testing?

deploy/Chart/values.yaml
```diff
@@ -21,6 +21,9 @@ global:
   #   url: "http://jaeger-collector.radius-monitoring.svc.cluster.local:9411/api/v2/spans"
   #
 
+  azureWorkloadIdentity:
```
kachawla, line 24: Would it make sense to add a comment here that it is disabled by default, hence the config value?

pkg/azure/armauth/auth.go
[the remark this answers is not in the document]
kachawla: Would a link to this doc help clarifying how this maps workload identity https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet#azure-identity-client-libraries?

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +103,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
...
-	c.credential = s
+		c.tokenCred = azCred
+		c.credential = s
 
-	c.refreshExpiry()
-	return nil
+		c.refreshExpiry()
+		return nil
+	default:
+		return nil
```
kachawla, line 178: Do we ever expect this to be hit? If not, then I think this should error out

pkg/cli/azure/provider.go
```diff
@@ -16,20 +16,33 @@ limitations under the License.
...
+// AzureCredentialKind - Azure credential kinds supported.
+type AzureCredentialKind string
+
 const (
 	// ProviderDisplayName is the text used in display for Azure.
-	ProviderDisplayName = "Azure"
+	ProviderDisplayName                                     = "Azure"
+	AzureCredentialKindWorkloadIdentity AzureCredentialKind = "WorkloadIdentity"
+	AzureCredentialKindServicePrincipal AzureCredentialKind = "ServicePrincipal"
 )
```
kachawla, line 27: Where is managed identity auth used https://github.com/radius-project/radius/blob/a6e749d185db8785e8a0710d8ba155da1d72ebed/pkg/azure/armauth/auth.go#L34?

pkg/cli/cmd/credential/register/azure/azure.go
```diff
@@ -14,176 +14,35 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package azure
+package register
```
kachawla, line 17: Shouldn't this be `azure` since this is still under the azure package?

pkg/cli/cmd/credential/register/azure/azure.go
```diff
@@ -14,176 +14,35 @@ See the License for the specific language governing permissions and
...
 
-// NewCommand creates an instance of the command and runner for the `rad credential create azure` command.
-//
-
-// NewCommand creates a new cobra command for registering an Azure cloud provider credential for a Radius installation,
-// which requires a service principal with the Contributor or Owner role assigned to the provided resource group.
-func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
-	runner := NewRunner(factory)
-
+func NewCommand(factory framework.Factory) *cobra.Command {
```
kachawla, line 27: Let's add function level comment

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return nil
+}
+
+// Run runs the `rad credential register azure wi` command.
+//
+
+// Run registers a credential for the Azure cloud provider in the Radius installation, updates the server-side
+// to add/change credentials. It returns an error if any of the steps fail.
+func (r *Runner) Run(ctx context.Context) error {
```
kachawla, line 150: Everything except credential properties is common between this and service principal, can we abstract that out?

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,62 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
+		if credentials.WorkloadIdentity == nil ||
+			credentials.WorkloadIdentity.ClientID == "" ||
+			credentials.WorkloadIdentity.TenantID == "" {
+			logger.Info("Azure workload identity credentials are not registered, skipping credentials configuration.")
+			return nil, nil
+		}
+
+		return credentials, nil
+	default:
 		logger.Info("Azure credentials are not registered, skipping credentials configuration.")
```
kachawla, line 151: Let's make the log message explicit here that provided credential kind is not supported, although I'm not sure what would cause this to happen.

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +120,62 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
+			configMap[azureClientIDParam] = credentials.ServicePrincipal.ClientID
+			configMap[azureClientSecretParam] = credentials.ServicePrincipal.ClientSecret
+			configMap[azureTenantIDParam] = credentials.ServicePrincipal.TenantID
+		}
+	case ucp_datamodel.AzureWorkloadIdentityCredentialKind:
+		if credentials.WorkloadIdentity != nil &&
+			credentials.WorkloadIdentity.ClientID != "" &&
+			credentials.WorkloadIdentity.TenantID != "" {
+			configMap[azureClientIDParam] = credentials.WorkloadIdentity.ClientID
+			configMap[azureTenantIDParam] = credentials.WorkloadIdentity.TenantID
```
kachawla, line 177: So looks like for workload identity a few additional properties are expected by terraform - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/aks_workload_identity#configuring-terraform-to-use-an-aks-workload-identity

```
use_aks_workload_identity = true
use_cli                   = false
```

kachawla (commented):
Few comments on what I've reviewed so far

pkg/cli/cmd/credential/register/azure/wi/workloadidentity_test.go
kachawla: @willdavsmith I know this is a copy of existing code.. but while you are at it can you move this comment into an issue, if this is still something in our plan? Leaving the comment here isn't very helpful.

pkg/cli/cmd/credential/show/objectformats.go
```diff
@@ -19,32 +19,70 @@ package show
...
-func credentialFormat(credentialType string) output.FormatterOptions {
+func credentialFormat(credentialType string, providers credential.ProviderCredentialConfiguration) output.FormatterOptions {
 	if strings.EqualFold(credentialType, "azure") {
-		return output.FormatterOptions{
-			Columns: []output.Column{
-				{
-					Heading:  "NAME",
-					JSONPath: "{ .Name }",
+		switch *providers.AzureCredentials.Kind {
+		case AzureServicePrincipalCredentialKind:
```
kachawla, line 36: I'd suggest returning an error for unexpected kind

pkg/cli/cmd/credential/show/objectformats.go
```diff
@@ -19,32 +19,70 @@ package show
 import (
 	"strings"
 
+	"github.com/radius-project/radius/pkg/cli/credential"
 	"github.com/radius-project/radius/pkg/cli/output"
 )
 
+const (
+	AzureServicePrincipalCredentialKind = "ServicePrincipal"
+	AzureWorkloadIdentityCredentialKind = "WorkloadIdentity"
```
kachawla, line 28: These are defined in multiple files, can they be consolidated?

pkg/cli/azure/provider.go
willdavsmith: I don't think we use this anymore, looks like it was added 3 years ago.

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
willdavsmith: I started to abstract it out and it ended up being more code and more complicated... For example I had to write an abstraction for the Runner on each of the credential types. Currently we're repeating code but I think it's not too bad - do you have suggestions or do you think it's okay to keep the repetition?

pkg/cli/cmd/credential/show/objectformats.go
willdavsmith: unified to be the ucp datamodel value

pkg/cli/cmd/credential/register/azure/wi/workloadidentity_test.go
willdavsmith: Talked to ryan - looks like this test was out of date. provider info should not be added to the workspace. I have cleaned up this test

pkg/cli/cmd/radinit/azure_test.go
```diff
@@ -82,6 +86,58 @@ func Test_enterAzureCloudProvider(t *testing.T) {
...
+	subscription := azure.Subscription{
+		Name: "test-subscription",
+		ID:   "test-subscription-id",
+	}
+
+	resourceGroup := armresources.ResourceGroup{
+		Name: to.Ptr("test-resource-group"),
+	}
+
+	// selectAzureSubscription
```
ytimocin, line 105: Is there a reason why the comments are camel case?

pkg/cli/cmd/radinit/cloud_test.go
```diff
@@ -29,16 +29,27 @@ import (
...
 			ClientID:     "test-client-id",
 			ClientSecret: "test-client-secret",
 			TenantID:     "test-tenant-id",
 		},
 	}
 
+	azureProviderWorkloadIdentity := azure.Provider{
+		SubscriptionID: "test-subscription-id",
+		ResourceGroup:  "test-resource-group",
+		CredentialKind: "WorkloadIdentity",
```
ytimocin, line 46: Can't we get these from the constant we created above (in a different file)?

pkg/cli/credential/azure_credential_management.go
```diff
@@ -45,39 +45,36 @@ type AzureCredentialManagementClientInterface interface {
...
-type CloudProviderStatus struct {
-	// Name is the name/kind of the provider. For right now this only supports Azure and AWS.
-	Name string
+	// Service Principal properties
+	ServicePrincipal *AzureServicePrincipalCredentialProperties
 
-	// Enabled is the enabled/disabled status of the provider.
-	Enabled bool
+	// Workload Identity properties
+	WorkloadIdentity *AzureWorkloadIdentityCredentialProperties
```
ytimocin, line 56: All property docs should start with the property name

pkg/recipes/terraform/config/providers/azure.go
```diff
@@ -119,30 +122,67 @@ func fetchAzureCredentials(ctx context.Context, azureCredentialsProvider credent
...
+
+			// Use AKS Workload Identity for Azure provider
+			// https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/aks_workload_identity#configuring-with-environment-variables
+			configMap[azureUseAKSWorkloadIdentityParam] = true
+			configMap[azureUseCLIParam] = false
```
ytimocin, line 184: Could these be set by the user? Like can we do `configMap[azureUseAKSWorkloadIdentityParam] = credentials.WorkloadIdentity.useAKSWI` or something like that?

pkg/ucp/credentials/azure.go
ytimocin: Don't we need to add better error messages here? I know this was always the case but should we udate these messages?

swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AzureCredential_List_WI.json
ytimocin, line 1: Are we also going to rename these files like the one below?

ytimocin (commented):
Looks good so far. Left a few more comments. Thanks for this big change.

pkg/cli/cmd/radinit/azure_test.go
willdavsmith: looks like they aren't very useful - removed

pkg/recipes/terraform/config/providers/azure.go
willdavsmith: I think I'm a bit confused - currently this will auto-detect which credentials are registered (WI or SP) and set the correct values in the provider based on that. I'm not sure what benefit adding this option would have, can you explain more?

deploy/Chart/templates/ucp/deployment.yaml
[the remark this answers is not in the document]
rynowak: Controller does not need to talk to Azure 👍

docs/contributing/contributing-code/contributing-code-tests/testing-local.md
```diff
@@ -55,7 +55,7 @@ The above steps will not configure the ability for Radius to talk with azure res
     go run ./cmd/rad/main.go env create radius-rg --namespace default
     go run ./cmd/rad/main.go env switch radius-rg
     go run ./cmd/rad/main.go env update radius-rg --azure-subscription-id <subscriptionId> --azure-resource-group <resourcegroupName>
-    go run ./cmd/rad/main.go credential register azure --client-id <appId> --client-secret <pwd> --tenant-id <tenantId>
+    go run ./cmd/rad/main.go credential register azure sp --client-id <appId> --client-secret <pwd> --tenant-id <tenantId>
```
rynowak, line 58:
```suggestion
    go run ./cmd/rad/main.go credential register azure sp --client-id <appId> --client-secret <clientSecret> --tenant-id <tenantId>
```

pkg/azure/armauth/auth.go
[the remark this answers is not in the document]
rynowak: Would be good to double-check when this code gets used... I think this might only be called when running Radius locally in the debugger. 

It might be good to think about whether we can remove it ...

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +104,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
 		return err
 	}
 
-	if s.ClientID == "" || s.ClientSecret == "" || s.TenantID == "" {
-		return errors.New("invalid azure service principal credential info")
-	}
+	switch s.Kind {
+	case datamodel.AzureServicePrincipalCredentialKind:
+		azureServicePrincipalCredential := s.ServicePrincipal
```
rynowak, line 109: Worth pulling this out into another function for readability?

pkg/azure/credential/ucpcredentials.go
```diff
@@ -102,48 +104,91 @@ func (c *UCPCredential) refreshCredentials(ctx context.Context) error {
...
+		c.credential = s
 
-	// Do not instantiate new client unless the secret is rotated.
-	if c.credential != nil && c.credential.ClientSecret == s.ClientSecret &&
-		c.credential.ClientID == s.ClientID && c.credential.TenantID == s.TenantID {
 		c.refreshExpiry()
 		return nil
-	}
+	case datamodel.AzureWorkloadIdentityCredentialKind:
+		azureWorkloadIdentityCredential := s.WorkloadIdentity
```
rynowak, line 145: same, might be worth pulling this into another function for readability

ytimocin (dismissed):
LGTM. Let's try this out. 🚀

ytimocin (dismissed):

ytimocin (dismissed):

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal.go
```diff
@@ -0,0 +1,191 @@
...
+
+	r.ClientID = clientID
+	r.ClientSecret = clientSecret
+	r.TenantID = tenantID
+
+	kubeContext, ok := r.Workspace.KubernetesContext()
+	if !ok {
+		return clierrors.Message("A Kubernetes connection is required.")
+	}
+	r.KubeContext = kubeContext
```
rynowak, line 150: Do you know why the kubecontext is required here? it looks to be unused.

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal.go
```diff
@@ -0,0 +1,191 @@
...
+	commonflags.AddOutputFlag(cmd)
+	commonflags.AddWorkspaceFlag(cmd)
+
+	cmd.Flags().String("client-id", "", "The client id or app id of an Azure service principal.")
+	_ = cmd.MarkFlagRequired("client-id")
+
+	cmd.Flags().String("client-secret", "", "The client secret or password of an Azure service principal.")
+	_ = cmd.MarkFlagRequired("client-secret")
+
+	cmd.Flags().String("tenant-id", "", "The tenant id of an Azure service principal.")
```
rynowak, line 80: Minor suggestion - use `cmd.Flags().StringVar(&r.TenantID, ...)`. Then you don't need to call `cmd.Flags().GetString("tenant-id")`. 

This is a simpler way to write the same code. Most of our CLI code was written before we knew about this feature.

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal_test.go
```diff
@@ -104,15 +101,6 @@ func Test_Validate(t *testing.T) {
...
 		},
-		{
-			Name: "Azure command without subscription",
-			Input: []string{
-				"--client-id", "abcd",
-				"--client-secret", "efgh",
-			},
-			ExpectedValid: false,
-			ConfigHolder:  framework.ConfigHolder{Config: configWithWorkspace},
-		},
```
rynowak, line 115: Looks like a bad copy-paste? and this was redundant with L99?

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
```diff
@@ -0,0 +1,180 @@
...
+	ucp "github.com/radius-project/radius/pkg/ucp/api/v20231001preview"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCommand creates an instance of the command and runner for the `rad credential create azure wi` command.
+//
+
+// NewCommand creates a new cobra command for registering an Azure cloud provider credential for a Radius installation,
+// which requires a service principal with the Contributor or Owner role assigned to the provided resource group.
```
rynowak, line 43: Looks like a very cool AI-generated comment 👍

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
rynowak: IMO: 

> prefer duplication over the wrong abstraction

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	r.ClientID = clientID
+	r.TenantID = tenantID
+
+	kubeContext, ok := r.Workspace.KubernetesContext()
+	if !ok {
+		return clierrors.Message("A Kubernetes connection is required.")
+	}
+	r.KubeContext = kubeContext
```
rynowak, line 140: Do we need the kubecontext? looks unused

pkg/cli/cmd/credential/show/objectformats.go
```diff
@@ -17,34 +17,71 @@ limitations under the License.
...
 
+	"github.com/radius-project/radius/pkg/cli/credential"
 	"github.com/radius-project/radius/pkg/cli/output"
+	"github.com/radius-project/radius/pkg/ucp/datamodel"
 )
 
 // credentialFormat function returns a FormatterOptions struct based on the credentialType parameter, which can
 // be either "azure" or "aws".
-func credentialFormat(credentialType string) output.FormatterOptions {
+func credentialFormat(credentialType string, providers credential.ProviderCredentialConfiguration) (output.FormatterOptions, error) {
```
rynowak, line 30: Is this function actually helping? It looks like we have to pass in the credential type, and now the credential type so we can switch between these four different options. It's essentially just accessing static data.  

Happy to for it to stay as-is or change it, do whatever seems most sensible.

pkg/cli/cmd/radinit/azure.go
```diff
@@ -57,38 +65,80 @@ func (r *Runner) enterAzureCloudProvider(ctx context.Context) (*azure.Provider,
...
+		tenantID, err := r.Prompter.GetTextInput(enterAzureWorkloadIdentityTenantIDPrompt, prompt.TextInputOptions{
+			Placeholder: enterAzureWorkloadIdentityTenantIDPlaceholder,
+			Validate:    prompt.ValidateUUIDv4,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		// Set the value for the Helm chart
+		options.SetValues = append(options.SetValues, "global.azureWorkloadIdentity.enabled=true")
```
rynowak, line 128: 👍

pkg/cli/credential/azure_credential_management.go
```diff
@@ -45,39 +45,36 @@ type AzureCredentialManagementClientInterface interface {
 	Delete(ctx context.Context, name string) (bool, error)
 }
 
-const (
-	AzureCredential     = "azure"
-	AzurePlaneName      = "azurecloud"
-	azureCredentialKind = "ServicePrincipal"
-)
+type AzureCredentialProperties struct {
```
rynowak, line 48: Add docs if these are used outside the package. Otherwise, make them non-exported

pkg/recipes/terraform/config/providers/azure.go
rynowak: Same. I'm not sure what this has to do with users.

We allow users to set whatever configuration they want for TF providers. We also fill it in for them for AWS and Azure.

pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go
```diff
@@ -79,15 +80,30 @@ func (c *CreateOrUpdateAzureCredential) Run(ctx context.Context, w http.Response
...
+		err = secret.SaveSecret(ctx, c.secretClient, secretName, newResource.Properties.AzureCredential)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return armrpc_rest.NewBadRequestResponse("Invalid Credential Kind"), nil
 	}
 
-	// Do not save the secret in metadata store.
-	newResource.Properties.AzureCredential.ClientSecret = ""
```
rynowak, line 89: Do we need to keep this code that blanks out the client secret? I think this was here to prevent the client secret from getting saved to the DB in the service principal case.

rynowak (commented):
This is looking really good overall. Left you some minor feedback and questions. No real concerns from me. 

This is a big feature 👍 really impressive.

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal.go
willdavsmith: looks like it was unused. removed

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal.go
willdavsmith: ooh snazzy. updated

pkg/cli/cmd/credential/register/azure/sp/serviceprincipal_test.go
willdavsmith: I think so

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
willdavsmith: 🤖

willdavsmith: I removed some of the extra fluff. comment should make more sense now

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
willdavsmith: removed

pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go
willdavsmith: I think so, I just moved it into the service principal switch because workload identity doesn't need it: https://github.com/radius-project/radius/blob/ff78c0ac5154a5b46063341bc1513ff9e902829c/pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go#L93

ytimocin (approved):

pkg/cli/cmd/credential/register/azure/wi/workloadidentity.go
kachawla: @willdavsmith that's fair. I'm fine leaving it as is.

kachawla (approved):
Thanks for addressing all the comments, looks good.

Since functional tests aren't updated to use workload identity, could you manually test that Terraform config works as expected after this gets merged? [Here's](https://github.com/radius-project/radius/blob/86b43976fa254ee64dca47133d0787d26b467779/test/functional-portable/corerp/cloud/resources/recipe_terraform_test.go#L48) a functional test that deploys to Azure through Terraform. Feel free to ping me if you need another test module.

PR #7643, opened 2024-05-30
Adding EnableWebsockets property to the gateway

### LINKED ISSUE #7590: Gateway does not properly support SignalR
### Steps to reproduce

- Create a container that exposes a SignalR hub.
- Create a (SPA) frontend container that hosts a web page which connects to it.
- Create a Gateway that exposes the SignalR hub and the Frontend.
Example yaml:

```
resource gateway 'Applications.Core/gateways@2023-10-01-preview' = {
  name: 'gateway'
  properties: {
    application: application 
    environment: environment
    routes: [
      {
        path: '/api' //REST API
        destination: 'http://${api.name}:${api.properties.container.ports.web.containerPort}'
      }
      {
        path: '/dispatchhub' //SignalR Hub
        destination: 'http://${api.name}:${api.properties.container.ports.web.containerPort}'
      }      
      {
        path: '/' //frontend index.html
        destination: 'http://${frontend.name}:${frontend.properties.container.ports.web.containerPort}'
      }
    ]
  }
}
```

### Observed behavior

- After 15 seconds, the server returns HTTP Status Code 504.
  - 504 Gateway Timeout (from service worker)
- The body will be: `upstream request timeout`
- The server Response header will be: `envoy`
- SignalR connection is lost

image
image

### Desired behavior

- I'd like SignalR to keep working indefinitely

### Workaround

- Connecting directly to the application container works, but is not great from a CORS perspective.

### rad Version

```
0.33.0    v0.33.0   0.33.0    6ffdc9980420c00f7af35b024dd0e62e598efef2
```

### Operating system

Windows 11 x64

### Additional context

- Envoy does support [wss](https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/websocket)
- Contour does so as [well](https://projectcontour.io/docs/1.24/config/websockets/), patching the `HTTPProxy` helps, but is not the way to go imo

[AB#12363](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12363)

ytimocin:
# Description
Based on the issue created, web sockets are not supported in our Gateway resource as of now. We added **EnableWebsockets** property to the Gateway Route Resource so that it can be set by Radius users. By default, it is set to false.

## Type of change
Fixes: #7590


pkg/corerp/renderers/gateway/render.go
```diff
@@ -37,6 +37,9 @@ import (
 	resources_kubernetes "github.com/radius-project/radius/pkg/ucp/resources/kubernetes"
 )
 
+const ErrSecretStoreNotFound = "secretStore resource %s not found"
```
rynowak, line 40: nit: does this need to be exported from the package? looks like it's only used by the unit tests.

Also, I think it would be better to avoid the prefix `Err`. Usually that's used when a variable is a static `error` value, in this case it's a format string.

pkg/corerp/renderers/gateway/render.go
```diff
@@ -366,6 +369,7 @@ func MakeRoutesHTTPProxies(ctx context.Context, options renderers.RenderOptions,
 							},
 						},
 						PathRewritePolicy: pathRewritePolicy,
+						EnableWebsockets:  true,
```
rynowak, line 372: Are there any downsides to this? what do the contour docs say?

ytimocin: Here is the link to the documentation: https://projectcontour.io/docs/main/config/websockets/#websockets.

Now that I think about this again, I think we will need to add a property to our Gateway resource so that setting **enableWebsockets** to true all the time doesn't affect other routes that don't need **websockets** enabled.

typespec/Applications.Core/gateways.tsp
```diff
@@ -107,6 +107,9 @@ model GatewayRoute {
 
   @doc("Optionally update the prefix when sending the request to the service. Ex - replacePrefix: '/' and path: '/myservice' will transform '/myservice/myroute' to '/myroute'")
   replacePrefix?: string;
+
+  @doc("Enables websocket support for the route.")
```
kachawla, line 111: Does it automatically default to false or does it need to be explicitly set like [this](https://github.com/radius-project/radius/blob/1c40cf3be4d12374740337bbd6f41c6cb9d59d03/typespec/Applications.Core/gateways.tsp#L53)?

```suggestion
  @doc("Enables websocket support for the route. Defaults to false.")
```

kachawla (commented):
I think the description still reflects the previous approach.. can you please update it with a link to the gist perhaps?

pkg/corerp/renderers/gateway/render_test.go
```diff
@@ -1499,6 +1499,7 @@ func validateContourHTTPRoute(t *testing.T, outputResources []rpv1.OutputResourc
 					},
 				},
 				PathRewritePolicy: expectedRewrite,
+				EnableWebsockets:  false,
```
kachawla, line 1502: Should we also add a unit test for scenario where `route.EnableWebsockets` is `true`?

ytimocin: I added a test by refactoring this file a little bit. We may need a better refactor in future.

typespec/Applications.Core/gateways.tsp
```diff
@@ -107,6 +107,9 @@ model GatewayRoute {
 
   @doc("Optionally update the prefix when sending the request to the service. Ex - replacePrefix: '/' and path: '/myservice' will transform '/myservice/myroute' to '/myroute'")
   replacePrefix?: string;
+
+  @doc("Enables websocket support for the route. Defaults to false.")
+  enableWebsockets?: boolean;
```
kachawla, line 112: Does it automatically default to false or does it need to be explicitly set like [this](https://github.com/radius-project/radius/blob/1c40cf3be4d12374740337bbd6f41c6cb9d59d03/typespec/Applications.Core/gateways.tsp#L53)?

kachawla (approved):

typespec/Applications.Core/gateways.tsp
ytimocin: Discussed offline. Created this follow-up issue: https://github.com/radius-project/radius/issues/7659.

PR #7646, opened 2024-05-30
Update purge workflow_dispatch logic

### LINKED ISSUE #7645: Purge test resources failed - Run ID: 9302711732
## Bug information 

This bug is generated automatically if the purge test resources workflow fails. For the further investigation, please visit [here](https://github.com/radius-project/radius/actions/runs/9302711732).

[AB#12424](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12424)

sk593:
# Description

The check for creating issues was creating issues even on successes. Tested this with manual runs and they're not creating issues 

## Type of change

Fixes: #7645

### DIFF
--- a/.github/workflows/purge-test-resources.yaml
+++ b/.github/workflows/purge-test-resources.yaml
@@ -105,7 +105,7 @@ jobs:
     name: Create issue for failing purge test resources run
     needs: [purge_ghcr_dev, purge_azure_resources]
     runs-on: ubuntu-latest
-    if: always() && ${{ github.event_name != 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
+    if: ${{ github.event_name != 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
     steps:
       - uses: actions/github-script@v7
         with:
--- a/.github/workflows/purge-test-resources.yaml
+++ b/.github/workflows/purge-test-resources.yaml
@@ -105,7 +105,7 @@ jobs:
     name: Create issue for failing purge test resources run
     needs: [purge_ghcr_dev, purge_azure_resources]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/purge-test-resources.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
     name: Create issue for failing purge test resources run
     needs: [purge_ghcr_dev, purge_azure_resources]
     runs-on: ubuntu-latest
-    if: always() && ${{ github.event_name != 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
+    if: always() && (${{ github.event_name != 'workflow_dispatch' }} && needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure')
```
ytimocin, line 108: I think `always()` needs to be removed here. Could you check the docs for the `always()` keyword?

sk593: I don't think `always()` was the issue since it was creating issues only on failures before. I think it was just the syntax of the conditional that was causing issues. These are the docs for `always()`: https://docs.github.com/en/actions/learn-github-actions/expressions#always. So it's used when we want this to run even if a job is cancelled

ytimocin: Could you explain to me the difference between the conditions here? I might be missing something here.

sk593: I think this should have been `always() && (${{ github.event_name != 'workflow_dispatch' }} && (needs.purge_ghcr_dev.result == 'failure' || needs.purge_azure_resources.result == 'failure'))`. But I did some more testing and it looks like this conditional might be causing issues because of how `always()` works on GH actions.

Unless I'm missing something, the conditional evaluated to `true&false` before and now it's evaluating to `true&true&false` which should still be false. so I'm actually not sure why it was working correctly before but failing now. Going to try with removing the `always()` as suggested and seeing if that fixes it

lakshmimsft (dismissed):

ytimocin (approved):

PR #7663, opened 2024-06-04
Update rad bicep download to use official bicep compiler

sk593:
# Description

Update rad bicep download to download the official bicep compiler

Screenshot 2024-06-03 at 10 45 00 PM
[downloading the official release 0.27.1 instead of our release version 0.34]

## Type of change

Fixes: #issue_number


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "bicep-types"]
+	path = bicep-types
+	url = https://github.com/Azure/bicep-types
```
rynowak, line 3: Is this something we need long-term? Adding a submodule means we have to update a lot of instructions.

Do you think there's a way we could decouple the repos so we don't need the bicep generate in the Radius repo?

build/generate.mk
```diff
@@ -116,7 +116,9 @@ generate-go: generate-mockgen-installed ## Generates go with 'go generate' (Mock
 generate-bicep-types: generate-node-installed ## Generate Bicep extensibility types
 	@echo "$(ARROW) Generating Bicep extensibility types from OpenAPI specs..."
 	@echo "$(ARROW) Build autorest.bicep..."
-	cd hack/bicep-types-radius/src/autorest.bicep; \
+	cd bicep-types/src/bicep-types; \
+	npm ci && npm run build; \
```
rynowak, line 120: Might be easier to use `--prefix` instead of changing directories? https://stackoverflow.com/questions/37078968/how-to-specify-the-path-of-package-json-to-npm

rynowak: Then you don't have to do math about home many `..`s you need.

.gitmodules
sk593: Sure, we could remove the `hack` folder into its own repository similar to how the `bicep-types-aws` repo is. So that repository would host the submodule instead of the main radius repo and we'd handle uploads to ACR through the new repo. 

The submodule should just require a `git submodule init && git submodule update` if the user wants to use it locally. I don't think it affects development unless someone is working with the bicep generator

pkg/cli/tools/binary_tools.go
```diff
@@ -123,58 +119,41 @@ func GetValidPlatform(currentOS, currentArch string) (string, error) {
...
+	resp, err := http.Get(binaryRepo + extension)
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
 
-	// Open the folder so we can mark it as executable
-	bicepBinary, err := os.Open(filepath)
+	// Write the body to file
+	_, err = io.Copy(bicepBinary, resp.Body)
```
rynowak, line 145: What would happen if...

- There's an existing version of bicep on disk.
- The existing copy on disk is larger than the new version.

pkg/cli/tools/binary_tools.go
```diff
@@ -17,21 +17,17 @@ limitations under the License.
...
-	"github.com/radius-project/radius/pkg/version"
-	"oras.land/oras-go/v2"
-	"oras.land/oras-go/v2/content/file"
-	"oras.land/oras-go/v2/registry/remote"
 )
 
 const (
 	// binaryRepo is the name of the remote bicep binary repository
-	binaryRepo = "ghcr.io/radius-project/radius/bicep/rad-bicep/"
+	binaryRepo = "https://github.com/Azure/bicep/releases/latest/download/"
```
ytimocin, line 30: Would getting the `latest` version all the time cause a problem? Could there be a breaking change that could affect us?

pkg/cli/tools/binary_tools.go
```diff
@@ -17,21 +17,17 @@ limitations under the License.
 package tools
 
 import (
-	"context"
 	"fmt"
+	"io"
+	"net/http"
```
rynowak, line 22: Does it make sense to rename this file? It's called `binary_tools.go` because it used to handle multiple tools (there used to be multiple things we downloaded). Should this go the bicep package and be renamed?

pkg/cli/tools/binary_tools.go
```diff
@@ -193,3 +172,16 @@ func getFilename(base string) (string, error) {
 		return "", fmt.Errorf("unsupported platform %s/%s", runtime.GOOS, runtime.GOARCH)
 	}
 }
+
+func getFileExtension() (string, error) {
```
rynowak, line 176: nit: these are filenames not file extensions. A file extension would be like `.exe`

pkg/cli/tools/binary_tools.go
```diff
@@ -193,3 +172,16 @@ func getFilename(base string) (string, error) {
...
+	switch runtime.GOOS {
+	case "darwin":
+		return "bicep-osx-x64", nil
+	case "linux":
+		return "bicep-linux-x64", nil
+	case "windows":
+		return "bicep-win-x64.exe", nil
```
rynowak, line 183: Do we need to handle OS architectures as well?

pkg/cli/tools/binary_tools.go
sk593: I think we'd want to always get the latest version of bicep. It is possible that there are breaking changes since this is now a dependency instead of something we own. But we will have dependency updates and testing as part of the merge (this will likely come in closer to the end) that will notify us if there's a breaking change.

pkg/cli/tools/binary_tools.go
sk593: This should only download the binary to .rad directory. So if there's an existing version of that file, it will get overridden but if there's another bicep binary somewhere else, I don't believe that will get updated. 

The size of the updated file would be that of the new version. os.Create() truncates the file if it exists so it should be the size of the new file after io.Copy()

Let me know if any of that seems wrong

pkg/cli/tools/binary_tools.go
sk593: That makes sense. Move to the bicep folder and renamed to `download_tools`

pkg/cli/tools/binary_tools.go
```diff
@@ -123,58 +119,41 @@ func GetValidPlatform(currentOS, currentArch string) (string, error) {
...
-	if version.IsEdgeChannel() {
-		tag = "latest"
-	}
-	_, err = oras.Copy(ctx, repo, tag, fs, tag, oras.DefaultCopyOptions)
+	// Get the data
+	resp, err := http.Get(binaryRepo + extension)
 	if err != nil {
 		return err
 	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
sk593: Added a check to error out if there's a 404

pkg/cli/tools/binary_tools.go
sk593: Updated

rynowak (approved):
Looks great!

I'm ok with downloading the latest Bicep for now. 

- The version of bicep we use inside `rad` won't affect the user's system overall, just `rad`.
- We only download Bicep when it's missing. If the user needs to use another version to work around a problem, they can fix it manually.

ytimocin (approved):

PR #7664, opened 2024-06-04
Update rad init to generate bicep config

sk593:
# Description

Update rad init to generate bicep configuration
Screenshot 2024-06-03 at 10 45 14 PM

## Type of change

Fixes: #issue_number


pkg/cli/setup/application.go
```diff
@@ -76,6 +88,17 @@ func ScaffoldApplication(directory string, name string) error {
...
+	_, err = os.Stat(bicepConfigFilepath)
+	if os.IsNotExist(err) {
+		err = os.WriteFile(bicepConfigFilepath, []byte(bicepConfigTemplate), 0644)
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
+
```
rynowak, line 101: Does the interactive display from `rad init` include this info in its output?

pkg/cli/setup/application.go
```diff
@@ -47,6 +47,18 @@ resource demo 'Applications.Core/containers@2023-10-01-preview' = {
...
+	bicepConfigTemplate = `{
+	"experimentalFeaturesEnabled": {
+		"extensibility": true,
+		"providerRegistry": true,
+		"dynamicTypeLoading": true
+	},
+	"providers": {
+		"radius": "br:biceptypes.azurecr.io/radius:latest",
+		"aws": "br:biceptypes.azurecr.io/aws:latest"
```
rynowak, line 59: Will we end up publishing these to the official ACR registry? Or will we have our own?

pkg/cli/setup/application_test.go
```diff
@@ -33,6 +33,7 @@ func Test_ScaffoldApplication_CreatesBothFiles(t *testing.T) {
 
 	require.FileExists(t, filepath.Join(directory, ".rad", "rad.yaml"))
 	require.FileExists(t, filepath.Join(directory, "app.bicep"))
+	require.FileExists(t, filepath.Join(directory, "bicepconfig.json"))
```
rynowak, line 36: Should we have a test that validates the content of `bicepconfig.json`? I know it's hardcoded right now, but I thinking about future us ⌛

pkg/cli/setup/application.go
sk593: The plan was to have them pushed to an ACR within OSS assets. Do you think it'd be better to have them in the official registry?

pkg/cli/setup/application.go
```diff
@@ -47,6 +47,18 @@ resource demo 'Applications.Core/containers@2023-10-01-preview' = {
...
+	}
+}`
```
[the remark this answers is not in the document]
alex-frankel: Nothing obvious to me, but @harshpatel17 can you also take a look?

pkg/cli/setup/application.go
rynowak: I don't have a strong opinion, I just wanted to know the plan 👍

pkg/cli/setup/application.go
sk593: It should. The Run command would return an error if scaffolding fails. Tested forcing an error to confirm that it does (this may not be the actual error, this was just a dummy message).
 
Screenshot 2024-06-06 at 1 45 10 PM

pkg/cli/setup/application_test.go
sk593: This should check that the generated file vs template are the same: https://github.com/radius-project/radius/blob/e2eae38d5d62429d371c604b9426cc3e81b83617/pkg/cli/setup/application_test.go#L58

Let me know if you meant something different

rynowak: This is fine for now 👍 It will get more complicated in the future, which is why I wanted some kind of test to exist.

pkg/cli/setup/application.go
harshpatel17: This looks good

pkg/cli/setup/application.go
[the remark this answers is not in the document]
harshpatel17: This looks good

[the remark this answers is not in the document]
rynowak: Thanks @harshpatel17

pkg/cli/setup/application.go
```diff
@@ -47,6 +50,18 @@ resource demo 'Applications.Core/containers@2023-10-01-preview' = {
...
 ` // Trailing newline intentional.
+
+	bicepConfigTemplate = `{
+	"experimentalFeaturesEnabled": {
+		"extensibility": true,
+		"providerRegistry": true,
+		"dynamicTypeLoading": true
+	},
+	"providers": {
+		"radius": "br:biceptypes.azurecr.io/radius:%s",
```
sk593, line 61: Originally had this set to "latest" but I updated this to correspond with the current version. So "latest" will only be used on edge versions and the current release version is used otherwise

rynowak: Good plan 👍

pkg/cli/setup/application.go
```diff
@@ -20,16 +20,19 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/radius-project/radius/pkg/version"
 )
 
 const (
-	appBicepTemplate = `import radius as radius
+	appBicepTemplate = `provider radius
```
rynowak, line 28: FYI it looks like this will change again to `extension radius`.

sk593: Got it, will update as part of this PR or in a follow up PR depending on when they release the next version with that change

pkg/cli/setup/application.go
```diff
@@ -20,16 +20,19 @@ import (
...
 const (
-	appBicepTemplate = `import radius as radius
+	appBicepTemplate = `extension radius
 
 @description('The Radius Application ID. Injected automatically by the rad CLI.')
 param application string
 
 resource demo 'Applications.Core/containers@2023-10-01-preview' = {
   name: 'demo'
+  location: 'global'
```
willdavsmith, line 35: is this required now?

willdavsmith (approved):
lgtm

pkg/cli/setup/application.go
sk593: oops no, its not. will remove this

PR #7669, opened 2024-06-06
Show correct environment and application for delete commands

### LINKED ISSUE #7542: `rad app delete` shows the wrong environment
### Steps to reproduce

1. Run `rad group list` to see your groups
2. Run `rad group switch other-group` to change your group
3. Run `rad app list` to see your apps
4. Run `rad env list` to see your environments within that group (Please see the screenshot below)
5. Run `rad app delete my-app` and see that the environment is default and not the app's environment

### Observed behavior

image

image
App details

### Desired behavior

`rad app delete app-name` should show the app's environment

### Workaround

_No response_

### rad Version

RELEASE VERSION BICEP COMMIT
0.32.0 v0.32.0 0.32.0 https://github.com/radius-project/radius/commit/1b8bd4ad7f7a3f9d88b1275b284f7357915a914e

### Operating system

MacOS, Sonoma 14.4.1, Apple Chip

### Additional context

_No response_

[AB#12058](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12058)

rynowak:
# Description

The existing prompt logic for application and resource deletion was relying the workspace configuration for the prompt data. This means a prompt like "would you like to delete application X from environment Z?" might print the wrong environment name!!! This is pretty bad because you might end up doing something you didn't want to and cause data loss.

The fix to do additional retrievals of the resource and potentially the application so that can give the correct prompt.

Also fixed a minor validation bug in `rad env create` where we could accept and ignore extra positional args.

## Type of change

Fixes: #7542

### DIFF
--- a/pkg/cli/cmd/app/delete/delete.go
+++ b/pkg/cli/cmd/app/delete/delete.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package list
+package delete
 
 import (
 	"context"
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/radius-project/radius/pkg/cli"
+	"github.com/radius-project/radius/pkg/cli/clients"
 	"github.com/radius-project/radius/pkg/cli/clierrors"
 	"github.com/radius-project/radius/pkg/cli/cmd/commonflags"
 	"github.com/radius-project/radius/pkg/cli/connections"
@@ -154,19 +155,32 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 // error if there is an issue with the connection or the prompt.
 func (r *Runner) Run(ctx context.Context) error {
 	// Prompt user to confirm deletion
+	client, err := r.ConnectionFactory.CreateApplicationsManagementClient(ctx, *r.Workspace)
+	if err != nil {
+		return err
+	}
+
+	app, err := client.GetApplication(ctx, r.ApplicationName)
+	if clients.Is404Error(err) {
+		r.Output.LogInfo("Application '%s' does not exist or has already been deleted.", r.ApplicationName)
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	environmentID, err := resources.ParseResource(*app.Properties.Environment)
+	if err != nil {
+		return err
+	}
+
 	if !r.Confirm {
-		confirmed, err := prompt.YesOrNoPrompt(fmt.Sprintf(deleteConfirmation, r.ApplicationName, r.EnvironmentName), prompt.ConfirmNo, r.InputPrompter)
+		confirmed, err := prompt.YesOrNoPrompt(fmt.Sprintf(deleteConfirmation, r.ApplicationName, environmentID.Name()), prompt.ConfirmNo, r.InputPrompter)
 		if err != nil {
 			return err
 		}
 		if !confirmed {
 			return nil
 		}
-	}
-
-	client, err := r.ConnectionFactory.CreateApplicationsManagementClient(ctx, *r.Workspace)
-	if err != nil {
-		return err
 	}
 
 	deleted, err := client.DeleteApplication(ctx, r.ApplicationName)
--- a/pkg/cli/cmd/env/create/create.go
+++ b/pkg/cli/cmd/env/create/create.go
@@ -52,7 +52,7 @@ func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
 		Long: `Create a new Radius Environment
 Radius Environments are prepared "landing zones" for Radius Applications.
 Applications deployed to an environment will inherit the container runtime, configuration, and other settings from the environment.`,
-		Args:    cobra.MinimumNArgs(1),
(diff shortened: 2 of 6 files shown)


pkg/cli/cmd/app/delete/delete.go
```diff
@@ -14,14 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package list
+package delete
```
rynowak, line 17: This was just wrong ... not sure how it was working before

kachawla (approved):

pkg/cli/cmd/env/create/create.go
```diff
@@ -52,7 +52,7 @@ func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
 		Long: `Create a new Radius Environment
 Radius Environments are prepared "landing zones" for Radius Applications.
 Applications deployed to an environment will inherit the container runtime, configuration, and other settings from the environment.`,
-		Args:    cobra.MinimumNArgs(1),
+		Args:    cobra.ExactArgs(1),
```
ytimocin, line 55: Can **group** be also an argument to this command?

pkg/cli/cmd/resource/delete/delete.go
```diff
@@ -169,3 +182,63 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		environmentID, err = convertToResourceID(resource.Properties["environment"])
+		if err != nil {
+			return resources.ID{}, resources.ID{}, err
+		}
+	}
+
+	if resource.Properties["application"] != nil {
+		applicationID, err = convertToResourceID(resource.Properties["application"])
+		if err != nil {
+			return resources.ID{}, resources.ID{}, err
```
ytimocin, line 209: Should we send the **environmentID** here or do we not send it because we are returning an error?

pkg/cli/cmd/resource/delete/delete.go
```diff
@@ -169,3 +182,63 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		return environmentID, applicationID, nil // Case 1 or Case 2
+	}
+
+	application, err := client.GetApplication(ctx, applicationID.String())
+	if clients.Is404Error(err) {
+		// Ignore 404s for this case, and just assume there is no application. The user is
+		// likely just doing cleanup and we don't want to block them.
+		return environmentID, resources.ID{}, nil
+	} else if err != nil {
+		return resources.ID{}, resources.ID{}, err
```
ytimocin, line 226: Should we also do `return environmentID, resources.ID{}, nil` here?

pkg/cli/cmd/resource/delete/delete.go
```diff
@@ -169,3 +182,63 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		applicationID, err = convertToResourceID(resource.Properties["application"])
+		if err != nil {
+			return resources.ID{}, resources.ID{}, err
+		}
+	}
+
+	// At this point we have the environment and application IDs **if** they were returned by
+	// the API. That covers case 1 & 2. Now we need to handle case 3, by doing an additional
+	// lookup.
+	if !environmentID.IsEmpty() {
```
ytimocin, line 216: Can the **applicationID** be also empty?

pkg/cli/cmd/app/delete/delete.go
```diff
@@ -154,8 +155,26 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
...
+	}
+
+	environmentID, err := resources.ParseResource(*app.Properties.Environment)
+	if err != nil {
+		return err
+	}
+
 	if !r.Confirm {
-		confirmed, err := prompt.YesOrNoPrompt(fmt.Sprintf(deleteConfirmation, r.ApplicationName, r.EnvironmentName), prompt.ConfirmNo, r.InputPrompter)
+		confirmed, err := prompt.YesOrNoPrompt(fmt.Sprintf(deleteConfirmation, r.ApplicationName, environmentID.Name()), prompt.ConfirmNo, r.InputPrompter)
```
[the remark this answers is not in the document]
rynowak: I don't want to print the application ID, that would be really long and hard for users to read.

pkg/cli/cmd/resource/delete/delete.go
rynowak: it's an error.

pkg/cli/cmd/resource/delete/delete.go
rynowak: Yes, this is valid (case 2 above).

pkg/cli/cmd/resource/delete/delete.go
rynowak: definitely not. For an error, the caller should not look at the other return values.

pkg/cli/cmd/env/create/create.go
rynowak: It's a flag, not a positional argument.

ytimocin (approved):

PR #7674, opened 2024-06-10
Adding TypeSpec check to Lint workflow and running the TypeSpec format

### LINKED ISSUE #7659: Consistency across Boolean properties
### Area for Improvement

We need to have consistency across the Boolean properties that we have. Right now, some properties are set to false by default. But bool gets false by default.

### Observed behavior

Inconsistency across bool properties of some resources.

### Desired behavior

Consistency across bool properties of some resources.

### Proposed Fix

We can delete setting the property to false.

### rad Version

edge

### Operating system

_No response_

### Additional context

image

`= false` settings can be deleted.

[AB#12455](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12455)

ytimocin:
# Description

1. Running `tsp format **/*.tsp`.
2. Adding `tsp format --check "**/*.tsp"` to the Lint workflow.

## Type of change
Fixes: #7659

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -67,6 +67,11 @@ jobs:
       - name: Run Helm linter
         run: |
           helm lint ${{ env.HELM_CHARTS_DIR }}
+      - name: Run TypeSpec format check
+        run: |
+          pushd typespec
+          npx tsp format --check "**/*.tsp"
+          popd
       - name: Run linter
         uses: golangci/golangci-lint-action@v6
         with:
--- a/docs/contributing/contributing-code/contributing-code-schema-changes/README.md
+++ b/docs/contributing/contributing-code/contributing-code-schema-changes/README.md
@@ -5,11 +5,16 @@ This page will explain the process to make a change to Radius' REST API (eg: add
 ## Step 1: Update TypeSpec and generate Bicep types and API client
 
 In order to update or create a new schema follow these steps:
-1. Create or update the applicable TypeSpec files (named after resource type) within the `typespec` directory in the root of the Radius repo
-2. Run `make generate` to generate the OpenAPI spec and API clients:
+
+1. Create or update the applicable TypeSpec files (named after resource type) within the `typespec` directory in the root of the Radius repo.
+1. Run `tsp format --check "**/*.tsp"` in the typespec folder to make sure that the format of added or updated files are good.
+1. You can run `tsp format **/*.tsp` to apply the formatting of TypeSpec compiler.
+1. Run `make generate` to generate the OpenAPI spec and API clients:
+
     ```bash
     make generate
     ```
+
     This will generate the OpenAPI spec and API client for all namespaces and run mockgen to generate mocks.
     <details>
     <summary>Alternately, if you would like to manually generate the OpenAPI spec and API client, follow these steps:</summary>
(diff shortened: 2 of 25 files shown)


.github/workflows/lint.yaml
```diff
@@ -67,6 +67,11 @@ jobs:
       - name: Run Helm linter
         run: |
           helm lint ${{ env.HELM_CHARTS_DIR }}
+      - name: Run TypeSpec format check
+        run: |
+          pushd typespec
+          npx tsp format --check "**/*.tsp"
+          popd
```
rynowak, line 74: What does the failure message look like when this fails? Is it something a user could figure out?

.github/workflows/lint.yaml
```diff
@@ -67,6 +67,11 @@ jobs:
       - name: Run Helm linter
         run: |
           helm lint ${{ env.HELM_CHARTS_DIR }}
+      - name: Run TypeSpec format check
+        run: |
+          pushd typespec
+          npx tsp format --check "**/*.tsp"
```
rynowak, line 73: Will this actually fail the step if it encounters an error? You might need `set -e`

typespec/Applications.Core/containers.tsp
```diff
@@ -224,7 +225,7 @@ model ConnectionProperties {
   source: string;
 
   @doc("default environment variable override")
-  disableDefaultEnvVars?: boolean = false;
+  disableDefaultEnvVars?: boolean;
```
rynowak, line 228: Very nice. The autoformat did this right?

.github/workflows/lint.yaml
```diff
@@ -67,6 +67,11 @@ jobs:
       - name: Run Helm linter
         run: |
           helm lint ${{ env.HELM_CHARTS_DIR }}
+      - name: Run TypeSpec format check
```
rynowak, line 70: Do we need to update any documentation for this? We're adding a new requirement for contributors.

typespec/Applications.Core/containers.tsp
ytimocin: I did this. The autoformat doesn't handle this case.

.github/workflows/lint.yaml
ytimocin: image

.github/workflows/lint.yaml
ytimocin: image

.github/workflows/lint.yaml
ytimocin: I will check this and update necessary docs.

.github/workflows/lint.yaml
rynowak: Great

.github/workflows/lint.yaml
rynowak: It might be helpful to do something like this. 

```
npx tsp format --check "**/*.tsp" || { echo "TypeSpec format check failed. Run 'npx tsp format **/*.tsp' and try again."; exit 1 }
```

docs/contributing/contributing-code/contributing-code-schema-changes/README.md
```diff
@@ -5,11 +5,16 @@ This page will explain the process to make a change to Radius' REST API (eg: add
...
 
 In order to update or create a new schema follow these steps:
-1. Create or update the applicable TypeSpec files (named after resource type) within the `typespec` directory in the root of the Radius repo
-2. Run `make generate` to generate the OpenAPI spec and API clients:
+
+1. Create or update the applicable TypeSpec files (named after resource type) within the `typespec` directory in the root of the Radius repo.
+1. Run `tsp format --check "**/*.tsp"` in the typespec folder to make sure that the format of added or updated files are good.
+1. You can run `tsp format **/*.tsp` to apply the formatting of TypeSpec compiler.
+1. Run `make generate` to generate the OpenAPI spec and API clients:
+
```
rynowak, line 13: 👍

typespec/Applications.Core/containers.tsp
rynowak: I think it's OK to do cleanup like this if we want to. We should be careful introducing rules that aren't enforced by tooling. Over time it turns into extra complexity for us. 

It might be a good idea to log an issue on typespec to suggest automatically removing `= false` from boolean values.

rynowak (dismissed):

typespec/Applications.Core/containers.tsp
ytimocin: Created this issue: https://github.com/microsoft/typespec/issues/3565

rynowak (approved):

PR #7685, opened 2024-06-14
Adding a check for `matchArtifact` being undefined in functional tests

### LINKED ISSUE #7668: Functional test runs triggered by a push to main fail
### Area for Improvement

Functional Testing

### Observed behavior

When a commit is pushed to main, this triggers a run of the functional tests. These function test runs always fail. 

Here's an example: https://github.com/radius-project/radius/actions/runs/9404321768/job/25902966666

### Desired behavior

I can think of two options for this:

- Don't run functional tests on push to main. I'm not sure why we need this.
- Fix the workflow so that it works on a push to main.

### Proposed Fix

_No response_

### rad Version

asdf

### Operating system

_No response_

### Additional context

_No response_

[AB#12476](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12476)

ytimocin:
# Description
Adding a check for **matchArtifact** being undefined in functional tests.

## Type of change
Fixes: #7668

### DIFF
--- a/.github/workflows/functional-test.yaml
+++ b/.github/workflows/functional-test.yaml
@@ -117,17 +117,17 @@ jobs:
       - name: Set up checkout target (pull_request)
         if: github.event_name == 'pull_request'
         run: |
-            echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
-            echo "CHECKOUT_REF=${{ github.ref }}" >> $GITHUB_ENV
-            echo "PR_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
+          echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
+          echo "CHECKOUT_REF=${{ github.ref }}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -18,18 +18,25 @@ runs:
...
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                artifact_id: matchArtifact.id,
+                archive_format: 'zip',
+            });
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/pr_number.zip`, Buffer.from(download.data));
+          } else {
+            console.log('No artifact named "pr_number" found');
+          }
```
rynowak, line 38: Should this fail if the PR number is not set? what should the behavior be?

.github/actions/download-pr-data-artifact/action.yaml
```diff
@@ -18,18 +18,25 @@ runs:
...
+    - name: "Unzip artifact"
       shell: bash
       run: unzip pr_number.zip
```
rynowak, line 41: Won't this fail on the next step?

PR #7693, opened 2024-06-18
Update build workflow to publish bicep types to ACR

sk593:
# Description

Update build workflow to publish bicep types to ACR

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -65,7 +65,7 @@ env:
   # URL to get source code for building the image
   IMAGE_SRC: https://github.com/radius-project/radius
 
-  # ACR url for bicep-types repository
+  # bicep-types ACR url for uploading Radius Bicep types
   BICEP_TYPES_ACR: 'biceptypes.azurecr.io'
 
 jobs:
@@ -265,8 +265,8 @@ jobs:
           echo ${{ secrets.GITHUB_TOKEN }} | helm registry login -u ${{ github.actor }} --password-stdin ${{ env.OCI_REGISTRY }}
           helm push ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}/radius-${{ env.CHART_VERSION }}.tgz oci://${{ env.OCI_REGISTRY }}/${{ env.OCI_REPOSITORY }}
 
-  generate-bicep-types:
-    name: Generate Radius Bicep types 
+  build-and-push-bicep-types:
+    name: Publish Radius bicep types to ACR
     runs-on: ubuntu-latest
     steps: 
       - name: Check out code
@@ -283,34 +283,26 @@ jobs:
           node-version: '18'
       - name: Generate Bicep extensibility types from OpenAPI specs
         run: |
-          make generate-bicep-types VERSION=testing
+          make generate-bicep-types VERSION=${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }}
       - name: Upload Radius Bicep types artifacts
         uses: actions/upload-artifact@v4
         with:
           name: radius-bicep-types
           path: ./hack/bicep-types-radius/generated
           if-no-files-found: error
-  
-  publish-bicep-types:
-    name: Publish Radius Bicep types to ACR 
-    needs: ['generate-bicep-types']
-    if: github.event_name == 'workflow_dispatch'
-    runs-on: ubuntu-latest
-    steps: 
-      - name: Check out code
-        uses: actions/checkout@v4
-      - name: Parse release version and set environment variables
-        run: python ./.github/scripts/get_release_version.py
       - name: 'Login via Azure CLI'
+        if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
         uses: azure/login@v1
         with:
           creds: ${{ secrets.BICEP_TYPES_AZURE_CREDENTIALS }}
       - name: Setup and verify bicep CLI
+        if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
         run: |
           curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
           chmod +x ./bicep
           bicep --version
       - name: Publish bicep types 
+        if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
         run: |
           bicep publish-provider ./hack/bicep-types-radius/generated/index.json --target br:${{ env.BICEP_TYPES_ACR }}/radius:${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }} --force
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -66,7 +66,7 @@ env:
   IMAGE_SRC: https://github.com/radius-project/radius
 
   # bicep-types ACR url for uploading Radius Bicep types
-  BICEP_TYPES_ACR: 'biceptypes.azurecr.io'
+  BICEP_TYPES_REGISTRY: 'biceptypes.azurecr.io'
 
 jobs:
   build-and-push-cli:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
           echo ${{ secrets.GITHUB_TOKEN }} | helm registry login -u ${{ github.actor }} --password-stdin ${{ env.OCI_REGISTRY }}
           helm push ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}/radius-${{ env.CHART_VERSION }}.tgz oci://${{ env.OCI_REGISTRY }}/${{ env.OCI_REPOSITORY }}
 
+  build-and-push-bicep-types:
+    name: Publish Radius bicep types to ACR
+    runs-on: ubuntu-latest
+        # Don't push on PR 
+    if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
```
sk593, line 272: The idea is that we only publish on merge to main or during a release in this workflow. The functional test workflow will also have a step to publish types to ACR and these will be tagged with the PR version. 

Open to suggestions if that doesn't sound like the right idea. Since the functional tests are where we'd consume the uploaded types (i.e. in all our test bicep templates), it made most sense to upload them as part of that workflow instead of here

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
...
+        with:
+          creds: ${{ secrets.BICEP_TYPES_AZURE_CREDENTIALS }}
+      - name: Setup and verify bicep CLI
+        run: |
+          curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+          chmod +x ./bicep
+          bicep --version
+      - name: Publish bicep types 
+        run: |
+          bicep publish-provider ./hack/bicep-types-radius/generated/index.json --target br:${{ env.ACR }}/radius:${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }} --force
```
sk593, line 305: Keeping up with the GHCR standards where the edge is tagged with "latest" instead of "edge"

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
...
+        with:
+          node-version: '18'
+      - name: Detecting node
+        run: |
+          echo "Detecting node..."
+          which node > /dev/null || { echo "node is a required dependency"; exit 1; }
+          echo "OK"
+      - name: Generate Bicep extensibility types from OpenAPI specs
+        run: |
+          make generate-bicep-types VERSION=${{ env.REL_VERSION }}
```
sk593, line 293: There's code in `generate.mk` and `generate.ts` that takes in a version string as an argument and uses that to correctly tag the index.json file. So we'll pass in the release version here (default value is latest)

.github/workflows/build.yaml
rynowak: > Open to suggestions if that doesn't sound like the right idea. Since the functional tests are where we'd consume the uploaded types (i.e. in all our test bicep templates), it made most sense to upload them as part of that workflow instead of here

We could do that, or we could keep the bicep types on disk.

For the functional tests, we'll need to update/generate a bicep config to either:

- Reference the specific version associated with the PR from the OCI registry (like we do with magpie)
- Reference a file on disk 

For the purposes of this PR, I think you have the right idea 👍

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
...
+          go-version: ${{ env.GOVER }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '18'
+      - name: Detecting node
+        run: |
+          echo "Detecting node..."
+          which node > /dev/null || { echo "node is a required dependency"; exit 1; }
+          echo "OK"
```
rynowak, line 290: Do we need this step immediately after we install node?

.github/workflows/build.yaml
rynowak: 👍

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
           echo ${{ secrets.GITHUB_TOKEN }} | helm registry login -u ${{ github.actor }} --password-stdin ${{ env.OCI_REGISTRY }}
           helm push ${{ env.ARTIFACT_DIR }}/${{ env.HELM_PACKAGE_DIR }}/radius-${{ env.CHART_VERSION }}.tgz oci://${{ env.OCI_REGISTRY }}/${{ env.OCI_REPOSITORY }}
 
+  build-and-push-bicep-types:
```
rynowak, line 268: I'd suggest refactoring this to move the "skip" logic down to the part where we publish to ACR. If you want to be really fancy you could use two separate jobs for it.

The reason is that we want to run as much of the workflow as possible on every commit. That way we can catch breaks as early as possible.

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
...
+      - name: Detecting node
+        run: |
+          echo "Detecting node..."
+          which node > /dev/null || { echo "node is a required dependency"; exit 1; }
+          echo "OK"
+      - name: Generate Bicep extensibility types from OpenAPI specs
+        run: |
+          make generate-bicep-types VERSION=${{ env.REL_VERSION }}
+      - name: 'Login via Azure CLI'
+        uses: azure/login@v1
```
rynowak, line 295: I'd also suggest uploading the bicep types output files to artifacts. That way we can inspect them easily if we need to.

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
```diff
@@ -256,7 +256,7 @@ async function buildTypeIndex(logger: ILogger, baseDir: string, version: string)
       types: readTypesJson(content),
     });
   }
-  const indexContent = await buildIndex(typeFiles,  log => logOut(logger, log), {name: "radius", version: version, isSingleton: false} as TypeSettings);
+  const indexContent = await buildIndex(typeFiles,  log => logOut(logger, log), {name: "radius", version: (version == "edge" ? "latest" : version), isSingleton: false} as TypeSettings);
```
rynowak, line 259: it would be nice if we could keep the logic about edge/latest out of this code and limit to the GH actions + makefile.

.github/workflows/build.yaml
sk593: The plan was to generate a temporary bicep config in the functional test folder that has the PR-specific provider reference during pipeline runs. So we upload to ACR with the PR-specific tag and then the tests will pull from that provider tag. 

There will be a bicep config that permanently lives in the radius folder, but this will mostly be for development purposes and will always reference the latest tag by default

.github/workflows/build.yaml
sk593: Not really, will remove

.github/workflows/build.yaml
sk593: Updated to only publish to ACR on main/releases. Everything else should run every time

.github/workflows/build.yaml
sk593: Added this step

hack/bicep-types-radius/src/generator/src/cmd/generate.ts
sk593: Makes sense. Removed the check from here and added it to the build workflow instead

.github/workflows/build.yaml
```diff
@@ -262,6 +265,45 @@ jobs:
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: '18'
+      - name: Generate Bicep extensibility types from OpenAPI specs
+        run: |
+          make generate-bicep-types VERSION=${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }}
+      - name: 'Login via Azure CLI'
+        uses: azure/login@v1
+        with:
+          creds: ${{ secrets.BICEP_TYPES_AZURE_CREDENTIALS }}
```
rynowak, line 290: Should we skip the login unless we're going to publish?

.github/workflows/build.yaml
```diff
@@ -65,6 +65,9 @@ env:
   # URL to get source code for building the image
   IMAGE_SRC: https://github.com/radius-project/radius
 
+  # ACR url for bicep file uploads
+  ACR: 'biceptypes.azurecr.io'
```
rynowak, line 69: Can you come up with a more descriptive name for this? It looks like my comment disappeared, but its still the same.

sk593: Ah, sorry I didnt see that comment. I updated to BICEP_TYPES_ACR, but I'm open to suggestions if something else seems better

.github/workflows/build.yaml
sk593: Updated. I did the same for the bicep cli download. Ideally, these would be moved into a new job that only runs with the `bicep publish` step. But since the jobs are independent, the generated files wouldn't be up to date with what's outputted in the make generate step. So these will just have an `if` condition on them

.github/workflows/build.yaml
rynowak: I'd suggest something like `BICEP_TYPES_REGISTRY`. It's more neutral.

rynowak (approved):

.github/workflows/build.yaml
sk593: Updated

PR #7702, opened 2024-06-20
Update validate bicep workflow

sk593:
# Description

Update validate bicep workflow 

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -283,7 +283,7 @@ jobs:
           node-version: '18'
       - name: Generate Bicep extensibility types from OpenAPI specs
         run: |
-          make generate-bicep-types VERSION=${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }}
+          make generate-bicep-types VERSION=${{ env.REL_CHANNEL == 'edge' && 'latest' || env.REL_CHANNEL }}
       - name: Upload Radius Bicep types artifacts
         uses: actions/upload-artifact@v4
         with:
@@ -304,7 +304,7 @@ jobs:
       - name: Publish bicep types 
         if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
         run: |
-          bicep publish-provider ./hack/bicep-types-radius/generated/index.json --target br:${{ env.BICEP_TYPES_REGISTRY }}/radius:${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }} --force
+          bicep publish-extension ./hack/bicep-types-radius/generated/index.json --target br:${{ env.BICEP_TYPES_REGISTRY }}/radius:${{ env.REL_CHANNEL == 'edge' && 'latest' || env.REL_CHANNEL }} --force
 
   publish-release:
     name: Publish GitHub Release
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -300,6 +300,7 @@ jobs:
         run: |
           curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
           chmod +x ./bicep
+          sudo mv ./bicep /usr/local/bin/bicep
           bicep --version
       - name: Publish bicep types 
         if: github.repository == 'radius-project/radius' && ((startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main'))
--- a/.github/workflows/validate-bicep.yaml
+++ b/.github/workflows/validate-bicep.yaml
@@ -37,12 +37,13 @@ jobs:
         uses: actions/checkout@v4
       - name: Parse release version and set environment variables
         run: python ./.github/scripts/get_release_version.py
-      - uses: oras-project/setup-oras@main
-      - name: Download rad-bicep
+      - name: Setup Azure CLI
+        run: curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+      - name: Setup and verify bicep CLI
         run: |
-          oras pull ghcr.io/radius-project/radius/bicep/rad-bicep/linux-x64:latest -o ./
-          chmod +x rad-bicep
-          ./rad-bicep --version
+          curl -Lo rad-bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+          chmod +x ./rad-bicep
+          bicep --version
       - name: Verify Bicep files
         run: ./build/validate-bicep.sh
         env:
--- a/.github/workflows/validate-bicep.yaml
+++ b/.github/workflows/validate-bicep.yaml
@@ -37,14 +37,13 @@ jobs:
         uses: actions/checkout@v4
       - name: Parse release version and set environment variables
         run: python ./.github/scripts/get_release_version.py
-      - name: Setup Azure CLI
-        run: curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
       - name: Setup and verify bicep CLI
         run: |
-          curl -Lo rad-bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
-          chmod +x ./rad-bicep
+          curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+          chmod +x ./bicep
+          sudo mv ./bicep /usr/local/bin/bicep
           bicep --version
       - name: Verify Bicep files
         run: ./build/validate-bicep.sh
         env:
-          BICEP_PATH: './rad-bicep'
+          BICEP_PATH: 'bicep'
--- a/.github/workflows/validate-bicep.yaml
+++ b/.github/workflows/validate-bicep.yaml
@@ -33,16 +33,14 @@ jobs:
     name: Validate Bicep Code
     runs-on: ubuntu-latest
     steps:
-      - name: Check out repo
-        uses: actions/checkout@v4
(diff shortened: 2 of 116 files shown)


build/validate-bicep.sh
```diff
@@ -5,6 +5,7 @@ then
     exit 1
 fi
 
+WARNING_MSG="WARNING: The following experimental Bicep features have been enabled: Extensibility, Dynamic type loading, Provider registry. Experimental features should be enabled for testing purposes only, as there are no guarantees about the quality or stability of these features. Do not enable these settings for any production usage, or your production environment may be subject to breaking."
```
sk593, line 8: Will look into if this can be disabled so it doesn't print on every build

sk593: For now if this is the output, we'll ignore it since it's not an error

sk593:
The Bicep compiler doesn't make assumptions about the location (rad CLI defaults to 'global') so we'll include that in our bicep templates

build/validate-bicep.sh
ytimocin: What if this message changes? How will we know that we need to update this?

test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
```diff
@@ -19,7 +19,7 @@ resource webapp 'Applications.Core/containers@2023-10-01-preview' = {
     container: {
       image: magpieimage
       env: {
-        DBCONNECTION: redis.connectionString()
+        DBCONNECTION: redis.listSecrets().connectionString
```
ytimocin, line 22: Was this wrong before?

ytimocin: Or is it a new feature of the provider?

ytimocin: Or is it a new feature of the provider?

test/testrecipes/test-bicep-recipes/resource-creation.bicep
```diff
@@ -8,6 +8,7 @@ var basename = context.resource.name
 // doing things this way to test the UCP functionality without using cloud resources.
 resource extender 'Applications.Core/extenders@2023-10-01-preview' = {
   name: '${basename}-created'
+  location: 'global'
```
ytimocin, line 11: Should this be a constant?

build/validate-bicep.sh
sk593: The workflow would start failing. I think the ideal situation would be to suppress the warning message if that's possible

test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
sk593: Yes, its a new feature so we now make function calls using listSecrets

bicepconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"experimentalFeaturesEnabled": {
+		"extensibility": true,
+		"providerRegistry": true,
+		"dynamicTypeLoading": true
+	},
+	"providers": {
+		"radius": "br:biceptypes.azurecr.io/radius:latest",
```
rynowak, line 8: How do you expect this to work for a PR?

build/validate-bicep.sh
rynowak: Possible we could turn this into a wildcard so we don't have to match the whole string? eg: `WARNING: The following experimental Bicep features *`

I'm wondering about whether the string is going to change and break our build.

test/functional-portable/corerp/cloud/resources/testdata/corerp-resources-extender-aws-s3-recipe.bicep
```diff
@@ -1,15 +1,16 @@
...
 param awsAccountId string
 param awsRegion string
 param registry string 
 param version string
+param location string = 'global'
 
 resource env 'Applications.Core/environments@2023-10-01-preview' = {
   name: 'corerp-resources-extenders-aws-s3-recipe-env'
-  location: 'global'
+  location: location
```
rynowak, line 13: Is there a reason why we need to specify the location here? We made a change to make location optional in Bicep for our resources.

rynowak: If location is not showing as optional, then that's a regression and we need to fix it.

test/functional-portable/corerp/noncloud/resources/testdata/containers/corerp-resources-friendly-container-version-1.bicep
rynowak: Yes, we're intentionally simplifying the design of the provider.

bicepconfig.json
sk593: The bicepconfig here is the default for the repo and it shouldn't change (so when you're doing development work, the templates will all compile). These will all use the latest tag. For PRs, we'll have a temporary bicepconfig that's generated with a PR-specific tag for workflows like functional testing, PR builds, etc. That'll be added in a separate PR

test/functional-portable/corerp/cloud/resources/testdata/corerp-resources-extender-aws-s3-recipe.bicep
sk593: Location has as required flag in the types files so this might be a regression in the generation logic. Will look into that and fix it. But our docs also have location as a required field so that probably needs to get updated too: https://docs.radapp.io/reference/resource-schema/overview/

rynowak: Whether or not it's required depends on which client you're talking about. The Radius API requires location to be set. Bicep will default it for you. One of the ways we tweak the Bicep experience to be simpler.

sk593: This was a casing bug in the generator. Fixed as part of the PR

ytimocin (dismissed):

.github/workflows/build.yaml
```diff
@@ -283,7 +283,7 @@ jobs:
           node-version: '18'
       - name: Generate Bicep extensibility types from OpenAPI specs
         run: |
-          make generate-bicep-types VERSION=${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }}
+          make generate-bicep-types VERSION=${{ env.REL_CHANNEL == 'edge' && 'latest' || env.REL_CHANNEL }}
```
willdavsmith, line 286: was this wrong before? if not why change over to `REL_CHANNEL`?

.github/workflows/validate-bicep.yaml
```diff
@@ -37,12 +37,13 @@ jobs:
         uses: actions/checkout@v4
       - name: Parse release version and set environment variables
         run: python ./.github/scripts/get_release_version.py
-      - uses: oras-project/setup-oras@main
-      - name: Download rad-bicep
+      - name: Setup Azure CLI
```
willdavsmith, line 40: why do we need azure CLI?

build/validate-bicep.sh
```diff
@@ -29,7 +30,7 @@ do
         exec 3>&-
     fi
     
-    if [[ ! $EXITCODE -eq 0 || ! -z $STDERR ]]
+    if [[ ! $EXITCODE -eq 0 || (! -z $STDERR && ! $STDERR == $WARNING_MSG* ) ]]
```
willdavsmith, line 33: did you test this? I don't speak bash-fu

willdavsmith (approved):
approved with minor questions

.github/workflows/build.yaml
sk593: yea it was, we pull from the version channel in the bicep config so the uploads should also be to the channel tag

build/validate-bicep.sh
sk593: yea, ran this through a workflow_dispatch on the branch and it passes

.github/workflows/validate-bicep.yaml
sk593: I'm pretty sure the Bicep CLI relies on the az CLI for certain commands so both are needed

.github/workflows/validate-bicep.yaml
```diff
@@ -33,17 +33,15 @@ jobs:
     name: Validate Bicep Code
     runs-on: ubuntu-latest
     steps:
+      - name: Setup and verify bicep CLI
```
sk593, line 36: there seems to be some lag between the bicep download and the verify step so moving this up to mitigate that

.github/workflows/validate-bicep.yaml
```diff
@@ -33,17 +33,15 @@ jobs:
...
+        run: |
+          curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+          chmod +x ./bicep
+          sudo mv ./bicep /usr/local/bin/bicep
+          bicep --version
       - name: Check out repo
         uses: actions/checkout@v4
-      - name: Parse release version and set environment variables
```
sk593, line 38: Don't think these variables were being used

ytimocin (approved):

PR #7722, opened 2024-07-01
Fix prompt text test occasional failure

### LINKED ISSUE #7721: Prompt text test occasionally fails
### Steps to reproduce

1. Run `make test`, following instructions at [contributing-code-tests](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-tests/README.md#unit-tests).
1. Observe the result of this test: `github.com/radius-project/radius/pkg/cli/prompt/text`. If the test passed, run it in isolation, without cached results. For example, run the test using this command: `go test -run ^Test_E2E$/^confirm_value$ github.com/radius-project/radius/pkg/cli/prompt/text -count=1` 

Eventually the test will fail when running it in isolation, especially if you are on a resource constrained machine, like GitHub Codespaces. The test failed for me when walking through the [initial setup instructions for contributing](https://github.com/radius-project/radius/tree/main/docs/contributing/contributing-code/contributing-code-prerequisites). When running on a fast desktop using the dev container, I was still able to get the test to fail, but it happens much less often, suggesting that there is a timing issue.

### Observed behavior

When the test fails, the output will look like this:

```bash
$ go test -run ^Test_E2E$/^confirm_value$ github.com/radius-project/radius/pkg/cli/prompt/text -count=1
--- FAIL: Test_E2E (0.05s)
    --- FAIL: Test_E2E/confirm_value (0.05s)
        text_test.go:92: Testing output:
            
        text_test.go:92: Testing output:
            
test prompt 
            
            > test placeholder
            
            (ctrl+c to quit)
        text_test.go:144: 
                Error Trace:    /workspaces/radius/pkg/cli/prompt/text/text_test.go:144
                Error:          Should be empty, but was test prompt
                                > a
                Test:           Test_E2E/confirm_value
FAIL
FAIL    github.com/radius-project/radius/pkg/cli/prompt/text    0.057s
FAIL
```

A passing test will report a result like this:
```bash
ok      github.com/radius-project/radius/pkg/cli/prompt/text    0.057s
```

### Desired behavior

The expected result is for the test to pass every time.

### Workaround

_No response_

### rad Version

Main branch, commit `13d50f7b02f77b868bebb9a3582953d3bc7c9f2d`.

### Operating system

Dev container, running in GitHub Codespaces and also the same dev container running on a desktop machine using Docker Desktop, with the source cloned to WSL (Ubuntu default).

### Additional context

You can force the test to fail every time by adding a `sleep` command before [pkg/cli/prompt/text/text_test.go](https://github.com/radius-project/radius/blob/main/pkg/cli/prompt/text/text_test.go#L138) (138). For example (starting on line 137):

```golang
tm.Type("abcd")
time.Sleep(1 * time.Second)
tm.Send(tea.KeyMsg{Type: tea.KeyEnter})
```

Suggested fix:

Later in the test method, there is this line, which is expecting the output to be an empty string. I think this line is incorrect and should be removed. The reason it usually passes is because the underlying TeaTest test model has not finished rendering the text output. The `time.Sleep` statement above shows how pausing execution gives the test model time to catch up and render the text. The Tea Test docs say that the final output will be available when the `FinalModel` method is run. This line below is run before `FinalModel` runs, which is why it usually, but not always, passes.
```golang
require.Empty(t, strings.TrimSpace(output)) // Output sometimes contains a single space.
```

In other words, I think the fix is simply to delete the line above, because the expected output is not actually an empty string, even though at that point in the code execution the console has not finished rendering and the output is often an empty string.

[AB#12635](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12635)

brooke-hamilton:
# Description

Fixes an issue in which the `confirm value` test in `/pkg/cli/prompt/text` sometimes fails due to the timing of the text rendering in the test model.

The test was checking the output before calling `TestModel.FinalModel()`, which resulted in an empty result most of the time, but not all of the time. The other assertions made after calling `FinalModel()` showed that the code was operating as designed, which means we can remove the interim check that tests for an empty output. This change simply removes the code that checks the output before calling `FinalModel()`.

## Type of change

This pull request fixes a bug in Radius and has an approved issue.

Fixes: #7721


rynowak:
Thanks @brooke-hamilton for digging in and fixing this. Taking a look....

pkg/cli/prompt/text/text_test.go
```diff
@@ -137,11 +137,7 @@ func Test_E2E(t *testing.T) {
 		tm.Type("abcd")
 		tm.Send(tea.KeyMsg{Type: tea.KeyEnter})
 		tm.WaitFinished(t, teatest.WithFinalTimeout(waitTimeout))
-		bts, err := io.ReadAll(tm.FinalOutput(t))
-		require.NoError(t, err)
 
-		output = normalizeOutput(bts)
-		require.Empty(t, strings.TrimSpace(output)) // Output sometimes contains a single space.
```
rynowak, line 144: BTW: what this is doing is verifying that the output is blank after the user confirms the prompt or cancels. This is intentional because we want to hide the display after the user is done.

However, if this is the thing that's causing the test to fail I don't think it's critical to cover in a test. We've seen lots of issues using the test framework (teatest) - so over time these tests have been simplified to try and keep them reliable. 

---

I tried your fix with:

```
go test -count=10000 -run=Test_E2E -failfast ./pkg/cli/prompt/text
```

and it worked for me.

rynowak: Can you also update [here](https://github.com/radius-project/radius/pull/7722/files#diff-a95b3c117c873821d9473cd582ee12f2ffccb36c71d5fd41b1a3897f0268d957L163-L167) and [here](https://github.com/radius-project/radius/pull/7722/files#diff-a95b3c117c873821d9473cd582ee12f2ffccb36c71d5fd41b1a3897f0268d957L116-L120) where we have the same test logic?

If it's wrong for this testcase, then it's probably wrong for the others.

rynowak (changes requested):
Thanks for digging into this!

pkg/cli/prompt/text/text_test.go
brooke-hamilton: > Can you also update [here](https://github.com/radius-project/radius/pull/7722/files#diff-a95b3c117c873821d9473cd582ee12f2ffccb36c71d5fd41b1a3897f0268d957L163-L167) and [here](https://github.com/radius-project/radius/pull/7722/files#diff-a95b3c117c873821d9473cd582ee12f2ffccb36c71d5fd41b1a3897f0268d957L116-L120) where we have the same test logic?
> 
> If it's wrong for this testcase, then it's probably wrong for the others.

Thank you for looking at this PR! 🥇 🚀 

I pushed changes to modify the other two tests. I also refactored to reduce some duplication, but I'm still learning this code so if you disagree with the style I'm happy to change it back to the previous test style.

> BTW: what this is doing is verifying that the output is blank after the user confirms the prompt or cancels. This is intentional because we want to hide the display after the user is done.

That makes a lot of sense for what to test. In the "confirm value" test case where we send "abcd" to the program, if we sleep long enough, the result of `TestModel.FinalOutput()` is this: `"test prompt\n> abcd            \n "`. 

> However, if this is the thing that's causing the test to fail I don't think it's critical to cover in a test. We've seen lots of issues using the test framework (teatest) - so over time these tests have been simplified to try and keep them reliable.

The test framework may need an additional feature to enable these tests. The behavior I observed is that when sending text to the test model, the `TestModel.out` reader is not synchronously updated with program output, I think because it is attached to the `stdout` in the program, which happens in the setup that is performed by `teatest`. (teatest.go, line 131.) The reader is not updated synchronously, which makes logical sense, but there ought to be a way to wait for the result. 
Running `teatest.WithFinalTimeout()`, `TestModel.FinalOutput()`, or `TestModel.FinalModel()` does not force the test to wait until the `stdout` is finished writing to the stream, which means that calling `TestModel.FinalOutput()` is not deterministic.

ytimocin (approved):
🚀

pkg/cli/prompt/text/text_test.go
rynowak: > The test framework may need an additional feature to enable these tests. The behavior I observed is that when sending text to the test model, the TestModel.out reader is not synchronously updated with program output, I think because it is attached to the stdout in the program, which happens in the setup that is performed by teatest. (teatest.go, line 131.) The reader is not updated synchronously, which makes logical sense, but there ought to be a way to wait for the result.

Thanks for digging into. I wondered if this was that was the case - specifically that `FinalOutput` is flaky. We've had a bit of trouble with these tests/this-framework. I'm glad that this code is testable at all, so it seems OK to work around some of the limitations!

rynowak (approved):
Changes look great, and a big improvement to the readability of the tests.

rynowak:
@ytimocin - can you help get this merged?

lakshmimsft (approved):

PR #7727, opened 2024-07-05
Updating and adding a timeout to the cleanup cluster step

ytimocin:
# Description
Updating and adding a timeout to the cleanup cluster step.

## Type of change

### DIFF
--- a/.github/scripts/cleanup-cluster.sh
+++ b/.github/scripts/cleanup-cluster.sh
@@ -45,7 +45,7 @@ fi
 # Delete all test namespaces.
 echo "delete all test namespaces"
 namespaces=$(kubectl get namespace |
-    grep -E '^kubernetes-interop-tutorial.*|^corerp.*|^default-.*|^radiusfunctionaltestbucket.*|^radius-test.*|^kubernetes-cli.*|^dpsb-.*|^dsrp-.*|^azstorage-workload.*|^dapr-serviceinvocation|^daprrp-rs-.*|^mynamespace.*|^demo.*|^tutorial-demo.*|^ms.+' |
+    grep -E '^kubernetes-interop-tutorial.*|^corerp.*|^test.*|^default-.*|^radiusfunctionaltestbucket.*|^radius-test.*|^kubernetes-cli.*|^dpsb-.*|^dsrp-.*|^azstorage-workload.*|^dapr-serviceinvocation|^daprrp-rs-.*|^mynamespace.*|^demo.*|^tutorial-demo.*|^ms.+' |
     awk '{print $1}')
 for ns in $namespaces; do
     if [ -z "$ns" ]; then
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -515,6 +515,7 @@ jobs:
             --yes --verbose
       - name: Clean up cluster
         if: always()
+        timeout-minutes: 60
         run: ./.github/scripts/cleanup-cluster.sh
   report-failure:
     name: Report test failure


.github/scripts/cleanup-cluster.sh
```diff
@@ -52,5 +52,14 @@ for ns in $namespaces; do
         break
     fi
     echo "deleting namespaces: $ns"
+
+    # Remove finalizers from the namespace if present
+    kubectl get namespace $ns -o json | jq '.spec.finalizers = []' >temp-namespace.json
```
ytimocin, line 57: Interestingly, UCP and Applications RP Pods were not healthy, and this caused finalizers to not finalize in some of the resources in some of the namespaces. And then this caused those namespaces to not be deleted.

UCP Pod Down -> Finalizers in some resources -> Namespaces couldn't be deleted -> Clean up step got stuck -> Long Running Workflow failed back to back

.github/workflows/long-running-azure.yaml
```diff
@@ -515,6 +515,7 @@ jobs:
             --yes --verbose
       - name: Clean up cluster
         if: always()
+        timeout-minutes: 60
```
ytimocin, line 518: Added timeout to clean up step.

.github/scripts/cleanup-cluster.sh
ytimocin, line 1: Reformatted the files

.github/scripts/cleanup-cluster.sh
```diff
@@ -52,5 +52,14 @@ for ns in $namespaces; do
...
+    kubectl replace --raw "/api/v1/namespaces/$ns/finalize" -f ./temp-namespace.json
+
+    # Attempt to delete the namespace
     kubectl delete namespace $ns --ignore-not-found=true
 done
+
+# Cleanup temporary file
+rm -f temp-namespace.json
```
lakshmimsft, line 65: if an error occurs above/script exists, is it possible that file is not cleaned up ?

.github/scripts/cleanup-cluster.sh
rynowak: Do you have logs from the failing pods? or from the Kubernetes `controller-manager` pod?

.github/scripts/cleanup-cluster.sh
```diff
@@ -52,5 +52,14 @@ for ns in $namespaces; do
         break
     fi
     echo "deleting namespaces: $ns"
+
+    # Remove finalizers from the namespace if present
+    kubectl get namespace $ns -o json | jq '.spec.finalizers = []' >temp-namespace.json
+    kubectl replace --raw "/api/v1/namespaces/$ns/finalize" -f ./temp-namespace.json
```
rynowak, line 58: Does `--force` help here? https://kubernetes.io/docs/reference/kubectl/generated/kubectl_delete/

Just thinking it might be easier if it works.

I'm also curious what the finalizers are. We have to be a little bit careful with this, because certain types in Kubernetes will manage cloud resources. If we delete the Kubernetes resource without running the finalizers, then the cloud resources leak.

.github/workflows/long-running-azure.yaml
rynowak: What's the goal of the timeout? That we'd be notified if we hit it?

sylvainsf: Default is 6 hours, so fail faster?

ytimocin: Exactly. To fail faster.

sylvainsf (approved):
Looks good to me!

PR #7738, opened 2024-07-15
implement irsa server side support

### LINKED ISSUE #7618: Add IRSA (workload identity) support for AWS cloud provider

Current State
Today when you register the AWS provider, it AWS Access Key ID and the AWS Secret Access Key of IAM role. This is how UCP connects to AWS in order to deploy Azure resources.

Desired State
We should also support IAM roles for service accounts which is the equivalent of workload identity for AWS 

### Acceptance criteria

1. Setup [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) on their cluster
2. Register their AWS provider with the IAM role
3. Deploy AWS resources 

### Additional context

_No response_

[AB#12397](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12397)

nithyatsu:
# Description

Add server side support for AWS IRSA. 
UCP handles AWS resource deployment and needs irsa suuport.
Terraform provider communicates with AWS directly and needs IRSA support too.

## Type of change

Partially Fixes: https://github.com/radius-project/radius/issues/7618


pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -127,7 +136,10 @@ func fetchAWSCredentials(ctx context.Context, awsCredentialsProvider credentials
 			return nil, nil
 		}
 	case ucp_datamodel.AWSIRSACredentialKind:
-		return nil, errors.New("AWS IRSA Credential is not supported yet.")
+		if credentials.IRSACredential == nil || credentials.IRSACredential.RoleARN == "" {
```
vishwahiremat, line 139:
```suggestion
		if credentials.IRSACredential == nil && credentials.IRSACredential.RoleARN == "" {
```
I wonder if this should be `and` operator here.

pkg/ucp/aws/ucpcredentialprovider.go
```diff
@@ -73,19 +86,65 @@ func (c *UCPCredentialProvider) Retrieve(ctx context.Context) (aws.Credentials,
...
-		AccessKeyID:     s.AccessKeyCredential.AccessKeyID,
-		SecretAccessKey: s.AccessKeyCredential.SecretAccessKey,
-		Source:          "radiusucp",
-		CanExpire:       true,
-		// Enables AWS SDK to fetch (rotate) access keys by calling Retrieve() after Expires.
-		Expires: time.Now().UTC().Add(c.options.Duration),
+	var value aws.Credentials
+	switch s.Kind {
+	case CredentialKindAccessKey:
+		if s.AccessKeyCredential == nil || s.AccessKeyCredential.AccessKeyID == "" || s.AccessKeyCredential.SecretAccessKey == "" {
```
vishwahiremat, line 92:
```suggestion
		if s.AccessKeyCredential == nil && (s.AccessKeyCredential.AccessKeyID == "" || s.AccessKeyCredential.SecretAccessKey == "") {
```
I think we will hit nil pointer here when `s.AccessKeyCredential==nil`

pkg/ucp/aws/ucpcredentialprovider.go
```diff
@@ -73,19 +86,65 @@ func (c *UCPCredentialProvider) Retrieve(ctx context.Context) (aws.Credentials,
...
+		value = aws.Credentials{
+			AccessKeyID:     s.AccessKeyCredential.AccessKeyID,
+			SecretAccessKey: s.AccessKeyCredential.SecretAccessKey,
+			Source:          "radiusucp",
+			CanExpire:       true,
+			Expires:         time.Now().UTC().Add(c.options.Duration),
+		}
+
+	case CredentialKindIRSA:
+		if s.IRSACredential == nil || s.IRSACredential.RoleARN == "" {
```
vishwahiremat, line 106:
```suggestion
		if s.IRSACredential == nil && s.IRSACredential.RoleARN == "" {
```
same here

pkg/ucp/aws/ucpcredentialprovider.go
```diff
@@ -33,6 +37,15 @@ var _ aws.CredentialsProvider = (*UCPCredentialProvider)(nil)
 const (
 	// DefaultExpireDuration is the default access key expiry duration.
 	DefaultExpireDuration = time.Minute * time.Duration(15)
+
+	// CredentialKind is IRSA
+	CredentialKindIRSA = "IRSA"
+	// CredentialKindAccessKey is AccessKey
```
vishwahiremat, line 43:
```suggestion
	// CredentialKind is AccessKey
```

vishwahiremat (approved):

pkg/ucp/aws/ucpcredentialprovider_test.go
```diff
@@ -53,23 +53,44 @@ func newMockProvider() *mockProvider {
...
+		cp = NewUCPCredentialProvider(p, DefaultExpireDuration)
+		p.fakeCredential.IRSACredential.RoleARN = ""
+
+		_, err = cp.Retrieve(context.TODO())
+		require.Error(t, err)
 	})
 
-	t.Run("valid credential", func(t *testing.T) {
-		p := newMockProvider()
+	t.Run("valid redential", func(t *testing.T) {
```
lakshmimsft, line 92: nit: credential

pkg/ucp/credentials/aws.go
```diff
@@ -71,6 +71,13 @@ func (p *AWSCredentialProvider) Fetch(ctx context.Context, planeName, name strin
 		default:
 			return nil, errors.New("invalid AWSAccessKeyCredentialProperties")
 		}
+	case *ucpapi.AwsIRSACredentialProperties:
+		switch c := p.Storage.(type) {
+		case *ucpapi.InternalCredentialStorageProperties:
+			storage = c
+		default:
+			return nil, errors.New("invalid AWSAccessKeyCredentialProperties")
+		}
```
lakshmimsft, line 80: is it possible to  combine both these?
eg. case *ucpapi.AwsAccessKeyCredentialProperties,  *ucpapi.AwsIRSACredentialProperties

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -43,6 +44,14 @@ const (
...
 	awsAccessKeyParam = "access_key"
 	awsSecretKeyParam = "secret_key"
+
+	// configs for AWS IRSA
+	// Ref: https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assuming-an-iam-role-using-a-web-identity
+	awsIRSAProvider = "assume_role_with_web_identity"
+	awsRoleARN      = "role_arn"
+	sessionName     = "session_name"
+	tokenFile       = "web_identity_token_file"
+	tokenFilePath   = "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"
```
kachawla, line 54: Could you add a comment about this value? Is it always constant or it depends on another configuration Radius sets? A link to relevant AWS docs would be helpful as well.

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -139,10 +151,33 @@ func (p *awsProvider) generateProviderConfigMap(credentials *credentials.AWSCred
...
+				config[awsSecretKeyParam] = credentials.AccessKeyCredential.SecretAccessKey
+			}
+
+		case ucp_datamodel.AWSIRSACredentialKind:
+			// Radius requests will first be routed to STS endpoint,
+			// where it will be validated and then the request to the specific service (such as S3) will be made using
+			// the bearer token from the STS response.
+			// Based on the https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html,
+			// STS endpoint should be region based, and in the same region as
+			// Radius instance to minimize latency associated with STS call and thereby improve performance.
```
kachawla, line 169: This comment seems out of place because the code below isn't making any calls to AWS, was this intended?

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -139,10 +151,33 @@ func (p *awsProvider) generateProviderConfigMap(credentials *credentials.AWSCred
...
+			// We should provide the user with ability to configure the STS endpoint region.
+			// For now, we are using the global STS endpoint, which is the default.
+			// Ref. https://github.com/radius-project/radius/issues/7747
+			if credentials.IRSACredential != nil && credentials.IRSACredential.RoleARN != "" {
+				config[awsIRSAProvider] = map[string]any{
+					awsRoleARN:  credentials.IRSACredential.RoleARN,
+					sessionName: "radius-terraform-" + uuid.New().String(),
```
kachawla, line 176: What is the significance of this session name? Could you share some details on the motivation for the format here?

pkg/ucp/aws/ucpcredentialprovider.go
```diff
@@ -73,19 +86,65 @@ func (c *UCPCredentialProvider) Retrieve(ctx context.Context) (aws.Credentials,
...
+				o.RoleSessionName = "radius-ucp-" + uuid.New().String()
+			},
+		))
+
+		value, err = credsCache.Retrieve(ctx)
+		if err != nil {
+			logger.Info(fmt.Sprintf("Failed to retrieve AWS Credential IRSA - %s", err.Error()))
+			return aws.Credentials{}, err
+		}
+		value.Source = "radiusucp"
```
kachawla, line 143: Where is "radiusucp" coming from?

pkg/recipes/terraform/config/providers/aws.go
nithyatsu: I have added "once we switch to region based STS endpoint, we should add the region to the config." to the comment. Hope that makes it clearer. issues/7747 is at a high level going to introduce a way to set region for both ucp and terraform, based on radius's location.

pkg/ucp/aws/ucpcredentialprovider.go
nithyatsu: I made it the same as the source we are setting for accesskey.. I understood it to be the source (holder) of credential and hence made it same.

pkg/ucp/credentials/aws.go
nithyatsu: nice one. But it does not like it when I tried, since in the next line we  have `switch c := p.Storage.(type)` and looks since now p can correspond to either of 2 values, it gives an error.

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -43,6 +44,16 @@ const (
 	awsRegionParam    = "region"
 	awsAccessKeyParam = "access_key"
 	awsSecretKeyParam = "secret_key"
+
```
lakshmimsft, line 47: do we want to add "radiusucp" , "radius-terraform-" etc as consts here?

nithyatsu: added

pkg/ucp/credentials/aws.go
lakshmimsft: let's pull common code into function and each switch case calls the function, eg:
func getStorageProperties(p any) (*InternalCredentialStorageProperties, error) {
	switch c := p.(type) {
	case *InternalCredentialStorageProperties:
		return c, nil
	default:
		return nil, errors.New("invalid AWS credential storage properties")
	}
}

pkg/ucp/credentials/types.go
```diff
@@ -56,3 +64,12 @@ type CredentialProvider[T any] interface {
 	// Fetch gets the credentials from secret storage.
 	Fetch(ctx context.Context, planeName, name string) (*T, error)
 }
+
+func getStorageProperties(p any) (*ucpapi.InternalCredentialStorageProperties, error) {
+	switch c := p.(type) {
+	case *ucpapi.InternalCredentialStorageProperties:
+		return c, nil
+	default:
+		return nil, errors.New("invalid AWS credential storage properties - field 'properties.storage' is not InternalCredentialStorageProperties")
```
lakshmimsft, line 73: Error msg is no longer specific to AWS

pkg/recipes/terraform/config/providers/aws.go
kachawla: I see, makes sense. We might want to consider operation id instead of uuid to map it to the corresponding request, doesn't need to happen in this PR because it will require changing API interface in a couple of places but something to look into as a follow up

kachawla (approved):

PR #7739, opened 2024-07-15
implement controllers for AWS IRSA

### LINKED ISSUE #7618: Add IRSA (workload identity) support for AWS cloud provider

nithyatsu:
# Description

Add convertor and controllers for the new IRSA credential type

## Type of change

Partially Fixes: https://github.com/radius-project/radius/issues/7618


swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AwsCredentials_List.json
kachawla, line 1: Did you mean remove these examples?

pkg/ucp/frontend/controller/credentials/azure/deleteazurecredential_test.go
```diff
@@ -132,7 +132,7 @@ func setupCredentialMocks(mockStorageClient store.MockStorageClient) {
 	datamodelCredential := datamodel.AzureCredential{
 		BaseResource: v1.BaseResource{},
 		Properties: &datamodel.AzureCredentialResourceProperties{
-			Kind: datamodel.AWSCredentialKind,
+			Kind: datamodel.AWSAccessKeyCredentialKind,
```
kachawla, line 135: I know this is unrelated to your PR, but do you know why we are testing with AWS kind for Azure credentials?

pkg/ucp/datamodel/credential.go
```diff
@@ -100,14 +102,30 @@ type AzureCredentialProperties struct {
 	WorkloadIdentity *AzureWorkloadIdentityCredentialProperties `json:"workloadIdentity,omitempty"`
 }
 
-// AWSCredentialProperties contains ucp AWS credential properties.
-type AWSCredentialProperties struct {
+// AWSAccessKeyCredentialProperties contains ucp AWS credential properties.
```
kachawla, line 105:
```suggestion
// AWSAccessKeyCredentialProperties contains ucp AWS access key credential properties.
```

pkg/ucp/credentials/types.go
```diff
@@ -34,6 +34,12 @@ const (
 
 	// AzureWorkloadIdentityCredentialKind represents the kind of Azure workload identity credential.
 	AzureWorkloadIdentityCredentialKind = ucp_dm.AzureWorkloadIdentityCredentialKind
+
+	// AWSAccessKeyCredentialKind represents the kind of AWS access key credential.
+	AWSAccessKeyCredentialKind = ucp_dm.AWSAccessKeyCredentialKind
+
+	// AWSIRSACredentialKind represents the kind of AWS IRSA credential.
+	AWSIRSACredentialKind = ucp_dm.AWSIRSACredentialKind
```
kachawla, line 42: Where are these constants used? Wondering why do we need to declare a new constant that references existing constant, instead of directly referencing the original definition.

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -41,6 +43,7 @@ const (
 	awsRegionParam    = "region"
 	awsAccessKeyParam = "access_key"
 	awsSecretKeyParam = "secret_key"
+	awsRoleARN        = "role_ARN"
```
kachawla, line 46: Thanks for figuring this out. Is this the format expected by Terraform? Curious to look at relevant docs if you have that handy.

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -118,9 +121,17 @@ func fetchAWSCredentials(ctx context.Context, awsCredentialsProvider credentials
...
-		logger.Info("AWS credentials are not registered, skipping credentials configuration.")
-		return nil, nil
+	switch credentials.Kind {
+	case ucp_datamodel.AWSAccessKeyCredentialKind:
+		if credentials.AccessKeyCredential == nil || credentials.AccessKeyCredential.AccessKeyID == "" || credentials.AccessKeyCredential.SecretAccessKey == "" {
+			logger.Info("AWS AccessKey credentials are not registered, skipping credentials configuration.")
+			return nil, nil
+		}
+	case ucp_datamodel.AWSIRSACredentialKind:
+		if credentials.IRSACredential == nil || credentials.IRSACredential.RoleARN == "" {
```
kachawla, line 131: Maybe I'm reading it wrong.. Shouldn't this be `and` operator instead of `or`? It will likely panic if we try to check `RoleArn` for nil `IRSACredential`. Same for the check above for access keys. This is also something that's not specific to this PR but just went unnoticed in previous iterations.


```suggestion
		if credentials.IRSACredential == nil && credentials.IRSACredential.RoleARN == "" {
```

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -132,9 +143,14 @@ func (p *awsProvider) generateProviderConfigMap(credentials *credentials.AWSCred
...
-		config[awsSecretKeyParam] = credentials.SecretAccessKey
+	if credentials != nil && credentials.Kind == ucp_datamodel.AWSAccessKeyCredentialKind && credentials.AccessKeyCredential != nil &&
+		credentials.AccessKeyCredential.AccessKeyID != "" && credentials.AccessKeyCredential.SecretAccessKey != "" {
+		config[awsAccessKeyParam] = credentials.AccessKeyCredential.AccessKeyID
+		config[awsSecretKeyParam] = credentials.AccessKeyCredential.SecretAccessKey
+	}
+
+	if credentials != nil && credentials.Kind == ucp_datamodel.AWSIRSACredentialKind && credentials.IRSACredential != nil && credentials.IRSACredential.RoleARN != "" {
+		config[awsRoleARN] = credentials.IRSACredential.RoleARN
 	}
```
kachawla, line 154: We can move `if credentials != nil` outside the two if loops, something like this:

```suggestion
if credentials != nil {
	if credentials.Kind == ucp_datamodel.AWSAccessKeyCredentialKind {
	    if credentials.AccessKeyCredential != nil &&
		credentials.AccessKeyCredential.AccessKeyID != "" && credentials.AccessKeyCredential.SecretAccessKey != "" {
		config[awsAccessKeyParam] = credentials.AccessKeyCredential.AccessKeyID
		config[awsSecretKeyParam] = credentials.AccessKeyCredential.SecretAccessKey
             }
	} else if credentials.Kind == ucp_datamodel.AWSIRSACredentialKind {
	        if credentials.IRSACredential != nil && credentials.IRSACredential.RoleARN != "" {
		config[awsRoleARN] = credentials.IRSACredential.RoleARN
           }
        } else {
         // Error out for unexpected credentials kind? Is this possible at this point?
        }
   }
```

pkg/recipes/terraform/config/providers/aws_test.go
```diff
@@ -45,24 +50,28 @@ type mockAWSCredentialsProvider struct {
...
+				AccessKeyID:     testAWSCredentials.AccessKeyCredential.AccessKeyID,
+				SecretAccessKey: testAWSCredentials.AccessKeyCredential.SecretAccessKey,
+			},
 		},
 	}
 }
 
 // Fetch returns mock AWS credentials for testing. It takes in a context, planeName and name and returns
 // an AWSCredential or an error if the credentials are empty.
+// TODO: update as part of IRSA support in Terraform provider
```
kachawla, line 64: Is this something you meant to handle in this PR?

swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AwsCredentials_List.json
nithyatsu: yes, since we renamed this to AwsCredentialsAccessKey_List.json and another similar for IRSA. But I have missed deleting these files in my previous tsp PR.

pkg/ucp/frontend/controller/credentials/azure/deleteazurecredential_test.go
nithyatsu: thats interesting.. looks like a mistake. Will fix this.

pkg/recipes/terraform/config/providers/aws.go
nithyatsu: https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assuming-an-iam-role-using-a-web-identity has the information. But I have not yet gotten to the terraform. I added this just so I can get the changes to compile with new model.

pkg/recipes/terraform/config/providers/aws_test.go
nithyatsu: No, I will add it as part of Terraform support for IRSA PR ( upcoming). I had to make some minimal changes just so I could compile with the changes to support the model.

pkg/ucp/api/v20231001preview/aws_credential_conversion.go
```diff
@@ -82,13 +82,47 @@ func (cr *AwsCredentialResource) getDataModelCredentialProperties() (*datamodel.
...
 			},
 			Storage: storage,
 		}, nil
+	case *AwsIRSACredentialProperties:
+		var storage *datamodel.CredentialStorageProperties
+
+		switch c := p.Storage.(type) {
+		case *InternalCredentialStorageProperties:
+			if c.Kind == nil {
+				return nil, &v1.ErrModelConversion{PropertyName: "$.properties", ValidValue: "not nil"}
```
rynowak, line 101: What's the right property path for this? It should be longer.

pkg/ucp/credentials/types.go
rynowak: This can be a nice pattern because it makes the package self-contained.

pkg/recipes/terraform/config/providers/aws.go
nithyatsu: I added the change. The else part is not neccessary, since if we PUT a credential with any kind other than these two then we will get an error 
```
$.properties.kind in body should be one of [AccessKey IRSA]"
```

pkg/ucp/credentials/types.go
nithyatsu: I added them to model azure, but looks like they are not needed. I removed both AWS and AZ constants that arent being currently used in our code.

pkg/ucp/frontend/controller/credentials/azure/deleteazurecredential_test.go
nithyatsu: chnaged to datamodel.AzureWorkloadIdentityCredentialKind

pkg/ucp/api/v20231001preview/aws_credential_conversion.go
nithyatsu: Updated it to .properties.storage.kind

pkg/ucp/credentials/types.go
nithyatsu: AZ constant was actually used in a test.. Removed just AWS constants that I introduced.

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -118,9 +121,17 @@ func fetchAWSCredentials(ctx context.Context, awsCredentialsProvider credentials
 		return nil, err
 	}
 
-	if credentials == nil || credentials.AccessKeyID == "" || credentials.SecretAccessKey == "" {
-		logger.Info("AWS credentials are not registered, skipping credentials configuration.")
-		return nil, nil
+	switch credentials.Kind {
+	case ucp_datamodel.AWSAccessKeyCredentialKind:
+		if credentials.AccessKeyCredential == nil || credentials.AccessKeyCredential.AccessKeyID == "" || credentials.AccessKeyCredential.SecretAccessKey == "" {
```
willdavsmith, line 126: what if `credentials` is nil here?

pkg/recipes/terraform/config/providers/aws.go
```diff
@@ -132,9 +143,18 @@ func (p *awsProvider) generateProviderConfigMap(credentials *credentials.AWSCred
...
+	if credentials != nil {
+		if credentials.Kind == ucp_datamodel.AWSAccessKeyCredentialKind {
+			if credentials.AccessKeyCredential != nil &&
+				credentials.AccessKeyCredential.AccessKeyID != "" && credentials.AccessKeyCredential.SecretAccessKey != "" {
+				config[awsAccessKeyParam] = credentials.AccessKeyCredential.AccessKeyID
+				config[awsSecretKeyParam] = credentials.AccessKeyCredential.SecretAccessKey
+			}
+		} else if credentials.Kind == ucp_datamodel.AWSIRSACredentialKind {
+			if credentials.IRSACredential != nil && credentials.IRSACredential.RoleARN != "" {
+				config[awsRoleARN] = credentials.IRSACredential.RoleARN
```
willdavsmith, line 155: is this how to enable irsa with terraform? can we link a terraform docs page here?

pkg/recipes/terraform/config/providers/aws_test.go
```diff
@@ -45,24 +50,28 @@ type mockAWSCredentialsProvider struct {
 func newMockAWSCredentialsProvider() *mockAWSCredentialsProvider {
```
willdavsmith, line 50: should we write unit tests for irsa here as well?

swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AWSCredential_CreateOrUpdate.json
willdavsmith, line 1: are the swagger changes coming in a different PR?

pkg/ucp/datamodel/credential.go
```diff
@@ -100,14 +102,30 @@ type AzureCredentialProperties struct {
...
+// AWSIRSACredentialProperties contains ucp AWS IRSA credential properties.
+type AWSIRSACredentialProperties struct {
+	// RoleARN contains aws role arn for irsa.
+	RoleARN string `json:"roleARN"`
+}
+
+// AWSCredentialProperties contains ucp AWS credential properties.
+type AWSCredentialProperties struct {
+	// Kind is the kind of AWS credential.
+	Kind string `json:"kind,omitempty"`
```
kachawla, line 122: With this update, `kind` here https://github.com/radius-project/radius/blob/3924c62fdc5207087c7c3a7db37263b45fcbe66b/pkg/ucp/datamodel/credential.go#L71 seems redundant. What's the motivation to keep both? I know this is done the same way for Azure credentials too, but just went unnoticed then.

pkg/ucp/credentials/types.go
kachawla: > This can be a nice pattern because it makes the package self-contained. Consumers of the package can work with a single import if its done right.

@rynowak constants were being copied from another package by referencing them, which defeats the purpose of making the package self-contained (if that's the goal). Ideally, I would abstract all of them out in a separate package rather than duplicating them if we really need same constants to be referenced from different packages.

kachawla: > AZ constant was actually used in a test.. Removed just AWS constants that I introduced.

@nithyatsu could you point me to the test that is using it? Trying to understand why datamodel package can't be referenced from the test since there aren't any tests within this credentials package. Can't seem to find references from github dev editor

pkg/recipes/terraform/config/providers/aws.go
kachawla: are you planning on completing it in a follow up PR or as a part of this PR? If a follow up PR is needed, then I'd recommend returning an error message that explicitly states that it is unsupported. Leaving it in half baked state might result into error messages which are difficult to make sense out of.

pkg/recipes/terraform/config/providers/aws_test.go
kachawla: I think this answers my question here: https://github.com/radius-project/radius/pull/7739#discussion_r1678790947. I'd suggest returning an error instead of setting just one property which will indirectly result into an error. If this is just temporary and we are planning to complete it before release, then I'm fine with either approach as long as we add clear comment about it.

pkg/recipes/terraform/config/providers/aws.go
kachawla: makes sense, thanks

pkg/ucp/datamodel/credential.go
```diff
@@ -100,14 +102,30 @@ type AzureCredentialProperties struct {
...
+	// AccessKeyCredential represents the access key credential properties.
+	AccessKeyCredential *AWSAccessKeyCredentialProperties `json:"accesskey,omitempty"`
+	// IRSA represents the irsa credential properties.
```
rynowak, line 125:
```suggestion
	// IRSACredential represents the irsa credential properties.
```

pkg/recipes/terraform/config/providers/aws_test.go
nithyatsu: yes, I am going to add UT to aws_test as part of PR that supports terraform.

swagger/specification/ucp/resource-manager/UCP/preview/2023-10-01-preview/examples/AWSCredential_CreateOrUpdate.json
nithyatsu: it got merged https://github.com/radius-project/radius/pull/7708 . I had missed removing files that are not relevant anymore. We have AWSCredential_CreateOrUpdateAccessKey.json and AWSCredential_CreateOrUpdateIRSA.json instead which went into the mentioned PR.

pkg/ucp/credentials/types.go
nithyatsu: /radius/pkg/azure/credential/ucpcredentials_test.go line 44
We just have defined all constants in the types.go, and instead of importing the datamodel package, imported just `sdk_cred "github.com/radius-project/radius/pkg/ucp/credentials"`

pkg/recipes/terraform/config/providers/aws_test.go
nithyatsu: I was thinking since we use it only in the case where the type IRSA credential we might be OK, but it makes sense to not take risk. let me see if I can remove this property as well, and just retain the changes that use the new type.

pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential.go
```diff
@@ -84,7 +84,9 @@ func (c *CreateOrUpdateAWSCredential) Run(ctx context.Context, w http.ResponseWr
 	}
 
 	// Do not save the secret in metadata store.
-	newResource.Properties.AWSCredential.SecretAccessKey = ""
+	if newResource.Properties.AWSCredential.Kind == datamodel.AWSAccessKeyCredentialKind {
+		newResource.Properties.AWSCredential.AccessKeyCredential.SecretAccessKey = ""
+	}
```
kachawla, line 89: Do we have an existing unit test for this check (to make sure we don't accidentally start storing secrets)?

pkg/ucp/datamodel/credential.go
nithyatsu: AWSCredentialResourceProperties is a UCPResource, and has its "kind", along with storage and the actual AWS IAM.  AWSCredentialProperties is the AWS IAM representation. But since we support 2 types of IAM ( access key and irsa), the kind in this structure helps us to navigate to the right kind of credential (AccessKeyCredential versus IRSACredential). We could just check which one is not nil, but it feels neat to have a kind.

pkg/recipes/terraform/config/providers/aws_test.go
nithyatsu: I removed the todo, since partial terraform changes were causing confusion like you and will mentioned in another comment. will revisit it as part of terraform PR

pkg/recipes/terraform/config/providers/aws.go
nithyatsu: updated code so we dont have this condition for now.

pkg/recipes/terraform/config/providers/aws.go
nithyatsu: removed terraform related changes ref. https://github.com/radius-project/radius/pull/7739#discussion_r1679825544

pkg/recipes/terraform/config/providers/aws_test.go
nithyatsu: same as https://github.com/radius-project/radius/pull/7739#discussion_r1679825544

pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential.go
nithyatsu: the tests in /radius/pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential_test.go use  func getAwsResponse() armrpc_rest.Response as the expected response, which makes sure the structure does not have the secret.

pkg/ucp/datamodel/credential.go
kachawla: Thanks for clarifying. Do we have a use case where we consume [UCP kind](https://github.com/radius-project/radius/blob/3924c62fdc5207087c7c3a7db37263b45fcbe66b/pkg/ucp/datamodel/credential.go#L71) in a way that switching to consuming the [inner kind on the properties](https://github.com/radius-project/radius/blob/3924c62fdc5207087c7c3a7db37263b45fcbe66b/pkg/ucp/datamodel/credential.go#L122) won't be possible or complicated?

pkg/recipes/terraform/config/providers/aws_test.go
kachawla: makes sense, thanks. I think the TODO didn't get added to the commit, you might need to re-add and push. :) Updates in `aws.go` are reflecting as expecting.

kachawla (dismissed):
Thanks for addressing all the feedback. Looks good, please make sure to remove the [TODO](https://github.com/radius-project/radius/pull/7739#discussion_r1679888536). I have one open [question](https://github.com/radius-project/radius/pull/7739#discussion_r1679880506) but it is not blocking for this PR.

pkg/ucp/datamodel/credential.go
nithyatsu: I am not sure I see any currently. @rynowak , @willdavsmith , is there a use case because of which we modeled kind in both AzureCredentialResourceProperties as well as AzureCredentialProperties  ?

kachawla (approved):

PR #7741, opened 2024-07-17
Adding rad install changes for AWS IRSA support

vishwahiremat:
# Description

Added changes in deployment yaml to enable `global.awsIRSA.enabled` to true it is set using `rad install kubernetes` command

## Type of change

Fixes: #issue_number


nithyatsu (approved):
LGTM

deploy/Chart/templates/rp/deployment.yaml
```diff
@@ -69,6 +69,10 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/config
+        {{- if eq .Values.global.awsIRSA.enabled true }}
+        - name: aws-iam-token
+          mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
```
nithyatsu, line 74: the parameter name should be global.aws.irsa.enabled , based on ryan's comment..

vishwahiremat: updated it!

deploy/Chart/values.yaml
```diff
@@ -26,6 +26,11 @@ global:
   azureWorkloadIdentity:
     enabled: false
 
+  # Configure global.aws.irsa.enabled=true to enable AWS IRSA.
+  # Disabled by default.
```
nithyatsu, line 30: this should be similar to
aws:
  irsa:
    enabled: true

lakshmimsft (dismissed):

deploy/Chart/templates/rp/deployment.yaml
```diff
@@ -69,6 +69,10 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/config
+        {{- if eq .Values.global.aws.irsa.enabled true }}
+        - name: aws-iam-token
+          mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
```
kachawla, line 74: @nithyatsu did we say during design discussion that IRSA can be enabled on non EKS clusters as well? Would this mount path still be applicable on other clusters?

deploy/Chart/templates/rp/deployment.yaml
```diff
@@ -69,6 +69,10 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/config
+        {{- if eq .Values.global.aws.irsa.enabled true }}
+        - name: aws-iam-token
```
kachawla, line 73:
```suggestion
        - name: aws-irsa-token
```

deploy/Chart/templates/rp/deployment.yaml
```diff
@@ -83,6 +87,15 @@ spec:
         - name: config-volume
           configMap:
             name: applications-rp-config
+        {{- if eq .Values.global.aws.irsa.enabled true }}
+        - name: aws-iam-token
```
kachawla, line 91:
```suggestion
        - name: aws-irsa-token
```

deploy/Chart/templates/rp/deployment.yaml
```diff
@@ -83,6 +87,15 @@ spec:
...
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: token
+                expirationSeconds: 86400
```
kachawla, line 96: @nithyatsu @vishwahiremat could you share details on how did we decide on this number?

deploy/Chart/templates/rp/deployment.yaml
nithyatsu: Hi Karishma, IRSA is not specific to eks clusters. But the first iteration of the feature is based on eks. 
We would have to make the token path configurable in order to support it on other clusters too. Would it be OK to log a ticket to cover this?

deploy/Chart/templates/rp/deployment.yaml
nithyatsu: Hi Karishma, the typical approach for IRSA was to use the pod identity webhook. (
ref. https://github.com/radius-project/design-notes/blob/3b3f66e7b6d340ef7beebcc4212fcd7129bf8b06/cli/2024-06-04-aws-irsa-support.md#alternatives-considered , https://github.com/aws/amazon-eks-pod-identity-webhook#amazon-eks-pod-identity-webhook)

The reason we did not choose that was since it wasnt compatible with multi-tenancy and needed service restarts to register credential. But, we did the exact same thing the webhook does. These are the configurations the webhook sets (https://github.dev/aws/amazon-eks-pod-identity-webhook#amazon-eks-pod-identity-webhook) and mentions it as a default value.

deploy/Chart/templates/ucp/deployment.yaml
```diff
@@ -63,6 +63,10 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/config
+        {{- if eq .Values.global.aws.irsa.enabled true }}
+        - name: aws-iam-token
```
lakshmimsft, line 67: here and on line 85, aws-irsa-token

lakshmimsft:
let's pls rerun tests. we haven't seen a clean run so far.

deploy/Chart/templates/rp/deployment.yaml
vishwahiremat: as discussed offline keeping it `aws-iam-token`

deploy/Chart/templates/rp/deployment.yaml
vishwahiremat: as discussed offline keeping it `aws-iam-token`

deploy/Chart/templates/ucp/deployment.yaml
vishwahiremat: we are not changing it, so keeping it as is

lakshmimsft:
LTGM. Pls get another approval as well.

lakshmimsft (approved):

deploy/Chart/templates/rp/deployment.yaml
kachawla: yeah that sounds good as long as it is intentional and being tracked

kachawla (approved):

deploy/Chart/templates/rp/deployment.yaml
nithyatsu: created https://github.com/radius-project/radius/issues/7774 for tracking this.

rynowak: Is this actually something we'd have to change for another cluster?

From what I understand, the AWS SDKS will use the value of [AWS_WEB_IDENTITY_TOKEN_FILE](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-envvars.html) to load the token. There's nothing special about this file path (or any other file path) as long as the environment variable points to the token on disk.

nithyatsu: In code, we added the token file location as a const: https://github.com/radius-project/radius/blob/007f7e68fea64a93103c18274aeefaacd3a6a217/pkg/recipes/terraform/config/providers/aws.go#L56

The mount path should work, but it felt odd to reference eks in path for a cluster that was not on eks. Or is it OK since IRSA is a AWS feature? Also, instead of const, should I have looked to set the env variable for the pods?

rynowak: IMO we should do the following:

- Set the env-var `AWS_WEB_IDENTITY_TOKEN_FILE` to `/var/run/secrets/eks.amazonaws.com/serviceaccount/token` in the helm chart when IRSA is enabled. This is also done by the webhook [here](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/bd3e26dba0be6d8f8ac26578defa94582b026324/pkg/handler/handler.go#L244).
- We can simplify the logic for Terraform because Terraform also uses `AWS_WEB_IDENTITY_TOKEN_FILE`. [link](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#container-credentials)

Then it doesn't matter what the path is, because it's not hardcoded.

nithyatsu: updated the ticket https://github.com/radius-project/radius/issues/7774 to implement this comment.

PR #7748, opened 2024-07-19
Adding multi arch container image generation for magpie and testrp

### LINKED ISSUE #7743: Publish magpiego as a multi-arch image
### Area for Improvement

Functional testing

### Observed behavior

`magpiego` is an image we use for functional testing. Right now this is only published as an amd64 image. 

For users on MacOS running the functional tests locally will fail with `exec format error` unless they have Rosetta installed. 

### Desired behavior

Functional testing just works. 

### Proposed Fix

Publish `magpiego` as a multi-architecture image.

### rad Version

aaaa

### Operating system

MacOS + M1 chip

### Additional context

Found while debugging with @superbeeny 

[AB#12849](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12849)

ytimocin:
# Description
- Removed `make docker-test-image-build` and `make docker-test-image-push` because they were creating single arch builds.
- Added **testrp** and **magpie** as part of multi arch builds: https://github.com/radius-project/radius/pull/7748/files#diff-267abdbaab3ee4a767a1905f07cb0033220b32d4a8c035502fe24dad37ee2e03R105-R106.
- Added **testrp** and **magpie** as part of multi arch builds and pushes: https://github.com/radius-project/radius/pull/7748/files#diff-e5017862a52d1de76ba232f290f53785637620580359895b37c593f181afc6c1R115-R119.
- Updated Dockerfiles for ucpd, applications-rp, and controller:
  - Removed unnecessary `TARGETPLATFORM:-linux/amd64` because that was throwing a warning stating that the TARGETPLATFORM is already `linux/amd64` by default if not provided.
 - Updated Dockerfiles for testrp and magpiego:
   - Made them more like the other Dockerfiles: ucpd, applications-rp, and controller.
 - Updated dependencies for magpiego (because I was there).
 - Reformatted a few files.

## Type of change
Fixes: #7743


build/docker.mk
```diff
@@ -108,9 +108,15 @@ $(foreach IMAGE,$(DOCKER_IMAGES),$(eval $(call generateDockerMultiArches,$(IMAGE
...
 
+# multi arch container image targets for magpie
+$(eval $(call generateDockerMultiArches,magpiego,./test/magpiego/,./test/magpiego/Dockerfile))
+
 # testrp comes from our test directory.
 $(eval $(call generateDockerTargets,testrp,./test/testrp/,./test/testrp/Dockerfile))
 
+# multi arch container image targets for testrp
+$(eval $(call generateDockerMultiArches,testrp,./test/testrp/,./test/testrp/Dockerfile))
+
```
rynowak, line 119: What do we need to do to make sure these targets get called?

.github/workflows/build.yaml
```diff
@@ -190,23 +190,20 @@ jobs:
           platforms: linux/amd64,linux/arm64,linux/arm/v7
       - name: Push container images (latest)
         run: |
-          make docker-test-image-build && make docker-test-image-push
```
rynowak, line 193: Have these targets been consolidated into `make docker-multi-arch-push`?

build/build.mk
```diff
@@ -87,23 +87,35 @@ define generatePlatformBuildTarget
 build-$(3)-$(1)-$(2):
   $(eval BINS_OUT_DIR_$(1)_$(2) := $(OUT_DIR)/$(1)_$(2)/$(BUILDTYPE_DIR))
 	@echo "$(ARROW) Building $(3) on $(1)/$(2) to $(BINS_OUT_DIR_$(1)_$(2))/$(3)$(BINARY_EXT)"
-	GOOS=$(1) GOARCH=$(2) go build \
+	cd $(4) && GOOS=$(1) GOARCH=$(2) go build \
```
rynowak, line 90: suggestion: rather than changing directories, pass the directory in to `go build`.

build/build.mk
```diff
@@ -87,23 +87,35 @@ define generatePlatformBuildTarget
...
+		-o $(CURDIR)/$(BINS_OUT_DIR_$(1)_$(2))/$(3)$(BINARY_EXT)
 endef
 
 # defines a target for each binary
 GOOSES := darwin linux windows
 GOARCHES := amd64 arm arm64
-BINARIES := docgen rad applications-rp ucpd controller
-$(foreach ITEM,$(BINARIES),$(eval $(call generateBuildTarget,$(ITEM),./cmd/$(ITEM))))
-$(foreach ARCH,$(GOARCHES),$(foreach OS,$(GOOSES),$(foreach ITEM,$(BINARIES),$(eval $(call generatePlatformBuildTarget,$(OS),$(ARCH),$(ITEM),./cmd/$(ITEM))))))
+BINARIES := docgen:./cmd/docgen \
```
rynowak, line 100: can you leave a comment here to document the syntax this expects?

build/build.mk
```diff
@@ -87,23 +87,35 @@ define generatePlatformBuildTarget
...
+	rad:./cmd/rad \
+	applications-rp:./cmd/applications-rp \
+	ucpd:./cmd/ucpd \
+	controller:./cmd/controller \
+	testrp:./test/testrp \
+	magpiego:./test/magpiego
```
rynowak, line 106: Some clarifying questions.. These are all good fodder for some improved comments for the existing logic.

So basically you're adding the test binaries and the paths to this data structure?

Then we use macros to parse these and generate a set of build targets dynamically.

Would be great to see an example. Each entry => these build targets.

build/build.mk
```diff
@@ -87,23 +87,35 @@ define generatePlatformBuildTarget
...
+$(eval ENTRYPOINT := $(shell echo $(1) | cut -d: -f2))
+endef
+
+$(foreach ITEM,$(BINARIES),$(eval $(call parseBinary,$(ITEM)) $(call generateBuildTarget,$(NAME),$(ENTRYPOINT))))
+
+$(foreach ARCH,$(GOARCHES),$(foreach OS,$(GOOSES),$(foreach ITEM,$(BINARIES),$(eval $(call parseBinary,$(ITEM)) $(call generatePlatformBuildTarget,$(OS),$(ARCH),$(NAME),$(ENTRYPOINT))))))
 
 # list of 'outputs' to build for all binaries
-BINARY_TARGETS:=$(foreach ITEM,$(BINARIES),build-$(ITEM))
+BINARY_TARGETS := $(foreach ITEM,$(BINARIES),$(eval $(call parseBinary,$(ITEM))) build-$(NAME))
```
rynowak, line 118: And this aggregates all of the build targets for the current architecture right?

rynowak: Do we need the equivalent for the "platform" targets?

build/docker.mk
```diff
@@ -23,7 +23,8 @@ IMAGE_SRC?=https://github.com/radius-project/radius
 # Generate a target for each image we define
 # Params:
 # $(1): the image name for the target
-# $(2): the Dockerfile path
+# $(2): additional arguments (if it is a dot, it means the current directory)
```
rynowak, line 26: Is this actually `additional arguments`? or is it the "context directory"? https://docs.docker.com/build/building/context/

build/docker.mk
```diff
@@ -57,6 +58,7 @@ endef
 define generateDockerMultiArches
 .PHONY: docker-multi-arch-build-$(1)
 docker-multi-arch-build-$(1): build-$(1)-linux-arm64 build-$(1)-linux-amd64 build-$(1)-linux-arm
+	@echo "docker-multi-arch-build-$(1): build-$(1)-linux-arm64 build-$(1)-linux-amd64 build-$(1)-linux-arm"
```
rynowak, line 61: Do we need to print this? how does it help users? it's just a list of targets.

build/docker.mk
```diff
@@ -97,50 +99,48 @@ configure-buildx:
...
+APPS_MAP := ucpd:./deploy/images/ucpd \
+	applications-rp:./deploy/images/applications-rp \
+	controller:./deploy/images/controller \
+	testrp:./test/testrp \
+	magpiego:./test/magpiego
+
+# Function to extract name and Dockerfile from the app string
+define parseApp
+$(eval NAME := $(shell echo $(1) | cut -d: -f1))
+$(eval DOCKERFILE_PATH := $(shell echo $(1) | cut -d: -f2))
```
rynowak, line 112: Nit, this is the directory containing the Dockerfile, not the path. The path would include `/Dockerfile`.

build/docker.mk
```diff
@@ -97,50 +99,48 @@ configure-buildx:
...
 
-# magpie comes from our test directory.
-$(eval $(call generateDockerTargets,magpiego,./test/magpiego/,./test/magpiego/Dockerfile))
+# This command will dynamically generate the targets for each image in the APPS_MAP list.
+$(foreach APP,$(APPS_MAP),$(eval $(call parseApp,$(APP)) $(call generateDockerTargets,$(NAME),.,$(DOCKERFILE_PATH)/Dockerfile,go)))
 
-# testrp comes from our test directory.
-$(eval $(call generateDockerTargets,testrp,./test/testrp/,./test/testrp/Dockerfile))
+# This command will dynamically generate the multi-arch targets for each image in the APPS_MAP list.
+$(foreach APP,$(APPS_MAP),$(eval $(call parseApp,$(APP)) $(call generateDockerMultiArches,$(NAME),.,$(DOCKERFILE_PATH)/Dockerfile)))
```
rynowak, line 119: Comments here are much much better compared to `build.mk`. Can you try to make that better too since you're changing it?

deploy/images/applications-rp/Dockerfile
```diff
@@ -1,22 +1,25 @@
-# Note: Using Alpine image to enable Git installation for Terraform module downloads.
+# Use Alpine image to enable Git installation for Terraform module downloads.
 # Switch to distroless when Terraform execution is moved to a separate container.
-FROM --platform=${TARGETPLATFORM:-linux/amd64} alpine:3.18
```
rynowak, line 3: Is this logic built in?

test/magpiego/Dockerfile
```diff
@@ -1,21 +1,14 @@
...
+WORKDIR /
 
-#command to start the magpie as a server to listen at port 3000
-CMD [ "./magpiego" ]
\ No newline at end of file
+# Copy the application binary for the specified architecture
+COPY ./linux_${TARGETARCH:-amd64}/release/magpiego /
+
+# Set the entrypoint to the application binary
+ENTRYPOINT ["/magpiego"]
```
rynowak, line 14: Thanks for simplifying all of this.

rynowak (commented):
Overall this looks right. Left some smaller comments and suggestions.

.github/workflows/build.yaml
ytimocin: Yes, they have.

build/docker.mk
ytimocin: No, this was for debugging purposes and I have forgotten this here. I removed it now.

build/build.mk
ytimocin: Added some comments. Let me know if those are clear. If not, I can add more.

build/build.mk
ytimocin: The next line creates the necessary targets for different platforms for all the items in the list.

deploy/images/applications-rp/Dockerfile
ytimocin: In CI/CD I saw this warning with this line: `RedundantTargetPlatform: Setting platform to predefined ${TARGETPLATFORM:-linux/amd64} in FROM is redundant as this is the default behavior (line 2)`. And I decided to remove the `platform`.

Also: https://docs.docker.com/reference/build-checks/redundant-target-platform/.

rynowak: Very cool 👍

build/build.mk
rynowak: What do you think of this idea? `go build ./my/directory` instead of `cd ./my/directory && go build`.

build/build.mk
```diff
@@ -87,23 +97,44 @@ define generatePlatformBuildTarget
...
-BINARIES := docgen rad applications-rp ucpd controller
-$(foreach ITEM,$(BINARIES),$(eval $(call generateBuildTarget,$(ITEM),./cmd/$(ITEM))))
-$(foreach ARCH,$(GOARCHES),$(foreach OS,$(GOOSES),$(foreach ITEM,$(BINARIES),$(eval $(call generatePlatformBuildTarget,$(OS),$(ARCH),$(ITEM),./cmd/$(ITEM))))))
 
-# list of 'outputs' to build for all binaries
-BINARY_TARGETS:=$(foreach ITEM,$(BINARIES),build-$(ITEM))
+# List of binaries to build.
+# Format: binaryName:binaryMainDirectory
+# Example: docgen:./cmd/docgen
+# Dockerfile should be present in the binaryMainDirectory.
```
rynowak, line 114: This doesn't seem right. We don't keep the dockerfile right next to the `main.go`.

test/magpiego/Dockerfile
```diff
@@ -1,21 +1,13 @@
-# Issues on M1 mac: https://github.com/docker/for-mac/issues/5831
-# it is a multistage docker creation for small magpie image
-FROM --platform=${TARGETPLATFORM:-linux/amd64} golang:1.22-alpine
-WORKDIR /usr/src/tmpapp
+FROM golang:1.22-alpine AS builder
```
rynowak, line 1: nit: can you remove `AS builder`? We're not going a two-phase build anymore.

rynowak (changes requested):
A few more really minor things, looking really good!

build/build.mk
ytimocin: Removed this one. My bad.

rynowak (approved):

PR #7758, opened 2024-07-23
Use local aws config in cli to get account and regions

vishwahiremat:
# Description

Today for `rad init --full` command while configuring `aws` provider, we are using `accesskey` and `secretkey` provided by user to get accountid and regions, but for `irsa` we cannot get the account id and region information.

So instead of using `accesskey` and `secretkey` to authenticate , updated it to use the local aws config to authenticate and get account id and regions informaton.

## Type of change

Fixes: #issue_number


pkg/cli/aws/client.go
```diff
@@ -43,13 +43,18 @@ type client struct{}
...
-	credentialsProvider := credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")
-	stsClient := sts.New(sts.Options{
-		Region:      region,
-		Credentials: credentialsProvider,
-	})
-	result, err := stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+func (c *client) GetCallerIdentity(ctx context.Context, region string) (*sts.GetCallerIdentityOutput, error) {
+	// Load the AWS SDK config and credentials
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile("default"))
+	if err != nil {
```
nithyatsu, line 49: should "default" be a const?

pkg/cli/aws/client.go
```diff
@@ -43,13 +43,18 @@ type client struct{}
...
 // GetCallerIdentity gets information about the provided credentials.
-func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKeyID string, secretAccessKey string) (*sts.GetCallerIdentityOutput, error) {
-	credentialsProvider := credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")
-	stsClient := sts.New(sts.Options{
-		Region:      region,
-		Credentials: credentialsProvider,
-	})
-	result, err := stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+func (c *client) GetCallerIdentity(ctx context.Context, region string) (*sts.GetCallerIdentityOutput, error) {
+	// Load the AWS SDK config and credentials
```
nithyatsu, line 47: do we need to update docs so the user knows by default rad cli will pick credentials from aws config file under "default" profile?

pkg/cli/aws/client.go
```diff
@@ -58,13 +63,19 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
-func (c *client) ListRegions(ctx context.Context, region string, accessKeyID string, secretAccessKey string) (*ec2.DescribeRegionsOutput, error) {
-	credentialsProvider := credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")
-	ec2Client := ec2.New(ec2.Options{
-		Region:      region,
-		Credentials: credentialsProvider,
-	})
-	result, err := ec2Client.DescribeRegions(ctx, &ec2.DescribeRegionsInput{})
+func (c *client) ListRegions(ctx context.Context, region string) (*ec2.DescribeRegionsOutput, error) {
+	// Load the AWS SDK config and credentials
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile("default"))
```
nithyatsu, line 68: should "default" be a const?

pkg/cli/cmd/radinit/aws.go
```diff
@@ -71,8 +71,8 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
 	}, nil
 }
 
-func (r *Runner) getAccountId(ctx context.Context, accessKeyID, secretAccessKey string) (string, error) {
-	callerIdentityOutput, err := r.awsClient.GetCallerIdentity(ctx, QueryRegion, accessKeyID, secretAccessKey)
+func (r *Runner) getAccountId(ctx context.Context) (string, error) {
+	callerIdentityOutput, err := r.awsClient.GetCallerIdentity(ctx, QueryRegion)
 	if err != nil {
 		return "", clierrors.MessageWithCause(err, "AWS credential verification failed.")
 	}
```
nithyatsu, line 78: here, would it make sense to add some detail, like "AWS credential verification failed. Please use `aws configure` to  configre credentials and then try again ? cc @Reshrahim

pkg/cli/aws/client.go
kachawla, line 68: Is `config.WithSharedConfigProfile("default")` for default config as well? Doesn't `LoadDefaultConfig` do that automatically? https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#loading-aws-shared-configuration

pkg/cli/aws/client.go
```diff
@@ -58,13 +63,19 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create an EC2 client
+	svc := ec2.NewFromConfig(cfg)
```
kachawla, line 74: what does `svc` stand for?

pkg/cli/aws/client.go
```diff
@@ -58,13 +63,19 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
+
+	// Call DescribeRegions
+	input := &ec2.DescribeRegionsInput{}
```
kachawla, line 77: What's the motivation for initializing empty input?

pkg/cli/aws/client.go
```diff
@@ -58,13 +63,19 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
+	// Load the AWS SDK config and credentials
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile("default"))
+	if err != nil {
+		return nil, err
+	}
+
+	// Create an EC2 client
+	svc := ec2.NewFromConfig(cfg)
+
+	// Call DescribeRegions
```
kachawla, line 76: nit: this comment isn't really providing any additional context

pkg/cli/aws/client.go
kachawla, line 49: Switching this to use local configuration is more nuanced than the call to list regions. The credentials registered with the environment do not necessarily need to belong to the same account as the account the user has setup locally, and it will certainly be different for all production scenarios. So I'd suggest making this a user input. Please bounce it by @Reshrahim @willtsai to confirm if there are any concerns with updating this user experience.

pkg/cli/aws/client.go
```diff
@@ -58,13 +63,19 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
 
 // ListRegions lists the AWS regions available (fetched from EC2.DescribeRegions API).
-func (c *client) ListRegions(ctx context.Context, region string, accessKeyID string, secretAccessKey string) (*ec2.DescribeRegionsOutput, error) {
-	credentialsProvider := credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")
-	ec2Client := ec2.New(ec2.Options{
-		Region:      region,
-		Credentials: credentialsProvider,
-	})
-	result, err := ec2Client.DescribeRegions(ctx, &ec2.DescribeRegionsInput{})
+func (c *client) ListRegions(ctx context.Context, region string) (*ec2.DescribeRegionsOutput, error) {
```
kachawla, line 66: Is the `region` parameter used anymore?

```suggestion
func (c *client) ListRegions(ctx context.Context) (*ec2.DescribeRegionsOutput, error) {
```

pkg/cli/aws/client.go
```diff
@@ -43,13 +43,18 @@ type client struct{}
...
-func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKeyID string, secretAccessKey string) (*sts.GetCallerIdentityOutput, error) {
-	credentialsProvider := credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")
-	stsClient := sts.New(sts.Options{
-		Region:      region,
-		Credentials: credentialsProvider,
-	})
-	result, err := stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+func (c *client) GetCallerIdentity(ctx context.Context, region string) (*sts.GetCallerIdentityOutput, error) {
+	// Load the AWS SDK config and credentials
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile("default"))
```
ytimocin, line 48: Can we ever use anything other than **default**?

pkg/cli/aws/client.go
```diff
@@ -43,13 +43,18 @@ type client struct{}
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create an STS client
```
ytimocin, line 53: What is an STS Client?

pkg/cli/aws/client.go
```diff
@@ -43,13 +43,18 @@ type client struct{}
...
+	svc := sts.NewFromConfig(cfg)
+
+	// Call GetCallerIdentity
```
ytimocin, line 56: nit: remove

ytimocin (changes requested):

pkg/cli/aws/client.go
vishwahiremat: Security Token Service  Client and it is generally used when short-term access is needed to privileged AWS resources

pkg/cli/aws/client.go
vishwahiremat: removed the `config.WithSharedConfigProfile`

pkg/cli/aws/client.go
vishwahiremat: removed the `config.WithSharedConfigProfile`

pkg/cli/aws/client.go
vishwahiremat: thanks for pointing this, removed config with sharedProfile

pkg/cli/aws/client.go
vishwahiremat: updated it to ec2Client

pkg/cli/aws/client.go
vishwahiremat: updated it

kachawla: the update doesn't seem to be reflected here

pkg/cli/aws/client.go
kachawla: Chatted offline with @vishwahiremat and @willtsai, agreed to go forward with the following approach -

- User provides irsa role to register
- Radius fetches aws account id for the credentials configured in users local aws config. This account could be different from the account id that the irsa role belongs to.
      - Ask user if this account id should be configured with the environment, or do they want to provide another account id.
- List region and selection experience stays the same, internally we will fetch it using locally configured aws credentials. It will fail if no local aws credentials are configured.
 
Here's a refresher on the current state:
- User provides aws access key and secret key
- Radius fetches corresponding account id using the provides access key and configures that on the environment
- Radius lists regions and asks the user to pick a region

pkg/cli/cmd/radinit/aws.go
```diff
@@ -27,15 +28,15 @@ import (
...
-
 	selectAWSRegionPrompt                 = "Select the region you would like to deploy AWS resources to:"
 	enterAWSIAMAcessKeyIDPrompt           = "Enter the IAM access key id:"
 	enterAWSIAMAcessKeyIDPlaceholder      = "Enter IAM access key id..."
 	enterAWSIAMSecretAccessKeyPrompt      = "Enter your IAM Secret Access Key:"
 	enterAWSIAMSecretAccessKeyPlaceholder = "Enter IAM secret access key..."
 	errNotEmptyTemplate                   = "%s cannot be empty"
+	confirmAWSAccountIDPromptFmt          = "Use account id '%v'?"
+	enterAWSAccountIDPrompt               = "Enter the account ID you want to use:"
+	enterAWSAccountIDPlaceholder          = "Enter the account ID you want to use.."
```
kachawla, line 39: did you mean to remove the placeholder?

kachawla, line 39: Could you please check that the format/phrasing is consistent with rest of the prompts?

```suggestion
	enterAWSAccountIDPrompt               = "Enter the account ID:"
	enterAWSAccountIDPlaceholder          = "Enter the account ID you want to use..."
```

pkg/cli/cmd/radinit/aws.go
```diff
@@ -53,12 +54,24 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
 	}
 
-	accountId, err := r.getAccountId(ctx, accessKeyID, secretAccessKey)
+	accountId, err := r.getAccountId(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	region, err := r.selectAWSRegion(ctx, QueryRegion, accessKeyID, secretAccessKey)
+	addingCloudProvider, err := prompt.YesOrNoPrompt(fmt.Sprintf(confirmAWSAccountIDPromptFmt, accountId), prompt.ConfirmYes, r.Prompter)
```
kachawla, line 62: this code is a bit difficult to understand, what is intended usage of `addingCloudProvider`? Isn't this decided based on the inputs for `confirmCloudProviderPrompt` and `selectCloudProviderPrompt` prompts?

pkg/cli/cmd/radinit/aws.go
```diff
@@ -53,12 +54,24 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
+	if err != nil {
+		return nil, err
```
kachawla, line 64: what would cause this to error out?

pkg/cli/cmd/radinit/aws.go
```diff
@@ -53,12 +54,24 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
+	}
+
+	if !addingCloudProvider {
+		accountId, err = r.Prompter.GetTextInput("Enter the account ID you want to use: ", prompt.TextInputOptions{Placeholder: enterAWSAccountIDPlaceholder})
```
kachawla, line 68: use the constant defined above for the prompt text

pkg/cli/cmd/radinit/aws.go
```diff
@@ -84,8 +97,8 @@ func (r *Runner) getAccountId(ctx context.Context, accessKeyID, secretAccessKey
 	return *callerIdentityOutput.Account, nil
 }
 
-func (r *Runner) selectAWSRegion(ctx context.Context, region, accessKeyID, secretAccessKey string) (string, error) {
-	listRegionsOutput, err := r.awsClient.ListRegions(ctx, region, accessKeyID, secretAccessKey)
+func (r *Runner) selectAWSRegion(ctx context.Context) (string, error) {
```
kachawla, line 100: just for easier context, can you please add a comment here that it uses local aws config to list regions?

pkg/cli/cmd/radinit/aws.go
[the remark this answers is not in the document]
vishwahiremat: moved accoutID related code into getAccountID

pkg/cli/cmd/radinit/aws.go
vishwahiremat: added it

pkg/cli/cmd/radinit/aws.go
vishwahiremat: updated it

pkg/cli/cmd/radinit/aws.go
vishwahiremat: error could be caused when creating a new Program in GetListInput function. https://github.com/radius-project/radius/blob/c62434ff63af925dabfa36f6a9a781454fcc322f/pkg/cli/prompt/prompt.go#L74-L77

pkg/cli/cmd/radinit/aws.go
vishwahiremat: i missed to change the variable name here, updated it. changed it from `addingCloudProvider ` -> `addingAccountID`

pkg/cli/aws/client.go
```diff
@@ -58,12 +62,16 @@ func (c *client) GetCallerIdentity(ctx context.Context, region string, accessKey
...
-		Credentials: credentialsProvider,
-	})
+func (c *client) ListRegions(ctx context.Context) (*ec2.DescribeRegionsOutput, error) {
+	// Load the AWS SDK config and credentials
+	cfg, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create an EC2 client
```
ytimocin, line 72: Not sure if we need these comments.

pkg/cli/cmd/radinit/aws.go
```diff
@@ -53,12 +54,24 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
+	// if err != nil {
+	// 	return nil, err
+	// }
+
+	// if !addAccountID {
+	// 	accountId, err = r.Prompter.GetTextInput(enterAWSAccountIDPrompt, prompt.TextInputOptions{Placeholder: enterAWSAccountIDPlaceholder})
+	// 	if err != nil {
+	// 		return nil, err
+	// 	}
+	// }
```
ytimocin, line 72: Is this supposed to be uncommented?

pkg/cli/cmd/radinit/aws.go
```diff
@@ -71,21 +84,36 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
 	}
 
-	return *callerIdentityOutput.Account, nil
+	accountID := *callerIdentityOutput.Account
+	addAccountID, err := prompt.YesOrNoPrompt(fmt.Sprintf(confirmAWSAccountIDPromptFmt, accountID), prompt.ConfirmYes, r.Prompter)
+	if err != nil {
+		return "", err
+	}
+
+	if !addAccountID {
```
ytimocin, line 103: Shouldn't this be true so that the user can enter account id? Like the logic here is that if `addAccountID` is false then ask the user to enter AWS account id? Am I missing something here?

pkg/cli/cmd/radinit/aws.go
```diff
@@ -71,21 +84,36 @@ func (r *Runner) enterAWSCloudProvider(ctx context.Context) (*aws.Provider, erro
...
+		}
+	}
+
+	return accountID, nil
 }
 
-func (r *Runner) selectAWSRegion(ctx context.Context, region, accessKeyID, secretAccessKey string) (string, error) {
-	listRegionsOutput, err := r.awsClient.ListRegions(ctx, region, accessKeyID, secretAccessKey)
+// selectAWSRegion prompts the user to select an AWS region from a list of available regions.
+// regions list is retrieved using the locally configured AWS account.
```
ytimocin, line 114:
```suggestion
// Region list is retrieved using the locally configured AWS account.
```

pkg/cli/aws/client.go
vishwahiremat: removed it

pkg/cli/cmd/radinit/aws.go
vishwahiremat: missed to removed thess, updated in the pr

pkg/cli/cmd/radinit/aws.go
vishwahiremat: we retrieve the account id using the local config , and in the yes or no prompt ask the user if they want to use the account id from local config, if user selects no then we propmpt to provide the account id.

ytimocin: We can have a more meaningful name instead of `addAccountID` then I guess.

ytimocin (dismissed):

ytimocin (approved):

PR #7760, opened 2024-07-24
Adding packages:write permission to the purge test resources workflow

ytimocin:
# Description
Adding `packages:write` permission to the purge test resources workflow.

## Type of change

### DIFF
--- a/.github/workflows/purge-test-resources.yaml
+++ b/.github/workflows/purge-test-resources.yaml
@@ -19,6 +19,7 @@ name: Purge test resources
 permissions:
   id-token: write # Required for requesting the JWT
   contents: read # Required for actions/checkout
+  packages: write # Required for reading package versions and deleting packages
 
 on:
   # Enable manual trigger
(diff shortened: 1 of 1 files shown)


.github/workflows/purge-test-resources.yaml
```diff
@@ -93,14 +96,16 @@ jobs:
...
             fi
           done <<< "$resource_groups"
+
       - name: Delete Azure resource groups
         run: |
           echo "## Deleting resource group list" >> $GITHUB_STEP_SUMMARY
           cat ${{ env.AZURE_RG_DELETE_LIST_FILE}} | while read line
           do
               echo " * $line" >> $GITHUB_STEP_SUMMARY
-              az group delete --resource-group $line --yes --verbose --no-wait
```
ytimocin, line 102: Removing the `--no-wait` because we are running into a throttling issue. GitHub blocks our requests because we are sending too many deletion requests.

rynowak: Why does this help? And why is Github involved in a call we make the Azure API?

ytimocin: My bad. Ignore this comment please :) I will update this PR.

sylvainsf (approved):
Looks good!

PR #7779, opened 2024-08-01
Update the registry in CLI functional test

ytimocin:
# Description
It uses localhost:5000 by default which only applies for the non-cloud functional tests. It breaks the long-running tests.

## Type of change

### DIFF
--- a/.github/actions/create-local-registry/action.yaml
+++ b/.github/actions/create-local-registry/action.yaml
@@ -96,6 +96,11 @@ runs:
         sudo cp $TEMP_CERT_DIR/certs/${{ inputs.registry-server }}/client.crt /usr/local/share/ca-certificates/${{ inputs.registry-name }}.crt
         sudo update-ca-certificates
 
+    - name: Add registry name to /etc/hosts
+      shell: bash
+      run: |
+        echo "127.0.0.1 ${{ inputs.registry-name }}" | sudo tee -a /etc/hosts
+
     - name: Create secure Docker registry
       if: ${{ inputs.secure == 'true' }}
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/create-local-registry/action.yaml
```diff
@@ -96,6 +96,11 @@ runs:
         sudo cp $TEMP_CERT_DIR/certs/${{ inputs.registry-server }}/client.crt /usr/local/share/ca-certificates/${{ inputs.registry-name }}.crt
         sudo update-ca-certificates
 
+    - name: Add registry name to /etc/hosts
+      shell: bash
+      run: |
+        echo "127.0.0.1 ${{ inputs.registry-name }}" | sudo tee -a /etc/hosts
+
```
lakshmimsft, line 103: question, do we then need to remove references to localhost, say here: https://github.com/radius-project/radius/blob/42d85feb2c2f2c32dfb961305b46f24dc2034f13/.github/workflows/functional-test-noncloud.yaml#L80C3-L80C24 because the registry name now points to localhost??

lakshmimsft: maybe it's good to leave it in for clarity

willdavsmith (approved):

.github/actions/create-local-registry/action.yaml
ytimocin: We still need localhost and radius-registry at different points as discussed offline. If we are doing a call to the registry from within the cluster then we need to use the registry name. Otherwise, we can use localhost.

lakshmimsft (approved):

PR #7784, opened 2024-08-05
Update location property flags in type generator

sk593:
# Description

location is not a required field in our Bicep templates since the DE will inject it if it's missing. it is a required field in our API specs so the generator has been treating it as required. this PR doesn't add the required flag if the field is location 

## Type of change

Fixes: #issue_number

### DIFF
--- a/hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
+++ b/hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
@@ -358,11 +358,13 @@ export function generateTypes(host: AutorestExtensionHost, definition: ProviderD
     return ObjectTypePropertyFlags.None;
   }
 
-  function parsePropertyFlags(putProperty: Property | undefined, getProperty: Property | undefined, propertyName: string) {
+  function parsePropertyFlags(putProperty: Property | undefined, getProperty: Property | undefined, propertyName?: string) {
     let flags = ObjectTypePropertyFlags.None;
 
-    if (putProperty && putProperty.required && propertyName !== 'location') {
-      flags |= ObjectTypePropertyFlags.Required;
+    if (putProperty && putProperty.required) {
+      if (!propertyName || propertyName !== 'location') {
+        flags |= ObjectTypePropertyFlags.Required;
+      }
     }
 
     if (putProperty && getProperty) {
@@ -465,7 +467,7 @@ export function generateTypes(host: AutorestExtensionHost, definition: ProviderD
       const propertyDefinition = parseType(putProperty?.schema, getProperty?.schema);
       if (propertyDefinition) {
         const description = (putProperty?.schema ?? getProperty?.schema)?.language.default?.description;
-        const flags = parsePropertyFlags(putProperty, getProperty, propertyName);
+        const flags = parsePropertyFlags(putProperty, getProperty);
         definitionProperties[propertyName] = createObjectProperty(propertyDefinition, flags, description);
       }
     }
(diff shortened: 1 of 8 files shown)


hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
```diff
@@ -358,10 +358,10 @@ export function generateTypes(host: AutorestExtensionHost, definition: ProviderD
     return ObjectTypePropertyFlags.None;
   }
 
-  function parsePropertyFlags(putProperty: Property | undefined, getProperty: Property | undefined) {
+  function parsePropertyFlags(putProperty: Property | undefined, getProperty: Property | undefined, propertyName: string) {
     let flags = ObjectTypePropertyFlags.None;
 
-    if (putProperty && putProperty.required) {
+    if (putProperty && putProperty.required && propertyName !== 'location') {
```
sk593, line 364: We could also check that the propertyType is StringType but I'm not sure how likely we'd be to run into a situation where a 'location' field isn't a string

rynowak: Would this apply to all fields name `location`? Is there a good way to apply this to only fields at the top level?

eg:

- `$.location` is updated to be optional
- `$.properties.foo[0].location` can be required

sk593: Updated to only apply this check if we're parsing a location field on the resource body. anything nested should still be marked as required

hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
```diff
@@ -68,7 +68,7 @@ export function generateTypes(host: AutorestExtensionHost, definition: ProviderD
       const propertyDefinition = parseType(putProperty?.schema, getProperty?.schema);
       if (propertyDefinition) {
         const description = (putProperty?.schema ?? getProperty?.schema)?.language.default?.description;
-        const flags = parsePropertyFlags(putProperty, getProperty);
+        const flags = parsePropertyFlags(putProperty, getProperty, propertyName);
```
rynowak, line 71: So does this work because you're only passing in `propertyName` for top level properties? I'm assuming that there are other calls to `parsePropertyFlags` that don't pass in `propertyName`...

I wonder if it would be clearer to have a parameter called `isTopLevel` (or similar). 

This all feels like it should have a comment explaining **what** and **why**. It would be easy to someone else to work on this code and change the behavior by mistake.

hack/bicep-types-radius/src/autorest.bicep/test/integration/generated/basic/test.rp1/2021-10-31/types.json
```diff
@@ -59,7 +59,7 @@
         "type": {
           "$ref": "#/0"
         },
-        "flags": 1,
```
rynowak, line 62: Do we have any tests where we can cover this updated behavior?

sk593: The integration tests check that the location field doesn't get flagged as required. We don't currently run those tests in workflows but we could update that if we want: https://github.com/radius-project/radius/blob/update-location-field/hack/bicep-types-radius/src/autorest.bicep/test/integration/integration.test.ts

hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
sk593: Yes, we only pass propertyName for top level properties. For other calls, we don't pass it in (i.e. https://github.com/radius-project/radius/blob/50394ac9c3ea2982419bbbf6e0255175759d221c/hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts#L470)

I'm not opposed to passing in another parameter to check if it's top level but since we'd need to pass in propertyName regardless to check if the field is 'location', I combined it into just one check for propertyName

rynowak (approved):

hack/bicep-types-radius/src/autorest.bicep/src/type-generator.ts
sk593: synced offline. A check for topLevel is already rolled into the check for propertyName so we'll just add a comment here explaining what's happening

Issue #7717, opened 2024-06-25
Azure policy `custom container allow` in long-running cluster


ytimocin:
### Steps to reproduce

We needed to recreate the long-running cluster and in the first run of the long-running test workflow I saw the following warning messages:

![image](https://github.com/radius-project/radius/assets/5220939/dc6d8248-e44f-4634-9d27-74c650ca8e7b)

This issue is to keep an eye on these warning messages. It doesn't seem to block the tests from running.

Link to the run: https://github.com/radius-project/radius/actions/runs/9669729901/job/26677021991.

### Observed behavior

Azure policy `custom container allow` warning messages after creating the cluster for the first time.

### Desired behavior

Not seeing the warning messages. Updating the bicep files in our repo, if possible, with the exceptions to the custom containers.

### rad Version

edge

### Would you like to support us?

- [ ] Yes, I would like to support you

[AB#12625](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12625)

nicolejms:
nothing is broken, just noise so closing

PR #7785, opened 2024-08-05
Update nginx image in LR test workflow

### LINKED ISSUE #4277: Idempotency Gap # 2: Handling of `propertyTransform` schema attribute
## Overview
Since Radius deployments follow a declarative approach, the concern of processing [CloudFormation Stacks](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-whatis-howdoesitwork.html) in an idempotent way is handled by UCP. UCPs update algorithm should consider the `propertyTransform` attribute to prevent false positive drifts on resource types. More info on this found here [Preventing false drift detection results for resource types
](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-model-false-drift.html)

### The Problem

`propertyTransform` attributes are specified using [JSONata](https://docs.jsonata.org/overview.html) and UCP needs to pick the transforms and apply them on the input before it can check for updates.

### Proposed Solution

Add UCP logic to run the JSONata transforms on input and match before doing the PATCH operation

[AB#4489](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/4489)

<adosync>

sk593:
# Description

Fixing issues with an old version of nginx during the LRT workflows during the publish TF test recipes step. `1.27` is the latest version released of nginx 

```
kubectl apply -f ./deploy/tf-module-server/resources.yaml -n radius-test-tf-module-server
Warning: [azurepolicy-k8sazurev2customcontainerallow-db27d7bfaf0671f6e28a] Container image nginx:1.25 for container nginx has not been allowed.
```

## Type of change

Fixes: #4277

### DIFF
--- a/deploy/tf-module-server/resources.yaml
+++ b/deploy/tf-module-server/resources.yaml
@@ -15,7 +15,7 @@ spec:
         app.kubernetes.io/name: tf-module-server
     spec:
       containers:
-      - image: nginx:1.25
+      - image: nginx:1.27
         name: nginx
         # nginx will serve files found in this directory.
         volumeMounts:


rynowak:
Is this failing because of the version we're using? or because we're pulling it from dockerhub?

sk593:
> Is this failing because of the version we're using? or because we're pulling it from dockerhub?

Based on the error output, it seems like the version is the issue. It's specific to the image, not to docker

ytimocin:
> > Is this failing because of the version we're using? or because we're pulling it from dockerhub?
> 
> Based on the error output, it seems like the version is the issue. The error message is specific to the image, not to docker

We get these errors when we are installing radius also. I think we may need to add the images we are going to use to a list (if there is one) so that we don't see the warnings. Please see this issue: https://github.com/radius-project/radius/issues/7717. cc/ @sk593 @rynowak

sk593:
Haven't seen this being reproduced anymore. Going to close this in favor of addressing here:  #7717

PR #7788, opened 2024-08-07
Threat model for the Controller component

ytimocin:
# Description
**Threat Model for the Controller Component.**
Diagram will be added.


pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -630,13 +630,17 @@ func (r *DeploymentReconciler) saveState(ctx context.Context, deployment *appsv1
 }
 
 func (r *DeploymentReconciler) findDeploymentsForRecipe(ctx context.Context, obj client.Object) []reconcile.Request {
-	recipe := obj.(*radappiov1alpha3.Recipe)
+	recipe, ok := obj.(*radappiov1alpha3.Recipe)
+	if !ok {
+		return []reconcile.Request{}
+	}
```
rynowak, line 636: :)

PR #7791, opened 2024-08-08
Removing double imports found by staticcheck

ytimocin:
# Description
Removing double imports found by staticcheck

## Type of change

### DIFF
--- a/pkg/cli/cmd/env/update/update_test.go
+++ b/pkg/cli/cmd/env/update/update_test.go
@@ -30,7 +30,6 @@ import (
 	"github.com/radius-project/radius/pkg/cli/framework"
 	"github.com/radius-project/radius/pkg/cli/output"
 	"github.com/radius-project/radius/pkg/cli/workspaces"
-	"github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
 	corerp "github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
 	"github.com/radius-project/radius/pkg/to"
 	"github.com/radius-project/radius/test/radcli"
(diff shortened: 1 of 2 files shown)


willdavsmith (approved):

lakshmimsft (approved):

pkg/cli/cmd/env/update/update_test.go
```diff
@@ -30,7 +30,6 @@ import (
 	"github.com/radius-project/radius/pkg/cli/framework"
 	"github.com/radius-project/radius/pkg/cli/output"
 	"github.com/radius-project/radius/pkg/cli/workspaces"
-	"github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
```
rynowak, line 33: Is this something we could turn on in `golanglint-ci`? I think this is a good change, I'm just wondering how we could automate it.

rynowak (approved):

pkg/cli/cmd/env/update/update_test.go
ytimocin: I found it by using `staticcheck ./...`. We can add that to a CI/CD we have.

rynowak: I did some research, it sounds like we should add staticcheck. I thought it was bundled into golanglint-ci but it isn't exactly, sounds like there is a history there.

ytimocin: Created this feature request: https://github.com/radius-project/radius/issues/7795. cc/ @rynowak

Issue #6917, opened 2023-12-05
Recipe support for private Bicep registries


AaronCrawfis:
## Current state

Today we allow OCI registries with anonymous pull to be used as Bicep registries for Recipes. For example, `ghcr.io/radius-project/recipes/local-dev/rediscaches:latest` is a public Recipe Bicep module: https://github.com/radius-project/recipes/pkgs/container/recipes%2Flocal-dev%2Frediscaches

## Desired state

We want to enable users to be able to use private OCI registries that are not open to the public. For example, an ACR or GHCR registry could be locked down to only certain users or accounts.

## Acceptance criteria

- [x] A user can create an OCI registry, such as an Azure Container Registry, that is private and can only be pulled from by allowed users/accounts
- [x] Recipes can be pushed to this OCI registry via `rad bicep publish`, after the user logs into the OCI registry via the registries login mechanism (`az acr login` for example).
- [x] The private Recipe can be added to an environment's list of Recipes
- [ ] A user can configure credentials for the OCI registry
- [ ] Radius uses the credentials to fetch Recipes
- [ ] If credentials are invalid for the registry, an anonymous pull is attempted. This prevents a user from registering credentials for `ghcr.io/myaccount` and then public recipes for `ghcr.io/radius-project` can no longer be pulled.
- [ ] Users can list configured credentials and view the username, but not the password
- [ ] Users update credentials to roll them when they are expired or need to be replaced
- [ ] Users can remove credentials for a specific OCI registry
- [ ] Nested Bicep modules can leverage the same set of credentials

[AB#10666](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10666)

AaronCrawfis:
Similar to #6911

shalabhms:
Triaged: similar parity #6911

PR #7796, opened 2024-08-09
Update LoadSecrets()

lakshmimsft:
# Description

1. Update LoadSecrets() to return data for all keys for corresponding secret store when no keys filter is provided.
2. Update return type to include SecretData{}
3. Updated existing LoadSecrets() to return data collated for multiple secret store ids
4. Update to helper function populateSecretData() and it's unit tests which populates secret data map returned.
5. Update functional test to include envSecrets input

## Type of change
- This pull request adds or changes features of Radius and has an approved issue (#6917).

Fixes: (Part of #6917)

### DIFF
--- a/pkg/recipes/configloader/secrets.go
+++ b/pkg/recipes/configloader/secrets.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	aztoken "github.com/radius-project/radius/pkg/azure/tokencredentials"
 	"github.com/radius-project/radius/pkg/corerp/api/v20231001preview"
+	"github.com/radius-project/radius/pkg/recipes"
 	"github.com/radius-project/radius/pkg/ucp/resources"
 )
 
@@ -36,13 +37,17 @@ type secretsLoader struct {
 }
 
 // LoadSecrets loads secrets from secret stores based on input map of provided secret store IDs and secret keys filter.
-// If the input secret keys filter for a secret store ID is nil or empty, it retrieves secret data for all keys for that secret store.
+// It returns a map with keys as secret store IDs and corresponding SecretData{}.
+// If the input secret keys filter for a secret store ID is nil or empty, it retrieves secret data for all keys for that secret store ID.
+// Eg: secretStoreKeysFilter = {"SecretStoreID1": nil} or secretKeysFilter = {"SecretStoreID1": []} will retrieve data for all secrets from "SecretStoreID1".
+// ---
 // When the secret keys filter is populated for a secret store ID, it retrieves secret data for the specified keys for the associated secret store ID.
 // The function returns a map of secret data, where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
-func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDKeysFilter map[string][]string) (secretData map[string]map[string]string, err error) {
-	secretData = make(map[string]map[string]string)
+// Eg; secretStoreKeysFilter = {"SecretStoreID1": ["secretkey1", "secretkey2"]} will retrieve data for only "secretkey1" and "secretkey2" keys from "SecretStoreID1".
+func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreKeysFilter map[string][]string) (secretData map[string]recipes.SecretData, err error) {
+	secretData = make(map[string]recipes.SecretData)
 
-	for secretStoreID, secretKeysFilter := range secretStoreIDKeysFilter {
+	for secretStoreID, secretKeysFilter := range secretStoreKeysFilter {
 		secretStoreResourceID, err := resources.ParseResource(secretStoreID)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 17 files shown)


pkg/recipes/configloader/secrets.go
```diff
@@ -56,32 +60,49 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
+	for _, secretKey := range secretKeysFilter {
+		secretDataValue, ok := secrets.Data[secretKey]
+		if !ok {
 			return nil, fmt.Errorf("a secret key was not found in secret store '%s'", secretStoreID)
 		}
+
+		if secretData[secretStoreID] == nil {
+			secretData[secretStoreID] = make(map[string]string)
+		}
+		secretData[secretStoreID][secretKey] = *secretDataValue.Value
```
lakshmimsft, line 105: ln96- ln105 is just ln76-ln82 refactored

pkg/recipes/configloader/secrets.go
```diff
@@ -35,10 +35,14 @@ type secretsLoader struct {
...
-// It returns a map of secret data, where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
-func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKeys map[string][]string) (secretData map[string]map[string]string, err error) {
-	for secretStoreID, secretKeys := range secretStoreIDResourceKeys {
+// LoadSecrets loads secrets from secret stores based on input map of provided secret store IDs and secret keys filter.
+// If the input secret keys filter for a secret store ID is nil or empty, it retrieves secret data for all keys for that secret store.
+// When the secret keys filter is populated for a secret store ID, it retrieves secret data for the specified keys for the associated secret store ID.
+// The function returns a map of secret data, where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
+func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDKeysFilter map[string][]string) (secretData map[string]map[string]string, err error) {
```
ytimocin, line 42: Why not just name this `filter` and give an example in the doc above?
```suggestion
func (e *secretsLoader) LoadSecrets(ctx context.Context, filter map[string][]string) (secretData map[string]map[string]string, err error) {
```

pkg/recipes/configloader/secrets.go
```diff
@@ -56,32 +60,49 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
 	return secretData, nil
 }
 
 // populateSecretData is a helper function to populate secret data from a secret store.
-func populateSecretData(secretStoreID string, secretKeys []string, secrets *v20231001preview.SecretStoresClientListSecretsResponse) (map[string]map[string]string, error) {
-	secretData := make(map[string]map[string]string)
-
+// It takes a secret store ID, a filter for secret keys, and a response containing the secrets.
+// If the secret keys filter is nil or empty, it retrieves data for all keys in the secret store.
+// It returns a map where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
```
ytimocin, line 80: Should start with what the function does and then give examples like you did here.

pkg/recipes/configloader/secrets.go
```diff
@@ -56,32 +60,49 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
+func populateSecretData(secretStoreID string, secretKeysFilter []string, secrets *v20231001preview.SecretStoresClientListSecretsResponse) (map[string]map[string]string, error) {
```
ytimocin, line 81:
```suggestion
func populateSecretData(id string, filter []string, secrets *v20231001preview.SecretStoresClientListSecretsResponse) (map[string]map[string]string, error) {
```

pkg/recipes/configloader/secrets.go
```diff
@@ -56,32 +60,49 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
+
+	for _, secretKey := range secretKeysFilter {
+		secretDataValue, ok := secrets.Data[secretKey]
+		if !ok {
 			return nil, fmt.Errorf("a secret key was not found in secret store '%s'", secretStoreID)
 		}
+
+		if secretData[secretStoreID] == nil {
+			secretData[secretStoreID] = make(map[string]string)
+		}
```
ytimocin, line 104: This one will be empty for sure, right? Why don't we move this to line 87 right after we initialize the slice? That would be more readable maybe?

pkg/recipes/configloader/secrets_test.go
```diff
@@ -19,8 +19,30 @@ func Test_populateSecretData(t *testing.T) {
 		expectedErrMsg  string
 	}{
 		{
-			name:       "success",
-			secretKeys: []string{"secretKey1", "secretKey2"},
+			name:       "success - data for input secretKey returned",
```
ytimocin, line 22: This is actually secretKey1, right?
```suggestion
			name:       "success - data for input secretKey1 returned",
```

pkg/recipes/configloader/secrets.go
vishwahiremat, line 42: return type should be updated to add the type of secret.
```
type secretData struct{
  type string
  Data map[string]string
}
```

pkg/recipes/configloader/secrets.go
```diff
@@ -56,32 +60,49 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
+
+	// If secretKeys is nil or empty, retrieve data for all keys
+	if len(secretKeysFilter) == 0 {
+		secretKeysFilter = make([]string, 0, len(secrets.Data))
+		for secretKey := range secrets.Data {
+			secretKeysFilter = append(secretKeysFilter, secretKey)
+		}
+	}
+
+	for _, secretKey := range secretKeysFilter {
```
vishwahiremat, line 96: i think this should be part of `else` block?

lakshmimsft: this is correct as is. if condition ln89 populates secretKeysFilter with data for all keys if empty as described.
the for loop ln96 populates the return type for the list of keys.

pkg/recipes/configloader/secrets.go
lakshmimsft: the return type is a map of maps. we do not want to return a map of keys (secretstoreids)with no data if no secret data exist. Keeping this as is for now.

pkg/recipes/configloader/secrets.go
lakshmimsft: keeping the more descriptive names

bicep-types
vishwahiremat, line 1: are these changes related to this PR?

pkg/recipes/configloader/secrets.go
```diff
@@ -55,33 +64,55 @@ func (e *secretsLoader) LoadSecrets(ctx context.Context, secretStoreIDResourceKe
...
+		Data: make(map[string]string),
+	}
+
+	// If secretKeysFilter is nil or empty, populate secretKeysFilter with all keys
+	if len(secretKeysFilter) == 0 {
+		secretKeysFilter = make([]string, 0, len(secrets.Data))
+		for secretKey := range secrets.Data {
+			secretKeysFilter = append(secretKeysFilter, secretKey)
+		}
 	}
```
vishwahiremat, line 108: can we do it this way instead of creating a secretKeyFilter map ?
```suggestion
	if len(secretKeysFilter) == 0 {
	        secretData.Data = secrets.Data
	        return secretData,nil
	}
```

lakshmimsft: secrets.Data is a map[string]*SecretValueProperties, different datatype from what is needed. keeping current code as discussed offline.

pkg/recipes/terraform/execute.go
```diff
@@ -225,7 +225,7 @@ func (e executor) setEnvironmentVariables(tf *tfexec.Terraform, options Options)
 		for secretName, secretReference := range recipeConfig.EnvSecrets {
 			// Extract secret value from the secrets input
 			if secretIDs, ok := options.Secrets[secretReference.Source]; ok {
-				if secretValue, ok := secretIDs[secretReference.Key]; ok {
+				if secretValue, ok := secretIDs.Data[secretReference.Key]; ok {
 					envVarUpdate = true
```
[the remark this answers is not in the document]
lakshmimsft: secretIDs is a struct and will not be nil. Renamed variable to secretData. There exists a unit test to check for secretData.Data when nil.

bicep-types
lakshmimsft: removed from PR.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-terraform-postgres.bicep
```diff
@@ -24,13 +24,15 @@ resource env 'Applications.Core/environments@2023-10-01-preview' = {
         providers: {
```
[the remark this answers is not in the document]
lakshmimsft: Discussed offline. The functional test needs to be updated with envSecrets block. This PR updates code that can affect running of functional test with envSecrets. We'll include this test in the PR.

vishwahiremat (approved):
lgtm

ytimocin (approved):

PR #7797, opened 2024-08-09
Adding a check to a cast to the Recipe object

ytimocin:
# Description
Adding a check to a cast to the Recipe object

## Type of change


pkg/controller/reconciler/deployment_reconciler.go
```diff
@@ -630,7 +630,10 @@ func (r *DeploymentReconciler) saveState(ctx context.Context, deployment *appsv1
 }
 
 func (r *DeploymentReconciler) findDeploymentsForRecipe(ctx context.Context, obj client.Object) []reconcile.Request {
-	recipe := obj.(*radappiov1alpha3.Recipe)
+	recipe, ok := obj.(*radappiov1alpha3.Recipe)
+	if !ok {
+		return []reconcile.Request{}
+	}
```
rynowak, line 636: It's unfortunate that there's no good way to return an error here. I agree that skipping is the most reasonable thing to do.

rynowak (approved):

PR #7803, opened 2024-08-12
Update credential precedence in bicepconfig.json

sk593:
# Description

Some workflows need Az authentication and some don't. This removes the default az auth in workflows anytime its not needed

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/functional-test-cloud.yaml
+++ b/.github/workflows/functional-test-cloud.yaml
@@ -411,7 +411,7 @@ jobs:
   
   tests:
     name: Run ${{ matrix.name }} functional tests
-    needs: [build, publish-test-bicep-types]
+    needs: [build]
     if: github.event_name == 'repository_dispatch' || (github.event_name == 'schedule' && github.repository == 'radius-project/radius') || github.event_name == 'workflow_run'
     strategy:
       fail-fast: true
--- a/.github/workflows/functional-test-cloud.yaml
+++ b/.github/workflows/functional-test-cloud.yaml
@@ -315,6 +315,30 @@ jobs:
           message: |
             :hourglass: Publishing Bicep Recipes for functional tests...
 
+      - name: Generate test bicepconfig.json
+        run: |
+          if [[ "${{ env.REL_VERSION }}" == "edge" ]]; then
+            RADIUS_VERSION="latest"
+          else
+            RADIUS_VERSION="${{ env.REL_VERSION }}"
+          fi
+          cat <<EOF > ./test/bicepconfig.json
+          {
+            "experimentalFeaturesEnabled": {
+              "extensibility": true,
+              "extensionRegistry": true,
+              "dynamicTypeLoading": true
+            },
+            "extensions": {
+              "radius": "br:${{ env.BICEP_TYPES_REGISTRY }}/test/radius:$RADIUS_VERSION",
+              "aws": "br:${{ env.BICEP_TYPES_REGISTRY }}/aws:latest"
+            },
+            "cloud": {
+              "credentialPrecedence": ["Environment"]
+            }
+          }
+          EOF
+
       - name: Publish Bicep Test Recipes
         run: |
           mkdir ./bin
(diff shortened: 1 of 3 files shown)


.github/workflows/functional-test-cloud.yaml
```diff
@@ -314,6 +314,64 @@ jobs:
           append: true
           message: |
             :hourglass: Publishing Bicep Recipes for functional tests...
+
```
sk593, line 317: Moving this into the build job so we make sure to upload any bicep changes before pushing recipes

.github/workflows/functional-test-cloud.yaml
```diff
@@ -314,6 +314,64 @@ jobs:
...
+          node-version: '18'
+
+      - name: Generate Bicep extensibility types from OpenAPI specs
+        run: |
+          make generate-bicep-types VERSION=${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }}
+
+      - name: Upload Radius Bicep types artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.name }}_radius_bicep_types
```
ytimocin, line 330: matrix is not reachable here

ytimocin (changes requested):

.github/workflows/functional-test-cloud.yaml
sk593: Updated

.github/workflows/functional-test-cloud.yaml
```diff
@@ -314,6 +314,64 @@ jobs:
...
+          curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-x64
+          chmod +x ./bicep
+          sudo mv ./bicep /usr/local/bin/bicep
+          bicep --version
+          
+      - name: Publish bicep types 
+        run: |
+          bicep publish-extension ./hack/bicep-types-radius/generated/index.json --target br:${{ env.BICEP_TYPES_REGISTRY }}/test/radius:${{ env.REL_VERSION == 'edge' && 'latest' || env.REL_VERSION }} --force
+
+      - name: Generate test bicepconfig.json
```
ytimocin, line 352: We have of two of these in this workflow. Is this expected?

.github/workflows/validate-bicep.yaml
```diff
@@ -44,6 +44,14 @@ jobs:
...
       - name: Check out repo
         uses: actions/checkout@v4
+      - name: Install jq
+        run: sudo apt-get install -y jq
+      - name: Modify bicepconfig.json
+        run: |
+          # Add credential preference to the bicepconfig.json file
+          jq '. + {"cloud": {"credentialPrecedence": ["Environment"]}}' bicepconfig.json > tmp.json && mv tmp.json bicepconfig.json
+      - name: Print updated bicepconfig.json
+        run: cat bicepconfig.json
```
ytimocin, line 54: Do we need this in production?

ytimocin (approved):
Do we need the step that prints out the details of the JSON file in production?

lakshmimsft (approved):
lgtm

.github/workflows/validate-bicep.yaml
sk593: No, this was the initial workaround since there were authentication issues with the ACR even though we have anonymous pulls. The issue is that when we publish to ACR, we have to authenticate so only have "Environment" authentication removed any Azure CLI auth. Ideally, we can remove this everywhere in the repo once the bug is fixed: https://github.com/radius-project/radius/issues/7804

.github/workflows/functional-test-cloud.yaml
sk593: yes, the files dont persist between jobs so we need it in both places. the first one is part of the workaround because we need to disable azure auth before we publish the bicep test recipes (this will eventually be removed once the auth bug is fixed) and the second is so we can test against the PR versions of the types (in the event that there are bicep changes in a PR)

kachawla (approved):

PR #7816, opened 2024-08-16
Add new secret types to Applications.Core/secretstores 

lakshmimsft:
# Description

Add new types to Applications.Core/secretstores (basicAuthentication, azureWorkloadIdentity, awsIRSA)
Update convertor, tests.
Update existing ValidateAndMutateRequest() in /pkg/corerp/frontend/controller/secretstores/kubernetes.go
to check if required secret keys exist for current secret type. Add to existing unit tests.

## Type of change

- This pull request fixes a bug in Radius and has an approved issue (#6917 ).

Fixes: Part of #6917


pkg/corerp/api/v20231001preview/secretstore_conversion.go
```diff
@@ -106,6 +106,12 @@ func toSecretStoreDataTypeDataModel(src *SecretStoreDataType) datamodel.SecretTy
 		return datamodel.SecretTypeGeneric
```
vishwahiremat, line 106: lets add validations here during conversion if the required keys are present for the particular type. ex: "clientID" and "tenantID" keys should be present for type "azureWorkloadIdentity" type

lakshmimsft: We discussed offline. Currently validations are included in https://github.com/radius-project/radius/blob/e5776243b89d8d6636cd42597e60e0ef9905efd6/pkg/corerp/frontend/controller/secretstores/kubernetes.go#L90 in line with other validations required for creating this resource. This function is accessed by controller here: https://github.com/radius-project/radius/blob/e5776243b89d8d6636cd42597e60e0ef9905efd6/pkg/corerp/setup/setup.go#L172
including @kachawla to confirm this is the right approach.

pkg/corerp/frontend/controller/secretstores/kubernetes_test.go
```diff
@@ -247,53 +252,109 @@ func TestGetOrDefaultEncoding(t *testing.T) {
...
-		r := resp.(*rest.BadRequestResponse)
-		require.Equal(t, "$.properties.type cannot change from 'generic' to 'certificate'.", r.Body.Error.Message)
-	})
-
-	t.Run("inherit resource id from existing resource", func(t *testing.T) {
-		oldResource := testutil.MustGetTestData[datamodel.SecretStore](testFileCertValueFrom)
-		newResource := testutil.MustGetTestData[datamodel.SecretStore](testFileCertValueFrom)
-		newResource.Properties.Resource = ""
-		resp, err := ValidateAndMutateRequest(context.TODO(), newResource, oldResource, nil)
+	tests := []struct {
```
vishwahiremat, line 255: 👍

vishwahiremat (dismissed):
lgtm

pkg/corerp/datamodel/secretstore.go
```diff
@@ -45,6 +45,12 @@ const (
 	SecretTypeGeneric SecretType = "generic"
 	// SecretTypeCert is the certificate secret type.
 	SecretTypeCert SecretType = "certificate"
+	// SecretTypeBasicAuthentication is the basicAuthentication secret type.
+	SecretTypeBasicAuthentication SecretType = "basicAuthentication"
+	// SecretTypeAzureWorkloadIdentity is the azureFederatedIdentity secret type.
+	SecretTypeAzureWorkloadIdentity SecretType = "azureWorkloadIdentity"
```
ytimocin, line 51:
```suggestion
	// SecretTypeAzureWorkloadIdentity is the azureWorkloadIdentity secret type.
	SecretTypeAzureWorkloadIdentity SecretType = "azureWorkloadIdentity"
```

pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_AWSIRSA.json
ytimocin, line 1: nit: pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_awsirsa.json

pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_AZWorkloadIdentity.json
ytimocin, line 1: nit: pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_azwi.json

pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_basicAuthentication.json
ytimocin, line 1: nit: pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_basicauth.json

pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_basicAuthentication_invalid.json
ytimocin, line 1: nit: pkg/corerp/frontend/controller/secretstores/testdata/secretstores_datamodel_basicauth_invalid.json

ytimocin (changes requested):

ytimocin (dismissed):

ytimocin (approved):

typespec/Applications.Core/environments.tsp
```diff
@@ -79,7 +79,7 @@ model RecipeConfigProperties {
   @doc("Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.")
   terraform?: TerraformConfigProperties;
 
-  @doc("Environment variables injected during Terraform Recipe execution for the recipes in the environment.")
+  @doc("Environment variables injected during recipe execution for the recipes in the environment.")
```
kachawla, line 82: @lakshmimsft what's causing this to change?

typespec/Applications.Core/secretstores.tsp
```diff
@@ -68,6 +68,15 @@ enum SecretStoreDataType {
 
   @doc("Certificate secret data type")
   certificate,
+
+  @doc("basicAuthentication type is used to represent username and password based authentication and the secretstore resource is expected to have the keys 'username' and 'password'.")
+  basicAuthentication,
+
+  @doc("azureWorkloadIdentity type is used to represent registry authentication using azure federated identity and the secretstore resource is expected to have the keys 'clientId' and 'tenantId'.")
```
kachawla, line 75: Do we need to call out "registry authentication" here? This resource is/can be referenced in multiple places with use cases being not limited to ACR. I'd also get PMs eyes on these user facing descriptions.

typespec/Applications.Core/secretstores.tsp
```diff
@@ -68,6 +68,15 @@ enum SecretStoreDataType {
...
+  azureWorkloadIdentity,
+
+  @doc("awsIRSA type is used to represent registry authentication using AWS IRSA(IAM Roles for Service accounts) and the secretstore resource is expected to have the keys 'roleARN'.")
```
kachawla, line 78: Same as above.

Couple of nits (can't add suggestions since the PR is merged):
- Missing space between `IRSA` and `(`
- `key` instead of `keys`

pkg/corerp/frontend/controller/secretstores/types.go
```diff
@@ -19,4 +19,11 @@ package secretstores
 const (
 	// ResourceTypeName is the resource type name for secret stores.
 	ResourceTypeName = "Applications.Core/secretStores"
+
+	// The following are possible required keys in a SecretStore depending on it's SecretType
+	RequiredUsername = "username"
+	RequiredPassword = "password"
+	RequiredClientId = "clientId"
+	RequiredTenantId = "tenantId"
+	RequiredRoleARN  = "roleARN"
```
kachawla, line 28: The prefix `Required` is confusing, maybe I'm not following the thought process? They are the expected keys in the resource, so in general for such variables the convention is to add a `key` suffix, `<key_name>Key`, so it would be `UsernameKey` for example.

pkg/corerp/frontend/controller/secretstores/types.go
```diff
@@ -19,4 +19,11 @@ package secretstores
 const (
 	// ResourceTypeName is the resource type name for secret stores.
 	ResourceTypeName = "Applications.Core/secretStores"
+
+	// The following are possible required keys in a SecretStore depending on it's SecretType
```
kachawla, line 23: nit: let's not use different comment formats within the same file (example above for `ResourceTypeName`).

https://tip.golang.org/doc/comment#var

pkg/corerp/frontend/controller/secretstores/kubernetes.go
```diff
@@ -75,8 +78,15 @@ func getOrDefaultEncoding(t datamodel.SecretType, e datamodel.SecretValueEncodin
 	return e, err
 }
 
+// Define a map of required keys for each SecretType
+var requiredKeys = map[datamodel.SecretType][]string{
```
kachawla, line 82: This seems to be only referenced from `ValidateAndMutateRequest` function so I'd assume it should be scoped to that. Could you tell me more about the motivation to define it here?

typespec/Applications.Core/environments.tsp
lakshmimsft: Correcting the comment to indicate environment variables schema is part of RecipeConfigProperties and is not intended to be specific to Terraform execution.

pkg/corerp/api/v20231001preview/secretstore_conversion.go
```diff
@@ -117,6 +123,12 @@ func fromSecretStoreDataTypeDataModel(src datamodel.SecretType) *SecretStoreData
 		return to.Ptr(SecretStoreDataTypeGeneric)
 	case datamodel.SecretTypeCert:
 		return to.Ptr(SecretStoreDataTypeCertificate)
+	case datamodel.SecretTypeBasicAuthentication:
+		return to.Ptr(SecretStoreDataTypeBasicAuthentication)
+	case datamodel.SecretTypeAzureWorkloadIdentity:
+		return to.Ptr(SecretStoreDataTypeAzureWorkloadIdentity)
+	case datamodel.SecretTypeAWSIRSA:
+		return to.Ptr(SecretStoreDataTypeAwsIRSA)
```
kachawla, line 131: Should a unit test be added?

typespec/Applications.Core/environments.tsp
kachawla: @lakshmimsft it is specific to Terraform though at this point. Unless we are actually injecting this env vars for non-Terraform recipes, we should indicate it is. As a user, I'd then set it assuming it works for bicep as well, which it doesn't.

lakshmimsft: This was in response to a question asked in a design meeting: https://github.com/radius-project/design-notes/pull/53#discussion_r1661395462
Perhaps the comment should be more clear, eg: Environment variables are injected during the execution of recipes within the environment, currently supported for Terraform recipes.

typespec/Applications.Core/secretstores.tsp
lakshmimsft: cc @vishwahiremat, @Reshrahim . referring to [link](https://github.com/radius-project/design-notes/pull/53/files#diff-12354a738cc93ef8bfee787de4ebb16befdb3c7dccce0df0a0f4e29e523b779dR288)

pkg/corerp/api/v20231001preview/secretstore_conversion.go
lakshmimsft: added in https://github.com/radius-project/radius/pull/7867

pkg/corerp/frontend/controller/secretstores/kubernetes.go
lakshmimsft: moved into function: https://github.com/radius-project/radius/pull/7867

pkg/corerp/frontend/controller/secretstores/types.go
lakshmimsft: updated in in https://github.com/radius-project/radius/pull/7867

typespec/Applications.Core/secretstores.tsp
vishwahiremat: I agree with not specifying `registry` here as it can be referenced elsewhere as well.
```suggestion
  @doc("azureWorkloadIdentity type is used to represent authentication using azure federated identity and the secretstore resource is expected to have the keys 'clientId' and 'tenantId'.")
```

Reshrahim:
```suggestion
  @doc("azureWorkloadIdentity type is used to represent registry authentication using azure workload identity and the secretstore resource is expected to have the keys 'clientId' and 'tenantId'.")
```

Reshrahim: I think it is references as Azure workload identity and not federated identity in the docs

kachawla: Thanks @Reshrahim. What are your thoughts on explicitly mentioned "registry" authentication? Secret store resource is not limited to supporting management of secrets only for the use case of registries.

Reshrahim: Yeah sorry din notice the comment above. I agree with your comment of calling it as authentication as the uses cases can be outside of registries

kachawla: Thanks for confirming @Reshrahim.

@vishwahiremat @lakshmimsft - could one of you please update this as per Reshma's feedback?

PR #7823, opened 2024-08-20
Support for secret store reference in Dapr components

### LINKED ISSUE #7704: Reference secrets in components for Dapr resources

According the to [Dapr documentation](https://docs.dapr.io/operations/components/component-secrets/), in order to reference secrets for the `spec.metadata` section within the components definition you need to set the `auth.secretStore` field to specify the name of the secret store that holds the secrets if you are looking to reference something outside of the Kubernetes secret store.

However, the `auth.secretStore` field is not currently supported in the Radius Dapr Resources `properties` (e.g. [Dapr pubsub properties](https://docs.radapp.io/reference/resource-schema/dapr-schema/pubsub/#properties]) and thus Radius users are not able to leverage this feature of Dapr components. This request is for this `auth.secretStore` field to be supported so that users may reference non-Kubernetes secrets for the Dapr `spec.metadata`.

### Acceptance criteria

- Support is added for the `auth.secretStore` field in Radius.
- Additional `property` for `auth.secretStore` in Dapr Resource schema 
- Documentation updated to reflect this.

### Additional context

_No response_

[AB#12577](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12577)

SoTrx:
# Description
Hey,

This PR adds support for building Dapr components with an optional reference to a Dapr secret store.

The new property is named _secretStoreComponentName_. This name is totally arbitrary but was chosen to try to be as clear as possible.

A secret reference can be created using a _secretKeyRef_ object, following the same [naming convention as Dapr](https://docs.dapr.io/operations/components/component-secrets/).

Example : 
```yaml
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: {
        secretKeyRef: {
            name: 'redisHost'
            key: 'redisHost'
        }

      }
      redisPassword: ''
    }
    secretStoreComponentName: daprSecretStore.name
  }
}
```

## Type of change

Fixes: #7704


rynowak:
Wow thanks for sending this. This is a big improvement. I'll take a look at this tomorrow. 

Normally we might ask for an design document for an API change, but I think we can skip that in this case:

- It's a very small change
- It has an approved issue that describes the API change.
- It's aligned with the naming and design of the existing Dapr API - which we're trying to match.

@radius-project/maintainers-radius holler if you have any concerns.

pkg/daprrp/processors/statestores/processor_test.go
```diff
@@ -199,6 +200,111 @@ func Test_Process(t *testing.T) {
...
+							"secretKeyRef": map[string]any{
+								"name": "secretStoreKey",
+								"key":  "secretStoreValue",
+							},
+						},
+					},
+				},
+				"auth": map[string]any{
+					"secretStore": secretStoreComponentName,
+				},
```
ytimocin, line 286: Is this `auth` generated by **Dapr** since we create a **Dapr** object with a property called **secretStoreComponentName**?

pkg/portableresources/renderers/dapr/generic.go
```diff
@@ -56,9 +57,24 @@ func ConstructDaprGeneric(daprGeneric DaprGeneric, namespace string, componentNa
...
+
+		switch value := v.(type) {
+
+		// Nested object, has to be a secret reference
+		case map[string]any:
+			if secretKeyRef, ok := value["secretKeyRef"]; ok {
+				yamlItem["secretKeyRef"] = secretKeyRef
+			} else {
+				return unstructured.Unstructured{}, v1.NewClientErrInvalidRequest(fmt.Sprintf("Invalid metadata value for key %s in Dapr component %s", k, componentName))
+			}
```
ytimocin, line 73: Could there be a case where it is not a secret reference even if it is a nested object? Do we need to do an additional check for this case to make sure that it is a secret reference?

typespec/Applications.Dapr/secretStores.tsp
```diff
@@ -45,10 +45,15 @@ model DaprSecretStoreResource
   name: ResourceNameString;
 }
 
+// This is to prevent a secret store from referencing another secret store
+// This would be valid according to Dapr component specs, but may confuse users
+@withoutOmittedProperties("secretStoreComponentName")
+model NonRedundantDaprResourceProperties is DaprResourceProperties;
```
ytimocin, line 51: Could there be a case where users can require authentication to a Secret Store?

ytimocin (commented):
This is a great addition. Thanks!

Should we add a functional test or add this to [an existing functional test](https://github.com/radius-project/radius/blob/main/test/functional-portable/daprrp/noncloud/resources/testdata/daprrp-resources-statestore-manual.bicep) so that we can test it out?

pkg/portableresources/renderers/dapr/generic.go
rynowak: You can see the schema of `Component` here: https://github.com/dapr/dapr/blob/master/charts/dapr/crds/components.yaml#L48

Note that this is the **output** we're generating. 

The reason why I'm mention it is that it shows what the schema is in Dapr that we're trying roughly match. The allowed keys are:

- `value`
- `secretKeyRef`
- `envRef`

`envRef` isn't relevant for us - it's a dev-time feature.

typespec/Applications.Dapr/secretStores.tsp
rynowak: I think we're assuming the Dapr secret store has already been created.

rynowak:
Hi @SoTrx - thanks for sending this PR. Looks like it works overall.

One question based on your example:

```bicep
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: {
        secretKeyRef: {
            name: 'redisHost'
            key: 'redisHost'
        }

      }
      redisPassword: ''
    }
    secretStoreComponentName: daprSecretStore.name
  }
}
```

It looks like the type of `metadata` after your change is `map from string to (string or object)`. This isn't ideal because we try to avoid designs where a field has an ambiguous type. Ideally this would be `map from string to object` in all cases. I think we probably made a mistake in the original design 😢 

If I compare to the Dapr schema of `Component` I'd describe this as `list of key-value-pairs where the key is a string and the value is an object`.

Do you think it makes sense for us to change the design?

Example:

```bicep
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: {
        secretKeyRef: {
            name: 'redisHost'
            key: 'redisHost'
        }
      }
      redisPassword: {
        value: ''
      }
    }
    secretStoreComponentName: daprSecretStore.name
  }
}
```

This would be a breaking change, and we'd need to update our existing code, but it's probably a better design. What do you think?

SoTrx:
Thank you for your reviews

@rynowak 
I believe your proposed change might be the best approach, as it would provide the most extensibility.

Going the other way and using strings only (with a magic prefix to reference secrets) might also seem tempting since it would be non-breaking. But this is a  worse design and may lead to parsing/escaping issues or conflict with [special values](https://docs.dapr.io/reference/resource-specs/component-schema/#templated-metadata-values) or even a company's own oddly defined [pluggable components](https://docs.dapr.io/developing-applications/develop-components/pluggable-components/pluggable-components-overview/).

```bicep
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: 'secretRef:://key,value'
      redisPassword: ''
    }
    secretStoreComponentName: daprSecretStore.name
  }
}
```

Another approach could be treating the secret as its own resource. However, aside from the added complexity, this might encourage users to attempt using secrets from multiple secret stores within a single component, which, as far as I'm aware, Dapr does not support.

So using only objects may be the best solution in my opinion.

------------

On a related note, if the goal is to match Dapr component schema, should we also change the `secretStoreComponentName` property to `auth.secretstore` to match what Dapr [uses]((https://docs.dapr.io/operations/components/component-secrets/) ?

Something like :

```bicep
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: {
        secretKeyRef: {
            name: 'redisHost'
            key: 'redisHost'
        }
      }
      redisPassword: {
        value: ''
      }
    }
    auth: {
      secretStore: daprSecretStore.name
    }
  }
}
```

---------

I'll make some functional tests with the curent implementation first as @ytimocin suggested before going forward with the changes we decide on.

pkg/daprrp/processors/statestores/processor_test.go
SoTrx: In this case, the auth.secretstore property is added [here](https://github.com/radius-project/radius/pull/7823/files/4b23f1c2bb1e783d06e3d11a34c2296dbb512e85#diff-6ae8441cbc7aeb3a6c3a1d2dc2e278e964ddf0f99649c07832c7bf133f436e7dR100-R107) if the Applications.Dapr/* contains the `secretStoreComponentName` property.
The auth.secretstore property comes from the [Dapr component spec](https://github.com/dapr/dapr/blob/master/charts/dapr/crds/components.yaml#L26)

SoTrx:
The current implementation follow this schema :
```bicep 
resource stateStore 'Applications.Dapr/stateStores@2023-10-01-preview' = {
  name: 'statestore'
  properties: {
    environment: environment
    application: application
    resourceProvisioning: 'manual'
    type: 'state.redis'
    version: 'v1'
    metadata: {
      redisHost: {
        secretKeyRef: {
            name: 'redisHost'
            key: 'redisHost'
        }
      }
      redisPassword: {
        value: ''
      }
    }
    secretStoreComponentName: daprSecretStore.name
  }
}
```
Should I change secretStoreComponentName to auth.secretStore as well? 

Let me know if you need any further adjustments

rynowak:
> Should I change secretStoreComponentName to auth.secretStore as well?
> 
> Let me know if you need any further adjustments

I'm on board with this 👍. I think it's good if we align to the Dapr component schema. Then it will be easy to update Radius as Dapr continues to add features.

Taking a look now! Thanks for working through this.

test/testrecipes/modules/redis-selfhost.bicep
```diff
@@ -44,12 +52,11 @@ resource redis 'apps/Deployment@v1' = {
...
-              'redis-cli'
-              '-h'
-              'localhost'
-              'MONITOR'
-            ]
+            args: concat(
+                      ['redis-cli'],
+                      password != '' ? ['-a', password] : [],
+                      ['-h', 'localhost', 'MONITOR']
+                  )
```
rynowak, line 59: nice 👍

rynowak (changes requested):
This is looking awesome overall. Can you make the schema changes?

typespec/Applications.Dapr/common.tsp
```diff
@@ -33,4 +33,7 @@ model DaprResourceProperties {
 
   @doc("Dapr component version")
   version?: string;
+
```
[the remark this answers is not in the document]
SoTrx: Sure! I'll also add types to the Metadata field in the Dapr components data model as well, if that’s okay with you. This might help prevent some "casting to any" errors.

rynowak (approved):
Wow this looks great! Thanks @SoTrx

ytimocin (approved):

kachawla:
@SoTrx Thank you for contributing. We have updated the [samples](https://github.com/radius-project/samples/pull/1757/files) to reflect these changes. Could you please update the [docs](https://github.com/radius-project/docs/tree/edge) corresponding to this feature? For example, [here](https://github.com/radius-project/docs/blob/edge/docs/content/guides/author-apps/dapr/overview/snippets/statestore.bicep#L37).

PR #7832, opened 2024-08-22
Remove assets.yaml

### LINKED ISSUE #7800: Remove/disable assets workflow
### Steps to reproduce

The assets workflow has started to fail. We don't use any of the assets that get uploaded so these should be disabled and/or removed: https://github.com/radius-project/radius/blob/main/.github/workflows/assets.yaml

### Observed behavior

_No response_

### Desired behavior

Either we disable and eventually delete the workflow or we update the workflow so it passes. We don't pull assets from Azure storage anymore so this shouldn't be needed

### Workaround

_No response_

### rad Version

GH actions runs 

### Operating system

_No response_

### Additional context

_No response_

[AB#12960](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12960)

lakshmimsft:
# Description

Cleaning up disabled workflow assets.yaml. 

## Type of change
- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius (#7800 ).

Fixes: #7800


ytimocin (approved):

kachawla (approved):
Just checking - are there any other related code paths as well that should be cleaned up as well?

kachawla:
> Just checking - are there any other related code paths as well that should be cleaned up as well?

@lakshmimsft

lakshmimsft:
@kachawla I had searched codebase for any reference to assets.yaml and found none.

PR #7835, opened 2024-08-23
improve purge workflow failure logging by upgrading to snok/container-retention v 3.0.0

nithyatsu:
# Description

updating to latest version of snok/container-retention seems to help with failures. 
ref: https://github.com/radius-project/radius/actions/runs/10532606785/job/29187086343
(check only delete unused container images, since it is a workflow_dispatch)

## Type of change

Fixes: #issue_number

### DIFF
--- a/.github/workflows/purge-test-resources.yaml
+++ b/.github/workflows/purge-test-resources.yaml
@@ -19,7 +19,9 @@ name: Purge test resources
 permissions:
   id-token: write # Required for requesting the JWT
   contents: read # Required for actions/checkout
-  packages: write # Required for reading package versions and deleting packages
+  packages: 
+    read: true # Required for reading package versions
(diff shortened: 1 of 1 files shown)


ghost (approved):

.github/workflows/purge-test-resources.yaml
```diff
@@ -41,13 +41,14 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'radius-project/radius'
     steps:
-      - name: Delete 'dev' containers older than a week
-        uses: snok/container-retention-policy@v2
+      - name: Delete 'dev' containers older 3 days
+        uses: snok/container-retention-policy@v3.0.0
```
ytimocin, line 45: v3 should be enough or do we want to fix it to a specific version (v3.0.0)?

PR #7846, opened 2024-08-27
Adding DELETE operation to the webhook rules in the config

ytimocin:
# Description
Added the DELETE op to the webhook rules in the webhook configuration. Here is where the **ValidateDelete** function is: https://github.com/radius-project/radius/blob/main/pkg/controller/reconciler/recipe_webhook.go#L75. We have the function but we don't have the DELETE operation added to the rules.

Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-rules.

## Type of change

### DIFF
--- a/deploy/Chart/templates/controller/validating-webhook-configuration.yaml
+++ b/deploy/Chart/templates/controller/validating-webhook-configuration.yaml
@@ -42,6 +42,7 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
+    - DELETE
     resources:
     - recipes
   sideEffects: None


lakshmimsft:
I believe that was intentional. There is no validation for Delete today. We implement the function in code since the RecipeWebhook implements the Validator interface [link](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator.go) , indicated [here](https://github.com/radius-project/radius/blob/7da87d435642f27e28a14729c051ef00632d50d8/pkg/controller/reconciler/recipe_webhook.go#L41)

ytimocin:
> I believe that was intentional. There is no validation for Delete today, so we wouldn't want to add DELETE operation. We implement the function in code since the RecipeWebhook implements the Validator interface [link](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator.go) , indicated [here](https://github.com/radius-project/radius/blob/7da87d435642f27e28a14729c051ef00632d50d8/pkg/controller/reconciler/recipe_webhook.go#L41)

We have the Delete in the reconciler: https://github.com/radius-project/radius/blob/main/pkg/controller/reconciler/recipe_reconciler.go#L311.

Shouldn't we check if the Recipe being deleted is a Portable resource like it is suggested in the ValidateDelete function https://github.com/radius-project/radius/blob/main/pkg/controller/reconciler/recipe_webhook.go#L75?

rynowak:
> Shouldn't we check if the Recipe being deleted is a Portable resource like it is suggested in the ValidateDelete function https://github.com/radius-project/radius/blob/main/pkg/controller/reconciler/recipe_webhook.go#L75?

As a thought experiment, can we try to answer two questions to examine your idea:

- How did invalid data get into the `Recipe`?
- What do we want to do given that the `Recipe` contains invalid data?


> I believe that was intentional. There is no validation for Delete today, so we wouldn't want to add DELETE operation. We implement the function in code since the RecipeWebhook implements the Validator interface [link](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/validator.go) , indicated [here](https://github.com/radius-project/radius/blob/7da87d435642f27e28a14729c051ef00632d50d8/pkg/controller/reconciler/recipe_webhook.go#L41)

@lakshmimsft is correct. The interface requires you to implement both functions. We didn't register `DELETE` for the webhook because it wouldn't do anything helpful.

Issue #7764, opened 2024-07-25
radius uninstall should provide a way for users to delete all data


nithyatsu:
Today, when we uninstall radius, we uninstall "control-plane", while retaining all data created so far. 
Sometimes this is not what we want and would like to have a really clean start. Today we can do this by deleting `radius-system` namespace using Kubernetes commands. 

It would help if radius provided the ability to uninstall/ reinstall with an option to clear all data.

[AB#12893](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12893)

lakshmimsft:
This has been completed. Thanks @lbzss!!

PR #7847, opened 2024-08-27
Add purge flag to delete radius related namespace

lbzss:
# Description

Add purge flag to delete radius related namespace, the purge flag should be set explicitly(--purge=true).

## Type of change

issue link #7764


pkg/cli/cmd/uninstall/kubernetes/kubernetes.go
```diff
@@ -97,6 +102,14 @@ func (r *Runner) Run(ctx context.Context) error {
 		return err
 	}
 
+	if r.Purge {
+		r.Output.LogInfo("Deleting namespace...")
+		if err := r.Kubernetes.DeleteNamespace(r.KubeContext); err != nil {
+			return err
+		}
+		r.Output.LogInfo("Radius was fully uninstalled. Any existing data have been removed.")
```
willdavsmith, line 110: @willtsai @Reshrahim what error message would you like to see here?

pkg/cli/cmd/uninstall/kubernetes/kubernetes.go
```diff
@@ -97,6 +102,14 @@ func (r *Runner) Run(ctx context.Context) error {
 		return err
 	}
 
+	if r.Purge {
+		r.Output.LogInfo("Deleting namespace...")
```
nithyats82, line 106: This could be a bit explicit, "Deleting `radius-system` namespace...". @willtsai / @Reshrahim what do you think?

pkg/cli/cmd/commonflags/flags.go
```diff
@@ -130,3 +130,8 @@ func AddAWSAccountFlag(cmd *cobra.Command) {
 func AddKubeContextFlagVar(cmd *cobra.Command, ref *string) {
 	cmd.Flags().StringVar(ref, "kubecontext", "", "The Kubernetes context to use, will use the default if unset")
 }
+
+// AddPurgeFlagVar adds a flag to the given command that allows the user to delete all Radius related resources.
+func AddPurgeFlagVar(cmd *cobra.Command, purge *bool) {
+	cmd.Flags().BoolVar(purge, "purge", false, "Delete all Radius related resources")
+}
```
rynowak, line 137: This could be defined inline in the command if it's only used in one place. The flags defined here are "common" because they are used in multiple packages.

pkg/cli/cmd/uninstall/kubernetes/kubernetes.go
rynowak: I think that would be a good improvement.

pkg/cli/cmd/uninstall/kubernetes/kubernetes_test.go
```diff
@@ -115,4 +116,46 @@ func Test_Run(t *testing.T) {
...
+			},
+			output.LogOutput{
+				Format: "Deleting namespace...",
+			},
+			output.LogOutput{
+				Format: "Radius was fully uninstalled. Any existing data have been removed.",
+			},
+		}
+		require.Equal(t, expectedWrites, outputMock.Writes)
+	})
```
rynowak, line 160: Nice 👍

pkg/cli/kubernetes/kubernetes.go
```diff
@@ -20,6 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/radius-project/radius/pkg/cli/helm"
+	errors2 "k8s.io/apimachinery/pkg/api/errors"
```
rynowak, line 24: Can you call this alias `apierrors` I think that's what we call it in a few other places. 

Other ideas are fine too, but `errors2` isn't very readable 👍

pkg/cli/kubernetes/kubernetes.go
```diff
@@ -117,6 +121,30 @@ func EnsureNamespace(ctx context.Context, client k8s.Interface, namespace string
 	return nil
 }
 
+// DeleteNamespace delete the specified namespace.
+func DeleteNamespace(ctx context.Context, client k8s.Interface, namespace string) error {
+	gracePeriodSeconds := int64(10)
```
rynowak, line 126: Is there a specific motivation for 10 seconds? If we don't have a reason to configure a different value we should stick with the default. 

Context: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_delete/

pkg/cli/kubernetes/kubernetes.go
```diff
@@ -117,6 +121,30 @@ func EnsureNamespace(ctx context.Context, client k8s.Interface, namespace string
...
+	if err := client.CoreV1().Namespaces().
+		Delete(ctx, namespace, metav1.DeleteOptions{GracePeriodSeconds: &gracePeriodSeconds}); err != nil {
+		return err
+	}
+
+	err := wait.PollUntilContextCancel(ctx, 2*time.Second, true, func(ctx context.Context) (done bool, err error) {
```
rynowak, line 132: Can you leave a comment explaining why this is needed? If this is important, we don't want someone else to come through and "simplify" it later.

pkg/cli/kubernetes/kubernetes.go
```diff
@@ -172,3 +201,15 @@ type Impl struct {
 func (i *Impl) GetKubeContext() (*api.Config, error) {
 	return kubeutil.LoadConfigFile("")
 }
+
+func (i *Impl) DeleteNamespace(kubeContext string) error {
+	output.LogInfo("Deleting namespace: %s", helm.RadiusSystemNamespace)
```
rynowak, line 206: Can you move this logging into the command itself? We try to avoid logging in generic/reusable codepaths in the CLI.

pkg/cli/kubernetes/kubernetes.go
```diff
@@ -117,6 +121,30 @@ func EnsureNamespace(ctx context.Context, client k8s.Interface, namespace string
 	return nil
 }
 
+// DeleteNamespace delete the specified namespace.
+func DeleteNamespace(ctx context.Context, client k8s.Interface, namespace string) error {
```
rynowak, line 125: Can you change this to a non-exported function unless there's a need to call it elsewhere? We want to have most code use the interface so we can test it.

pkg/cli/kubernetes/kubernetes_test.go
```diff
@@ -49,6 +50,19 @@ func TestEnsureNamespace(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestDeleteNamespace(t *testing.T) {
+	namespace := "radius-test"
+	f := k8sfake.NewSimpleClientset(&v1.Namespace{ObjectMeta: meta_v1.ObjectMeta{Name: namespace}})
+
+	ctx := context.TODO()
```
rynowak, line 57: Use `context.Background()` instead. https://stackoverflow.com/questions/74239074/context-todo-or-context-background-which-one-should-i-prefer

rynowak (changes requested):
Thanks @lbzss - this looks like it's on the right track 👍  I left some minor feedback.

pkg/cli/cmd/commonflags/flags.go
```diff
@@ -130,3 +130,8 @@ func AddAWSAccountFlag(cmd *cobra.Command) {
 func AddKubeContextFlagVar(cmd *cobra.Command, ref *string) {
 	cmd.Flags().StringVar(ref, "kubecontext", "", "The Kubernetes context to use, will use the default if unset")
 }
+
+// AddPurgeFlagVar adds a flag to the given command that allows the user to delete all Radius related resources.
+func AddPurgeFlagVar(cmd *cobra.Command, purge *bool) {
+	cmd.Flags().BoolVar(purge, "purge", false, "Delete all Radius related resources")
```
rynowak, line 136: I'd suggest for the description:

`Delete all data stored by Radius.`

`Resources` is a highly overload term in the context of Radius 👍

rynowak (approved):
This looks great! Thanks @lbzss

PR #7895, opened 2024-09-07
support deployment of an application to a scope "scope1" using environment in a different scope "scope2"

### LINKED ISSUE #7520: rad deploy should be able to use environments in different groups to deploy application

nithyatsu:
# Description

Today, rad deploy can use only the environments that are in same scope as where the application is being deployed to (either default scope or one specified by -g). 
It should be able to use environments in different groups to deploy application. 

## Type of change

Fixes: #7520

### DIFF
--- a/pkg/cli/clivalidation.go
+++ b/pkg/cli/clivalidation.go
@@ -101,10 +101,34 @@ func RequireEnvironmentName(cmd *cobra.Command, args []string, workspace workspa
 	return environmentName, err
 }
 
+// RequireEnvironmentNameOrID checks if an environment name or ID is provided as a flag or as a default environment in the workspace,
+// and returns an error if neither is present. It also handles any errors that occur while parsing the resource.
+func RequireEnvironmentNameOrID(cmd *cobra.Command, args []string, workspace workspaces.Workspace) (string, error) {
+	environmentNameOrID, err := cmd.Flags().GetString("environment")
+	if err != nil {
+		return "", err
+	}
+
+	// We store the environment id in config
+	if environmentNameOrID == "" && workspace.Environment != "" {
+		environmentNameOrID = workspace.Environment
+	}
+
+	if environmentNameOrID == "" && workspace.IsEditableWorkspace() {
+		// Setting a default environment only applies to editable workspaces
+		return "", fmt.Errorf("no environment name or ID provided and no default environment set, " +
+			"either pass in an environment name or ID or set a default environment by using `rad env switch`")
+	} else if environmentNameOrID == "" {
+		return "", fmt.Errorf("no environment name or ID provided, pass in an environment name or ID")
+	}
+
+	return environmentNameOrID, err
+}
+
 // DidSpecifyEnvironmentName checks if an environment name is provided as a flag
 func DidSpecifyEnvironmentName(cmd *cobra.Command, args []string) bool {
 	environmentName, err := cmd.Flags().GetString("environment")
-	return err == nil && environmentName != ""
+	return err == nil && environmentName != "" && !strings.HasPrefix(environmentName, resources.SegmentSeparator)
 }
 
 // RequireKubeContext is used by commands that need a kubernetes context name to be specified using -c flag or has a default kubecontext
--- a/pkg/cli/cmd/deploy/deploy.go
+++ b/pkg/cli/cmd/deploy/deploy.go
@@ -194,7 +194,7 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
-		// If the environment doesn't exist, but the user specified it's name as
+		// If the environment doesn't exist, but the user specified its name or resource id as
 		// a command-line option, return an error
 		if cli.DidSpecifyEnvironmentName(cmd, args) {
 			return clierrors.Message("The environment %q does not exist in scope %q. Run `rad env create` first. You could also provide the environment ID if the environment exists in a different group.", r.EnvironmentNameOrID, r.Workspace.Scope)
--- a/pkg/cli/cmd/deploy/deploy.go
+++ b/pkg/cli/cmd/deploy/deploy.go
@@ -87,6 +87,10 @@ rad deploy myapp.bicep --environment production
 # deploy using a specific environment and resource group
 rad deploy myapp.bicep --environment production --group mygroup
 
+# deploy using an environment ID and a resource group. The application will be deployed in mygroup scope, using the specified environment.
+# use this option if the environment is in a different group.
+rad deploy myapp.bicep --environment /planes/radius/local/resourcegroups/prod/providers/Applications.Core/environments/prod --group mygroup
+
 # specify a string parameter
 rad deploy myapp.bicep --parameters version=latest
 
@@ -123,12 +127,12 @@ type Runner struct {
 	Deploy            deploy.Interface
 	Output            output.Interface
 
-	ApplicationName string
-	EnvironmentName string
-	FilePath        string
-	Parameters      map[string]map[string]any
-	Workspace       *workspaces.Workspace
-	Providers       *clients.Providers
+	ApplicationName     string
+	EnvironmentNameOrID string
+	FilePath            string
+	Parameters          map[string]map[string]any
+	Workspace           *workspaces.Workspace
+	Providers           *clients.Providers
 }
 
 // NewRunner creates a new instance of the `rad deploy` runner.
@@ -166,7 +170,7 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
 	// does not exist.
 	workspace.Scope = scope
 
-	r.EnvironmentName, err = cli.RequireEnvironmentName(cmd, args, *workspace)
+	r.EnvironmentNameOrID, err = cli.RequireEnvironmentNameOrID(cmd, args, *workspace)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


pkg/cli/clivalidation.go
```diff
@@ -101,10 +101,34 @@ func RequireEnvironmentName(cmd *cobra.Command, args []string, workspace workspa
...
+		environmentNameOrID = workspace.Environment
+	}
+
+	if environmentNameOrID == "" && workspace.IsEditableWorkspace() {
+		// Setting a default environment only applies to editable workspaces
+		return "", fmt.Errorf("no environment name or ID provided and no default environment set, " +
+			"either pass in an environment name or ID or set a default environment by using `rad env switch`")
+	} else if environmentNameOrID == "" {
+		return "", fmt.Errorf("no environment name or ID provided, pass in an environment name or ID")
+	}
```
ytimocin, line 123: Does this mean that the workspace doesn't also have an environment? Should we add that to the error message?

pkg/cli/clivalidation.go
```diff
@@ -101,10 +101,34 @@ func RequireEnvironmentName(cmd *cobra.Command, args []string, workspace workspa
...
+
+	return environmentNameOrID, err
+}
+
 // DidSpecifyEnvironmentName checks if an environment name is provided as a flag
 func DidSpecifyEnvironmentName(cmd *cobra.Command, args []string) bool {
 	environmentName, err := cmd.Flags().GetString("environment")
-	return err == nil && environmentName != ""
+	return err == nil && environmentName != "" && !strings.HasPrefix(environmentName, resources.SegmentSeparator)
```
ytimocin, line 131: Does this apply to both just the environment name and also the resource id for the environment?

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -87,6 +87,10 @@ rad deploy myapp.bicep --environment production
 # deploy using a specific environment and resource group
 rad deploy myapp.bicep --environment production --group mygroup
 
+# deploy using an environment ID and a resource group. The application will be deployed in mygroup scope, using the specified environment.
+# use this option if the environment is in a different group.
+rad deploy myapp.bicep --environment /planes/radius/local/resourcegroups/prod/providers/Applications.Core/environments/prod --group mygroup
```
ytimocin, line 92: Will we be able to pass in just a name? It says here like that: https://github.com/radius-project/radius/pull/7895/files#diff-bf4e917446c73565ee6967f28bdd8e619c0d4fa40b215f195b840c021dd36ce9R104.

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -183,17 +187,17 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
...
-	env, err := client.GetEnvironment(cmd.Context(), r.EnvironmentName)
+	env, err := client.GetEnvironment(cmd.Context(), r.EnvironmentNameOrID)
 	if err != nil {
 		// If the error is not a 404, return it
 		if !clients.Is404Error(err) {
 			return err
 		}
 
-		// If the environment doesn't exist, but the user specified it as
+		// If the environment doesn't exist, but the user specified it's name as
```
ytimocin, line 197:
```suggestion
		// If the environment doesn't exist, but the user specified its name or resource id as
```

pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -53,7 +54,7 @@ func Test_Validate(t *testing.T) {
 			},
 			ConfigureMocks: func(mocks radcli.ValidateMocks) {
 				mocks.ApplicationManagementClient.EXPECT().
-					GetEnvironment(gomock.Any(), radcli.TestEnvironmentName).
+					GetEnvironment(gomock.Any(), "/planes/radius/local/resourceGroups/test-resource-group/providers/Applications.Core/environments/test-environment").
```
ytimocin, line 57: Should we have another test just with the name instead of the resource id?

pkg/cli/clivalidation.go
nithyatsu: The condition where workspace is editable and not set with env is checked at https://github.com/nithyatsu/radius/blob/1577a407a7079c53d56090882de2b74ceb2e99b8/pkg/cli/clivalidation.go#L120

This is because some of the cli code has the ability to make a fall back workspace to get to the server side  (this just injects kubectx = default kubernetes context) and continue to work if config.yaml is not found. If this is the workspace we are dealing with, it is not editable. Then we cannot set env.  L120 is the more common case.

pkg/cli/clivalidation.go
nithyatsu: No, it checks only for environment name. Since now we might be dealing with an id, it is updated to make sure we dont have the SegmentSeperator character that suggests this is an id.

pkg/cli/cmd/deploy/deploy.go
nithyatsu: Yes, we can still pass a name. In that case, the current group is searched for the environment (like before). But if we want to use an environment in a different group, then we would have to specify the ID so that we know the group.

pkg/cli/cmd/deploy/deploy_test.go
nithyatsu: I did not add it since most of the tests use a name and not ID for testing..

nithyatsu: sorry, I just noticed I covered it. They are "rad deploy - valid with environment"  and "rad deploy - valid with env ID".

pkg/cli/clivalidation.go
ytimocin: Do we need `DidSpecifyEnvironmentID`?

pkg/cli/cmd/deploy/deploy.go
ytimocin: Why do we need the resource group?

ytimocin: Can we also just do `rad deploy myapp.bicep --environment /planes/radius/local/resourcegroups/prod/providers/Applications.Core/environments/prod`?

ytimocin: Also, in the example the resource group in the environment ID is different than the resource group that is passed via a flag.

We should be more specific in these examples.

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -183,17 +187,17 @@ func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
...
 		if !clients.Is404Error(err) {
 			return err
 		}
 
-		// If the environment doesn't exist, but the user specified it as
+		// If the environment doesn't exist, but the user specified its name or resource id as
 		// a command-line option, return an error
 		if cli.DidSpecifyEnvironmentName(cmd, args) {
-			return clierrors.Message("The environment %q does not exist in scope %q. Run `rad env create` first.", r.EnvironmentName, r.Workspace.Scope)
+			return clierrors.Message("The environment %q does not exist in scope %q. Run `rad env create` first. You could also provide the environment ID if the environment exists in a different group.", r.EnvironmentNameOrID, r.Workspace.Scope)
```
ytimocin, line 200: Can we only provide an **environmentID** when we are deploying to another group?

nithyatsu: Yes. If not, it will enter the codeblock `if cli.DidSpecifyEnvironmentName` and ppint the error message indicating the environment isnt found in group and we should specify the id if it exists in a different group.

pkg/cli/cmd/deploy/deploy.go
nithyatsu: ```rad deploy myapp.bicep --environment /planes/radius/local/resourcegroups/prod/providers/Applications.Core/environments/prod --group mygroup```

This command deploys the application to group `mygroup` using environment  `/planes/radius/local/resourcegroups/prod/providers/Applications.Core/environments/prod`. 
--group flag existed already and specifies the group to which the application should be deployed to. There wasnt any ability to use an environment from another group though. But if we think about it, environments contain information used for deploying apps, and user should be able to use environments from any group.

pkg/cli/clivalidation.go
nithyatsu: No, there wasn't necessity for it.

ytimocin (approved):

PR #7907, opened 2024-09-09
Add workflow condition in cloud functional tests

### LINKED ISSUE #7782: `TypeError: Cannot read properties of undefined (reading 'id')` in `functional-test-cloud` workflow in main
### Steps to reproduce

In main branch, when the `functional-test-cloud` workflow runs, we get this error: `TypeError: Cannot read properties of undefined (reading 'id')`.

Here is an example workflow: https://github.com/radius-project/radius/actions/runs/10220061614/job/28279647756.

### Observed behavior

`functional-test-cloud` workflow fails on main.

### Desired behavior

`functional-test-cloud` workflow runs successfully on main.

### Workaround

_No response_

### rad Version

edge

### Operating system

_No response_

### Additional context

_No response_

[AB#12933](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/12933)

lakshmimsft:
# Description

Ref: https://github.com/radius-project/radius/pull/7892
Based on logs seen for errors in the past week,
eg. https://github.com/radius-project/radius/actions/runs/10753879482/job/29823820614

Adding condition to workflow to run when github.event.workflow_run.conclusion == 'success' when event is workflow_run.
Keeping additional debug logs in place for the next few days to monitor any subsequent failures. 
Another PR to remove logs will be sent closer to end of this week if no failures are detected.

## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius (#7782).

Fixes: #7782

### DIFF
--- a/.github/workflows/functional-test-cloud.yaml
+++ b/.github/workflows/functional-test-cloud.yaml
@@ -97,7 +97,10 @@ jobs:
   build:
     name: Build Radius for test
     runs-on: ubuntu-latest
-    if: github.event_name == 'repository_dispatch' || (github.event_name == 'schedule' && github.repository == 'radius-project/radius') || github.event_name == 'workflow_run' || github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'repository_dispatch' ||
+        (github.event_name == 'schedule' && github.repository == 'radius-project/radius') ||
+        (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') ||
+        github.event_name == 'workflow_dispatch'
     env:
       DE_IMAGE: "ghcr.io/radius-project/deployment-engine"
       DE_TAG: "latest"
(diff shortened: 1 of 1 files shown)


ytimocin:
The problem is here: https://github.com/radius-project/radius/blob/cce1fe68891739c7e66181770edba82c277ba765/.github/actions/download-pr-data-artifact/action.yaml#L21. When this action is called from not a pull request then the **matchArtifact** is null and, that is why, the **matchArtifact.id** throws an error: https://github.com/radius-project/radius/blob/cce1fe68891739c7e66181770edba82c277ba765/.github/actions/download-pr-data-artifact/action.yaml#L27.

@lakshmimsft

lakshmimsft:
@ytimocin, so, we want to restrict the downstream workflow (cloud functional tests) to run on valid approved PRs for the event is workflow_run. All the errored workflow runs the past few days were on branches that had a 'failure' conclusion status (so far i looked, it was set to failure for either timing out in approvals or deploy errors on the workflow), There were not valid active PRs and these branches were deleted (in the above example) multiple weeks ago. 
The downstream functional cloud test workflow should not run attempt to run on these failed upstream workflow runs.  As you have pointed out, the artifacts will not exist. 

We still want to continue to monitor and see if we still get this error when the conclusion is a 'success' and debug why the artifacts will not exist in those cases. Let me know if that makes sense or we should change approach.

ytimocin (approved):
We can give it a try.

PR #7937, opened 2024-09-18
Separating purge old images and purge Azure test resources workflows

ytimocin:
# Description
1. Separating purge old images and purge Azure test resources workflows.
2. Updating **snok** version and params in the purge old images workflow.

## Type of change

### DIFF
--- a/.github/workflows/purge-azure-test-resources.yaml
+++ b/.github/workflows/purge-azure-test-resources.yaml
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ------------------------------------------------------------
 
-name: Purge test resources
+name: Purge Azure test resources
 
 permissions:
   id-token: write # Required for requesting the JWT
@@ -36,23 +36,8 @@ env:
   ACTION_LINK: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
 jobs:
-  purge_ghcr_dev:
-    name: Delete old unused container images
-    runs-on: ubuntu-latest
-    if: github.repository == 'radius-project/radius'
-    steps:
-      - name: Delete 'dev' containers older than a week
-        uses: snok/container-retention-policy@v3.0.0
-        with:
-          account: radius-project
-          token: ${{ secrets.GH_RAD_CI_BOT_PAT }}
-          image-names: dev/*
-          image-tags: "pr-*"
-          cut-off: 1w
-          keep-n-most-recent: 1
-
   purge_azure_resources:
-    name: Azure resources clean-ups
+    name: Clean up unused Azure resources
     runs-on: ubuntu-latest
     if: github.repository == 'radius-project/radius'
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/purge-azure-test-resources.yaml
```diff
@@ -108,7 +93,7 @@ jobs:
 
   create_issue_on_failure:
     name: Create issue for failing purge test resources run
-    needs: [purge_ghcr_dev, purge_azure_resources]
+    needs: [purge_azure_resources]
```
nithyatsu, line 96: Are we choosing to not create issue when purging containers fails since it seems to be just flaky?

ytimocin: Nope, I just separated these two steps. They do different things.

ytimocin:
> Just trying to understand, are we trying to separate the workflows because it is the purging old container images that fails quite often?

Nope. They do different things. One of them purges Azure resources and the other one deletes old images. Like how we have `purge AWS resources` separately, I think these should be separate too.

One of them may fail and the other one may not. They are different workflows. They should have been separate in the first place.

nithyatsu:
LGTM

lakshmimsft:
Question: Both these workflows will run at the same time. Is it better if we separate their run times? It seems most likely that should have no impact, but checking anyways.

.github/workflows/purge-old-images.yaml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: Delete 'dev' containers older than a week
+        uses: snok/container-retention-policy@v3.0.0
+        with:
+          account: radius-project
+          token: ${{ secrets.GH_RAD_CI_BOT_PAT }}
+          image-names: dev/*
+          image-tags: "pr-*"
+          cut-off: 1w
+          keep-n-most-recent: 1
```
lakshmimsft, line 32: question: why do we keep 1? if we're cutting off older than a week?

ytimocin: I seriously don't have any strong opinions on this one. Not also sure what happens if there are 0 images in a package. What do you think? Any reason we shouldn't keep 1?

lakshmimsft: The package should still exist, just no images. Overall keeping one image version alive should not be a problem.  We can go ahead.

lakshmimsft (dismissed):

kachawla (commented):
What's the benefit of separating it into two workflows?

lakshmimsft (dismissed):

.github/workflows/purge-old-images.yaml
```diff
@@ -0,0 +1,48 @@
+name: Purge old images
```
kachawla, line 1: Let's give it a more explicit name/description. One suggestion below, feel free to rephrase it.

```suggestion
name: Purge test container images
```

kachawla:
> What's the benefit of separating it into two workflows?

Saw this was already captured in https://github.com/radius-project/radius/pull/7937#issuecomment-2359081395

ytimocin:
> What's the benefit of separating it into two workflows?

I think these can be some reasons and benefits:
- They do different things, and I don't think there is no point of keeping them in the same workflow.
- We have a separate workflow for `purge AWS test resources` and I guess we should keep our workflow separation consistent.
- We can have better visibility on which one fails more, and we can focus on fixing that one.
- Our workflows could benefit from being simple. Like the name of this workflow was `purge test resources`. I don't think this tells us what that workflow does in a good way. But `purge AWS test resources` is a good name for example. I believe we can follow the same pattern in our workflows.

kachawla (approved):

PR #7938, opened 2024-09-18
Update the timeout of the clean-up cluster step to 1h

ytimocin:
# Description
Update the timeout of the clean-up cluster step to 1h.

## Type of change

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -382,7 +382,9 @@ jobs:
         env:
           RESOURCE_GROUP: ${{ env.AZURE_TEST_RESOURCE_GROUP }}
       - name: Clean up cluster
-        run: ./.github/scripts/cleanup-cluster.sh
+        if: always()
+        timeout-minutes: 60
+        run: ./.github/scripts/cleanup-long-running-cluster.sh
       - name: Download Bicep
         run: |
           chmod +x ./bin/rad
@@ -512,7 +514,7 @@ jobs:
       - name: Clean up cluster
         if: always()
         timeout-minutes: 60
-        run: ./.github/scripts/cleanup-cluster.sh
+        run: ./.github/scripts/cleanup-long-running-cluster.sh
   report-failure:
     name: Report test failure
     needs: [build, tests]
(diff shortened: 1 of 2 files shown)


vishwahiremat (approved):

.github/workflows/long-running-azure.yaml
```diff
@@ -512,7 +514,7 @@ jobs:
       - name: Clean up cluster
```
sk593, line 514: Do we need to have two cleanup cluster steps? These both seem to be part of the test job

ytimocin: One cleans up before the tests and the other one after the tests. This is to be extra defensive. We want to have a clean start before running the tests and want to make sure we delete everything after we run the tests.

sk593 (approved):

.github/workflows/long-running-azure.yaml
sk593: makes sense. PR lgtm then

lakshmimsft:
@ytimocin  I am curious if this will alleviate the issue. If we are terminating the script in an hour and there are still resources/finalizers than need clean up, will they be cleaned up while the finalizers are still attached during the second run on the script? Based on the overlapping runs for multiple days, it seems like multiple runs of the script did not help clean it up.
let's try this. If it works, great, if not, maybe after the first run of the script, we can cycle through all resources in the namespace with 'Terminating' state and log them, so we narrow down what the finalizers are waiting on.

lakshmimsft (approved):

ytimocin:
> @ytimocin I am curious if this will alleviate the issue. If we are terminating the script in an hour and there are still resources/finalizers than need clean up, will they be cleaned up while the finalizers are still attached during the second run on the script? Based on the overlapping runs for multiple days, it seems like multiple runs of the script did not help clean it up. Atleast we will not wait for the script to complete over an hour, so that is good. let's try this. If it works, great, if not, maybe after the first run of the script, we can cycle through all resources in the namespace with 'Terminating' state and log them, so we narrow down what the finalizers are waiting on.

The problem with 6-hour timeout is that it doesn't do clean-up during that time. It just waits for the finalizers to finalize. But that doesn't happen. If the workflow waited for 6 days, it wouldn't also happen. The problem is something else.

I wanted to reduce the time to 1h so that we don't wait 6 hours for a run to be finalized.

Thanks for the feedback and the approval @lakshmimsft!

PR #7942, opened 2024-09-19
Add dapr-sts* namespaces to be deleted to the clean-up workflow

ytimocin:
# Description

This new namespace was added a few weeks ago and wasn't added to the cleanup-cluster.sh. I saw this when checking the long-running cluster. Adding this namespace.

## Type of change

### DIFF
--- a/.github/scripts/cleanup-long-running-cluster.sh
+++ b/.github/scripts/cleanup-long-running-cluster.sh
@@ -45,7 +45,7 @@ fi
 # Delete all test namespaces.
 echo "delete all test namespaces"
 namespaces=$(kubectl get namespace |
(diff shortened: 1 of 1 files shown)


lakshmimsft (dismissed):

lakshmimsft (approved):

kachawla (commented):
Thanks for updating this @ytimocin. This doesn't seem a reliable approach though. Would there be an issue if we delete all except radius-system namespace? That way we won't have to maintain a hardcoded list.

ytimocin:
> Thanks for updating this @ytimocin. This doesn't seem a reliable approach though. Would there be an issue if we delete all except radius-system namespace? That way we won't have to maintain a hardcoded list.

I agree with you @kachawla. We need to find a better approach for this. I will create an issue on this one tomorrow.

kachawla:
> > Thanks for updating this @ytimocin. This doesn't seem a reliable approach though. Would there be an issue if we delete all except radius-system namespace? That way we won't have to maintain a hardcoded list.
> 
> I agree with you @kachawla. We need to find a better approach for this. I will create an issue on this one tomorrow.

Thanks for looking into this @ytimocin! Created https://github.com/radius-project/radius/issues/7966

PR #7949, opened 2024-09-20
Improve validation and simplify database tests

rynowak:
# Description

This is an improvement to the Radius datastore package that improves the validation and enforces some useful invariants for the design.

Our datastore package is design to support multiple implementations across different technologies (ranging from KV-stores to document-stores to RDBMSes). We have a few of these implementations today but they are inconsistent in how they work.

This change:

- Improves and consolidates validation for the Query() function.
- Adds a new invariant that ResourceType must be provided for a Query().
- Removes tests for cases that are no longer possible.

The new invariant (resource type must be provided for a query) was enforced for the CosmosDB store, but not for the ETC.d or API Server stores. As a result the CosmosDB store was more limited. I'm working on some cleanup in the area and so I wanted to simplify the interface towards the bare-minimum number of requirements. The result is that we can delete tests and functionality, and it will be easier to add new store implementations in the future.

In general our access patterns support the new invariant. One place in code that had to be touched was the `/planes` endpoint in UCP. This would not have worked correctly using the existing CosmosDB store.

## Type of change

### DIFF
--- a/pkg/ucp/api/v20231001preview/awsplane_conversion_test.go
+++ b/pkg/ucp/api/v20231001preview/awsplane_conversion_test.go
@@ -41,7 +41,7 @@ func Test_AWSPlane_ConvertVersionedToDataModel(t *testing.T) {
 					TrackedResource: v1.TrackedResource{
 						ID:       "/planes/aws/aws",
 						Name:     "aws",
-						Type:     "System.AWS/planes",
+						Type:     datamodel.AWSPlaneResourceType,
 						Location: "global",
 						Tags: map[string]string{
 							"env": "dev",
@@ -87,7 +87,7 @@ func Test_AWSPlane_ConvertDataModelToVersioned(t *testing.T) {
 			expected: &AwsPlaneResource{
 				ID:       to.Ptr("/planes/aws/aws"),
 				Name:     to.Ptr("aws"),
-				Type:     to.Ptr("System.AWS/planes"),
+				Type:     to.Ptr(datamodel.AWSPlaneResourceType),
 				Location: to.Ptr("global"),
 				Tags: map[string]*string{
 					"env": to.Ptr("dev"),
--- a/pkg/ucp/frontend/controller/planes/listplanes.go
+++ b/pkg/ucp/frontend/controller/planes/listplanes.go
@@ -54,28 +54,43 @@ func (e *ListPlanes) Run(ctx context.Context, w http.ResponseWriter, req *http.R
 	serviceCtx := v1.ARMRequestContextFromContext(ctx)
 	logger := ucplog.FromContextOrDiscard(ctx)
 
-	query := store.Query{
-		RootScope:    serviceCtx.ResourceID.String(),
-		IsScopeQuery: true,
+	// The plane objects are all stored separately (by plane type). We need to query each type separately.
+	planeTypes := []string{
+		"aws",
+		"azure",
+		"radius",
 	}
-	logger.Info(fmt.Sprintf("Listing planes in scope %s", query.RootScope))
-	result, err := e.StorageClient().Query(ctx, query)
+
+	objs := []store.Object{}
+	for _, planeType := range planeTypes {
+		query := store.Query{
+			RootScope:    serviceCtx.ResourceID.String(),
+			ResourceType: planeType,
+			IsScopeQuery: true,
+		}
+
+		logger.Info(fmt.Sprintf("Listing planes of type %s in scope %s", query.ResourceType, query.RootScope))
+		result, err := e.StorageClient().Query(ctx, query)
+		if err != nil {
+			return nil, err
+		}
+
+		objs = append(objs, result.Items...)
+	}
+
+	planes, err := e.createResponse(ctx, objs)
 	if err != nil {
 		return nil, err
 	}
-	listOfPlanes, err := e.createResponse(ctx, result)
-	if err != nil {
-		return nil, err
-	}
-	var ok = armrpc_rest.NewOKResponse(listOfPlanes)
+	var ok = armrpc_rest.NewOKResponse(planes)
 	return ok, nil
 }
 
-func (p *ListPlanes) createResponse(ctx context.Context, result *store.ObjectQueryResult) (*v1.PaginatedList, error) {
+func (p *ListPlanes) createResponse(ctx context.Context, objs []store.Object) (*v1.PaginatedList, error) {
 	serviceCtx := v1.ARMRequestContextFromContext(ctx)
 	items := v1.PaginatedList{}
 
-	for _, item := range result.Items {
+	for _, item := range objs {
 		var plane datamodel.GenericPlane
 		err := item.As(&plane)
 		if err != nil {
--- a/pkg/ucp/frontend/controller/planes/listplanes_test.go
+++ b/pkg/ucp/frontend/controller/planes/listplanes_test.go
@@ -41,14 +41,9 @@ func Test_ListPlanes(t *testing.T) {
 
 	url := "/planes?api-version=2023-10-01-preview"
 
-	query := store.Query{
-		RootScope:    "/planes",
-		IsScopeQuery: true,
-	}
-
-	testPlaneId := "/planes/radius"
-	testPlaneName := "radius"
-	testPlaneType := "planes"
+	testPlaneId := "/planes/aws"
+	testPlaneName := "aws"
+	testPlaneType := datamodel.AWSPlaneResourceType
 
 	planeData := datamodel.AWSPlane{
 		BaseResource: v1.BaseResource{
@@ -62,7 +57,11 @@ func Test_ListPlanes(t *testing.T) {
 		Properties: datamodel.AWSPlaneProperties{},
 	}
 
-	mockStorageClient.EXPECT().Query(gomock.Any(), query).Return(&store.ObjectQueryResult{
+	mockStorageClient.EXPECT().Query(gomock.Any(), store.Query{
+		RootScope:    "/planes",
+		ResourceType: "aws",
+		IsScopeQuery: true,
+	}).Return(&store.ObjectQueryResult{
 		Items: []store.Object{
 			{
 				Metadata: store.Metadata{},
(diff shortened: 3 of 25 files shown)


pkg/ucp/api/v20231001preview/awsplane_conversion_test.go
```diff
@@ -41,7 +41,7 @@ func Test_AWSPlane_ConvertVersionedToDataModel(t *testing.T) {
 					TrackedResource: v1.TrackedResource{
 						ID:       "/planes/aws/aws",
 						Name:     "aws",
-						Type:     "System.AWS/planes",
+						Type:     datamodel.AWSPlaneResourceType,
```
rynowak, line 44: While working on this feature I noticed that we didn't have a single definition for these values, so I added one. This is just a refactor and won't affect behaviors.

pkg/ucp/frontend/controller/planes/listplanes.go
```diff
@@ -54,28 +54,43 @@ func (e *ListPlanes) Run(ctx context.Context, w http.ResponseWriter, req *http.R
...
-	result, err := e.StorageClient().Query(ctx, query)
-	if err != nil {
-		return nil, err
+
+	objs := []store.Object{}
+	for _, planeType := range planeTypes {
+		query := store.Query{
+			RootScope:    serviceCtx.ResourceID.String(),
+			ResourceType: planeType,
+			IsScopeQuery: true,
```
rynowak, line 69: This is the one place where we did a "multi-type query". eg: the only place besides tests where I had to change the code.

pkg/ucp/frontend/controller/planes/listplanes_test.go
```diff
@@ -71,6 +70,18 @@ func Test_ListPlanes(t *testing.T) {
...
+		ResourceType: "azure",
+		IsScopeQuery: true,
+	}).Return(&store.ObjectQueryResult{}, nil)
+
+	mockStorageClient.EXPECT().Query(gomock.Any(), store.Query{
+		RootScope:    "/planes",
+		ResourceType: "radius",
+		IsScopeQuery: true,
+	}).Return(&store.ObjectQueryResult{}, nil)
+
```
rynowak, line 84: Updated test code to reflect the fact that three queries are done.

pkg/ucp/store/apiserverstore/apiserverclient.go
```diff
@@ -105,11 +105,9 @@ func (c *APIServerClient) Query(ctx context.Context, query store.Query, options
...
 		return nil, &store.ErrInvalid{Message: "invalid argument. 'ctx' is required"}
 	}
-	if query.RootScope == "" {
-		return nil, &store.ErrInvalid{Message: "invalid argument. 'query.RootScope' is required"}
-	}
-	if query.IsScopeQuery && query.RoutingScopePrefix != "" {
-		return nil, &store.ErrInvalid{Message: "invalid argument. 'query.RoutingScopePrefix' is not supported for scope queries"}
+	err := query.Validate()
+	if err != nil {
+		return nil, &store.ErrInvalid{Message: fmt.Sprintf("invalid argument. Query is invalid: %s", err.Error())}
```
rynowak, line 110: Logic for validating a query has been consolidated.

pkg/ucp/store/apiserverstore/apiserverclient.go
```diff
@@ -535,10 +533,8 @@ func readEntry(entry *ucpv1alpha1.ResourceEntry) (*store.Object, error) {
 
 	obj := store.Object{
 		Metadata: store.Metadata{
-			ID:          entry.ID,
-			ETag:        entry.ETag,
-			APIVersion:  entry.APIVersion,
-			ContentType: entry.ContentType,
```
rynowak, line 541: These fields are unused. There's no sense storing extra data that's never read.

pkg/ucp/store/apiserverstore/apiserverclient_test.go
```diff
@@ -225,79 +225,79 @@ func Test_APIServer_Client(t *testing.T) {
 
 		// In this test we're going to **similuate** a hash collision and verify that it is saved correctly.
 		//
-		// Let's PRETEND that shared.BasicResource1ID and shared.BasicResource2ID result in the same
+		// Let's PRETEND that shared.BasicResource2ID and shared.BasicResource3ID result in the same
```
rynowak, line 228: This test needed a rework because it relied on a "multi-type query". I made changes here so that it's still covering the same case, but with two resources of the same type.

pkg/ucp/store/client.go
```diff
@@ -18,8 +18,19 @@ package store
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"regexp"
 )
 
+// jsonPropertyPattern is the pattern for a valid JSON property name.
+var jsonPropertyPattern = "[a-zA-Z$_][a-zA-Z0-9$_]*"
```
rynowak, line 27: The "field query" feature was not documented anywhere in code and the inputs were not validated. This change documents the **current** behavior and validates that the fields are specified using the correct syntax.

pkg/ucp/store/client.go
```diff
@@ -77,8 +88,55 @@ type Query struct {
 	Filters []QueryFilter
 }
 
+// Validate validates the Query.
+func (q Query) Validate() error {
+	var err error
+	if q.RootScope == "" {
+		err = errors.Join(err, &ErrInvalid{Message: "RootScope is required"})
```
rynowak, line 95: For reviewers - `errors.Join` is relatively new in go and useful for places like this. Instead of reporting the first error, we can report all of them.

test/ucp/storetest/shared.go
```diff
@@ -37,13 +37,15 @@ const (
 	ResourcePath2       = "System.Resources/resourceType2/resource2"
 	ResourcePath3       = "System.Resources/resourceType2/Resource3"
 	NestedResourcePath1 = "System.Resources/resourceType1/resource1/nestedType/nested1"
+	NestedResourcePath2 = "System.Resources/resourceType1/resource1/nestedType/nested2"
+	NestedResourcePath3 = "System.Resources/resourceType1/resource2/nestedType/nested3"
+	NestedResourcePath4 = "System.Resources/resourceType1/resource2/nestedType/nested4"
```
rynowak, line 42: Now that queries are limited to a single type, I needed to add some more resources to make sure we're testing the same logic.

test/ucp/storetest/shared.go
```diff
@@ -397,87 +425,65 @@ func RunTest(t *testing.T, client store.StorageClient, clear func(t *testing.T))
...
-			filters := []store.QueryFilter{{Field: "value", Value: "1"}}
-			objs, err := client.Query(ctx, store.Query{RootScope: ResourceGroup1Scope, Filters: filters})
-			require.NoError(t, err)
-			expected := []store.Object{
-				obj1,
-			}
-			CompareObjectLists(t, expected, objs.Items)
-		})
-
-		t.Run("query_resources_at_resource_group_scope_with_prefix", func(t *testing.T) {
```
rynowak, line 419: You'll notice a lot of tests were removed - about 40% of the test cases for Query. That's a **good thing**. We removed the need for a lot of these.

pkg/ucp/frontend/controller/planes/listplanes.go
```diff
@@ -54,28 +54,43 @@ func (e *ListPlanes) Run(ctx context.Context, w http.ResponseWriter, req *http.R
 	serviceCtx := v1.ARMRequestContextFromContext(ctx)
 	logger := ucplog.FromContextOrDiscard(ctx)
 
-	query := store.Query{
-		RootScope:    serviceCtx.ResourceID.String(),
-		IsScopeQuery: true,
+	// The plane objects are all stored separately (by plane type). We need to query each type separately.
+	planeTypes := []string{
```
ytimocin, line 58: Should we put this in the `types.go` or something like that and call it **availablePlaneTypes**?

test/ucp/storetest/shared.go
```diff
@@ -81,13 +86,36 @@ var Data2 = map[string]any{
...
 }
-
-var NestedData1 = map[string]any{
+var Data3 = map[string]any{
 	"value": "3",
 	"properties": map[string]any{
 		"resource": "3",
 	},
 }
+var NestedData1 = map[string]any{
```
ytimocin, line 95: any reason why these variables start with capital case?

ytimocin (dismissed):

pkg/ucp/store/object.go
```diff
@@ -23,10 +23,8 @@ import (
 type ETag = string
 
 type Metadata struct {
-	ID          string
-	ETag        ETag
-	APIVersion  string
-	ContentType string
+	ID   string
+	ETag ETag
```
lakshmimsft, line 27: Can you pls give more info on this change?

test/ucp/storetest/shared.go
rynowak: These are exported from the package, because they are used by other tests. Everything in `storetest` is designed to be common between the tests for all of our implementations.

pkg/ucp/store/object.go
rynowak: Sure. The `APIVersion` and `ContentType` fields have been part of the database implementation for a long time. `APIVersion` is redundant with a field in the datamodel. `ContentType` has always been unused. 

You'll notice that I only had to update test code to remove this - and one of the DB implementations. Nothing reads these fields.

pkg/ucp/frontend/controller/planes/listplanes.go
rynowak: If we end up with a second location where we need this list then sure.

rynowak:
Currently investigating the functional test failure.

rynowak:
@ytimocin updated

lakshmimsft (approved):

rynowak:
Thanks @lakshmimsft !

rynowak:
@ytimocin if you have additional feedback I can address in a follow-up

PR #7950, opened 2024-09-20
Add an in-memory data store

rynowak:
# Description

Implements an in-memory data store

This change implements an in-memory version of our data store interface. This is provided for testing and development purposes.

The in-memory store uses our data store compliance tests so it will behave consistently with other implementations. This will be simpler to use than mocks, which we use frequently in tests. This will be faster to use than the ETC.d implementation which we frequently use in integration tests.

**Note: this pull request depends on https://github.com/radius-project/radius/pull/7949 and will be rebased once that change is merged.**

## Type of change

### DIFF
--- a/pkg/ucp/store/client.go
+++ b/pkg/ucp/store/client.go
@@ -33,10 +33,52 @@ var fieldRegex = regexp.MustCompile(fmt.Sprintf(`^(%s)(\.%s)*$`, jsonPropertyPat
 
 //go:generate mockgen -typed -destination=./mock_storageClient.go -package=store -self_package github.com/radius-project/radius/pkg/ucp/store github.com/radius-project/radius/pkg/ucp/store StorageClient
 
+// StorageClient is the interface for persisting and querying resource data.
+//
+// The StorageClient is purpose-built to work with resource data and understands concepts like
+// scopes, resource types, and resource ids. This is a higher level abstraction than a generic
+// key-value store, but low-level enough to support multiple implementation strategies.
+//
+// The StorageClient provides a optimistic concurrency control using ETags. Callers that want
+// to enforce OCC should provide the ETag value in the options when calling Save or Delete.
+//
+// The StorageClient may return the errors ErrNotFound, ErrInvalid, and ErrConcurrency.
+//
+// - Callers should handle ErrNotFound on Get, Save, and Delete operations.
+// - Callers should handle ErrConcurrency when using ETags.
+// - Callers should not handle ErrInvalid as it represents a programming error.
+//
+// When using ETags, the Save or Delete operation will fail with ErrConcurrency (rather than ErrNotFound)
+// if the underlying resource has been deleted.
 type StorageClient interface {
+	// Query executes a query against the data store and returns the results.
+	//
+	// Queries must provide a root scope and a resource type. Other fields are optional.
 	Query(ctx context.Context, query Query, options ...QueryOptions) (*ObjectQueryResult, error)
+
+	// Get retrieves a single resource from the data store by its resource id.
+	//
+	// Get will return ErrNotFound if the resource is not found.
 	Get(ctx context.Context, id string, options ...GetOptions) (*Object, error)
+
+	// Delete removes a single resource from the data store by its resource id.
+	//
+	// Delete will return ErrNotFound if the resource is not found.
+	// When providing an ETag, Delete will return ErrConcurrency if the resource has been
+	// modified OR deleted since the ETag was retrieved.
 	Delete(ctx context.Context, id string, options ...DeleteOptions) error
+
+	// Save persists a single resource to the data store. Same is a logical PUT
+	// operation and will either create a new entry or update the existing entry.
+	//
+	// Save operations must set the ID field of the obj parameter.
+	// The ETag field of the obj parameter is read-only and will be updated by the Save operation.
+	//
+	// Use the options to pass an ETag if you want to enforce optimistic concurrency control.
+	//
+	// Save will return ErrNotFound if the resource is not found.
+	// When providing an ETag, Save will return ErrConcurrency if the resource has been
+	// modified OR deleted since the ETag was retrieved.
 	Save(ctx context.Context, obj *Object, options ...SaveOptions) error
 }
 
--- a/pkg/ucp/store/inmemory/client.go
+++ b/pkg/ucp/store/inmemory/client.go
@@ -1,0 +1,289 @@
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inmemory
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/radius-project/radius/pkg/ucp/resources"
+	"github.com/radius-project/radius/pkg/ucp/store"
+	"github.com/radius-project/radius/pkg/ucp/store/storeutil"
+	"github.com/radius-project/radius/pkg/ucp/util/etag"
+	"golang.org/x/exp/maps"
+)
+
+var _ store.StorageClient = (*Client)(nil)
+
+// Client is an in-memory implementation of store.StorageClient.
+type Client struct {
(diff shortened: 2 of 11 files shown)


pkg/ucp/api/v20231001preview/awsplane_conversion_test.go
```diff
@@ -41,7 +41,7 @@ func Test_AWSPlane_ConvertVersionedToDataModel(t *testing.T) {
 					TrackedResource: v1.TrackedResource{
 						ID:       "/planes/aws/aws",
 						Name:     "aws",
-						Type:     "System.AWS/planes",
+						Type:     datamodel.AWSPlaneResourceType,
```
ytimocin, line 44: I think this requires a rebase when the other PR gets merged in.

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+
+// NewClient creates a new in-memory store client.
+func NewClient() *Client {
+	return &Client{
+		mutex:     sync.Mutex{},
+		resources: map[string]entry{},
+	}
+}
+
+// Get implements store.StorageClient.
```
ytimocin, line 84: Do the docs need an update?

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+		return err
+	}
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	config := store.NewDeleteConfig(options...)
+
+	entry, ok := c.resources[strings.ToLower(normalized.String())]
+	if !ok && config.ETag != "" {
```
ytimocin, line 143: Could you explain this? If entry is not found and `config.ETag` is not empty, why is it a Concurrency error? If entry is found and ETag is not the most recent then I think that should be a concurrency issue.

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+			continue
+		} else if !query.ScopeRecursive && entry.rootScope != storeutil.NormalizePart(query.RootScope) {
+			continue
+		}
+
+		// Check resource type.
+		resourceType, err := storeutil.NormalizeResourceType(query.ResourceType)
+		if err != nil {
+			return nil, err
+		}
```
ytimocin, line 183: We can do this outside the for loop, right?

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	config := store.NewSaveConfig(options...)
+
+	entry, ok := c.resources[strings.ToLower(normalized.String())]
+	if !ok && config.ETag != "" {
+		return &store.ErrConcurrency{}
+	} else if ok && config.ETag != "" && config.ETag != entry.obj.ETag {
```
ytimocin, line 235: Can we have a case where we have the resource (ok is true) and etags are also equal?

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+
+	obj.ETag = etag.New(raw)
+	entry.obj = *obj
+	c.resources[strings.ToLower(normalized.String())] = entry
+
+	return nil
+}
+
+// Clear can be used to clear all stored data.
+func (c *Client) Clear() {
```
ytimocin, line 257: This can be useful for 409 errors :)

pkg/ucp/api/v20231001preview/awsplane_conversion_test.go
rynowak: That's the idea 👍

pkg/ucp/store/inmemory/client.go
rynowak: Is there something interesting you want to be documented here? I don't think it's useful to copy-paste things.

pkg/ucp/store/inmemory/client.go
rynowak: First, this isn't new behavior. The tests for the data store have been around for a long time. Second, I think it's the least surprising behavior and therefore the most correct thing to do.

Any operation that uses ETags is an example of *optimistic concurrency control*. Here's a relevant snippet from the [wikipedia description](https://en.wikipedia.org/wiki/Optimistic_concurrency_control).

> While running, transactions use data resources without acquiring locks on those resources. Before committing, each transaction verifies that no other transaction has modified the data it has read. If the check reveals conflicting modifications, the committing transaction rolls back and can be restarted.

This supports the following pattern:

- An actor reads the current state of the resource (including it's ETag).
- It applies some business logic based on the state of the resource and decides to delete the resource.
- It uses the ETag so that the delete operation will **ONLY** be persisted if the resource is unchanged.

There are three possible cases:

- The resource was changed (and has a new ETag).
- The resource was deleted.
- The resource is unchanged and will be deleted.

The behavior of the datastore collapses the first two cases into a single one.

Keep in mind that most code will not use ETags.

pkg/ucp/store/inmemory/client.go
rynowak: We could. I'm not worried about optimizing the behavior of the in-memory store unless we have a reason to.

pkg/ucp/store/inmemory/client.go
rynowak: I don't get it, but if you're psyched them so am I 😆

pkg/ucp/store/inmemory/client.go
rynowak: Yes 😁

rynowak:
@ytimocin @lakshmimsft - this one is ready to review now.

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,262 @@
...
+//
+// The entry would be:
+//
+//	rootScope: "/planes/radius/local/resourceGroups/my-rg"
+//	resourceType: "Applications.Test/testType1/testType2"
+//	routingScope: "Applications.Test/testType1/my-resource/testType2/my-child-resource"
+//
+// All fields are compared case-insensitively.
+type entry struct {
+	obj store.Object
```
lakshmimsft, line 64: nit: // obj holds the resource object.

rynowak:
@lakshmimsft @ytimocin - updated

pkg/ucp/store/inmemory/client.go
rynowak: We talked offline and agreed not to duplicate the StorageClient documentation. I also saw that `StorageClient` docs were missing, so I wrote some.

pkg/ucp/store/inmemory/client.go
```diff
@@ -0,0 +1,289 @@
...
+// - Add leading/trailing slashes.
+//
+// Here's an example:
+//
+//	resource ID: "/planes/radius/local/resourceGroups/my-rg/providers/Applications.Test/testType1/my-resource/testType2/my-child-resource"
+//
+// The entry would be:
+//
+//	rootScope: "/planes/radius/local/resourcegroups/my-rg/"
+//	resourceType: "/applications.test/testtype1/testtype2/"
```
ytimocin, line 64:
```suggestion
//	resourceType: "/applications.test/testtype1/my-resource/"
```

ytimocin: This is, I guess, about the child resource types, right?

- **Applications.Test** is the provider.
- **testType1** is the resource type under that provider.
- **my-resource** is the name of the resource of the type mentioned above.
- **testType2** is the child resource type under the **testType1** resource type.
- **my-child-resource** is the name of the resource of type **testType2**.

ytimocin (approved):

pkg/ucp/store/client.go
```diff
@@ -33,10 +33,52 @@ var fieldRegex = regexp.MustCompile(fmt.Sprintf(`^(%s)(\.%s)*$`, jsonPropertyPat
...
 //go:generate mockgen -typed -destination=./mock_storageClient.go -package=store -self_package github.com/radius-project/radius/pkg/ucp/store github.com/radius-project/radius/pkg/ucp/store StorageClient
 
+// StorageClient is the interface for persisting and querying resource data.
+//
+// The StorageClient is purpose-built to work with resource data and understands concepts like
+// scopes, resource types, and resource ids. This is a higher level abstraction than a generic
+// key-value store, but low-level enough to support multiple implementation strategies.
+//
+// The StorageClient provides a optimistic concurrency control using ETags. Callers that want
+// to enforce OCC should provide the ETag value in the options when calling Save or Delete.
```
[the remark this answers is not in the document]
rynowak: Get doesn't update any state, so there's nothing to enforce.

pkg/ucp/store/inmemory/client.go
rynowak: The resource type is the resource type. It doesn't contain the resource name. 

This is not meant to be documentation for what a resource id **is** this is documentation about how the in-memory data store works.

PR #7961, opened 2024-09-23
Implement PostgreSQL datastore

rynowak:
# Description

This change implements our datastore abstraction using PostgreSQL. This is an easier and better option for users to manage than our current set of datastores. The next pull-request in this series will hook it up to Radius so we can start using it. Ultimately we can replace the current default of Kubernetes CRD-store for both dev and deployed configurations in a few steps.

Also included in this pull-request is a set of new makefile commands (`make db-*`) to make it easier to work with the local database. As always run `make` to see the command-line help.

## Type of change


rynowak:
Still TODO for me:

- Add some developer-focused documentation.
- Wire up to our configuration files.
- Add `make db-init` to our pipelines for testing.

Makefile
```diff
@@ -17,4 +17,4 @@
 ARROW := \033[34;1m=>\033[0m
 
 # order matters for these
```
rynowak, line 19: For reviewers:

- db.sql.text is the best place to start to understand the schema. This works the same way as our existing data store implementations, but it's using SQL. 
- postgresclient.go has the actual implementation.
- postgresclient_test.go is where we hook up the existing tests.

deploy/init-db/init-db.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
rynowak, line 1: For reviewers: This file is passed in to the postgres container when running locally, and used to initialize databases, users, and tables.

UCP and Applications-RP will each have separate databases. We **DO NOT** want cross-component database reads.

go.mod
```diff
@@ -50,6 +51,7 @@ require (
 	github.com/hashicorp/hc-install v0.8.0
 	github.com/hashicorp/terraform-config-inspect v0.0.0-20240607080351-271db412dbcb
 	github.com/hashicorp/terraform-exec v0.21.0
+	github.com/jackc/pgx/v5 v5.7.1
```
rynowak, line 54: This is what I added. From what I can see this is the best-maintained postgres driver for Go. The other changes to this file were done by `go mod tidy`.

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,414 @@
...
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/radius-project/radius/pkg/to"
+	"github.com/radius-project/radius/pkg/ucp/resources"
+	"github.com/radius-project/radius/pkg/ucp/store"
+	"github.com/radius-project/radius/pkg/ucp/store/storeutil"
+	"github.com/radius-project/radius/pkg/ucp/util/etag"
+)
+
+type PostgresAPI interface {
```
rynowak, line 36: Oops will add doc comments.

run-postgres.sh
```diff
@@ -0,0 +1,18 @@
+#! /usr/bin/env bash
+set -e
```
rynowak, line 2: Will remove. This got moved to the makefile.

build/db.mk
```diff
@@ -0,0 +1,118 @@
...
+
+.PHONY: db-dependency-docker-running
+db-dependency-docker-running:
+	@if [ ! docker info > /dev/null 2>&1 ]; then \
+		echo "Docker is not installed or not running. Please install docker and try again."; \
+		exit 1; \
+	fi;
+
+.PHONY: db-postgres-running
+db-postgres-running:
```
rynowak, line 112: note: add dependency on `db-dependency-docker-running`

build/db.mk
```diff
@@ -0,0 +1,118 @@
...
+		psql \
+		"postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@localhost:5432"
+
+
+.PHONY: db-reset
+db-reset: db-postgres-running ## Reset the local PostgresSQL database
+	@echo "$(ARROW) Resetting local PostgresSQL database"
+	@echo ""
+	@echo "Resetting ucp resources..."
+	@DOCKER_CLI_HINTS=false docker exec \
```
rynowak, line 87: Note for reviewers... Docker has ads now 😢 `DOCKER_CLI_HINTS=false` turns it off.

build/db.mk
```diff
@@ -0,0 +1,118 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------
+
+POSTGRES_USER=postgres
+POSTGRES_PASSWORD=ucprulz!
+POSTGRES_CONTAINER_NAME=radius-postgres
+
```
nithyatsu, line 20: is the password here because this is v1 of postgres addition?

deploy/init-db/init-db.sh
nithyatsu: I understand not wanting to have cross-component database reads.
But today, we have it this way, correct? 
Does this mean going forward, UCP datastore will contain data like planes where are only the ones that are handled by RP will be stored in Applications-RP's instance of db? Also, do scope queries work with postgres identical to etcd?

build/db.mk
rynowak: This is for running a local PostgreSQL container for testing. This will replace where we currently store data in Kubernetes `radius-testing` in dev/test workflows.

deploy/init-db/init-db.sh
rynowak: > But today, we have it this way, correct?

Today it's possible but discouraged. There's nothing stopping us except code review. 

> Does this mean going forward, UCP datastore will contain data like planes and ucp resource-groups where as only the ones that are handled by RP will be stored in Applications-RP's instance of db? 

Correct. 

Keep in mind it's a single database server, with separate usernames and databases within the server. 

> Also, do scope queries work with postgres identical to etcd?

Yes. The tests are shared between all of the datastore implementations.

deploy/init-db/db.sql.txt
```diff
@@ -0,0 +1,62 @@
...
+    --
+    -- We ensure a leading and trailing slash on the components of the resource id for ease of comparison.
+    --
+    -- id -> "/planes/radius/local/resourcegroups/rg1/providers/applications.core/applications/my-app/"
+    -- resource_type -> "/applications.core/applications/"
+    -- root_scope -> "/planes/radius/local/resourcegroups/rg1/"
+    -- routing_scope -> "/applications.core/applications/my-app/"
+
+    -- resource id used as key.
+    id TEXT PRIMARY KEY NOT NULL,
```
lakshmimsft, line 19: Can we discuss using the resource_id as a primary key in PostGres versus say a UUID while creating a separate unique index for the resource id field??
Concerns are length of id and string type for primary key where size will increase and performance will be degraded over time

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,414 @@
...
+	Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error)
```
lakshmimsft, line 37: We were moving away from `interface{}` and using `any`. Can you clarify what the reason is for keeping interface{}

run-postgres.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo 'Starting database...'
+
+docker run \
+  --name ucp-postgres \
+  --env POSTGRES_PASSWORD=ucp-rulez \
+  --publish 5432:5432 \
+  --init \
+  --tty \
+  postgres
```
lakshmimsft, line 14: should this be: "ghcr.io/radius-project/mirror/postgres:latest"

deploy/init-db/db.sql.txt
rynowak: Sounds like a good topic. 

I started with the resource id as the PK because it supports all of our access patterns. Some of the prototyping that I did actually added a UUID, which would be suitable as a PK.

The UUID is helpful because it allows you to tell the difference between different instances of a resource with the same name (over time). [Kubernetes does this](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).  

Take a look a the Go APIs of the datastore - they map 1:1 with the operations exposed by the RP.

- Resource List (some scope) -> Query resources by type + scope
- Resource Get (some id) -> Get resource by id
- Resource Put (some id) -> Save resource by id
- Resource Delete (some id) -> Delete resource by id

If we used in a synthetic key like a UUID we'd have to reckon with how we map from resource id -> UUID. We'd have a UUID per-resource but it would almost never be used for SQL statements. We'd still need an index on the id and would be primarily leveraging that.

Any ideas?

rynowak: Keep in mind we're not going to be using RDBMS features like joins. We're treating PostgreSQL like a document store.

pkg/ucp/store/postgres/postgresclient.go
rynowak: Copilot did this 👍 will fix.

deploy/init-db/db.sql.txt
superbeeny: Do we have an idea on what the maximum table size will be?

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,414 @@
...
+		var err error
+		query.ResourceType, err = storeutil.NormalizeResourceType(query.ResourceType)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// NOTE: building SQL by concatenating strings is hard to do safely and should be avoided.
+	// If you need to work on this code MAKE SURE you use SQL parameters
+	// for any user input.
```
superbeeny, line 220: Agreed! 👍

superbeeny (commented):
Looks good and gives me a better idea of your coding styles and standards.

deploy/init-db/db.sql.txt
rynowak: I'd imagine a 100k or so would be a **really** **really** large amount of data for a single deployment of Radius. You can think of one row = one resource in Radius. 

Postgres is one of a few options we'll support. You can imagine another use case where the user chooses CosmosDB or DynamoDB or something else that has sharding built-in natively. There are good natural ways to use resource ids for sharding.

---

Totally up for debate .. some additional context from me.

- I wanted to avoid doing DDL dynamically. This would be required if we sharded on resource type.
- I think Postgres is a good **default** choice for Radius. Users that are really invested might choose something that can scale better.

lakshmimsft: The table size being limited, resource id as a primary key is a reasonable approach.  
Question on 'Adding in a UUID in the future', if resource id is a primary key, we will always have only 1 value of UUID in the table at any given time. Unless we further build out and store history/audit of data updates, I am not seeing where/how it will be useful. Is that a possible future usecase ?

rynowak: The reason I linked the Kuberentes `uid` feature is to give you can idea of the usecase. 

It allows the control-plane to know the difference between difference instances of the resource that are created at different times. Example:

- A user creates resource `A`, it gets uuid: `12E67073-1BE4-4935-80AC-6C7D02BE5D8A`
- A user deletes resource `A`
- A user creates resource `A`, it gets uuid: `6FE3910B-3185-43EC-82C3-70902FEF3421`

It allows for a form of opportunistic concurrency control that's based on resource identity rather than resource state.

pkg/ucp/store/postgres/postgresclient.go
rynowak: FYI for reviewers. `Query.Filters` is executed on the client side rather than pushed to the server. I did this for compatibility with the existing code, because the testing done for `Query.Filters` is minimal. 

This comment is relevant because we **will** need to build SQL dynamically to do that 😢.

run-postgres.sh
rynowak: Added this to the makefile since this script is being deleted.

deploy/init-db/db.sql.txt
rynowak: So it sounds like we've largely bottomed-out on this topic... with the consensus that we'll use the resource id as the PK. 

Does that sound right? Anything else we should consider?

rynowak:
Updated and addressed feedback so far.

pkg/ucp/dataprovider/options.go
```diff
@@ -37,6 +37,8 @@ type StorageProviderOptions struct {
 
 	// InMemory configures options for the in-memory store. Will be ignored if another store is configured.
 	InMemory InMemoryOptions `yaml:"inmemory,omitempty"`
+
+	PostgreSQL PostgreSQLOptions `yaml:"postgresql,omitempty"`
```
lakshmimsft, line 41: pls add comment~: // PostgreSQLOptions configures options for connecting to a PostgreSQL database. Will be ignored if another store is configured.

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	return nil
+}
+
+// For pagination we need to encode the timestamp in the token.
+//
+// We use ISO8601/RFC3339 format which postgres understands and can be used for comparison.
+// We also add microseconds to the timestamp to ensure uniqueness.
+
```
lakshmimsft, line 404: nit: extra line above function

pkg/ucp/store/postgres/postgresclient_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+var _ PostgresAPI = (*LoggingAPI)(nil)
+
+type LoggingAPI struct {
+	t    *testing.T
+	pool *pgxpool.Pool
+}
+
+// Exec implements PostgresAPI.
+func (l *LoggingAPI) Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error) {
```
lakshmimsft, line 69: we need to update to `any` in test code

pkg/ucp/store/postgres/postgresclient_test.go
```diff
@@ -0,0 +1,87 @@
...
+		t.Logf("Database reset ... %d rows deleted", tag.RowsAffected())
+	}
+
+	// The actual test logic lives in a shared package, we're just doing the setup here.
+	shared.RunTest(t, client, clear)
+}
+
+var _ PostgresAPI = (*LoggingAPI)(nil)
+
+type LoggingAPI struct {
```
lakshmimsft, line 63: does LoggingAPI need to be a public type? currently it is not used outside of this test file.

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+)
+SELECT
+CASE
+	WHEN EXISTS (SELECT 1 FROM updated) THEN 'Success'
+	WHEN EXISTS (SELECT 1 FROM resources WHERE id = $1) THEN 'ErrConcurrency'
+	ELSE 'ErrNotFound'
+END AS result;`
+
+	args := []any{
+		storeutil.NormalizePart(normalized.String()),
```
lakshmimsft, line 359: i was wondering why we needed to normalize again since on ln323 we normalize it. Checked locally and saw that for a valid resource id the effect would be we update it with lowercase. Cool.

rynowak: Do you think it would make sense to change some of the naming of these functions? `NormalizeResourceID` and `NormalizePart` aren't doing the same thing.

pkg/ucp/dataprovider/options.go
rynowak: Adding

pkg/ucp/store/postgres/postgresclient.go
rynowak: Renamed to `ConvertScopeIDToResourceID`.

pkg/ucp/store/postgres/postgresclient_test.go
rynowak: It does not!

rynowak:
@lakshmimsft - updated

build/db.mk
```diff
@@ -0,0 +1,119 @@
...
+.PHONY: db-shell
+db-shell: db-postgres-running ## Open a shell to the local PostgresSQL database
+	@echo "$(ARROW) Connecting to local PostgresSQL database..."
+	@DOCKER_CLI_HINTS=false docker exec \
+		--interactive \
+		--tty \
+		$(POSTGRES_CONTAINER_NAME) \
+		psql \
+		"postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@localhost:5432"
+
```
ytimocin, line 81: nit: double lines

deploy/init-db/db.sql.txt
```diff
@@ -0,0 +1,62 @@
...
+    routing_scope TEXT NOT NULL,
+    
+    -- etag used for optimistic concurrency control.
+    etag TEXT NOT NULL,
+
+    -- timestamp is used to implement cursor-based pagination (see below).
+    created_at TIMESTAMP (6) WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+
+    -- resource_data stores the resource data.
+    resource_data JSONB NOT NULL
```
ytimocin, line 42: We can even do JSON queries in future: https://www.postgresql.org/docs/current/functions-json.html#SQLJSON-QUERY-FUNCTIONS.

Like, give me all the resources that belongs to this application/environment.

deploy/init-db/db.sql.txt
```diff
@@ -0,0 +1,62 @@
...
+-- > "resource_type" = "/applications.core/applications/"
+-- > "root_scope" = "/planes/radius/local/resourcegroups/my-rg"
+-- > "routing_scope" = NULL
+--
+-- 'created_at' is used with ORDER BY to sort the output, so we can implement cursor-based pagination.
+--
+-- 1) For the initial query, we won't specify a cursor value.
+-- 2) For the next query, we will specify the cursor value as the last created_at value from the previous
+--    query, which allows us to skip the records that were already returned.
+CREATE INDEX idx_resource_query ON resources (resource_type, root_scope, routing_scope, created_at);
```
ytimocin, line 62: Should we create separate indexes based on the query patterns? Based on what you said above, I believe having two separate indexes might be better:
- `resource_type, and root_scope` => INDEX (resource_type, root_scope, created_at)
- `resource_type, root_scope, and routing_scope` => (resource_type, root_scope, routing_scope, created_at)

What do you think?

deploy/init-db/init-db.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo "Creating database and user for $RESOURCE_PROVIDER"
+    psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+        CREATE USER $RESOURCE_PROVIDER WITH PASSWORD '$POSTGRES_PASSWORD';
+        CREATE DATABASE $RESOURCE_PROVIDER;
+        GRANT ALL PRIVILEGES ON DATABASE $RESOURCE_PROVIDER TO $RESOURCE_PROVIDER;
+EOSQL
+done
+
+# Create tables within those databases
+for RESOURCE_PROVIDER in "${RESOURCE_PROVIDERS[@]}"; do
```
ytimocin, line 20: What if we have resource provider as a column in the **resources** table? Would any of the resources created by these two separate resource providers have the same resource id by any chance? This would make adding a custom resource provider easier in my opinion.

1. **resource_providers** table.
2. **resources** table with an additional column of **resource_provider**.
3. Adding a new **resource provider** would just be an addition to the **resource_providers** table.

Otherwise, are we going to create a new database for each resource provider?

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+END AS result;`
+
+	args := []any{storeutil.NormalizePart(converted.String())}
+
+	if config.ETag != "" {
+		// NOTE: we want to report ErrConcurrency for all failure cases here. This is what the tests do.
+		sql = `
+WITH deleted AS (
+	DELETE FROM resources
+	WHERE id = $1 AND etag = $2
```
ytimocin, line 111: We can make the **sql** variable an **fmt.Sprintf** and just update the WHERE part of the query instead of writing two almost identical queries.

ytimocin: I see that they also return two different results. Ignore my comment above.

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+
+	args := []any{
+		// If ScopeRecursive is false, the RootScope must match exactly.
+		// If ScopeRecursive is true, the RootScope must be a prefix of the stored RootScope.
+		storeutil.NormalizePart(query.RootScope),
+		query.ScopeRecursive,
+		resourceType,
+		routingScopePrefixFilter, // RoutingScopePrefix is optional and always treated as as prefix.
+		timestampFilter,          // Optional for pagination.
+		limitFilter,              // NOTE: Postgres allows LIMIT to be set with a NULL value to mean no limit.
```
ytimocin, line 247: Should the search be case insensitive?

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result.Items) < config.MaxQueryItemCount && config.MaxQueryItemCount > 0 {
+		// No more rows, so no need for pagination.
+		return &result, nil
+	}
+
+	if timestamp != nil {
```
ytimocin, line 291: UUID can also be good for internal operations. Like instead of using the timestamp for the pagination, we can use the UUID of the last row.

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+	//
+	// The key behavior difference is that if an etag is provided, we should not perform inserts, only updates.
+
+	// This is the more complex query that handles "upserts". It does not process etags.
+	sql := `
+WITH updated AS (
+	INSERT INTO resources (id, original_id, resource_type, root_scope, routing_scope, etag, resource_data)
+	VALUES ($1, $2, $3, $4, $5, $6, $7)
+	ON CONFLICT (id) 
+	DO UPDATE SET resource_data = $7
```
ytimocin, line 348: Just thinking out loud if we could benefit from having **updated_at** 🧐

pkg/ucp/store/postgres/postgresclient.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	return nil
+}
+
+// createPaginationToken converts a timestamp to a base64 encoded string.
+//
+// We use ISO8601/RFC3339 format which postgres understands and can be used for comparison.
+// We also add microseconds to the timestamp to ensure uniqueness.
+func (p *PostgresClient) createPaginationToken(timestamp time.Time) (string, error) {
```
ytimocin, line 404: Now I see that we don't need UUID for pagination.

deploy/init-db/db.sql.txt
rynowak: Yes, that's the idea. If you look at the file `postgresclient.go` you'll see a comment to that effect.

deploy/init-db/db.sql.txt
rynowak: Lemme run this through an explain plan and I'll share the results. There's a cost to extra indexes because they have to updated on writes. This should be easy to figure out with some data.

deploy/init-db/init-db.sh
rynowak: > Would any of the resources created by these two separate resource providers have the same resource id by any chance?

No this cannot happen. 

I don't see the advantage of an extra column. We actually want an RBAC boundary so that RP's can't reach each-other's data. 

> Otherwise, are we going to create a new database for each resource provider?

Yes. Right now we have 3, soon we'll have 4. I think that's reasonable.

pkg/ucp/store/postgres/postgresclient.go
rynowak: I'm really averse to dynamically generating SQL. So not doing it until it's an absolute requirement.

pkg/ucp/store/postgres/postgresclient.go
rynowak: Search of what? Resource id are case-insensitive. That's how they work. The code has numerous places where the behavior is explained. Please let me know if you have suggestions about how this could be improved.

pkg/ucp/store/postgres/postgresclient.go
rynowak: Is this just proposing an alternative?

pkg/ucp/store/postgres/postgresclient.go
rynowak: We can add that when we have a requirement for it.

pkg/ucp/store/postgres/postgresclient.go
rynowak: 👍 The timestamp-cursor based approach is pretty well understood. I'd like to stick with that unless there's a good reason to change.

pkg/ucp/store/storeutil/id.go
```diff
@@ -164,14 +164,14 @@ func NormalizeResourceID(parsed resources.ID) (resources.ID, error) {
 	return resources.ID{}, fmt.Errorf("invalid resource id: %s", parsed.String())
 }
 
-// NormalizeResourceType normalizes the resource type to be consistent between scopes and resources.
+// ConvertScopeTypeToResourceType normalizes the resource type to be consistent between scopes and resources.
 // See comments on NormalizeResourceID for full context.
```
lakshmimsft, line 168: upd: // See comments on ConvertScopeIDToResourceID for full context.

pkg/ucp/store/postgres/postgresclient_test.go
lakshmimsft: interface{} ->any still needs to be updated

deploy/init-db/init-db.sh
ytimocin: So, if a Radius user creates a new resource provider, will that create a new DB schema?

ytimocin: I am talking about the user defined resource providers and resource types.

ytimocin (dismissed):

deploy/init-db/db.sql.txt
rynowak: @lakshmimsft and I did an experiment here 😁 

 I wanted to make sure that both forms of query (with routing scope and without) can hit the index. If they can't then that's a good reason to add a second index like you proposed.

First, I generated 20k rows of synthetic data - 10k for a top-level resource and 10k for a child resource.

## Generating 10k top-level resources

```sql
INSERT INTO resources (id, original_id, resource_type, root_scope, routing_scope, etag, resource_data)
SELECT 
  '/planes/radius/local/resourcegroups/default/providers/applications.core/containers/c' || numbers.id || '/' AS id,
  '/planes/radius/local/resourceGroups/default/providers/Applications.Core/containers/c' || numbers.id AS original_id,
  '/applications.core/containers/' as resource_type,
  '/planes/radius/local/resourcegroups/default/' as root_scope,
  '/applications.core/containers/c' || numbers.id || '/' as routing_scope,
  'etag' || numbers.id  as etag,
  '{}' as resource_data
FROM generate_series(1, 10000) AS numbers (id);
```


## Generating 10k child resources

```sql
INSERT INTO resources (id, original_id, resource_type, root_scope, routing_scope, etag, resource_data)
SELECT 
  '/planes/radius/local/resourcegroups/default/providers/applications.core/containers/c/' || numbers.id || '/identities/a/' AS id,
  '/planes/radius/local/resourceGroups/default/providers/Applications.Core/containers/c' || numbers.id || '/identities/a' AS original_id,
  '/applications.core/containers/identities/' as resource_type,
  '/planes/radius/local/resourcegroups/default/' as root_scope,
  '/applications.core/containers/c' || numbers.id || '/identities/a/' as routing_scope,
  'etag' || numbers.id  as etag,
  '{}' as resource_data
FROM generate_series(1, 10000) AS numbers (id);
```

Next, I took our client code and translated it into some candidate queries. This is what we're going to explain. I tested each query to make sure it was returning results, before running the query through explain.

These queries correspond with the resource provider "list" operations.

## Query top-level resources

```sql
SELECT original_id, etag, resource_data, created_at 
FROM resources
WHERE ((root_scope = '/planes/radius/local/resourcegroups/default/') OR (FALSE AND (root_scope LIKE '/planes/radius/local/resourcegroups/default/' || '%'))) AND 
	resource_type = '/applications.core/containers/' AND 
	((routing_scope LIKE NULL || '%') OR NULL IS NULL) AND 
	(created_at > NULL::TIMESTAMP OR NULL IS NULL)
ORDER BY created_at ASC
LIMIT 10;

```

## Query child resources


```sql

SELECT original_id, etag, resource_data, created_at 
FROM resources
WHERE ((root_scope = '/planes/radius/local/resourcegroups/default/') OR (FALSE AND (root_scope LIKE '/planes/radius/local/resourcegroups/default/' || '%'))) AND 
	resource_type = '/applications.core/containers/identities/' AND 
	((routing_scope LIKE '/applications.core/containers/c10/identities/a/' || '%') OR '/applications.core/containers/c10/identities/a/' IS NULL) AND 
	(created_at > NULL::TIMESTAMP OR NULL IS NULL)
ORDER BY created_at ASC
LIMIT 10;
```

## Findings

First of all, 20k rows wasn't enough data. At 20k rows PostreSQL did table scans and ignored the indices. Next we tried going to 2m rows.

Secondly, our current index has fields that are used in `ORDER BY` or with the `LIKE` operator. We get better results by taking out `routing_scope` and `created_at`.

```sql
CREATE INDEX idx_resource_query ON resources (resource_type, root_scope, routing_scope, created_at);
```

The explain-plan for a top-level resource query:

```txt
 Limit  (cost=153798.23..153799.40 rows=10 width=113)
   ->  Gather Merge  (cost=153798.23..250907.38 rows=832306 width=113)
         Workers Planned: 2
         ->  Sort  (cost=152798.21..153838.59 rows=416153 width=113)
               Sort Key: created_at
               ->  Parallel Seq Scan on resources  (cost=0.00..143805.29 rows=416153 width=113)
                     Filter: ((root_scope = '/planes/radius/local/resourcegroups/default/'::text) AND (resource_type = '/applications.core/containers/'::text))
```

Notice that the index is not used. This is the primary operation we should be concerned about having good performance.

----

The explain-plan for a child resource query:

```txt
 Limit  (cost=6.46..6.46 rows=1 width=113)
   ->  Sort  (cost=6.46..6.46 rows=1 width=113)
         Sort Key: created_at
         ->  Index Scan using idx_resource_query on resources  (cost=0.68..6.45 rows=1 width=113)
               Index Cond: ((resource_type = '/applications.core/containers/identities/'::text) AND (root_scope = '/planes/radius/local/resourcegroups/default/'::text))
               Filter: (routing_scope ~~ '/applications.core/containers/c10/identities/a/%'::text)
```

The operation is much less common and will need to handle much less data. The index is used, but `created_at` is sorted outside the index. `routing_scope` is used from the index, but it's a filter rather than a lookup due to the like.

## Better option

Here's what I decided to go with:

```sql
CREATE INDEX idx_resource_query ON resources (resource_type, root_scope);
```

pkg/ucp/store/storeutil/id.go
rynowak: Fixed, thanks

pkg/ucp/store/postgres/postgresclient_test.go
rynowak: Fixed

rynowak:
@ytimocin @lakshmimsft updated

lakshmimsft (approved):

build/db.mk
```diff
@@ -0,0 +1,117 @@
...
+		docker logs $(POSTGRES_CONTAINER_NAME); \
+		echo ""; \
+		echo "Restarting PostgresSQL container $(POSTGRES_CONTAINER_NAME)"  \
+		docker start $(POSTGRES_CONTAINER_NAME) 1> /dev/null; \
+	else \
+		docker run \
+			--detach \
+			--name $(POSTGRES_CONTAINER_NAME) \
+			--publish 5432:5432 \
+			--env POSTGRES_USER=$(POSTGRES_USER) \
```
nithyatsu, line 42: could this command fail? what if 5432 on host is not available?

rynowak: Yes. It if fails it fails :) Do you think we should try to do something about it?

PR #7963, opened 2024-09-25
Add shellcheck extension to dev container

brooke-hamilton:
# Description

Adds the shell check extension to the dev container. See https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck.

## Type of change

Fixes: N/A

@rynowak FYI


rynowak (dismissed):

brooke-hamilton:
> Anything that can catch me being forgetful before the push is awesome, along the same lines how do folk feel about pre-commit and post-receive hooks?

@superbeeny thank you for the comment! 🎸  I created issue #8022 to track this. In the issue I described adding [shellcheck](https://github.com/koalaman/shellcheck) to the PR validation checks, and I'm thinking that using the VS Code extension + the PR checks is enough, but if you think we should also configure pre- and post-commit checks in git lets continue the discussion.

superbeeny:
> > Anything that can catch me being forgetful before the push is awesome, along the same lines how do folk feel about pre-commit and post-receive hooks?
> 
> I'm a fan of pre-commit hooks, but also understand that they optional and a lot of folk will ignore them 😆.
> 
> I'm not sure I've seen a post-receive hook in the wild. What do you use them for?

Let them ignore them at their peril! I'd like a hook to run `make generate` before committing 🤣 I haven't used post-receive in a long time, but would use them for notifications and any other server side nonsense

rynowak:
> Let them ignore them at their peril! I'd like a hook to run make generate before committing 🤣 I haven't used post-receive in a long time, but would use them for notifications and any other server side nonsense

If you want to add something, send us a PR 😆 As long as it's optional, that seems like a good experiment to try.

ytimocin (approved):

PR #7973, opened 2024-09-30
Use `ubuntu-latest` instead of a hosted 4-core runner

jeefy:
# Description

Heya! Friendly neighborhood CNCF staffer here. The ubuntu-latest runner is already a 4-core runner, so using a hosted 4-core runner is unnecessary (and actually costs money). This PR just swaps any 4-core hosted runners to use ubuntu-latest. There should be zero change or impact, functionally this is a SKU change.

Please reach out to me in the CNCF/K8s slack(s) (jeefy) or via email (jsica@linuxfoundation.org) with any questions. 

## Type of change
Replaces `ubuntu-latest-m` (which is a pay-for 4-core hosted runner) to `ubuntu-latest` (which is the free 4-core runner with no real spec difference)


lakshmimsft:
Thank you @jeefy !
But it looks like we hit a error running out of disk space with this update consistently. (Following are errors in different runs:)
https://github.com/radius-project/radius/actions/runs/11110614862/job/30941040772?pr=7973
image

https://github.com/radius-project/radius/actions/runs/11110614862/job/30941845181?pr=7973
image

We may have to stay with `ubuntu-latest-m` for now until we can explore possibilities like cleaning up after steps or other optimizations.
cc @rynowak

rynowak:
@jeefy - any ideas? We're using `kind` in the test pipelines to stand up Kubernetes, and then running tests on that cluster.

ytimocin:
`ubuntu-latest-m` has 16 GB RAM. Not sure if `ubuntu-latest` also has that much.

That could be the reason why one of our functional tests is failing.

jeefy:
They have the same amount of RAM, but the disk space is different (free runner: 14GB, non-free... 140GB)

I'm going to update the PR to use one of our hosted runners since it has more disk.

jeefy:
Don't know why the job isn't getting picked up, and I watched other jobs outside of this repo during this. I'll have to investigate further. Thanks!

rynowak:
> Don't know why the job isn't getting picked up, and I watched other jobs outside of this repo during this. I'll have to investigate further. Thanks!

Any ideas?

lakshmimsft:
We are continuing with `ubuntu-latest-m` because with the suggested alternative, `ubuntu-latest`, we encounter disk space issues when running tests. This issue can be revisited once we have confirmed a viable alternative.

PR #8001, opened 2024-10-17
Add DeploymentTemplate controller for GitOps support

willdavsmith:
# Description

* Add `rad bicep generate-kubernetes-manifest` CLI
* Add DeploymentTemplate controller
* Add DeploymentResource controller

## Type of change

https://github.com/radius-project/radius/issues/6689

## Contributor checklist

- [x] An overview of proposed schema changes is included in a linked GitHub issue.
- [x] A design document PR is created in the [design-notes repository](https://github.com/radius-project/design-notes/), if new APIs are being introduced.
- [x] If applicable, design document has been reviewed and approved by Radius maintainers/approvers.
- [x] A PR for the [samples repository](https://github.com/radius-project/samples) is created, if existing samples are affected by the changes in this PR.
- [x] A PR for the [documentation repository](https://github.com/radius-project/docs) is created, if the changes in this PR affect the documentation or any user facing updates are made.
- [x] A PR for the [recipes repository](https://github.com/radius-project/recipes) is created, if existing recipes are affected by the changes in this PR.


deploy/Chart/templates/controller/deployment.yaml
```diff
@@ -72,7 +72,23 @@ spec:
           mountPath: {{ .Values.global.rootCA.mountPath }}
           readOnly: true
         {{- end }}
+        - name: work-dir
```
sk593, line 75: opened a PR here for the bicep container work: https://github.com/radius-project/radius/pull/8002

pkg/cli/bicep/deployment_parameters.go
```diff
@@ -19,32 +19,23 @@ package bicep
...
 )
 
 // ParameterParser is used to parse the parameters as part of the `rad deploy` command. See the docs for `rad deploy` for examples
 // of what we need to support here.
 type ParameterParser struct {
-	FileSystem fs.FS
-}
-
-type OSFileSystem struct {
+	FileSystem afero.Fs
```
willdavsmith, line 32: instead of writing a filesystem abstraction I added this `afero` library. Seems to be widely used, I replaced the `OSFileSystem` with it. LMK if I should revert

pkg/cli/bicep/deployment_parameters_test.go
```diff
@@ -56,15 +57,24 @@ func Test_ParseParameters_Overwrite(t *testing.T) {
 		"key3=value3",
 	}
 
+	// Initialize the in-memory filesystem
+	fs := afero.NewMemMapFs()
```
willdavsmith, line 61: This is just re-writing the test with afero instead of fs

bicep-types
sk593, line 1: is this out of date with main or were there changes here?

pkg/cli/bicep/deployment_parameters.go
sk593: are there added functionalities that we get using `afero` or is the main benefit that's it's an abstraction layer over `os`?

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,405 @@
...
+	- A file containing multiple parameters using the ARM JSON parameter format (see below)
+	- A file containing a single value in JSON format
+	- A key-value-pair passed in the command line
+	
+	When passing multiple parameters in a single file, use the format described here:
+	
+		https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files
+	
+	You can specify parameters using multiple sources. Parameters can be overridden based on the 
+	order the are provided. Parameters appearing later in the argument list will override those defined earlier.
```
sk593, line 67:
```suggestion
	order they are provided. Parameters appearing later in the argument list will override those defined earlier.
```

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,405 @@
...
+		`,
+		Example: `
+# Generate a DeploymentTemplate Custom Resource from a Bicep file.
+rad bicep generate-kubernetes-manifest app.bicep --parameters @app.bicepparam --parameters tag=latest --outfile app.yaml
```
sk593, line 71: why is it `@app.bicepparam`

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,405 @@
...
+		Bicep:             factory.GetBicep(),
+		ConnectionFactory: factory.GetConnectionFactory(),
+		ConfigHolder:      factory.GetConfigHolder(),
+		Deploy:            factory.GetDeploy(),
+		Output:            factory.GetOutput(),
+	}
+}
+
+// Validate validates the inputs of the rad bicep generate-kubernetes-manifest command.
+func (r *Runner) Validate(cmd *cobra.Command, args []string) error {
```
sk593, line 117: A lot of the code in this file seems to be similar to the logic in `deploy.go`. Does it make sense to have shared functions between these two files since they're doing similar work?

pkg/controller/api/radapp.io/v1alpha3/deploymentresource_types.go
```diff
@@ -0,0 +1,101 @@
...
+	DeploymentResourcePhraseFailed DeploymentResourcePhrase = "Failed"
+
+	// DeploymentResourcePhraseDeleting indicates that the Deployment Resource is being deleted.
+	DeploymentResourcePhraseDeleting DeploymentResourcePhrase = "Deleting"
+
+	// DeploymentResourcePhraseDeleted indicates that the Deployment Resource has been deleted.
+	DeploymentResourcePhraseDeleted DeploymentResourcePhrase = "Deleted"
+)
+
+// +kubebuilder:object:root=true
```
sk593, line 76: is this meant to be here?

brooke-hamilton (changes requested):
I ran `make generate build test` and got 8 failures.

```bash
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run/Create_basic_DeploymentTemplate (0.00s)
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run/Create_DeploymentTemplate_with_template_content (0.00s)
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run/Create_DeploymentTemplate_with_Azure_scope (0.00s)
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run/Create_DeploymentTemplate_with_AWS_scope (0.00s)
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run (0.00s)
=== FAIL: pkg/controller/reconciler Test_DeploymentResourceReconciler_Basic (10.01s)
=== FAIL: pkg/controller/reconciler Test_DeploymentTemplateReconciler_Basic (0.01s)
=== FAIL: pkg/controller/reconciler Test_DeploymentTemplateReconciler_FailureRecovery (0.00s)
```

Here's one example:
```
=== FAIL: pkg/cli/cmd/bicep/generatekubernetesmanifest Test_Run/Create_DeploymentTemplate_with_Azure_scope (0.00s)
Creating DeploymentTemplate YAML file
    generatekubernetesmanifest_test.go:396: 
                Error Trace:    /workspaces/radius/pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go:396
                Error:          Not equal: 
                                expected: "apiVersion: radapp.io/v1alpha3\nkind: DeploymentTemplate\nmetadata:\n  name: azure.yaml\n  namespace: radius-system\nspec:\n  parameters: '{}'\n  providerConfig: '{\"radius\":{\"type\":\"radius\",\"value\":{\"scope\":\"/planes/radius/local/resourceGroups/test-resource-group\"}},\"az\":{\"type\":\"azure\",\"value\":{\"scope\":\"/subscriptions/test-subId/resourceGroups/test-rg\"}},\"deployments\":{\"type\":\"Microsoft.Resources\",\"value\":{\"scope\":\"/planes/radius/local/resourceGroups/test-resource-group\"}}}'\n  template: '{}'\n"
                                actual  : "apiVersion: radapp.io/v1alpha3\nkind: DeploymentTemplate\nmetadata:\n  name: azure.yaml\n  namespace: radius-system\nspec:\n  parameters: '{}'\n  providerConfig: '{\"radius\":{\"type\":\"radius\",\"value\":{\"scope\":\"/planes/radius/local/resourceGroups/test-resource-group\"}},\"az\":{\"type\":\"azure\",\"value\":{\"scope\":\"/subscriptions/test-subId/resourceGroups/test-rg\"}},\"deployments\":{\"type\":\"Microsoft.Resources\",\"value\":{\"scope\":\"/planes/radius/local/resourceGroups/test-resource-group\"}}}'\n  repository: azure.yaml\n  template: '{}'\n"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -8,2 +8,3 @@
                                   providerConfig: '{"radius":{"type":"radius","value":{"scope":"/planes/radius/local/resourceGroups/test-resource-group"}},"az":{"type":"azure","value":{"scope":"/subscriptions/test-subId/resourceGroups/test-rg"}},"deployments":{"type":"Microsoft.Resources","value":{"scope":"/planes/radius/local/resourceGroups/test-resource-group"}}}'
                                +  repository: azure.yaml
                                   template: '{}'
                Test:           Test_Run/Create_DeploymentTemplate_with_Azure_scope
    --- FAIL: Test_Run/Create_DeploymentTemplate_with_Azure_scope (0.00s)
```

brooke-hamilton (dismissed):
Running `make generate` results in these files being changed.

```
deploy/Chart/crds/radius/radapp.io_deploymentresources.yaml
deploy/Chart/crds/radius/radapp.io_deploymenttemplates.yaml
deploy/Chart/crds/radius/radapp.io_recipes.yaml
deploy/Chart/crds/ucpd/ucp.dev_queuemessages.yaml
deploy/Chart/crds/ucpd/ucp.dev_resources.yaml
```

For example, ` controller-gen.kubebuilder.io/version: v0.16.4` changes to `controller-gen.kubebuilder.io/version: v0.16.0`

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: I changed the logic here, there's less duplication now

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: This is the ARMJSON parameter format: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files#deploy-template-with-parameter-file.

Same behavior as `rad deploy`

pkg/cli/bicep/deployment_parameters.go
willdavsmith: Good find! I added it here because the stdlib `fs` does not include a filesystem abstraction with `CreateFile`. The benefit of adding `afero` was to not have to implement and maintain this ourselves. Since we already indirectly depend on this library, and that it's a pretty widespread dependency, I'd say we should be okay to use it. Thoughts? @brooke-hamilton

pkg/controller/api/radapp.io/v1alpha3/deploymentresource_types.go
willdavsmith: I believe so. `DeploymentResource` has its own controller.

willdavsmith:
> Running `make generate` results in these files being changed.
> 
> ```
> deploy/Chart/crds/radius/radapp.io_deploymentresources.yaml
> deploy/Chart/crds/radius/radapp.io_deploymenttemplates.yaml
> deploy/Chart/crds/radius/radapp.io_recipes.yaml
> deploy/Chart/crds/ucpd/ucp.dev_queuemessages.yaml
> deploy/Chart/crds/ucpd/ucp.dev_resources.yaml
> ```
> 
> For example, ` controller-gen.kubebuilder.io/version: v0.16.4` changes to `controller-gen.kubebuilder.io/version: v0.16.0`

This depends on your local `controller-gen` CLI version. I downgraded my own to match `v0.16.0` so we shouldn't see this anymore

pkg/controller/reconciler/deploymenttemplate_reconciler_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+	go func() {
+		err := mgr.Start(ctx)
+		require.NoError(t, err)
+	}()
+
+	return radius, mgr.GetClient()
+}
+
+func Test_DeploymentTemplateReconciler_Basic(t *testing.T) {
```
willdavsmith, line 89: these tests are a bit lacking at the moment. I'm improving them now

pkg/cli/bicep/deployment_parameters.go
willdavsmith: @rynowak thoughts on this?

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	_, err = f.Write(deploymentTemplateYaml)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// generateProviderConfig generates a ProviderConfig object based on the given scopes.
+func (r *Runner) generateProviderConfig() (providerConfig sdkclients.ProviderConfig) {
```
sk593, line 251: Are the type values correct here? Should they not be "AzureResourceManager", "AWS", and "Radius"?

pkg/cli/cmd/bicep/generatekubernetesmanifest/testdata/aws/aws.yaml
```diff
@@ -0,0 +1,74 @@
...
+kind: DeploymentTemplate
+metadata:
+  name: aws.bicep
+  namespace: radius-system
+spec:
+  parameters: {}
+  providerConfig: |-
+    {
+      "radius": {
+        "type": "radius",
```
sk593, line 11: same question as above for the provider types. are these the correct type values?

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
+	controller_runtime "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Test_DeploymentTemplate(t *testing.T) {
```
sk593, line 48: Would it make sense to add a cloud test case? Thinking about the providerConfig options and if we'd want coverage there. Not strictly needed, but just a thought

deploy/Chart/templates/controller/rbac.yaml
```diff
@@ -52,6 +56,20 @@ rules:
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - source.toolkit.fluxcd.io
+  resources:
+  - gitrepositories/status
+  verbs:
+  - get
```
rynowak, line 72: Is this part of this PR or something we need for the next one?

pkg/cli/bicep/deployment_parameters.go
rynowak: What's the savings of using this library? I don't have fundamental concerns about it, I just can't tell from the diff what's going to change.

rynowak: I'd also really like to remove viper as a dependency. We've got a lot of workarounds in place for for behaviors that aren't helpful to us.

pkg/cli/bicep/deployment_parameters_test.go
```diff
@@ -56,15 +57,24 @@ func Test_ParseParameters_Overwrite(t *testing.T) {
...
+
+	// Create the "many.json" file with the specified content
+	err := afero.WriteFile(fs, "many.json", []byte(`{ "parameters": { "key1": { "value": { "someValue": true } }, "key2": { "value": "overridden-value" } } }`), 0644)
+	if err != nil {
+		t.Fatalf("Failed to create many.json: %v", err)
```
rynowak, line 66: Use `require`

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+		Args: cobra.ExactArgs(1),
+		RunE: framework.RunCommand(runner),
+	}
+
+	commonflags.AddWorkspaceFlag(cmd)
+	commonflags.AddResourceGroupFlag(cmd)
+	commonflags.AddEnvironmentNameFlag(cmd)
+	commonflags.AddParameterFlag(cmd)
+
+	cmd.Flags().String("outfile", "", "Path of the generated DeploymentTemplate yaml file.")
```
rynowak, line 76: Let's pick a more obvious name for this. We can't use `output` because that's taken by `-o`. We should avoid abbreviations in parameters.

Some ideas:

- `output-file`. 
- `destination-file`. 
- others?

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	r.FilePath = args[0]
+
+	var err error
+	r.DeploymentScope, err = cmd.Flags().GetString("deployment-scope")
+	if err != nil {
+		return err
+	}
+
+	if r.DeploymentScope == "" {
+		r.DeploymentScope = "/planes/radius/local/resourceGroups/default"
```
rynowak, line 125: Can we avoid hardcoding this? We **really** don't want users deploying a multiple applications to a single resource group.

I'd suggest making resource group a required argument.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+		"apiVersion": "radapp.io/v1alpha3",
+		"metadata": map[string]any{
+			"name":      fileName,
+			"namespace": "radius-system",
+		},
+		"spec": map[string]any{
+			"template":       string(marshalledTemplate),
+			"parameters":     params,
+			"providerConfig": string(marshalledProviderConfig),
+			"rootFileName":   fileName,
```
rynowak, line 220: Something I need to dig into, what is `rootFileName` for? This feels like we're taking one piece of data and using it for a different purpose that *feels* related but may not be related.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	for k, v := range parameters {
+		params[k] = v["value"].(string)
+	}
+
+	deploymentTemplate := map[string]any{
+		"kind":       "DeploymentTemplate",
+		"apiVersion": "radapp.io/v1alpha3",
+		"metadata": map[string]any{
+			"name":      fileName,
+			"namespace": "radius-system",
```
rynowak, line 214: Is there a reason why this is hardcoded to `radius-system`? We discourage people for giving developers permission on that namespace.

I think the right thing is to omit this field. It's hard for us to choose the namespace for a user, so make it their choice.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+			"rootFileName":   fileName,
+		},
+	}
+
+	return deploymentTemplate, nil
+}
+
+// createDeploymentTemplateYAMLFile creates a DeploymentTemplate YAML file with the given content.
+func (r *Runner) createDeploymentTemplateYAMLFile(deploymentTemplate map[string]any) error {
+	fmt.Println("Creating DeploymentTemplate YAML file")
```
rynowak, line 229: use the CLI logger or don't output :)

We're historically had a lot of maintainability problems with output that can't be tested.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+
+	"github.com/radius-project/radius/pkg/cli/bicep"
+	"github.com/radius-project/radius/pkg/cli/cmd/commonflags"
+	"github.com/radius-project/radius/pkg/cli/connections"
+	"github.com/radius-project/radius/pkg/cli/deploy"
+	"github.com/radius-project/radius/pkg/cli/framework"
+	"github.com/radius-project/radius/pkg/cli/output"
+	sdkclients "github.com/radius-project/radius/pkg/sdk/clients"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
```
rynowak, line 36: Can we use `"gopkg.in/yaml.v3"` instead?

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
rynowak: This should match what `rad deploy` does

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go
```diff
@@ -0,0 +1,699 @@
...
+                  }
+                }
+              }
+            }
+          },
+          "type": "Applications.Core/environments@2023-10-01-preview"
+        }
+      }
+    }
+		`
```
rynowak, line 229: Can this moved into a file inside `testdata/`? It's kinda ugly to maintain JSON inside Go and there's an easy solution.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go
```diff
@@ -0,0 +1,699 @@
...
+                  },
+                  "type": "Microsoft.Storage/storageAccounts"
+                }
+              }
+            }
+          },
+          "type": "Microsoft.Resources/deployments"
+        }
+      }
+    }
```
rynowak, line 654: Please move all of these inputs to files :)

Some of these tests feel redundant. Can we have less to maintain?

pkg/cli/cmd/bicep/generatekubernetesmanifest/testdata/aws/aws.yaml
rynowak: Comparison with the `rad deploy` code is the right way to answer this question.

pkg/controller/api/radapp.io/v1alpha3/deploymentresource_types.go
```diff
@@ -0,0 +1,104 @@
...
+type DeploymentResourceSpec struct {
+	// Id is the resource Id.
+	Id string `json:"id,omitempty"`
+
+	// ProviderConfig specifies the scope for resources
+	ProviderConfig string `json:"providerConfig,omitempty"`
+
+	// RootFileName is the name of the Bicep file that
+	// `bicep build` is run on.
+	RootFileName string `json:"rootFileName,omitempty"`
```
rynowak, line 33: What do we use this for?

rynowak: I don't remember it showing up in the design doc :)

pkg/controller/api/radapp.io/v1alpha3/deploymentresource_types.go
```diff
@@ -0,0 +1,104 @@
...
+type DeploymentResourceStatus struct {
+	// Id is the resource Id.
+	Id string `json:"id,omitempty"`
+
+	// ProviderConfig specifies the scope for resources
+	ProviderConfig string `json:"providerConfig,omitempty"`
+
+	// RootFileName is the name of the Bicep file that
+	// `bicep build` is run on.
+	RootFileName string `json:"rootFileName,omitempty"`
```
rynowak, line 46: It looks like all of the fields from the spec also appear in the status. What does this do for us?

pkg/controller/api/radapp.io/v1alpha3/deploymenttemplate_types.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Parameters is the ARM JSON parameters for the template.
+	Parameters string `json:"parameters,omitempty"`
+
+	// ProviderConfig specifies the scope for resources
+	ProviderConfig string `json:"providerConfig,omitempty"`
+
+	// RootFileName is the name of the Bicep file that
+	// `bicep build` is run on.
+	RootFileName string `json:"rootFileName,omitempty"`
```
rynowak, line 55: Same question as DeploymentResource, why do we duplicate the fields from spec into the status?

pkg/controller/reconciler/const.go
```diff
@@ -47,4 +47,13 @@ const (
...
 	RecipeFinalizer = "radapp.io/recipe-finalizer"
+
+	// DeploymentTemplateFinalizer is the name of the finalizer added to DeploymentTemplates.
+	DeploymentTemplateFinalizer = "radapp.io/deployment-template-finalizer"
+
+	// DeploymentResourceFinalizer is the name of the finalizer added to DeploymentResources.
+	DeploymentResourceFinalizer = "radapp.io/deployment-resource-finalizer"
+
+	// RadiusSystemNamespace is the name of the system namespace where Radius resources are stored.
+	RadiusSystemNamespace = "radius-system"
```
rynowak, line 58: 👀

pkg/controller/reconciler/util.go
```diff
@@ -280,3 +280,19 @@ func createOrUpdateContainer(ctx context.Context, radius RadiusClient, container
 
 	return nil, nil
 }
+
+func generateDeploymentResourceName(resourceId string) string {
+	resourceBaseName := strings.Split(resourceId, "/")[len(strings.Split(resourceId, "/"))-1]
```
rynowak, line 285: Can we use the resource ID parser for this? It's hard to reason about where this data came from, and whether we expect it to be valid.

For example if I gave you the string `""` this would cause a panic. Any manual string parsing code like this needs a lot of testing, so we should avoid it.

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+		return r.reconcileDelete(ctx, &deploymentResource)
+	}
+
+	logger.Info("Resource is in desired state.", "resourceId", deploymentResource.Spec.Id)
+
+	deploymentResource.Status.Phrase = radappiov1alpha3.DeploymentResourcePhraseReady
+	deploymentResource.Status.ProviderConfig = deploymentResource.Spec.ProviderConfig
+	deploymentResource.Status.RootFileName = deploymentResource.Spec.RootFileName
+	deploymentResource.Status.Id = deploymentResource.Spec.Id
+	err = r.Client.Status().Update(ctx, &deploymentResource)
```
rynowak, line 121: I'd like to understand why we're copying all of the state over here.

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+	logger := ucplog.FromContextOrDiscard(ctx)
+
+	if deploymentResource.Status.Operation.OperationKind == radappiov1alpha3.OperationKindDelete {
+		providerConfig := sdkclients.ProviderConfig{}
+		err := json.Unmarshal([]byte(deploymentResource.Spec.ProviderConfig), &providerConfig)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to unmarshal providerConfig: %w", err)
+		}
+
+		poller, err := r.Radius.Resources(providerConfig.Deployments.Value.Scope, deploymentResourceType).ContinueDeleteOperation(ctx, deploymentResource.Status.Operation.ResumeToken)
```
rynowak, line 141: For this step, it looks like we're resuming the operation, and we need the resource ID of the deployment to do that. I think we're already capturing that in `status.id` so we don't also need the provider config?

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// Since we're going to reconcile, update the observed generation.
+	//
+	// We don't want to do this if we're in the middle of an operation, because we haven't
+	// fully processed any status changes until the async operation completes.
+	deploymentResource.Status.ObservedGeneration = deploymentResource.Generation
+
+	// NOTE: The following is a workaround for Radius API behavior. Since deleting
+	// an application or environment can leave hanging resources, we need to make sure to
+	// delete these resources before deleting the application or environment.
```
rynowak, line 240: Is this relevant?

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// Check other resources that depend on this resource.
+	// List all DeploymentResource objects in the same namespace
+	// that have the same rootFileName.
+	deploymentResourceList := &radappiov1alpha3.DeploymentResourceList{}
+	err := r.Client.List(ctx, deploymentResourceList, client.InNamespace(deploymentResource.Namespace), client.MatchingFields{rootFileNameField: deploymentResource.Spec.RootFileName})
```
rynowak, line 246: Is this how we're associating a DT with its DRs? 

We should be using owner references for this. https://book.kubebuilder.io/reference/watching-resources/secondary-owned-resources

rynowak: Or is this a workaround for ordering constraints?

rynowak: If that's the case we should still user owner reference, but without the automatic cascading delete: https://book.kubebuilder.io/reference/watching-resources/secondary-owned-resources

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	appsCount := 0
+	envsCount := 0
+	otherCount := 0
+	for _, dr := range deploymentResourceList.Items {
+		if dr.Status.Phrase == radappiov1alpha3.DeploymentResourcePhraseDeleted {
+			continue
+		}
+		if strings.Contains(dr.Spec.Id, "Applications.Core/applications") {
```
rynowak, line 258: Please use the resource ID parser and compare the types. Types need to be compared case-insensitively.

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	if strings.Contains(deploymentResource.Spec.Id, "Applications.Core/environments") {
+		if otherCount > 0 {
+			logger.Info("Resource is an environment, being used by another resource.", "resourceId", deploymentResource.Spec.Id)
+			return ctrl.Result{Requeue: true, RequeueAfter: r.requeueDelay()}, nil
+		}
+	}
+
+	poller, err := r.startDeleteOperation(ctx, deploymentResource)
```
rynowak, line 283: It looks like this code us directly manipulating the state of the DR resource, and handling its lifecycle. We should talk about this, I was expecting something different. 

What I had in mind for when a delete happens on a DT:

- DT controller **initiates** the delete of the DR through the K8s API and uses a **watch** to observe completion.
- DR controller is responsible for the deletion of underlying resource **owned** by the DR.

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,374 @@
...
+			return ctrl.Result{Requeue: true, RequeueAfter: r.requeueDelay()}, nil
+		}
+	}
+
+	if strings.Contains(deploymentResource.Spec.Id, "Applications.Core/environments") {
+		if otherCount > 0 {
+			logger.Info("Resource is an environment, being used by another resource.", "resourceId", deploymentResource.Spec.Id)
+			return ctrl.Result{Requeue: true, RequeueAfter: r.requeueDelay()}, nil
+		}
+	}
```
rynowak, line 281: Can you extract this logic into a function so we can unit-test it?

pkg/controller/reconciler/deploymentresource_reconciler_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	deployment := makeDeploymentResource(name, TestDeploymentResourceID)
+	err = client.Create(ctx, deployment)
+	require.NoError(t, err)
+
+	// Deployment will update after operation completes
+	status := waitForDeploymentResourceStateReady(t, client, name)
+	require.Equal(t, TestDeploymentResourceID, status.Id)
+
+	err = client.Delete(ctx, deployment)
```
rynowak, line 113: We definitely need to test some of the novel delete behaviors. I'll be looking to see if that's tested as part of the DT controller.

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,546 @@
...
+			outputResourceList := resp.Properties["outputResources"].([]any)
+			for _, resource := range outputResourceList {
+				outputResource := resource.(map[string]any)
+				outputResources = append(outputResources, outputResource["id"].(string))
+			}
+
+			// Compare outputResources with existing DeploymentResources
+			// if is present in deploymentTemplate.Status.OutputResources but not in outputResources, delete it
+			// if is not present in deploymentTemplate.Status.OutputResources but is in outputResources, create it
+			// if is present in both, do nothing
```
rynowak, line 179: 👍

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,546 @@
...
+						}
+
+						// Create the DeploymentResource
+						err = r.Client.Create(ctx, deploymentResource)
+						if err != nil {
+							return ctrl.Result{}, err
+						}
+					}
+				}
+			}
```
rynowak, line 221: Can you extract this logic into a function we can unit-test?

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,546 @@
...
+						Spec: radappiov1alpha3.DeploymentResourceSpec{
+							Id:             outputResourceId,
+							ProviderConfig: deploymentTemplate.Spec.ProviderConfig,
+							RootFileName:   deploymentTemplate.Spec.RootFileName,
+						},
+					}
+
+					if controllerutil.AddFinalizer(deploymentResource, DeploymentResourceFinalizer) {
+						// Add the DeploymentTemplate as the owner of the DeploymentResource
+						if err := controllerutil.SetControllerReference(deploymentTemplate, deploymentResource, r.Scheme); err != nil {
```
rynowak, line 210: I see you've already got an owner reference here. We should be using that for lookups instead of the `rootFileName`.

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,546 @@
...
+		}
+
+		// If we get here, the operation was a success. Update the status and continue.
+		//
+		// NOTE: we don't need to save the status here, because we're going to continue reconciling.
+		deploymentTemplate.Status.Operation = nil
+		deploymentTemplate.Status.OutputResources = outputResources
+		deploymentTemplate.Status.Template = deploymentTemplate.Spec.Template
+		deploymentTemplate.Status.Parameters = string(stringifiedSpecParameters)
+		deploymentTemplate.Status.Resource = providerConfig.Deployments.Value.Scope + "/providers/" + deploymentResourceType + "/" + deploymentTemplate.Name
```
rynowak, line 260: I'd like to dig into why we need to save these things to status

pkg/controller/reconciler/deploymenttemplate_reconciler_test.go
```diff
@@ -0,0 +1,401 @@
...
+		err := client.Get(ctx, name, current)
+		if apierrors.IsNotFound(err) {
+			return true
+		}
+
+		logger.Logf("DeploymentTemplate.Status: %+v", current.Status)
+		return false
+
+	}, deploymentTemplateTestWaitDuration, deploymentTemplateTestWaitInterval, "DeploymentTemplate still exists")
+}
```
rynowak, line 401: What's your plan for testing the more complex deletion scenarios?

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
rynowak: I'd prefer not if we can avoid it. The controller doesn't call any cloud APIs directly, just the DE. 

If we find bugs that are unique to this interaction we can reconsider.

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := testcontext.New(t)
+			opts := rp.NewRPTestOptions(t)
+
+			name := tc.name
+			namespace := tc.namespace
+
+			template, err := afero.ReadFile(afero.NewOsFs(), tc.templateFilePath)
```
rynowak, line 92: I'm not a huge fan of how viral `afero` is becoming in this PR. It's pretty easy to use `testdata` for inputs in tests, and temp directories for outputs.

rynowak (commented):
Left some feedback. I think there are some opportunities to simplify, but this is on track 👍

pkg/cli/bicep/deployment_parameters.go
willdavsmith: I removed `afero` as a dependency and wrote a simple filesystem abstraction that we can use

deploy/Chart/templates/controller/rbac.yaml
willdavsmith: part of the next one. I removed it for now

pkg/cli/bicep/deployment_parameters_test.go
willdavsmith: fixed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: I agree - I went with `destination-file`

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: agreed - I added resource group as a required argument

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: fixed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: I omitted this field

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go
willdavsmith: fixed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go
willdavsmith: fixed

pkg/controller/reconciler/const.go
willdavsmith: removed

pkg/controller/reconciler/util.go
willdavsmith: changed this to use the resourceid parser

pkg/controller/reconciler/deploymentresource_reconciler.go
willdavsmith: fixed

pkg/controller/reconciler/deploymenttemplate_reconciler.go
willdavsmith: fixed

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,546 @@
...
+func (r *DeploymentTemplateReconciler) requeueDelay() time.Duration {
+	delay := r.DelayInterval
+	if delay == 0 {
+		delay = PollingDelay
+	}
+
+	return delay
+}
+
+func ParseDeploymentScopeFromProviderConfig(providerConfig any) (string, error) {
```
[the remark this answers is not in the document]
willdavsmith: yes, I use it in the functional tests

pkg/controller/reconciler/deploymenttemplate_reconciler_test.go
willdavsmith: I added a couple of new functional tests, one that deploys a recipe with kubernetes resources via recipe. do you think we should expand these tests? any ideas for what to add?

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
willdavsmith: I removed `afero` and added a small filesystem abstraction that we can use instead.

pkg/cli/filesystem/osfs.go
```diff
@@ -0,0 +1,41 @@
...
+// It uses the methods from the os package to interact with the filesystem.
+type OSFileSystem struct{}
+
+var _ FileSystem = (*OSFileSystem)(nil)
+
+func NewOSFS() *OSFileSystem {
+	return &OSFileSystem{}
+}
+
+func (osfs OSFileSystem) Create(name string) (fs.File, error) {
```
sk593, line 18: Should we have unit testing for this abstraction? There is testing for`MemMapFS`

pkg/cli/filesystem/filesystem.go
```diff
@@ -0,0 +1,15 @@
+package filesystem
+
+import (
+	"io/fs"
+)
+
+// FileSystem is an interface that defines the methods needed to interact with a file system.
```
sk593, line 7: Could you explain the use cases for OSFS and MemMapFS? It seems like MemMapFS is only used for testing while OSFS is used when generating the manifest files

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -0,0 +1,358 @@
...
+			Parameters:     parameters,
+			ProviderConfig: providerConfig,
+		},
+	}
+
+	return deploymentTemplate
+}
+
+func waitForDeploymentTemplateReady(t *testing.T, ctx context.Context, name types.NamespacedName, client controller_runtime.WithWatch, initialVersion string) (*radappiov1alpha3.DeploymentTemplate, error) {
+	// Based on https://gist.github.com/PrasadG193/52faed6499d2ec739f9630b9d044ffdc
```
sk593, line 272: Can you add comments for what this code does

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
sk593: Might be missing something but it looks like deploy uses different values? 
https://github.com/radius-project/radius/blob/876ea07671b18506ecf316d73da9f7f5d4255287/pkg/cli/deployment/deploy.go#L138
https://github.com/radius-project/radius/blob/876ea07671b18506ecf316d73da9f7f5d4255287/pkg/sdk/clients/providerconfig.go#L21

Could you clarify why these don't match @willdavsmith?

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -0,0 +1,358 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
+	controller_runtime "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Test_DeploymentTemplate_Basic(t *testing.T) {
```
sk593, line 52:
```suggestion
func Test_DeploymentTemplate_Env(t *testing.T) {
```

so it lines up with the file name?

pkg/controller/reconciler/shared_test.go
```diff
@@ -43,8 +45,8 @@ const (
 	recipeTestControllerDelayInterval = time.Millisecond * 100
 )
 
-func createEnvironment(radius *mockRadiusClient, name string) {
-	id := fmt.Sprintf("/planes/radius/local/resourceGroups/%s/providers/Applications.Core/environments/%s", name, name)
+func createEnvironment(radius *mockRadiusClient, resourceGroup, name string) {
+	id := fmt.Sprintf("/planes/radius/local/resourceGroups/%s/providers/Applications.Core/environments/%s", resourceGroup, name)
```
sk593, line 49: Just confirming, did we add the RG because we create one if it doesn't exist as part of the deployment template controller?

pkg/controller/reconciler/testdata/deploymenttemplate-withresources.json
```diff
@@ -0,0 +1,37 @@
...
+      "properties": {
+        "name": "env",
+        "location": "global",
+        "properties": {
+          "compute": {
+            "kind": "kubernetes",
+            "resourceId": "self",
+            "namespace": "default"
+          }
+        }
```
sk593, line 33: Do we need to test for different provider configs?

pkg/controller/reconciler/deploymenttemplate_reconciler.go
```diff
@@ -0,0 +1,566 @@
...
+			// if is present in deploymentTemplate.Status.OutputResources but not in outputResources, delete it
+			// if is not present in deploymentTemplate.Status.OutputResources but is in outputResources, create it
+			// if is present in both, do nothing
+
+			existingOutputResources := make(map[string]bool)
+			for _, resource := range deploymentTemplate.Status.OutputResources {
+				existingOutputResources[resource] = true
+			}
+
+			newOutputResources := make(map[string]bool)
```
lakshmimsft, line 188: I did not see a test for this functionality (could it be pulled out)?

pkg/controller/reconciler/deploymentresource_reconciler.go
```diff
@@ -0,0 +1,393 @@
+/*
+Copyright 2024 The Radius Authors.
```
lakshmimsft, line 2: Some files have this as 2024 while most have 2023. May not be of consequence but good to be consistent.

pkg/cli/filesystem/filesystem.go
```diff
@@ -0,0 +1,15 @@
+package filesystem
```
lakshmimsft, line 1: Missing copyright text in some new files

lakshmimsft:
Just a thought: Is it possible to have a multi-step test where where we deploy a template file and check it's updated and then we deploy an updated file for the same template with updated resources and check for correct update? That may be a way to test the updates to outputresources unless there was another way that exists and I missed it.

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
willdavsmith: Updated to use these values instead

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	FileSystem      filesystem.FileSystem
+	Group           string
+	FilePath        string
+	Parameters      map[string]map[string]any
+	DestinationFile string
+	AzureScope      string
+	AWSScope        string
+}
+
+// NewRunner creates a new instance of the `rad deploy` runner.
```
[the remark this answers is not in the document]
willdavsmith: fixed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+rad bicep generate-kubernetes-manifest app.bicep --parameters @app.bicepparam --parameters tag=latest --destination-file app.yaml --resource-group default
+		`,
+		Args: cobra.ExactArgs(1),
+		RunE: framework.RunCommand(runner),
+	}
+
+	commonflags.AddResourceGroupFlag(cmd)
+	commonflags.AddParameterFlag(cmd)
+
+	cmd.Flags().StringP("destination-file", "d", "", "Path of the generated DeploymentTemplate yaml file.")
```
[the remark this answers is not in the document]
willdavsmith: updated the description: "Path of the generated DeploymentTemplate yaml file created by running this command."

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return err
+	}
+
+	// If the destination file is not provided, use the base name of the file with a .yaml extension
+	if r.DestinationFile == "" {
+		r.DestinationFile = strings.TrimSuffix(filepath.Base(r.FilePath), filepath.Ext(r.FilePath)) + ".yaml"
+	}
+
+	if filepath.Ext(r.DestinationFile) != ".yaml" {
```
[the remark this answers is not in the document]
willdavsmith: updated to accept both

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = r.createDeploymentTemplateYAMLFile(deploymentTemplate)
+	if err != nil {
+		return err
+	}
+
+	// Print the path to the file
```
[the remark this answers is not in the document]
willdavsmith: removed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go
```diff
@@ -0,0 +1,285 @@
...
+
+	return deploymentTemplate, nil
+}
+
+// createDeploymentTemplateYAMLFile creates a DeploymentTemplate YAML file with the given content.
+func (r *Runner) createDeploymentTemplateYAMLFile(deploymentTemplate map[string]any) error {
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+
+	// Set the indentation to 2 spaces
```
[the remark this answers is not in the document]
willdavsmith: fixed

pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest_test.go
```diff
@@ -0,0 +1,197 @@
...
+			Input:         []string{"app.bicep", "-g", "default"},
+			ExpectedValid: true,
+			ValidateCallback: func(t *testing.T, r framework.Runner) {
+				runner := r.(*Runner)
+				require.Equal(t, "default", runner.Group)
+			},
+		},
+		{
+			Name:          "rad bicep generate-kubernetes-manifest - valid with group long flag",
+			Input:         []string{"app.bicep", "--group", "default"},
```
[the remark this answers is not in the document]
willdavsmith: added the test

pkg/cli/filesystem/memmapfs_test.go
```diff
@@ -0,0 +1,83 @@
...
+	require.True(t, fs.Exists(fileName))
+}
+
+func TestMemMapFileSystem_Open(t *testing.T) {
+	fs := NewMemMapFileSystem()
+	fileName := "testfile"
+
+	_, _ = fs.Create(fileName)
+
+	file, err := fs.Open(fileName)
```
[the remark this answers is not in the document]
willdavsmith: no, because they aren't actually opening a file, they're just adding content to a `map` in memory. This is useful for testing

pkg/cli/filesystem/osfs.go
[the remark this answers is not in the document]
willdavsmith: These are just a wrapper over the `os` functions so I don't think unit tests are necessary.

pkg/cli/filesystem/osfs.go
willdavsmith: These are just a wrapper over the os functions so I don't think unit tests are necessary.

pkg/cli/filesystem/filesystem.go
willdavsmith: That's correct. In the real world (when a user runs the command), they expect to have the file generated for them on their OS filesystem (so we use OSFS). Since we don't want to have a whole OS filesystem for unit tests we can just use the MemMapFS.

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
willdavsmith: done

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
willdavsmith: done

pkg/controller/reconciler/shared_test.go
willdavsmith: yep, also just for clarity

pkg/controller/reconciler/testdata/deploymenttemplate-withresources.json
willdavsmith: for now I think this is okay. it'll take some work to add these tests, we can do it later.

willdavsmith: This PR is already large enough :)

pkg/cli/filesystem/filesystem.go
willdavsmith: fixed

pkg/controller/reconciler/deploymentresource_reconciler.go
willdavsmith: I'll just put 2024 for now since this is the same as the other reconciler files

pkg/controller/reconciler/deploymenttemplate_reconciler.go
willdavsmith: added a test

willdavsmith:
> Is it possible to have a multi-step test where where we deploy a template file and check it's updated and then we deploy an updated file for the same template with updated resources and check for correct update? That may be a way to test the updates to outputresources unless there was another way that exists and I missed it.

yep, just added one! good callout

brooke-hamilton (dismissed):

pkg/cli/deployment/deploy.go
```diff
@@ -49,7 +49,7 @@ const (
 
 type ResourceDeploymentClient struct {
 	RadiusResourceGroup string
-	Client              *sdkclients.ResourceDeploymentsClient
+	Client              sdkclients.ResourceDeploymentsClient
```
willdavsmith, line 52: I changed this from a struct to an interface (since I use the same type in deploymenttemplate controller). This makes testing way easier.

pkg/cli/deployment/deploy_test.go
```diff
@@ -54,8 +54,6 @@ func Test_GetProviderConfigs(t *testing.T) {
 func Test_GetProviderConfigsWithAzProvider(t *testing.T) {
 	resourceDeploymentClient := ResourceDeploymentClient{
 		RadiusResourceGroup: "testrg",
-		Client:              &sdkclients.ResourceDeploymentsClient{},
-		OperationsClient:    &sdkclients.ResourceDeploymentOperationsClient{},
```
willdavsmith, line 58: These weren't actually doing anything in the test. We should write tests using the interface like I do in the deploymenttemplate controller unit tests.

pkg/controller/reconciler/deployment_client.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"fmt"
+
+	aztoken "github.com/radius-project/radius/pkg/azure/tokencredentials"
+	"github.com/radius-project/radius/pkg/sdk"
+	sdkclients "github.com/radius-project/radius/pkg/sdk/clients"
+)
+
+type DeploymentClient interface {
```
willdavsmith, line 28: I added this DeploymentClient interface (similar to the existing RadiusClient interface) because the deploymenttemplate controller needs to access the UCP deployment client. Having this interface makes it possible to write out mocks for testing.

pkg/controller/reconciler/deploymentresource_reconciler.go
willdavsmith: I removed most of the state, just left the ResourceID now.

pkg/controller/reconciler/mock_radius_client_test.go
```diff
@@ -75,6 +76,13 @@ func (rc *mockRadiusClient) Update(exec func()) {
 	exec()
 }
```
willdavsmith, line 77: The changes in this file are for clarity. I was getting very confused when adding a new client, so I made it explicit that these operations are for the RadiusClient

pkg/controller/reconciler/radius_client.go
```diff
@@ -29,15 +27,6 @@ import (
 	"github.com/radius-project/radius/pkg/ucp/resources"
 )
```
willdavsmith, line 29: The changes in this file are only for clarity. functionality is unaffected

pkg/recipes/driver/bicep.go
```diff
@@ -72,7 +72,7 @@ type BicepOptions struct {
 
 type bicepDriver struct {
 	ArmClientOptions *arm.ClientOptions
-	DeploymentClient *clients.ResourceDeploymentsClient
+	DeploymentClient clients.ResourceDeploymentsClient
```
willdavsmith, line 75: Swapping to use interface instead of struct

pkg/controller/reconciler/deploymenttemplate_reconciler_test.go
```diff
@@ -0,0 +1,871 @@
...
+
+	// Complete the delete operation on the DeploymentResource.
+	mockDeploymentClient.CompleteOperation(dependencyStatus.Operation.ResumeToken, nil)
+
+	// Wait for the DeploymentTemplate to be deleted.
+	waitForDeploymentResourceDeleted(t, k8sClient, dependencyName)
+	waitForDeploymentTemplateStateDeleted(t, k8sClient, namespacedName)
+}
+
+func Test_DeploymentTemplateReconciler_OutputResources(t *testing.T) {
```
willdavsmith, line 593: @lakshmimsft here

pkg/sdk/clients/resourcedeploymentsclient.go
```diff
@@ -175,3 +188,57 @@ func (client *ResourceDeploymentsClient) createOrUpdateCreateRequest(ctx context
 	req.Raw().Header["Accept"] = []string{"application/json"}
 	return req, runtime.MarshalAsJSON(req, parameters)
 }
+
+// ContinueCreateOperation continues a create operation given a resume token.
+func (client *ResourceDeploymentsClientImpl) ContinueCreateOperation(ctx context.Context, resumeToken string) (*runtime.Poller[ClientCreateOrUpdateResponse], error) {
+	return runtime.NewPollerFromResumeToken[ClientCreateOrUpdateResponse](resumeToken, *client.pipeline, nil)
+}
+
+func (client *ResourceDeploymentsClientImpl) Delete(ctx context.Context, resourceID, apiVersion string) (*runtime.Poller[ClientDeleteResponse], error) {
```
lakshmimsft, line 197: Pls add comment for function.

pkg/controller/reconciler/deployment_client.go
```diff
@@ -0,0 +1,82 @@
...
+	ContinueCreateOperation(ctx context.Context, resumeToken string) (Poller[sdkclients.ClientCreateOrUpdateResponse], error)
+	Delete(ctx context.Context, resourceID, apiVersion string) (Poller[sdkclients.ClientDeleteResponse], error)
+	ContinueDeleteOperation(ctx context.Context, resumeToken string) (Poller[sdkclients.ClientDeleteResponse], error)
+}
+
+type DeploymentClientImpl struct {
+	connection sdk.Connection
+}
+
+func NewDeploymentClient(connection sdk.Connection) *DeploymentClientImpl {
```
lakshmimsft, line 43: comments needed for exported type/functions etc

pkg/controller/api/radapp.io/v1alpha3/deploymentresource_types.go
```diff
@@ -42,9 +39,6 @@ type DeploymentResourceStatus struct {
 
 	// Phrase indicates the current status of the Deployment Resource.
 	Phrase DeploymentResourcePhrase `json:"phrase,omitempty"`
-
```
sk593: Was this unused before?

willdavsmith: I can't tell exactly what this is referring to but if it's the ProviderConfig, yes it's unused now that I'm using the DeploymentsClient instead of the RadiusClient.

pkg/controller/reconciler/deployment_client.go
willdavsmith: fixed

pkg/sdk/clients/resourcedeploymentsclient.go
willdavsmith: added

pkg/controller/reconciler/deployment_client.go
```diff
@@ -0,0 +1,95 @@
...
+
+// DeploymentClient is an interface for interacting with
+// UCP ResourceDeploymentsClient.
+type DeploymentClient interface {
+	ResourceDeployments() ResourceDeploymentsClient
+}
+
+// ResourceDeploymentsClient is an interface for interacting
+// with UCP Deployments.
+type ResourceDeploymentsClient interface {
```
sk593, line 36: I'm a little confused by the naming choice. We already have a `ResourceDeploymentsClient` used in the SDK clients package: https://github.com/radius-project/radius/blob/72de369155eeb9bcbcf79e9ca68a2cc994e34749/pkg/controller/reconciler/deployment_client.go#L32

Based on the code, it seems like these are both deployment clients that serve UCP (although the one linked above serves Azure and UCP) so why are we redefining the interface? If both `ResourceDeploymentsClient are meant to serve different purposes or serve different deployments (Azure/UCP vs deployment resources/templates), I think we should rename this for clarity.

pkg/controller/reconciler/deployment_client.go
```diff
@@ -0,0 +1,95 @@
...
+// NewDeploymentClient creates a new DeploymentClient
+// with the given connection.
+func NewDeploymentClient(connection sdk.Connection) *DeploymentClientImpl {
+	return &DeploymentClientImpl{connection: connection}
+}
+
+var _ DeploymentClient = (*DeploymentClientImpl)(nil)
+
+// CreateOrUpdate creates or updates a deployment.
+func (rdc *ResourceDeploymentsClientImpl) CreateOrUpdate(ctx context.Context, parameters sdkclients.Deployment, resourceID, apiVersion string) (Poller[sdkclients.ClientCreateOrUpdateResponse], error) {
```
sk593, line 57: Is `rdc *ResourceDeploymentsClientImpl` the correct value? Should it not be 'DeploymentClientImpl'?

pkg/controller/reconciler/deployment_client.go
```diff
@@ -0,0 +1,95 @@
...
+
+// ContinueDeleteOperation continues a delete operation.
+func (rdc *ResourceDeploymentsClientImpl) ContinueDeleteOperation(ctx context.Context, resumeToken string) (Poller[sdkclients.ClientDeleteResponse], error) {
+	return rdc.inner.ContinueDeleteOperation(ctx, resumeToken)
+}
+
+var _ ResourceDeploymentsClient = (*ResourceDeploymentsClientImpl)(nil)
+
+// ResourceDeploymentsClientImpl is an implementation of ResourceDeploymentsClient.
+type ResourceDeploymentsClientImpl struct {
```
sk593, line 79: Can you explain why we're choosing to wrap the `ResourceDeploymentsClient` from sdkclients and also defined a new `ResourceDeploymentsClient` in this file? I'm not sure I see the 2 distinct use cases here

pkg/controller/reconciler/deployment_client.go
willdavsmith: They do the same thing - I think we could consolidate them. I think I would rather handle this together with other changes later to move the mocks outside of `reconciler` package though.

pkg/controller/reconciler/deployment_client.go
willdavsmith: I think this is right. `ResourceDeploymentsClientImpl` is the type that has all of the actual functions, `DeploymentClientImpl` just has an accessor (`ResourceDeployments()`)

pkg/controller/reconciler/deployment_client.go
willdavsmith: They're the same - we can merge them. For now, I wanted to keep the code in the `reconciler` package to match the `RadiusClient`. We can consolidate after

pkg/recipes/controllerconfig/config.go
```diff
@@ -44,7 +44,7 @@ type RecipeControllerConfig struct {
 	ConfigLoader configloader.ConfigurationLoader
 
 	// DeploymentEngineClient is the client for interacting with the deployment engine.
-	DeploymentEngineClient *clients.ResourceDeploymentsClient
+	DeploymentEngineClient clients.ResourceDeploymentsClient
```
willdavsmith, line 47: This is struct -> interface, shouldn't affect functionality

sk593 (approved):
:shipit:

PR #8016, opened 2024-10-24
Setting cli prompt width to the user's terminal width

### LINKED ISSUE #5467: Set prompt width to the user's terminal width
Today we hardcode the width of a text prompt:

https://github.com/project-radius/radius/blob/main/pkg/cli/prompt/text/text.go#L38

We should instead have a smart approach where we read the user's terminal window set the prompt width based on that instead.

Today if a user's terminal window is below 20 (40 in the current PR), then the window flashes. Since 40 is still a small width this wouldn't be hit too much

[AB#7647](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/7647)

vishwahiremat:
# Description

- Getting the terminal size when the prompt model is created and setting the width to be terminal width.

## Type of change

Fixes: https://github.com/radius-project/radius/issues/5467

## Contributor checklist


pkg/cli/prompt/text/text.go
```diff
@@ -74,7 +77,12 @@ func NewTextModel(prompt string, options TextModelOptions) Model {
 	// so it will be blocked. This means you can't type `prod-aws` which is a valid name.
 	ti := textinput.New()
 	ti.Focus()
-	ti.Width = 40
+	width, err := getTerminalSize()
+	if err != nil {
+		return Model{}, err
+	}
+
+	ti.Width = width
```
rynowak, line 85: Do you know what the behavior is if we don't set the width? https://github.com/charmbracelet/bubbles/blob/master/textinput/textinput.go#L118

Does it autoscale?

rynowak: I agree that hardcoding is not the right thing.

pkg/cli/prompt/text/text.go
```diff
@@ -84,7 +92,7 @@ func NewTextModel(prompt string, options TextModelOptions) Model {
 		options:   options,
 		prompt:    prompt,
 		textInput: ti,
-	}
+	}, nil
```
rynowak, line 95: Can you look into this as an approach? https://github.com/charmbracelet/bubbletea/blob/main/examples/progress-download/tui.go#L44

The framework should automatically send us `WindowSizeMsg`

pkg/cli/prompt/text/text.go
vishwahiremat: tested it by not setting the width and it automatically sets the width to teminal width and it autoscales as well. I am going to update the changes to remove the setting of width property.

rynowak: Great 👍 simple solutions are good.

pkg/cli/prompt/text/text.go
```diff
@@ -148,7 +151,7 @@ func (m Model) View() string {
 	view := &strings.Builder{}
 	view.WriteString(m.prompt)
```
rynowak, line 152: Do we need to implement wrapping for the prompt? or is that done automatically?

rynowak (approved):
:shipit: looks great! Thanks for finding and fixing the other cases.

PR #8019, opened 2024-10-31
Pin dependencies for typespec compilation

rynowak:
# Description

This change pins the dependencies we use for the TypeSpec compiler. Since these were set to latest, it's easy to operate on a configuration we don't support. TypeSpec moves fast and breaks things, so being certain about the dependencies is important.

Also adds the package.json to depdendabot so we can get PRs for updates.

## Type of change

## Contributor checklist


typespec/package.json
```diff
@@ -3,15 +3,15 @@
...
-    "@typespec/versioning": "latest"
+    "@typespec/compiler": "^0.60.0",
+    "@azure-tools/typespec-autorest": "^0.46.0",
+    "@azure-tools/typespec-azure-core": "^0.46.0",
+    "@azure-tools/typespec-azure-resource-manager": "^0.46.0",
+    "@azure-tools/typespec-providerhub": "^0.46.0",
+    "@typespec/http": "^0.60.0",
+    "@typespec/openapi": "^0.60.0",
+    "@typespec/rest": "^0.60.0",
+    "@typespec/versioning": "^0.60.0"
```
rynowak, line 14: PTAL at the `package-lock.json` to verify. We're still installing the same versions, but the version ranges we consider **allowed** are constrained.

sk593 (dismissed):

.github/dependabot.yml
```diff
@@ -28,3 +28,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "npm"
```
brooke-hamilton, line 31: 🥇 for adding dependabot configuration.

brooke-hamilton (commented):
LGTM: Running `make generate` in the dev container results in zero changes to the generated code. 🚀

lakshmimsft (approved):

typespec/package.json
```diff
@@ -3,15 +3,15 @@
...
-    "@typespec/compiler": "latest",
-    "@azure-tools/typespec-autorest": "latest",
-    "@azure-tools/typespec-azure-core": "latest",
-    "@azure-tools/typespec-azure-resource-manager": "latest",
-    "@azure-tools/typespec-providerhub": "latest",
-    "@typespec/http": "latest",
-    "@typespec/openapi": "latest",
-    "@typespec/rest": "latest",
-    "@typespec/versioning": "latest"
+    "@typespec/compiler": "^0.60.0",
```
brooke-hamilton, line 6: I read this as meaning we are ok with any version from `0.60.0` up to, but not including, `1.0.0`. Is that correct?

rynowak: I think you're right. These should be `~0.60.0`

lakshmimsft: this would be more restrictive then. I thought typically minor versions are meant to be backward compatible but this approach would mean greater control in how we handle updates.

rynowak: More restrictive is good. Semver basically has no rules for `0.X.Y` versions.

ytimocin (dismissed):

lakshmimsft (approved):

lakshmimsft (dismissed):
I tried to dismiss my own review/approval but looks like I cannot in Github. We're updating to ~0.60.0 as mentioned above

superbeeny (approved):
LGTM

brooke-hamilton (approved):
@rynowak looks good: rebuild dev container --> `make generate` --> zero file changes.

ytimocin (approved):

PR #8050, opened 2024-11-12
[UDT PR 6/N] Add default controller to async controller registry

rynowak:
# Description

This change adds a "default" controller capability to the controller registry used by the async worker. This is needed for the dynamic-rp. Because the dynamic-rp is extensible, it processes many different types of operations, it's not possible for us to statically register all of these. Instead, the dynamic-rp will use this new default capability to register a single controller with dynamic functionality.

## Type of change

## Contributor checklist


pkg/armrpc/asyncoperation/worker/registry.go
```diff
@@ -65,14 +70,46 @@ func (h *ControllerRegistry) Register(ctx context.Context, resourceType string,
...
 	}
 
-	return nil
+	return h.createDefault(ctx, operationType)
+}
+
+func (h *ControllerRegistry) createDefault(ctx context.Context, operationType v1.OperationType) (ctrl.Controller, error) {
+	if h.defaultFactory == nil {
+		return nil, nil
+	}
```
ytimocin, line 101: Isn't this function more like **getDefault**? If **defaultFactory** is nil, then we return nil for the controller. So, to me, it is more like **getDefault** that is called from the **Get** function if no controller is registered.

rynowak: Is this feedback about the naming or behavior? I don't have a strong opinion about the name 😆

ytimocin: It is about the name of the function not the behavior. It seems like it doesn't create the default but get the existing default.

rynowak:
Functional test failure appeared to be contention in azure on a storage account. If it happens again I'll disable it.

rynowak:
@ytimocin - updated

ytimocin (approved):

PR #8053, opened 2024-11-12
Unskipping tea tests and adding protection around possible nil models

ytimocin:
# Description
Un-skipping tea tests and adding protection around possible nil models.

## Type of change
Fixes: #issue_number

## Contributor checklist

### DIFF
--- a/pkg/cli/cmd/deploy/deploy_test.go
+++ b/pkg/cli/cmd/deploy/deploy_test.go
@@ -527,7 +527,6 @@ func Test_Run(t *testing.T) {
 	})
 
 	t.Run("Deployment with missing parameters", func(t *testing.T) {
-		//t.Skip()
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
--- a/pkg/cli/cmd/radinit/display_test.go
+++ b/pkg/cli/cmd/radinit/display_test.go
@@ -35,23 +35,11 @@ var (
 )
 
 func Test_summaryModel(t *testing.T) {
-	t.Skip("Test is flaky. See: https://github.com/radius-project/radius/issues/8044")
-
 	waitForRender := func(t *testing.T, reader io.Reader) string {
 		normalized := ""
 		teatest.WaitFor(t, reader, func(bts []byte) bool {
 			normalized = stripansi.Strip(strings.ReplaceAll(string(bts), "\r\n", "\n"))
 			return strings.Contains(normalized, strings.Trim(summaryFooter, "\n"))
-		}, teatest.WithDuration(waitTimeout))
-
-		return normalized
-	}
-
-	waitForEmpty := func(t *testing.T, reader io.Reader) string {
-		normalized := ""
-		teatest.WaitFor(t, reader, func(bts []byte) bool {
-			normalized = stripansi.Strip(strings.ReplaceAll(string(bts), "\r\n", "\n"))
-			return !strings.Contains(normalized, strings.Trim(summaryFooter, "\n"))
 		}, teatest.WithDuration(waitTimeout))
 
 		return normalized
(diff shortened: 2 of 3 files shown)


pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -527,7 +527,6 @@ func Test_Run(t *testing.T) {
 	})
 
 	t.Run("Deployment with missing parameters", func(t *testing.T) {
-		//t.Skip()
```
rynowak, line 530: Why is it a good idea to do this? Do we have a theory about why this was failing? What was the fix?

pkg/cli/cmd/radinit/display_test.go
```diff
@@ -50,6 +48,11 @@ func Test_summaryModel(t *testing.T) {
 	waitForEmpty := func(t *testing.T, reader io.Reader) string {
 		normalized := ""
 		teatest.WaitFor(t, reader, func(bts []byte) bool {
+			if bts == nil {
```
rynowak, line 51: Do we have a theory about why this would happen? Why is this the right fix?

pkg/cli/prompt/text/text_test.go
```diff
@@ -122,35 +122,52 @@ func Test_E2E(t *testing.T) {
...
 
 	t.Run("confirm default", func(t *testing.T) {
-		t.Skip("This test is intermittently failing in linux_amd64: https://github.com/radius-project/radius/issues/7670")
 		tm := setup(t)
-
 		tm.Send(tea.KeyMsg{Type: tea.KeyEnter})
+		tm.WaitFinished(t, teatest.WithFinalTimeout(waitTimeout))
+
+		finalModel := tm.FinalModel(t)
+		require.NotNil(t, finalModel, "Final model should not be nil")
```
rynowak, line 130: Can you please explain why this is a good fix given the failure pattern from the original issue? https://github.com/radius-project/radius/issues/7670

ytimocin: I believe that there is sometimes a very short delay of model being updated. I am still in the process of proving this. The reason for my thinking is that we get a nil pointer exception, and we don't get it all the time. I think, but am not sure yet, that waiting for the model to be completely finished may fix the issue.

As mentioned above, I am working on learning more about the tea library and proving my point. :)

rynowak: > I believe that there is sometimes a very short delay of model being updated

That's my theory as well. The test is flaky, which means the root cause is probably a data-race.

pkg/cli/cmd/radinit/display_test.go
```diff
@@ -66,12 +54,18 @@ func Test_summaryModel(t *testing.T) {
...
-		tm.Send(tea.KeyMsg{Type: key})
+		// Press the given key
+		tm.Send(tea.KeyMsg{
+			Type: key,
+		})
 
-		// Wait for final render and exit.
-		tm.WaitFinished(t, teatest.WithFinalTimeout(waitTimeout))
-		waitForEmpty(t, tm.FinalOutput(t))
+		if err := tm.Quit(); err != nil {
```
ytimocin, line 62: Please see example tests from the original repo: https://github.com/charmbracelet/x/blob/20117e9c8cd5ad229645f1bca3422b7e4110c96c/exp/teatest/app_test.go#L34. We can discuss about them.

pkg/cli/prompt/text/text_test.go
ytimocin: These are example tests from the original repo: https://github.com/charmbracelet/x/blob/20117e9c8cd5ad229645f1bca3422b7e4110c96c/exp/teatest/app_test.go#L34.

pkg/cli/cmd/radinit/display_test.go
rynowak: This is a different test though right? We're now explicitly telling the model to quit.

ytimocin: I added a comment below this line. I turns out that the **FinalModel** only returns once the program has finished running or when it times out. Please [see the official repo doc](https://github.com/charmbracelet/x/blob/20117e9c8cd5ad229645f1bca3422b7e4110c96c/exp/teatest/teatest.go#L220). That is why we call **tm.Quit()** before **tm.FinalModel()**.

Is there a reason why we shouldn't call the quit function? I believe if we don't call it, we can't get the final model unless we wait for it to timeout.

Anything I may be missing here? cc/ @vishwahiremat

pkg/cli/cmd/radinit/display_test.go
```diff
@@ -47,16 +45,6 @@ func Test_summaryModel(t *testing.T) {
...
-	waitForEmpty := func(t *testing.T, reader io.Reader) string {
-		normalized := ""
-		teatest.WaitFor(t, reader, func(bts []byte) bool {
-			normalized = stripansi.Strip(strings.ReplaceAll(string(bts), "\r\n", "\n"))
-			return !strings.Contains(normalized, strings.Trim(summaryFooter, "\n"))
-		}, teatest.WithDuration(waitTimeout))
-
-		return normalized
-	}
-
```
vishwahiremat, line 59: I am trying to understand why is this not needed.

ytimocin: `tm.Quit` is taking care of this part now.

vishwahiremat (approved):

PR #8126, opened 2024-12-12
Simplify database interface

rynowak:
# Description

This change simplifes our database provider interface by removing the 'name' parameter.

Prior to this change, each database client was used to store and retrieve a single resource type's data. Every piece of code had to be aware of the resource type associated with the client, and make sure to get a different client for each resource type it interacted with.

This was complicated further by the fact that not all of our database providers implemented this restriction.

---

After removing this restriction, our code can be simplified all over the place. A lot of code that used to need the `dataprovider.DataStoreProvider` now just needs the `store.StorageClient`.

This change pushes the simplification as far as possible. I also added validation to places where we initialize controllers, because it was too easy to initialize a controller with missing data.

## Type of change

## Contributor checklist


pkg/armrpc/asyncoperation/worker/registry.go
```diff
@@ -74,41 +71,40 @@ func (h *ControllerRegistry) Register(ctx context.Context, resourceType string,
...
 // The default controller will be used when Get is called with an operation type that has no registered controller.
 // The default controller will not be cached by the registry.
-func (h *ControllerRegistry) RegisterDefault(ctx context.Context, factoryFn ControllerFactoryFunc, opts ctrl.Options) error {
+func (h *ControllerRegistry) RegisterDefault(factoryFn ControllerFactoryFunc, opts ctrl.Options) error {
 	h.ctrlMapMu.Lock()
 	defer h.ctrlMapMu.Unlock()
 
+	// Note: we can't call opts.Validate() here because we don't know the resource type yet.
+	if opts.StorageClient == nil {
+		return fmt.Errorf("invalid controller options: .StorageClient is required")
```
ytimocin, line 80:
```suggestion
		return fmt.Errorf("invalid controller options: StorageClient is required")
```

pkg/armrpc/frontend/controller/controller.go
```diff
@@ -69,6 +66,28 @@ type Options struct {
 	StatusManager sm.StatusManager
 }
 
+func (o Options) Validate() error {
+	var err error
+	if o.Address == "" {
+		err = errors.Join(err, errors.New(".Address is required"))
```
ytimocin, line 72: why do we need dots

pkg/ucp/dataprovider/storageprovider.go
```diff
@@ -19,70 +19,122 @@ package dataprovider
...
+
+// GetStorageClient returns a storage client for the given resource type.
+func (p *DataStorageProvider) GetClient(ctx context.Context) (store.StorageClient, error) {
+	// Guarantee single initialization.
+	p.init.RLock()
+	result := p.result
+	p.init.RUnlock()
+
+	if result.client == nil && result.err == nil {
+		result = p.initialize(ctx)
```
ytimocin, line 88: It might be my bad but where do we use this result down the function?

pkg/armrpc/asyncoperation/controller/controller.go
```diff
@@ -44,6 +41,22 @@ type Options struct {
...
+	if o.ResourceType == "" {
+		err = errors.Join(err, errors.New("ResourceType is required"))
+	}
+
+	// KubeClient and GetDeploymentProcessor are not used by the majority of the code, so they
+	// are not validated here.
+
+	return err
+}
+
```
lakshmimsft, line 59: missing unit test for new function

pkg/armrpc/asyncoperation/worker/registry.go
rynowak: I did it this because it's consistent with the other error messages in validate. It has a leading period because go doesn't like error messages that begin with a capital letter.

pkg/armrpc/asyncoperation/worker/worker_runoperation_test.go
```diff
@@ -513,9 +494,13 @@ func TestRunOperation_CancelContext(t *testing.T) {
 
 	worker := New(Options{}, nil, tCtx.testQueue, nil)
 
+	// This test has a race condition with the worker loop trying to make an operation status
+	// as failed. We can't use a mock because that might happen after the mock is destroyed (on test completion).
+	//
+	// Instead we use the in-memory store.
```
rynowak, line 500: This took me a few hours to chase down 😢

lakshmimsft:
will there be separate PR for name updates proposed here: https://gist.github.com/rynowak/0e625521404326eb194adc98155a736b

rynowak:
Functional test failure is due to applications rp crashing at startup. I'll dig into it.

pkg/ucp/dataprovider/storageprovider.go
rynowak: You're right, this should be used instead of reading the struct field.

pkg/armrpc/asyncoperation/controller/controller.go
rynowak: Added

pkg/ucp/dataprovider/storageprovider.go
```diff
@@ -18,71 +18,119 @@ package dataprovider
...
+
+// DataStorageProviderFromClient creates a new instance of the DataStorageProvider struct with the given client.
+//
+// This will always return the given client and will not attempt to create a new one. This can be used for testing
+// with mocks.
+func DataStorageProviderFromClient(client store.StorageClient) *DataStorageProvider {
+	return &DataStorageProvider{result: result{client: client}}
+}
+
+// GetStorageClient returns a storage client for the given resource type.
```
lakshmimsft, line 74: Comment needs update since resource type is not an input for client generated

rynowak: 👍

lakshmimsft (approved):

PR #8161, opened 2024-12-20
Add async operation support to dynamic-rp

### LINKED ISSUE #6688: User-Defined Resource Types

Today, there are a fixed set of portable Radius resources:

- Applications.Dapr/stateStores
- Applications.Dapr/pubSubBrokers
- Applications.Datastores/redisCaches
- Applications.Datastores/sqlDatabases
- ...

These resources have a fixed schema and it allows teams to only select from a pre-defined list.

What's missing is a way for a team at an organization like Contoso to model their own schema and abstractions to work with.

For example, Contoso may want a resource type like:

```bicep
resource storage 'Contoso.Storage/database' = {
  name: 'inventory'
  properties: {...}
}
```

This would allow developers to select from resource types that are custom to Contoso. This also doesn't have to be limited to storage/messaging resources. Custom types could also encapsulate entire services or dependencies such as:

```bicep
resource storage 'Contoso.Monitoring/diagnostics' = {
  name: 'inventory'
  properties: {...}
}
```

## Acceptance criteria

- Users can define their own types + schema
- These resources can be modeled in Bicep, acting as a key/value store for the properties
- Recipes should work with the resources, allowing IT operators to define Recipe templates that spin up the actual backing infrastructure.

## Additional context

[AB#10284](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10284)
```[tasklist]
### Tasks
```

rynowak:
# Description

This change adds the async operationStatus and operationResult handlers to the dynamic-rp API. This change is significant, because like all functionality in dynamic-rp, it's dynamic and needs to work for any UDT resource provider rather than a fixed namespace.

## Type of change

Fixes: #6688 

## Contributor checklist

### DIFF
--- a/pkg/armrpc/asyncoperation/worker/worker.go
+++ b/pkg/armrpc/asyncoperation/worker/worker.go
@@ -267,7 +267,12 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 			result.SetFailed(armErr, false)
 		}
 
-		logger.Info("Operation returned", "success", result.Error == nil, "provisioningState", result.ProvisioningState(), "err", result.Error)
+		// We need the if/else here to prevent a panic inside the logger.
+		if result.Error == nil {
+			logger.Info("Operation returned", "success", "true", "provisioningState", result.ProvisioningState())
+		} else {
+			logger.Info("Operation returned", "success", "false", "provisioningState", result.ProvisioningState(), "err", result.Error)
+		}
 
 		// There are two cases when asyncReqCtx is canceled.
 		// 1. When the operation is timed out, w.completeOperation will be called in L186
--- a/pkg/dynamicrp/frontend/operations.go
+++ b/pkg/dynamicrp/frontend/operations.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package frontend
+
+import (
+	"net/http"
+	"strings"
+
+	v1 "github.com/radius-project/radius/pkg/armrpc/api/v1"
+	"github.com/radius-project/radius/pkg/armrpc/frontend/controller"
+	"github.com/radius-project/radius/pkg/armrpc/frontend/server"
+	"github.com/radius-project/radius/pkg/armrpc/rest"
+	"github.com/radius-project/radius/pkg/ucp/resources"
+)
+
+// dynamicOperationHandler returns an http.Handler that can instantiate and run a controller.Controller for a dynamic resource.
+//
+// Usually when we register a route, we know up-front the resource type that will be handled by the controller.
+// In the dynamic-rp use-case we don't know. We need to dynamically determine the resource type based on the URL.
+//
+// For example:
+//
+// Route: /planes/radius/{planeName}/providers/{providerNamespace}/locations/{locationName}/operationResults/{operationID}
+// URL: /planes/radius/myplane/providers/Applications.Example/locations/global/operationResults/1234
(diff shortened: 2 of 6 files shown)


pkg/armrpc/asyncoperation/worker/worker.go
```diff
@@ -267,7 +267,12 @@ func (w *AsyncRequestProcessWorker) runOperation(ctx context.Context, message *q
 			result.SetFailed(armErr, false)
 		}
 
-		logger.Info("Operation returned", "success", result.Error == nil, "provisioningState", result.ProvisioningState(), "err", result.Error)
+		// We need the if/else here to prevent a panic inside the logger.
+		if result.Error == nil {
+			logger.Info("Operation returned", "success", "true", "provisioningState", result.ProvisioningState())
+		} else {
+			logger.Info("Operation returned", "success", "false", "provisioningState", result.ProvisioningState(), "err", result.Error)
+		}
```
rynowak, line 275: I found this while debugging. It's not breaking anything, but is causing ugliness in the logs.

rynowak: The actual problem is that `v1.ErrorDetails` implements `error` through a value, rather than a pointer. I would like to change that, but it could be a very impactful change.

pkg/dynamicrp/frontend/operations.go
```diff
@@ -0,0 +1,88 @@
...
+// Resource Type: Applications.Example/operationResults
+//
+// # OR
+//
+// Route: /planes/radius/{planeName}/resourceGroups/my-rg/providers/{providerNamespace}/{resourceType}/{resourceName}
+// URL: /planes/radius/myplane/resourceGroups/my-rg/providers/Applications.Example/customService/my-service
+// Resource Type: Applications.Example/customService
+//
+// This code ensures that the controller will be provided with the correct resource type.
+func dynamicOperationHandler(method v1.OperationMethod, baseOptions controller.Options, factory func(opts controller.Options) (controller.Controller, error)) http.HandlerFunc {
```
rynowak, line 48: See comments for the explanation of why we need this. AFAIK this is a requirement only dynamic-rp has.

pkg/dynamicrp/integrationtest/dynamic/operations_test.go
```diff
@@ -0,0 +1,79 @@
...
+	operation := &statusmanager.Status{
+		AsyncOperationStatus: v1.AsyncOperationStatus{
+			ID:     operationStatusID,
+			Name:   operationName,
+			Status: v1.ProvisioningStateUpdating,
+		},
+	}
+
+	err = databaseClient.Save(ctx, &database.Object{Data: operation, Metadata: database.Metadata{ID: operationStatusID}})
+	require.NoError(t, err)
```
rynowak, line 71: One of the cool things about the integration-testing approach is that we can just manipulate the database if we need to simulate something tricky for a test 👍

pkg/armrpc/asyncoperation/worker/worker.go
ytimocin: Should we create an issue for this?

pkg/dynamicrp/frontend/operations.go
```diff
@@ -0,0 +1,88 @@
...
+		// Copy the options and initalize them dynamically for this type.
+		opts := baseOptions
+		opts.ResourceType = id.Type()
+
+		// Special case the operation status and operation result types.
+		//
+		// This is special-casing that all of our resource providers do to store a single data row for both operation statuses and operation results.
+		if strings.HasSuffix(strings.ToLower(opts.ResourceType), "locations/operationstatuses") || strings.HasSuffix(strings.ToLower(opts.ResourceType), "locations/operationresults") {
+			opts.ResourceType = id.ProviderNamespace() + "/operationstatuses"
+		}
```
ytimocin, line 72: In either case, why do we set the Resource Type to **operationStatuses**?

pkg/armrpc/asyncoperation/worker/worker.go
rynowak: I was planning to just fix it.

pkg/dynamicrp/frontend/operations.go
rynowak: The comments right above here explain this, please make a suggestion if it's not clear.

ytimocin: I just checked that we only have one for both which is **Applications.Core/operationStatuses**. Thanks.

ytimocin (approved):

PR #8170, opened 2024-12-26
Add retry logic to UCP `GetAWSResourceWithPost` handler

### LINKED ISSUE #7352: Add resiliency to the GET operation of AWS resource deployments

Sometimes in the functional tests (and potentially for users), we see a failed AWS deployment due to a GET call returning 404. We should investigate why the call fails and potentially add retries to resolve this. 

Example: https://github.com/radius-project/radius/actions/runs/8311733036/job/22745893415
It looks like the deployment fails because it cannot find the bucket after it is created. Checking the console will show that the bucket is created. (it's left hanging)

My hypothesis is that AWS takes a minute to register the bucket as created sometimes.

### Acceptance criteria

* Retries for the GET call have been added

### Additional context

Related issue: https://github.com/radius-project/radius/issues/5963

[AB#11554](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/11554)

willdavsmith:
# Description

We've seen flaky functional test failures with AWS S3: https://github.com/radius-project/radius/issues/5963

This PR adds retries to the handler that I think is causing this 404 error.

* Add `pkg/retry` directory for standard retries
* Use `pkg/retry` in UCP `GetAWSResourceWithPost` handler

## Type of change

Fixes: https://github.com/radius-project/radius/issues/7352

## Contributor checklist


pkg/retry/retry.go
```diff
@@ -0,0 +1,80 @@
...
+	BackoffStrategy retry.Backoff
+}
+
+// Retryer is a utility for retrying functions.
+type Retryer struct {
+	config *RetryConfig
+}
+
+// NewNoOpRetryer creates a new Retryer that does not retry.
+func NewNoOpRetryer() *Retryer {
```
willdavsmith, line 28: this is useful for testing. using this retryer should be the same functionality as we have today.

pkg/ucp/frontend/controller/awsproxy/createorupdateawsresource.go
```diff
@@ -125,7 +125,6 @@ func (p *CreateOrUpdateAWSResource) Run(ctx context.Context, w http.ResponseWrit
 
 	if existing {
 		// Get resource type schema
-
```
willdavsmith, line 128: extra space

pkg/ucp/frontend/controller/awsproxy/createorupdateawsresource.go
```diff
@@ -87,7 +87,7 @@ func (p *CreateOrUpdateAWSResource) Run(ctx context.Context, w http.ResponseWrit
 	}
 
 	cloudControlOpts := []func(*cloudcontrol.Options){CloudControlRegionOption(region)}
-	cloudFormationOpts := []func(*cloudformation.Options){CloudFormationWithRegionOption(region)}
```
willdavsmith, line 90: I renamed this function to match the `cloudcontrol` version

pkg/ucp/frontend/controller/awsproxy/getawsresourcewithpost_test.go
```diff
@@ -74,7 +75,7 @@ func Test_GetAWSResourceWithPost(t *testing.T) {
 		CloudControl:   testOptions.AWSCloudControlClient,
 		CloudFormation: testOptions.AWSCloudFormationClient,
 	}
-	awsController, err := NewGetAWSResourceWithPost(armrpc_controller.Options{DatabaseClient: testOptions.DatabaseClient}, awsClients)
+	awsController, err := NewGetAWSResourceWithPost(armrpc_controller.Options{DatabaseClient: testOptions.DatabaseClient}, awsClients, retry.NewNoOpRetryer())
```
willdavsmith, line 78: this should match the functionality we have today, i.e. these tests should pass with no other changes.

pkg/ucp/frontend/controller/awsproxy/getawsresourcewithpost.go
```diff
@@ -44,13 +46,15 @@ var _ armrpc_controller.Controller = (*GetAWSResourceWithPost)(nil)
 type GetAWSResourceWithPost struct {
 	armrpc_controller.Operation[*datamodel.AWSResource, datamodel.AWSResource]
 	awsClients ucpaws.Clients
+	retryer    *retry.Retryer
 }
 
 // NewGetAWSResourceWithPost creates a new GetAWSResourceWithPost controller with the given options and AWS clients.
-func NewGetAWSResourceWithPost(opts armrpc_controller.Options, awsClients ucpaws.Clients) (armrpc_controller.Controller, error) {
+func NewGetAWSResourceWithPost(opts armrpc_controller.Options, awsClients ucpaws.Clients, retryer *retry.Retryer) (armrpc_controller.Controller, error) {
```
willdavsmith, line 53: we could consider adding the retryer to the other ucp awsproxy routes too, either in the future or this PR. I wanted to get some feedback first

rynowak:
I'm sure this is written up somewhere already, but I'd like to understand the failure pattern that we're addressing with this change.

Is it something like this? (please help me fill in the blanks)

- A PUT operation is initiated. 
- The PUT operation succeeds asynchronously.
- A GET operation against the same resource then fails with a 404.
- At some point the future, the same GET operation will succeed if retried, without initiating any additional operations. 

The background context is that any multi-regional control plane is eventually consistent. Azure/ARM/Bicep has a similar eventually consistent behavior underneath it (see notes above), and it's mostly hidden from users via the deployment engine.

willdavsmith:
> I'm sure this is written up somewhere already, but I'd like to understand the failure pattern that we're addressing with this change.
> 
> Is it something like this? (please help me fill in the blanks)
> 
> * A PUT operation is initiated.
> * The PUT operation succeeds asynchronously.
> * A GET operation against the same resource then fails with a 404.
> * At some point the future, the same GET operation will succeed if retried, without initiating any additional operations.
> 
> The background context is that any multi-regional control plane is eventually consistent. Azure/ARM/Bicep has a similar eventually consistent behavior underneath it (see notes above), and it's mostly hidden from users via the deployment engine.

This is exactly what I think is happening. I noticed that in the cases that I was investigating, the resource was actually created but returned a 404 on this route during deployment. My understanding of the DE is that it will perform a PUT operation, monitor the operation, and then do a GET at the end, where it calls UCP (getawsresourcewithpost handler) and returns a 404 because AWS says the resource doesn't exist yet. My hope is that adding retries here will make this situation more reliable without too much overhead. We can verify that it works if we see this issue less in the future, but until then, this is pretty much an educated guess as to what's happening and the solution.

go.mod
```diff
@@ -139,6 +139,7 @@ require (
 	github.com/sagikazarmark/locafero v0.6.0 // indirect
 	github.com/sagikazarmark/slog-shim v0.1.0 // indirect
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 // indirect
+	github.com/sethvargo/go-retry v0.3.0 // indirect
```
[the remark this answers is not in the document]
ytimocin: Last commit seems to be from 6 months ago. Just wondering if that could be an issue.

pkg/retry/retry.go
```diff
@@ -0,0 +1,80 @@
...
+	if config != nil {
+		if config.BackoffStrategy != nil {
+			retryConfig.BackoffStrategy = config.BackoffStrategy
+		}
+	} else {
+		retryConfig.BackoffStrategy = retry.NewExponential(defaultInterval)
+	}
+
+	return &Retryer{
+		config: retryConfig,
```
ytimocin, line 68: Can **retryConfig** ever be just empty? Like **config** is not nil but **config.BackOffStrategy** is?

ytimocin: Is that an okay case?

pkg/retry/retry_test.go
```diff
@@ -0,0 +1,83 @@
...
+	expectedBackoffStrategy := goretry.NewConstant(time.Second * 1)
+	expectedBackoffStrategy = goretry.WithMaxRetries(0, expectedBackoffStrategy)
+
+	require.IsType(t, expectedBackoffStrategy, retryer.config.BackoffStrategy)
+}
+
+func TestNewRetryer(t *testing.T) {
+	config := &RetryConfig{
+		BackoffStrategy: goretry.NewConstant(1 * time.Second),
+	}
```
ytimocin, line 45: We can test this with a **RetryConfig** that has a nil **BackOffStrategy**.

go.mod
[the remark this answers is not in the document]
willdavsmith: I think it should be okay. The code is straightforward and has no dependencies, so hopefully there shouldn't need to be too many new commits.

pkg/retry/retry.go
willdavsmith: good find - I changed the code to not allow this case. now the retryer will use the default configuration in this case. I also added a test

pkg/retry/retry_test.go
willdavsmith: added a test

ytimocin (approved):

PR #8190, opened 2025-01-05
Implement fix for failures caused in test logging

rynowak:
# Description

Logging in tests has an intentional limitation. Any logging performed after the end of a test will cause the test to fail.

Unfortunately for us, this is a really hard limitation. We do a lot of asynchronous processing, and we can't always guarantee that processing is complete when a test completes. See https://github.com/radius-project/radius/actions/runs/12591118184/job/35093755888 for an example failure. This is especially bad because it's never a consistent repro.

This approach is based on the solution that Kubernetes uses: https://github.com/kubernetes/klog/pull/337/files#diff-9c4e6a71ada6fa5d0ee80e05296e5b418351025a21b2f9dce9595da245272313 . They wrote their own logging package, so they can make the fix inside of it. We're using logr (which was created by the same author), so we have to make the fix outside of the the logging package, but it's the same fundamental idea.

## Type of change

## Contributor checklist


rynowak:
@ytimocin - can you help review?

test/testcontext/testcontext.go
```diff
@@ -48,12 +49,60 @@ func Wrap(t *testing.T, ctx context.Context) (context.Context, context.CancelFun
...
+}
+
+var _ testr.TestingT = (*testSink)(nil)
+
+// Stop is called when the test terminates, this will disable logging.
+func (t *testSink) Stop() {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	t.stopped = true
```
ytimocin, line 87: Do we need to do anything other than setting this to true? Like setting the wrapper to nil or something like that?

ytimocin (approved):

PR #8196, opened 2025-01-06
Fix for picking the right helm chart version during release cut

### LINKED ISSUE #8195: Helm install picks up incorrect helm chart version during release cut
### Steps to reproduce

```
curl -sSL https://github.com/radius-project/radius/releases/download/v0.42.0-rc1/rad_linux_amd64 -o rad
chmod +x ./rad
./rad install kubernetes
```

### Observed behavior

You should see the following output: 
```
Installing Radius version v0.42.0-rc1 to namespace: radius-system...
Error: failed to run Radius Helm install, err: 
cannot re-use a name that is still in use
Helm output:
getting history for release radius
creating 1 resource(s)
creating 1 resource(s)
beginning wait for 2 resources with timeout of 1m0s
Clearing REST mapper cache
creating 1 resource(s)
creating 19 resource(s)
beginning wait for 19 resources with timeout of 10m0s
Deployment is not ready: radius-system/bicep-de. 0 out of 1 expected pods are ready
Deployment is not ready: radius-system/bicep-de. 0 out of 1 expected pods are ready
Deployment is not ready: radius-system/bicep-de. 0 out of 1 expected pods are ready
```

Investigating the pod logs shows that the install is using Helm chart version `0.42.42-pr-6071` instead of the desired `0.42.0-rc1`. 

### Desired behavior

Helm doesn't pick up chart versions based on the latest published version and chooses the latest chart based on semver. In this case, `0.42.42-pr-6071` is being chosen over `0.42.0-rc1`. This is causing failures because the version of the Radius on that Helm chart is old. 

To find the correct helm chart version in our code, we search for the release major value but wildcard the patch value. We need to update the versioning logic during the Helm install to handle a few different cases:
1. If there is a pre-release version (i.e. -rc1, -rc2, etc), we want to find a helm chart version that is an exact match 
2. If there is a full release version, we want to find a helm chart version that matches the release value but wildcards the patch value (in case we have a patch release)
3. For dev builds, the release version will always be `0.42.42-dev` so this value should be hardcoded as the version. 

Relevant code: https://github.com/radius-project/radius/blob/fa1078fcbd140e133a45a5222019d79f5bbe37cb/pkg/cli/helm/helm.go#L90 

### Workaround

_No response_

### rad Version

0.42.0-rc1

### Operating system

_No response_

### Additional context

_No response_

[AB#13983](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/13983)

vishwahiremat:
# Description

To find the correct helm chart version in our code, we search for the release major value but wildcard the patch value. We need to update the versioning logic during the Helm install to handle a few different cases:
1. If there is a pre-release version (i.e. -rc1, -rc2, etc), we want to find a helm chart version that is an exact match 
2. If there is a full release version, we want to find a helm chart version that matches the release value but wildcards the patch value (in case we have a patch release)
3. For dev builds, the release version will always be `0.42.42-dev` so this value should be hardcoded as the version. 

## Type of change

Fixes: https://github.com/radius-project/radius/issues/8195

## Contributor checklist


pkg/cli/helm/cluster.go
```diff
@@ -51,7 +52,13 @@ func NewDefaultClusterOptions() ClusterOptions {
 	// If this is an edge build, we'll use the latest available.
 	chartVersion := version.ChartVersion()
 	if !version.IsEdgeChannel() {
-		chartVersion = fmt.Sprintf("~%s", version.ChartVersion())
+		// When the chart version is the final release, we should use the ~ operator to ensure we fetch the latest patch version.
```
ytimocin, line 55: Is there a way to add unit tests to this function?

vishwahiremat: looks like we cannot add a test because the charVersion values are injected here
https://github.com/radius-project/radius/blob/7ac2c22be56c1b9b5ba6abfec63c3313894b379f/build/build.mk#L56
 and we cannot test for different versions as we hardcode the default version info to edge.

sk593: Can we isolate the chart version logic into a separate function and test that separately?

vishwahiremat: I think we still cannot because when we do `version.ChartVersion()` we get the version info from the default info i.e https://github.com/radius-project/radius/blob/907c6b32deceba64a63f41e402764157fb740aec/pkg/version/version.go#L20-L26

so we cannot test it for different verions ex, `-rc1` or `-dev`

pkg/cli/helm/cluster.go
```diff
@@ -51,7 +52,13 @@ func NewDefaultClusterOptions() ClusterOptions {
 	// If this is an edge build, we'll use the latest available.
 	chartVersion := version.ChartVersion()
 	if !version.IsEdgeChannel() {
-		chartVersion = fmt.Sprintf("~%s", version.ChartVersion())
```
willdavsmith, line 54: this looks like the culprit of the -rc versioning bug. good find!

pkg/cli/helm/cluster.go
willdavsmith: since this is so complicated to test, can we file an issue to improve testability here? for example, we could extract these variables into function args and use the defaults here.

pkg/cli/helm/cluster.go
```diff
@@ -51,7 +52,13 @@ func NewDefaultClusterOptions() ClusterOptions {
...
+		// When the chart version is the final release, we should use the ~ operator to ensure we fetch the latest patch version.
+		// For the pre release or dev builds, we should use the exact version.
+		ver, _ := semver.NewVersion(chartVersion)
+		preRelease := ver.Prerelease()
```
willdavsmith, line 58: have you tested this manually to see if it works?

pkg/cli/helm/cluster.go
```diff
@@ -51,7 +52,13 @@ func NewDefaultClusterOptions() ClusterOptions {
 	// If this is an edge build, we'll use the latest available.
 	chartVersion := version.ChartVersion()
 	if !version.IsEdgeChannel() {
-		chartVersion = fmt.Sprintf("~%s", version.ChartVersion())
+		// When the chart version is the final release, we should use the ~ operator to ensure we fetch the latest patch version.
+		// For the pre release or dev builds, we should use the exact version.
```
willdavsmith, line 56: dev builds shouldn't use this codepath because of the conditional above

willdavsmith (changes requested):
added a couple of questions

pkg/cli/helm/cluster.go
vishwahiremat: I have unit tested it manually by changing the default version from `pkg/version.go` to the pre release versions.

pkg/cli/helm/cluster.go
vishwahiremat: will create an issue for it!

pkg/cli/helm/cluster.go
vishwahiremat: makes sense, i believe for dev builds it was always looking for an exact match. Updated the comment

willdavsmith (approved):

PR #8276, opened 2025-01-22
Fix failing non-cloud functional tests by using built-in caching on the setup-go action

brooke-hamilton:
# Description

This PR is to remediate frequent failures of the [scheduled functional test (noncloud)](https://github.com/radius-project/radius/blob/main/.github/workflows/functional-test-noncloud.yaml) workflow. We are having issues with the caching logic. A test failure example is [here](https://github.com/radius-project/radius/actions/runs/12912790930/job/36008466625#step:8:31). 

This PR updates the workflow to remove the caching steps from the workflow and replace it with a new feature of the `actions/setup-go@v5` action, in which caching of Go dependencies is enabled by setting the `cache` parameter to `true`. See [the docs](https://github.com/actions/setup-go/blob/main/docs/adrs/0000-caching-dependencies.md) for details.

I also tested using the `ubuntu-latest` test runner, under the assumption that the non-cloud tests will have fewer artifact outputs and not need the larger disk on `ubuntu-latest-m`.

Changes:

* [`.github/workflows/functional-test-noncloud.yaml`](diffhunk://#diff-b7af1cc5c43a85f1f7d316c6e244726bb14b0b34b59512f60deea7abd498b90cL182-R182): Enabled caching for Go dependencies by setting the `cache` parameter to `true` in the `setup-go` action, and removed the manual cache path setup and cleanup steps to streamline the workflow.
* [`.github/workflows/functional-test-noncloud.yaml`](diffhunk://#diff-b7af1cc5c43a85f1f7d316c6e244726bb14b0b34b59512f60deea7abd498b90cL129-R129): Updated the operating system from `ubuntu-latest-m` to `ubuntu-latest`. Testing for this workflow in a fork did not encounter disk space limits we have seen on other workflows where we tested `ubuntu-latest`.
 
## Type of change

- This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius (issue link optional).

Fixes: #8274 

## Contributor checklist
Please verify that the PR meets the following requirements, where applicable:

- [ ] An overview of proposed schema changes is included in a linked GitHub issue.
- [ ] A design document PR is created in the [design-notes repository](https://github.com/radius-project/design-notes/), if new APIs are being introduced.
- [ ] If applicable, design document has been reviewed and approved by Radius maintainers/approvers.
- [ ] A PR for the [samples repository](https://github.com/radius-project/samples) is created, if existing samples are affected by the changes in this PR.
[...]

PR #8147, opened 2024-12-17
Disabling cache in the setup-go steps of CI/CD

ytimocin:
# Description

Because the cache already exists on the GitHub runners we use, disabling cache in the setup-go steps of our CI/CD pipelines is expected solve the issue of seeing bunch of logs that say "**cannot open: file exists**".

The official documentation of setup-go and the part that is relevant: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs. This points out that setup-go does caching for us. If the cache is already there, we will get all the logs in our workflows.

There is no disadvantage of this approach, as far as I know, since the cache is already there. I actually ran into this [comment](https://github.com/actions/setup-go/issues/403#issuecomment-2299976549) that says that disabling cache in setup-go step improved the speed of the said step. We can always update our workflows if we see any other disadvantages.

Our runners: https://github.com/radius-project/radius/actions/runners.

References:
- https://github.com/actions/setup-go/issues/403
- https://github.com/actions/setup-go/issues/403#issuecomment-2299976549
- https://github.com/actions/setup-go/issues/314#issuecomment-1379132756

## Type of change
- This pull request fixes a bug in Radius and has an approved issue (issue link required).
Fixes: #7790 

## Contributor checklist
Please verify that the PR meets the following requirements, where applicable:

- [ ] An overview of proposed schema changes is included in a linked GitHub issue.
- [ ] A design document PR is created in the [design-notes repository](https://github.com/radius-project/design-notes/), if new APIs are being introduced.
- [ ] If applicable, design document has been reviewed and approved by Radius maintainers/approvers.
- [ ] A PR for the [samples repository](https://github.com/radius-project/samples) is created, if existing samples are affected by the changes in this PR.
- [ ] A PR for the [documentation repository](https://github.com/radius-project/docs) is created, if the changes in this PR affect the documentation or any user facing updates are made.
- [ ] A PR for the [recipes repository](https://github.com/radius-project/recipes) is created, if existing recipes are affected by the changes in this PR.

PR #8280, opened 2025-01-23
Set cache option on setup-go action in workflows

brooke-hamilton:
# Description

This PR sets the `cache: true` option on the `setup-go` actions in the workflows. Previously we used a common caching technique to cache the go install and modules folders, but this feature is now built into the [`setup-go` action](https://github.com/actions/setup-go/blob/main/docs/adrs/0000-caching-dependencies.md). This change was tested with PR #8276. Following that, this PR updates the remaining workflows to use the new caching feature.

The default value for the `cache` option is `true` on the `setup-go@v5` action. This PR explicitly sets the value to `true` on the action where it is not specified so that the option will be obvious, even though it will not change the behavior of the action for cases when the option was not set.

- `build.yaml`: Explicitly set cache option to true, which is the same as the default value.
- `functional-test-cloud.yaml`: Changed the cache option from false to true and removed the original caching actions.
- `lint.yaml`: Explicitly set cache option to true, which is the same as the default value.
- `long-running-azure.yaml`: Explicitly set cache option to true, which is the same as the default value.
- `publish-docs.yaml`: Explicitly set cache option to true, which is the same as the default value.

> NOTE: This PR is created from a branch in this repo instead of a fork from an external repo to allow us to run the cloud test workflows.

## Type of change

Related to issue: #8274
Related to PR #8276

## Contributor checklist


superbeeny (approved):
LGTM, thanks for finding this

.github/actions/setup-rad-cli/action.yaml
ytimocin, line 1: Is this not being used anymore?

.github/workflows/build.yaml
```diff
@@ -101,6 +101,7 @@ jobs:
         with:
           go-version: ${{ env.GOVER }}
           cache-dependency-path: go.sum
+          cache: true
```
ytimocin, line 104: I think this is true by default and does not need to be set explicitly.

ytimocin: When this wasn't added, which means that it was true, we were getting the same error.

ytimocin: I added some information here: https://github.com/radius-project/radius/pull/8147. Maybe you can find the missing link.

.github/actions/setup-rad-cli/action.yaml
brooke-hamilton: I could not find any references to this action in the `radius-project` GitHub org.

.github/workflows/build.yaml
brooke-hamilton: > I think this is true by default and does not need to be set explicitly.

Yes you are right: setting `cache: true` does not change the behavior because it is true by default. I put it here to be explicit because in some of our other workflows we had additional caching logic that was created before the caching feature was available on the `setup-go` action. I removed the additional caching logic from `functional-test-cloud.yaml`, and from the non-cloud tests in a previous PR #8280. Setting this value makes it clear that we are doing caching, in case someone looks at the workflow and thinks "why aren't we doing caching like we had before?" But if you prefer to not have this value set explicitly I will remove it. 

> When this wasn't added, which means that it was true, we were getting the same error.

This workflow wasn't having an error, but the `functional-test-cloud.yaml` workflow was having occasional errors with cache conflicts, so I removed the `cache@v4` action, which I think is now unnecessary due to caching being built into the `setup-go` action. 

> I added some information here: #8147. Maybe you can find the missing link.

I think that PR made sense where we did caching with the `cache@v4` action, but I still saw cache file conflicts with that step, so with this PR we can try using the built-in caching with the `setup-go` action. If we don't see good results we can revert, but so far with PR #8280 things seem OK.

However, one caveat on what I said above, after PR #8280 we are still seeing caching errors with the `setup-go` action. Here is [an example](https://github.com/radius-project/radius/actions/runs/12949863383/job/36121450834#step:7:50). I found two related issues [here](https://github.com/actions/setup-go/issues/506) and [here](https://github.com/actions/setup-go/issues/498). They recommend updating our go version to > 1.23.0. We need to standardize our go version on the version set in `go.mod` anyway (because each workflow has a hard coded version that is separate from `go.mod`). So let's discuss which version and do that in a separate PR.

What do you think about this plan?
1. Remove separate caching steps that used `cache@v4` and set the `setup-go` action property `cache: true`. (this pr)
2. Update our `go.mod` to use a version greater than 1.23.0 and remove hard coded go versions from the workflow. (next pr)
3. If we still see caching errors after doing the above two steps, set `cache: false` on the `setup-go` action.

I really appreciate your feedback on this - I know you have a lot of knowledge in this area having spent significant cycles on it. ❤️

ytimocin: The plan looks good to me. Thanks for the detailed explanation.

ytimocin (approved):

kachawla (approved):

PR #8283, opened 2025-01-23
Restore bicep artifacts before starting tests

### LINKED ISSUE #8184: Flaky test: Failed to restore AWS bicep artifact
## Bug information 

This issue is automatically generated if the scheduled long running test fails. The Radius long running test operates on a schedule of every 2 hours everyday. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](https://github.com/radius-project/radius/actions/runs/12597393259).

[AB#13960](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/13960)

sk593:
# Description

This change restores Bicep artifacts before running any tests. This will avoid any restoration errors while the tests are running as seen in #8184
## Type of change

- This pull request fixes a bug in Radius and has an approved issue #8184 

Fixes: #8184 

## Contributor checklist

- [ ]


brooke-hamilton (dismissed):
I tested this locally - 🚀 🛰️

brooke-hamilton (dismissed):

.github/workflows/functional-test-cloud.yaml
```diff
@@ -758,6 +758,12 @@ jobs:
             }
           }
           EOF
+      - name: Restore Bicep artifacts before running functional tests
+        run: |
+          # Restore Radius Bicep types
+          bicep build ./test/functional-portable/corerp/cloud/resources/testdata/corerp-azure-connection-database-service.bicep
+          # Restore AWS Bicep types 
+          bicep build ./test/functional-portable/corerp/cloud/resources/testdata/aws-s3-bucket.bicep
```
ytimocin, line 766: Do the tests need the files generated by these two commands to be able to run successfully? If so, should we add those files to the repository itself?

I think that to be able to run these tests locally, this is also something we need to run. Is that right?

sk593: No, they're not needed. They're an output of the build command but we're more interested in the artifacts being restored in this step. 

Yes to the second question. The `bicep build` command gets run when we deploy bicep templates. `bicep build` automatically restores the artifacts which might be why we've been seeing flaky restoration failures. This change is added so the bicep artifacts are restored before the tests start and `bicep build` is run

brooke-hamilton (dismissed):

.github/workflows/functional-test-cloud.yaml
```diff
@@ -758,6 +758,12 @@ jobs:
             }
           }
           EOF
+      - name: Restore Bicep artifacts before running functional tests
+        run: |
+          # Restore Radius Bicep types
+          bicep restore ./test/functional-portable/corerp/cloud/resources/testdata/corerp-azure-connection-database-service.bicep --force
+          # Restore AWS Bicep types 
+          bicep restore ./test/functional-portable/corerp/cloud/resources/testdata/aws-s3-bucket.bicep --force
```
kachawla, line 766: Haven't look into this much in depth, can you help me understand why we need to provide path to specific test files? And why just these two?

sk593: This is the command being used: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-cli#restore

The error we're getting is that the bicep artifacts can't be restored before the bicep build command starts running. This command forces a restore before any builds start so the artifacts can be downloaded successfully. The exact files chosen to run the command on aren't necessarily important but we need 1 that uses the Radius extension and 1 that uses the AWS extension so we can restore both bicep artifacts.

sk593: The build command automatically calls the restore command, but since we're seeing flaky failures with this, we'll restore first and then rely on the build command

.github/workflows/functional-test-noncloud.yaml
```diff
@@ -374,7 +374,7 @@ jobs:
           BICEP_RECIPE_TAG_VERSION: ${{ env.REL_VERSION }}
           TEMP_CERT_DIR: ${{ steps.create-local-registry.outputs.temp-cert-dir }}
           SSL_CERT_FILE: ${{ steps.create-local-registry.outputs.temp-cert-dir }}/certs/${{ env.LOCAL_REGISTRY_SERVER }}/client.crt
-      
+
```
ytimocin, line 377: 👍

.github/workflows/functional-test-cloud.yaml
kachawla: Makes sense, thanks for sharing the details. Could you add your note about the kind of files needed in the comments? It will help in directing us which files can be used here if we ever were to delete these two files.

`The exact files chosen to run the command on aren't necessarily important but we need 1 that uses the Radius extension and 1 that uses the AWS extension so we can restore both bicep artifacts.`

kachawla (approved):

PR #8286, opened 2025-01-23
remove hardcoded list of resource types from cli.

nithyatsu:
# Description

Since we now register resource provider manifests at startup and are in process of enabling UDT end to end, we should not use static resource type lists. 

This PR removes static resource type lists in cli code. 

## Type of change

Part of https://github.com/radius-project/radius/issues/6688

## Contributor checklist

    - [ ] Yes
    - [X ] Not applicable
    - [ ] Yes
    - [X ] Not applicable
    - [ ] Yes
    - [ X] Not applicable
    - [ ] Yes
    - [X ] Not applicable
    - [ ] Yes
    - [X ] Not applicable
    - [ ] Yes
    - [ X] Not applicable


pkg/cli/cmd/resource/create/create.go
```diff
@@ -48,7 +48,7 @@ Resources are the primary entities that make up applications.
 Input can be passed via the -f flag to specify a file name.`,
 		Example: `
 # Create a resource (from file)
-rad resource create 'Applications.Core/containers' mycontainer -f /path/to/input.json`,
+rad resource create 'applications.core/containers' mycontainer -f /path/to/input.json`,
```
ytimocin, line 51: Do we only support the lower case now? Why not both?

pkg/cli/cmd/resource/list/list.go
```diff
@@ -45,18 +47,18 @@ func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
 		Short: "Lists resources",
 		Long:  "List all resources of specified type",
 		Example: `
-sample list of resourceType: containers, gateways, pubSubBrokers, extenders, mongoDatabases, rabbitMQMessageQueues, redisCaches, sqlDatabases, stateStores, secretStores
+sample list of resourceType: applications.core/containers, applications.core/gateways, applications.dapr/daprPubSubBrokers, applications.core/extenders, applications.datastores/mongoDatabases, applications.messaging/rabbitMQMessageQueues, applications.datastores/redisCaches, applications.datastores/sqlDatabases, applications.dapr/daprStateStores, applications.dapr/daprSecretStores
```
ytimocin, line 50: Why are resource providers all lower case but some resource types camel case?

ytimocin (changes requested):

pkg/cli/cmd/resource/delete/delete.go
```diff
@@ -52,10 +52,10 @@ func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
 		Short: "Delete a Radius resource",
 		Long:  "Deletes a Radius resource with the given name",
 		Example: `
-sample list of resourceType: containers, gateways, daprPubSubBrokers, extenders, mongoDatabases, rabbitMQMessageQueues, redisCaches, sqlDatabases, daprStateStores, daprSecretStores
+sample list of resourceType: applications.core/containers, applications.core/gateways, applications.dapr/daprPubSubBrokers, applications.core/extenders, applications.datastores/mongoDatabases, applications.messaging/rabbitMQMessageQueues, applications.datastores/redisCaches, applications.datastores/sqlDatabases, applications.dapr/daprStateStores, applications.dapr/daprSecretStores
 
 # Delete a container named orders
-rad resource delete containers orders`,
+rad resource delete applications.core/containers orders`,
```
kachawla, line 58: Why are the examples in all lowercase? From unit tests it looks like, it is not required..

pkg/cli/cmd/resource/delete/delete_test.go
```diff
@@ -44,7 +44,7 @@ func Test_Validate(t *testing.T) {
 	testcases := []radcli.ValidateInput{
 		{
 			Name:          "Valid Delete Command",
```
kachawla, line 46: Can we add a test for delete with unsupported resource type?

pkg/cli/cmd/resource/list/list.go
```diff
@@ -145,6 +147,17 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	parts := strings.Split(r.ResourceType, "/")
+	if len(parts) != 2 {
+		return clierrors.Message("Invalid resource type %q. Expected format: '<provider>/<type>'", r.ResourceType)
+	}
+	resourceProviderNamespace := parts[0]
+	resourceTypeSuffix := parts[1]
+	_, err = common.GetResourceTypeDetails(ctx, resourceProviderNamespace, resourceTypeSuffix, client)
+	if err != nil {
+		return err
+	}
```
kachawla, line 159: Why is this needed in addtion to [RequireResourceType](https://github.com/radius-project/radius/blob/afa6423787fcf91acc6f88240e32aef74dcb924a/pkg/cli/cmd/resource/list/list.go#L120)?

pkg/cli/cmd/resource/list/list_test.go
```diff
@@ -95,15 +97,6 @@ func Test_Validate(t *testing.T) {
...
 			},
 		},
-		{
-			Name:          "List Command with ambiguous args",
-			Input:         []string{"secretStores"},
-			ExpectedValid: false,
-			ConfigHolder: framework.ConfigHolder{
-				ConfigFilePath: "",
-				Config:         configWithWorkspace,
-			},
```
kachawla, line 105: What did we change to remove this scenario?

test/functional-portable/datastoresrp/noncloud/resources/simulated_environment_test.go
```diff
@@ -82,10 +83,16 @@ func Test_Deployment_SimulatedEnv_BicepRecipe(t *testing.T) {
...
+				if strings.EqualFold(*resources[0].Type, "Applications.Datastores/mongoDatabases") {
+					require.Equal(t, mongoDBName, *resources[0].Name)
+					require.Equal(t, "Applications.Core/containers", *resources[1].Type)
+					require.Equal(t, containerName, *resources[1].Name)
+				} else {
+					require.Equal(t, "Applications.Core/containers", *resources[0].Type)
+					require.Equal(t, containerName, *resources[0].Name)
+					require.Equal(t, mongoDBName, *resources[1].Name)
+					require.Equal(t, "Applications.Datastores/mongoDatabases", *resources[1].Type)
+				}
```
kachawla, line 95: This is interesting.. how was the order guaranteed before?

pkg/cli/clients/management.go
```diff
@@ -804,6 +773,76 @@ func (amc *UCPApplicationsManagementClient) GetResourceProviderSummary(ctx conte
 	return response.ResourceProviderSummary, nil
 }
 
+// ListAllResourceTypes lists all resource types in all resource providers in the configured scope.
+func (amc *UCPApplicationsManagementClient) ListAllResourceTypesNames(ctx context.Context, planeName string) ([]string, error) {
```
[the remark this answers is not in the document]
nithyatsu: added.

pkg/cli/clients/management.go
```diff
@@ -804,6 +773,76 @@ func (amc *UCPApplicationsManagementClient) GetResourceProviderSummary(ctx conte
...
+	resourceProviderSummaries, err := amc.ListResourceProviderSummaries(ctx, planeName)
+	if err != nil {
+		return []string{}, err
+	}
+
+	resourceTypeNames := []string{}
+	for _, summary := range resourceProviderSummaries {
+
+		resourceTypes := summary.ResourceTypes
```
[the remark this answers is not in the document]
nithyatsu: added if summary.Name != nil && resourceTypes != nil  checks to avoid crash, although we should not be able to have entry with nil name or types since these fields are required.

pkg/cli/clients/management_test.go
```diff
@@ -50,6 +111,19 @@ func Test_Resource(t *testing.T) {
 		}
 	}
 
+	createResourceAndRPClient := func(wrapped genericResourceClient, wrappedRP resourceProviderClient) *UCPApplicationsManagementClient {
```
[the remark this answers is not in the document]
nithyatsu: I made it createResourceAndResourceProviderClient. Not sure if thats any better. /radius/pkg/cli/clients/management_mocks.go should give an idea. WE are basically creating a client that can use  both the resource mocks and resourceprovider mocks.

pkg/cli/clients/management_test.go
```diff
@@ -50,6 +111,19 @@ func Test_Resource(t *testing.T) {
...
+		return &UCPApplicationsManagementClient{
+			RootScope: testScope,
+			genericResourceClientFactory: func(scope string, resourceType string) (genericResourceClient, error) {
+				return wrapped, nil
```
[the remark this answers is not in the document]
nithyatsu: I think it indicates a mock of one of the types in management_mocks.go that is to be "wrapped" into a UCPApplicationManagementClient instance.

pkg/cli/clients/management_test.go
```diff
@@ -162,20 +236,14 @@ func Test_Resource(t *testing.T) {
 	})
 
 	t.Run("ListResourcesInApplication", func(t *testing.T) {
-		mock := NewMockgenericResourceClient(gomock.NewController(t))
-		client := createClient(mock)
+		mockResourceClient := NewMockgenericResourceClient(gomock.NewController(t))
+		mockResourceProviderClient := NewMockresourceProviderClient(gomock.NewController(t))
```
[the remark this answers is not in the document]
nithyatsu: I dont really know. But it is tied back to mock generated file with that naming, hence I dont want to edit. Please let me know if you are ok with that. ref: pkg/cli/clients/mock_management_wrapped_clients.go

pkg/cli/clients/management_test.go
```diff
@@ -185,20 +253,15 @@ func Test_Resource(t *testing.T) {
...
-					NewListByRootScopePager(gomock.Any()).
-					Return(pager([]generated.GenericResourcesClientListByRootScopeResponse{{GenericResourcesList: generated.GenericResourcesList{NextLink: to.Ptr("0")}}}))
-			}
-		}
+		client := createResourceAndRPClient(mockResourceClient, mockResourceProviderClient)
+
+		mockResourceProviderClient.EXPECT().NewListProviderSummariesPager("local", gomock.Any()).Return(pager(resourceProviderSummaryPages))
+		mockResourceClient.EXPECT().
+			NewListByRootScopePager(gomock.Any()).
+			Return(pager(listPages)).AnyTimes()
```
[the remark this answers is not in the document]
nithyatsu: NewListByRootScopePager uses gomock.Any()  in all places in file because that parameter corresponds to options, which is not really used in any way today.

pkg/cli/cmd/resource/create/create.go
nithyatsu: No, the types are stored normalized. I was just trying keep it consistent in docs. we soecify only the type name for creation so I have updated this text.
```
nithya@MacBook-Pro radius % rad resource-provider show applications.core  
RESOURCE           TYPE      STATE
Applications.Core 
```

pkg/cli/cmd/resource/list/list.go
nithyatsu: resource types must be camel case. both lower and upper cases work for resource providers while we query, for example rad resource-provider show applications.core. But manifests and data storage are in the form Applications.Core. I will revert my texts to remove confusion.

pkg/cli/cmd/resource/delete/delete.go
nithyatsu: yes, it is not required. It will work either way.

pkg/cli/cmd/resource/delete/delete_test.go
nithyatsu: looks like all the rad resource commands will need a test for resource type and provider's existence before processing the request. Is it OK to create a ticket to address this as another task?

pkg/cli/cmd/resource/list/list.go
nithyatsu: the user enters a fully qualified name, liek applications.core/container, which is checked for correctness. But a lot of api calls need one of the provider or type. So we are having to split it again. Mostly this call is used in rad resource commands and I would like to talk a bit more about it, to discuss introducing another function RequireFullyQualifiedResourceType which can return in desired format.

test/functional-portable/datastoresrp/noncloud/resources/simulated_environment_test.go
nithyatsu: I am not sure .. but when I debuged I noticed only the order was different when it failed.

pkg/cli/clivalidation.go
```diff
@@ -296,28 +295,11 @@ func RequireResourceType(args []string) (string, error) {
...
-		if strings.EqualFold(supportedType, resourceTypeName) || strings.EqualFold(resourceType, resourceTypeName) {
-			foundTypes = append(foundTypes, resourceType)
-		}
-	}
-	if len(foundTypes) == 1 {
-		return foundTypes[0], nil
-	} else if len(foundTypes) > 1 {
-		return "", fmt.Errorf("multiple resource types match '%s'. Please specify the full resource type and try again:\n\n%s\n",
-			resourceTypeName, strings.Join(foundTypes, "\n"))
+	if !strings.Contains(resourceTypeName, "/") {
```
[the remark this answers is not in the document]
nithyatsu: yes. Rest of the check has to happen in "Run", since we cant test statically for anything other than the format rp/rt

pkg/cli/clients/clients.go
```diff
@@ -239,6 +239,9 @@ type ApplicationsManagementClient interface {
 	// DeleteResourceType deletes a resource type in the configured scope.
 	DeleteResourceType(ctx context.Context, planeName string, providerNamespace string, resourceTypeName string) (bool, error)
 
+	// ListAllResourceTypesNames lists the names of all resource types in the configured scope.
+	ListAllResourceTypesNames(ctx context.Context, planeName string) ([]string, error)
```
ytimocin, line 243: I believe that this should be named as **ListAllResourceTypeNames**.

pkg/cli/clients/management.go
```diff
@@ -74,6 +74,14 @@ var (
 	}
 )
 
+var (
+	ExcludeResourceTypesList = []string{
```
ytimocin, line 78: I believe this should be named as `ExcludedResourceTypeList`.

pkg/cli/clients/management.go
```diff
@@ -1029,3 +1068,12 @@ func (amc *UCPApplicationsManagementClient) captureResponse(ctx context.Context,
...
 }
+
+func inStringSlice(s string, slice []string) bool {
+	for _, v := range slice {
+		if strings.EqualFold(s, v) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
ytimocin: We may not need this function at all if we use the function mentioned above.

pkg/cli/cmd/resource/list/list_test.go
nithyatsu: the validation cannot check for whether the type is valid in Validation anymore, since it is dynamically tested.

pkg/cli/cmd/resource/delete/delete.go
nithyatsu: I have reverted the lowercasing since it has caused so much confusion.

pkg/cli/cmd/resource/list/list.go
lakshmimsft: should this be in Validate() ??

pkg/cli/clients/management.go
```diff
@@ -685,7 +654,7 @@ func (amc *UCPApplicationsManagementClient) DeleteResourceGroup(ctx context.Cont
 	return response.StatusCode != 204, nil
 }
 
-// ListResourceProviders lists all resource providers in the configured scope.
+// ListResourceProviders lists all resource providers in the configured plane.
```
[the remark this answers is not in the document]
kachawla: @nithyatsu got it, thanks for clarifying. Do you want to update it for rest of the functions you are updating as well (`ListAllResourceTypesNames`, `ListResourcesInApplication`, `ListResourcesInEnvironment`)?

pkg/cli/clients/management.go
```diff
@@ -804,6 +753,75 @@ func (amc *UCPApplicationsManagementClient) GetResourceProviderSummary(ctx conte
...
+	resourceProviderSummaries, err := amc.ListResourceProviderSummaries(ctx, planeName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list resource provider summaries: %v", err)
+	}
+	resourceTypeNames := []string{}
+	for _, resourceProvider := range resourceProviderSummaries {
+		resourceProviderName := *resourceProvider.Name
+		for typeName, _ := range resourceProvider.ResourceTypes {
+			fullResourceName := resourceProviderName + "/" + typeName
+			if !slices.Contains(ExcludedResourceTypesList, fullResourceName) {
```
kachawla, line 767: `slice.Contains` does a case sensitive check, which we don't enforce in the value passed for `resourcetype` flag through cli today. Are we intentionally changing this behavior going forward?

pkg/cli/clients/management.go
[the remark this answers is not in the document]
kachawla: I see. If this list is only intended for resources that shouldn't be listed for an application, then I think the implementation is doing more than what's intended. Correct me if this is not the case, but wouldn't it exclude application from env resources when `[ListResourcesInEnvironment](https://github.com/radius-project/radius/blob/16b05893206c8971c2a3a50d04617f89c9411bb8/pkg/cli/clients/management.go#L802)` is used?

pkg/cli/clivalidation.go
```diff
@@ -270,22 +269,22 @@ func RequireResource(cmd *cobra.Command, args []string) (resourceType string, re
 	return results[0], results[1], nil
 }
 
-// RequireResourceTypeAndName checks if the provided arguments contain a resource type and name, and returns them if they
+// RequireFullyQualifiedResourceTypeAndName checks if the provided arguments contain a resource type and name, and returns them if they
 // are present. If either is missing, an error is returned.
-func RequireResourceTypeAndName(args []string) (string, string, error) {
+func RequireFullyQualifiedResourceTypeAndName(args []string) (string, string, string, error) {
```
kachawla, line 274: This will change user experience for all `rad resource` commands for built in types. Have we reviewed this user facing change?

pkg/cli/clivalidation.go
```diff
@@ -295,29 +294,31 @@ func RequireResourceType(args []string) (string, error) {
...
-		supportedType := strings.Split(resourceType, "/")[1]
-		supportedTypes = append(supportedTypes, supportedType)
-		//check to see if the resource type is the correct short or long name.
-		if strings.EqualFold(supportedType, resourceTypeName) || strings.EqualFold(resourceType, resourceTypeName) {
-			foundTypes = append(foundTypes, resourceType)
-		}
+	return resourceTypeName, nil
+}
+
+func RequireFullyQualifiedResourceType(args []string) (string, string, error) {
```
kachawla, line 304: Let's add a function level comment.

pkg/cli/clivalidation.go
```diff
@@ -270,22 +269,22 @@ func RequireResource(cmd *cobra.Command, args []string) (resourceType string, re
...
-// RequireResourceTypeAndName checks if the provided arguments contain a resource type and name, and returns them if they
+// RequireFullyQualifiedResourceTypeAndName checks if the provided arguments contain a resource type and name, and returns them if they
 // are present. If either is missing, an error is returned.
-func RequireResourceTypeAndName(args []string) (string, string, error) {
+func RequireFullyQualifiedResourceTypeAndName(args []string) (string, string, string, error) {
 	if len(args) < 2 {
-		return "", "", errors.New("no resource type or name provided")
+		return "", "", "", errors.New("no resource type or name provided")
 	}
-	resourceType, err := RequireResourceType(args)
```
kachawla, line 279: Should [RequireResourceType](https://github.com/radius-project/radius/blob/16b05893206c8971c2a3a50d04617f89c9411bb8/pkg/cli/clivalidation.go#L290) be deleted if it is no longer used?

pkg/cli/cmd/resource/delete/delete_test.go
kachawla: I meant adding a unit test for the new condition that this PR is introducing - failure if `container` is specified instead of `Applications.Core/container` for example. If you mean creating another task for functional test then yes that makes sense. But unit tests should be added as a part of this PR since we are changing the expected behavior.

pkg/cli/cmd/resource/list/list.go
```diff
@@ -145,6 +148,11 @@ func (r *Runner) Run(ctx context.Context) error {
 
 	var resourceList []generated.GenericResource
 
+	_, err = common.GetResourceTypeDetails(ctx, r.ResourceProviderNameSpace, r.ResourceTypeSuffix, client)
+	if err != nil {
+		return err
+	}
```
kachawla, line 154: What is the purpose this extra Get call?

pkg/cli/cmd/resource/list/list_test.go
kachawla: ah so this test was testing for unsupported resource types?

pkg/cli/clients/management.go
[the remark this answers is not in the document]
kachawla: @ytimocin @nithyatsu The main difference though is that slices.Contains does case sensitive check. Until now we didn't enforce case sensitivity on resource types. I'm not sure if it is a good idea to start enforcing it now.

pkg/cli/clients/management.go
superbeeny: Do we want to do something along the lines of a `toLower` to remove the case sensitivity, tbh it's one of my pet peeves... especially with url context paths

pkg/cli/clients/management_test.go
```diff
@@ -39,6 +39,81 @@ const (
...
+var (
+	resourceProviderSummaryPages = []ucp.ResourceProvidersClientListProviderSummariesResponse{
+		{
+			PagedResourceProviderSummary: ucp.PagedResourceProviderSummary{
+				Value: []*ucp.ResourceProviderSummary{
+					{
+						Name: to.Ptr("Applications.Test1"),
+						ResourceTypes: map[string]*ucp.ResourceProviderSummaryResourceType{
+							"resourceType1": {
+								APIVersions: map[string]map[string]any{
```
superbeeny, line 51: We're repeating `2025-01-01` in a few places, does it make sense to extract this into a constant?

pkg/cli/clients/management.go
nithyatsu: Missed that about slices.Contains. Adding toLower.

pkg/cli/clivalidation.go
```diff
@@ -295,29 +294,31 @@ func RequireResourceType(args []string) (string, error) {
 
 	resourceTypeName := args[0]
 
-	// Allow any fully-qualified resource type.
 	if strings.Contains(resourceTypeName, "/") {
-		return resourceTypeName, nil
+		return "", fmt.Errorf("`%s` is not a valid resource type name. Please specify the resource type name. ex: `containers`", resourceTypeName)
```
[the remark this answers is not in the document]
nithyatsu: Only for resource-type create command, we use the RequireResourceType funtion and expect only the resource type, not fqdn, as specified in this ticket: https://dev.azure.com/azure-octo/Incubations/_workitems/edit/13729 .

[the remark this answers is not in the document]
kachawla: Ah I missed that. Could be a bit confusing, but let's revisit that later. For now, can you please update the function level comment to reflect what the valid resource type is for the scope of the validation in this function? https://github.com/radius-project/radius/blob/16b05893206c8971c2a3a50d04617f89c9411bb8/pkg/cli/clivalidation.go#L286

pkg/cli/clivalidation.go
nithyatsu: WE have it used in just the resource-type create command.

pkg/cli/cmd/resource/list/list.go
nithyatsu: This is to check if the resourceprovider/ resourcetype exists. before creating a resource, since we cannot now check against a static list.

pkg/cli/cmd/resource/list/list_test.go
nithyatsu: It was for ambigious case since we have secretStore in both Applications.Core and Applications.Dapr. Earlier, we could just specify "container" in most comands and it would infer that this is applications.core/container. This test ensures is user said secretStores, we call out the ambiguity since its provided by 2 providers. But now since we enforce the fully qualified name, we wont run into this scenario.

pkg/cli/clivalidation.go
Reshrahim: Me and Nithya had a discussion about this. I think this should be a fully qualified name across all commands as there could be types with same name under different namespaces/resource providers. ` rad resource create` already does this.

nithyatsu: confirmed with @Reshrahim about the UX changes.

pkg/cli/clients/management_test.go
nithyatsu: added const.

pkg/cli/clients/management.go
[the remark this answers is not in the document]
nithyatsu: I see the usage in multiple places, and now I am a bit confused. I updated ListAllResourceTypesNames, CreateOrUpdateResourceType and DeleteResourceType. ListResourcesInApplication, ListResourcesInEnvironment would work on a resource group, so I Thinking perhaps its OK to leave it as it is..

pkg/cli/cmd/resource/list/list_test.go
kachawla: makes sense, thanks for clarifying.

pkg/cli/clients/management.go
```diff
@@ -57,20 +52,13 @@ type UCPApplicationsManagementClient struct {
...
-		dapr_ctrl.DaprSecretStoresResourceType,
-		dapr_ctrl.DaprPubSubBrokersResourceType,
-		dapr_ctrl.DaprConfigurationStoresResourceType,
-		ext_ctrl.ResourceTypeName,
-		gtwy_ctrl.ResourceTypeName,
-		cntr_ctrl.ResourceTypeName,
-		sstr_ctrl.ResourceTypeName,
+	ExcludedResourceTypesList = []string{
+		"microsoft.resources/deployments",
+		"ppplications.core/applications",
```
kachawla, line 60:
```suggestion
		"applications.core/applications",
```

pkg/cli/clients/management.go
```diff
@@ -57,20 +52,13 @@ type UCPApplicationsManagementClient struct {
...
-		ds_ctrl.SqlDatabasesResourceType,
-		dapr_ctrl.DaprStateStoresResourceType,
-		dapr_ctrl.DaprSecretStoresResourceType,
-		dapr_ctrl.DaprPubSubBrokersResourceType,
-		dapr_ctrl.DaprConfigurationStoresResourceType,
-		ext_ctrl.ResourceTypeName,
-		gtwy_ctrl.ResourceTypeName,
-		cntr_ctrl.ResourceTypeName,
-		sstr_ctrl.ResourceTypeName,
+	ExcludedResourceTypesList = []string{
```
kachawla, line 58: Can you please call out in the comment that any future resource added here should be lowercase to support the way we are doing the string comparison?

pkg/cli/cmd/resource/list/list.go
kachawla, line 154: Given that the purpose of this call is to ensure that the resource type exists, should we return different explicit error message for 404 vs rest of the errors?

pkg/cli/clivalidation.go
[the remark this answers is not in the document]
nithyatsu: I spoke to @Reshrahim  about the changes. I have made a note to sync again for next release to make sure this is called out.

pkg/cli/clivalidation.go
```diff
@@ -296,28 +295,11 @@ func RequireResourceType(args []string) (string, error) {
...
-			foundTypes = append(foundTypes, resourceType)
-		}
-	}
-	if len(foundTypes) == 1 {
-		return foundTypes[0], nil
-	} else if len(foundTypes) > 1 {
-		return "", fmt.Errorf("multiple resource types match '%s'. Please specify the full resource type and try again:\n\n%s\n",
-			resourceTypeName, strings.Join(foundTypes, "\n"))
+	if !strings.Contains(resourceTypeName, "/") {
+		return "", fmt.Errorf("'%s' is not a valid resource type. Please specify the fully qualified resource type in format `resource-provider/resource-type` and try again", resourceTypeName)
```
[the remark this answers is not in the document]
nithyatsu: The check for existence of resource provider/resource type has been moved to Run method to make sure we are operating on existing resource types.

pkg/cli/cmd/resource/delete/delete_test.go
nithyatsu: Ah ok. Will add this.

pkg/cli/clivalidation.go
```diff
@@ -296,28 +295,11 @@ func RequireResourceType(args []string) (string, error) {
...
-		//check to see if the resource type is the correct short or long name.
-		if strings.EqualFold(supportedType, resourceTypeName) || strings.EqualFold(resourceType, resourceTypeName) {
-			foundTypes = append(foundTypes, resourceType)
-		}
-	}
-	if len(foundTypes) == 1 {
-		return foundTypes[0], nil
-	} else if len(foundTypes) > 1 {
-		return "", fmt.Errorf("multiple resource types match '%s'. Please specify the full resource type and try again:\n\n%s\n",
-			resourceTypeName, strings.Join(foundTypes, "\n"))
```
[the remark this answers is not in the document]
nithyatsu: I think I misunderstood the question. In cli, each command has a validate and a run method. The original thinking from Ryan's time  was to perform static validations in validate. The checks that would need api cals would be in run. This is why I did not introduce that check in clivalidation.

pkg/cli/clients/management.go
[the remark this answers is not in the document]
nithyatsu: Yes. But I think it would match current behavior, since the resource type list we used in old code did not have applications.
```
 ResourceTypesList = []string{
		ds_ctrl.MongoDatabasesResourceType,
		msg_ctrl.RabbitMQQueuesResourceType,
		ds_ctrl.RedisCachesResourceType,
		ds_ctrl.SqlDatabasesResourceType,
		dapr_ctrl.DaprStateStoresResourceType,
		dapr_ctrl.DaprSecretStoresResourceType,
		dapr_ctrl.DaprPubSubBrokersResourceType,
		dapr_ctrl.DaprConfigurationStoresResourceType,
		ext_ctrl.ResourceTypeName,
		gtwy_ctrl.ResourceTypeName,
		cntr_ctrl.ResourceTypeName,
		sstr_ctrl.ResourceTypeName,
	}
)
```

pkg/cli/cmd/resource/list/list.go
nithyatsu: In this case, we use common package, which in turn uses the applicationmanagementclient. So we get a cli error for 404 if provider or type is not found. Therefore we dont have to make it a clierror in this block. 
```
func GetResourceTypeDetails(ctx context.Context, resourceProviderName string, resourceTypeName string, client clients.ApplicationsManagementClient) (ResourceType, error) {
	resourceProvider, err := client.GetResourceProviderSummary(ctx, "local", resourceProviderName)
	if clients.Is404Error(err) {
		return ResourceType{}, clierrors.Message("The resource provider %q was not found or has been deleted.", resourceProviderName)
	} else if err != nil {
		return ResourceType{}, err
	}
:
if idx < 0 {
		return ResourceType{}, clierrors.Message("Resource type %q not found in resource provider %q.", resourceTypeName, *resourceProvider.Name)
	}
```

pkg/cli/clients/management.go
nithyatsu: added comment on lowercase.

kachawla (dismissed):

pkg/cli/clients/management.go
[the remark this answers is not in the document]
kachawla: Ahh! I see now how the current implementation is explicitly calling doing another list for application resource  https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/env/delete/delete.go#L158.. weird, but not in the scope of this PR to change. Thanks for sharing the details.

kachawla (approved):

PR #8298, opened 2025-01-24
Update long running test workflow to verify manifests are registered

lakshmimsft:
# Description

Apply same verification check as was done for cloud and non-cloud tests to check for registered manifests in long running test workflow.

## Type of change

## Contributor checklist

    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -404,6 +404,50 @@ jobs:
           rad install kubernetes --reinstall \
             --chart ${{ env.RADIUS_CHART_LOCATION }} \
             --set rp.image=${{ env.CONTAINER_REGISTRY }}/applications-rp,rp.tag=${{ env.REL_VERSION }},dynamicrp.image=${{ env.CONTAINER_REGISTRY }}/dynamic-rp,dynamicrp.tag=${{ env.REL_VERSION }},controller.image=${{ env.CONTAINER_REGISTRY }}/controller,controller.tag=${{ env.REL_VERSION }},ucp.image=${{ env.CONTAINER_REGISTRY }}/ucpd,ucp.tag=${{ env.REL_VERSION }}
+      
+          echo "*** Verify manifests are registered ***"
+          rm -f registermanifest_logs.txt
+          # Find the pod with container "ucp"
+          POD_NAME=$(
+            kubectl get pods -n radius-system \
+              -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.spec.containers[*].name}{"\n"}{end}' \
+            | grep "ucp" \
+            | head -n1 \
+            | cut -d" " -f1
+          )
+          echo "Found ucp pod: $POD_NAME"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/long-running-azure.yaml
```diff
@@ -404,6 +404,50 @@ jobs:
           rad install kubernetes --reinstall \
             --chart ${{ env.RADIUS_CHART_LOCATION }} \
             --set rp.image=${{ env.CONTAINER_REGISTRY }}/applications-rp,rp.tag=${{ env.REL_VERSION }},dynamicrp.image=${{ env.CONTAINER_REGISTRY }}/dynamic-rp,dynamicrp.tag=${{ env.REL_VERSION }},controller.image=${{ env.CONTAINER_REGISTRY }}/controller,controller.tag=${{ env.REL_VERSION }},ucp.image=${{ env.CONTAINER_REGISTRY }}/ucpd,ucp.tag=${{ env.REL_VERSION }}
+      
+          echo "*** Verify manifests are registered ***"
```
ytimocin, line 408: What I am trying to understand is why we need this step. I will try to explain below:

- Adding initializer as a service here: https://github.com/radius-project/radius/blob/main/pkg/ucp/server/server.go#L49
- Running the host with all the services needed: https://github.com/radius-project/radius/blob/main/cmd/ucpd/cmd/root.go#L72
- If any one of the services has an error, shut down for UCP is initiated: https://github.com/radius-project/radius/blob/main/pkg/components/hosting/run.go#L46

If this is the case why do we need to check all the logs of UCP to see if there is an error with the initializer. We are specifically looking for this error: **"Service initializer terminated with error"**. If that is the case, that should mean that one of the services that UCP depends on was terminated. That should trigger the termination of UCP.

Please help me understand why this wouldn't make the UCP pod fail when doing the `rad install`.

- `rad install` is run
- if one of the services of UCP fails, then UCP will fail to get to READY state

Is this wrong?

ytimocin: Are we running an extra command after initializing the `initializer` service?

lakshmimsft: The initializer service waits for UCP server to be up and calls apis to add data for the manifests. 
wait for ucp: https://github.com/radius-project/radius/blob/de991734859fa9568fd99e1d4ccf649fe225e086/pkg/ucp/initializer/service.go#L110
The pod up, ucp is receiving requests to add these manifests. 
The error we're seeing is we're trying to add groups/resources before these manifests have completed registering.

ytimocin (approved):

kachawla:
Sample tests are failing, does this need to be added to test samples workflow as well?

PR #8317, opened 2025-01-26
Update to skip cleanup of built-in resources in Long Running Tests

lakshmimsft:
# Description

Update to skip cleanup of built-in resources in Long Running Tests. 
https://github.com/radius-project/radius/issues/8357

## Type of change

## Contributor checklist

    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -404,8 +404,9 @@ jobs:
           rad install kubernetes --reinstall \
             --chart ${{ env.RADIUS_CHART_LOCATION }} \
             --set rp.image=${{ env.CONTAINER_REGISTRY }}/applications-rp,rp.tag=${{ env.REL_VERSION }},dynamicrp.image=${{ env.CONTAINER_REGISTRY }}/dynamic-rp,dynamicrp.tag=${{ env.REL_VERSION }},controller.image=${{ env.CONTAINER_REGISTRY }}/controller,controller.tag=${{ env.REL_VERSION }},ucp.image=${{ env.CONTAINER_REGISTRY }}/ucpd,ucp.tag=${{ env.REL_VERSION }}
-      
-          echo "*** Verify manifests are registered ***"
+      - name: Verify manifests are registered
+        if: env.SKIP_BUILD != 'true'
+        run: |
           rm -f registermanifest_logs.txt
           # Find the pod with container "ucp"
           POD_NAME=$(
@@ -448,6 +449,16 @@ jobs:
             echo "Manifests not registered after 3 minutes."
             exit 1
           fi
+      - name: Create a list of resources not to be deleted
+        if: env.SKIP_BUILD != 'true'
+        run: |
+          kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" > skip-delete-resources-list.txt
+      - name: Save list of resources not to be deleted
+        if: env.SKIP_BUILD != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: skip-delete-resources-list.txt
+          key: skip-delete-resources-list-file
       - name: Configure Radius test workspace
         run: |
           set -x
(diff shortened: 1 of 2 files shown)


.github/workflows/long-running-azure.yaml
```diff
@@ -553,10 +565,23 @@ jobs:
             --subscription ${{ secrets.AZURE_SUBSCRIPTIONID_TESTS }} \
             --name ${{ env.AZURE_TEST_RESOURCE_GROUP }} \
             --yes --verbose
+      - name: Restore skip-resource-list
+        if: always()
+        uses: actions/cache/restore@v4
+        with:
+          path: skip-resource-list.txt
+          key: skip-resource-list-file      π
```
ytimocin, line 573:
```suggestion
          key: skip-resource-list-file
```

.github/workflows/long-running-azure.yaml
```diff
@@ -448,6 +449,17 @@ jobs:
             echo "Manifests not registered after 3 minutes."
             exit 1
           fi
+      - name: Generate skip-resource-list
+        if: env.SKIP_BUILD != 'true'
+        run: |
+          echo "Generating skip-resource-list..."
+          kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" > skip-resource-list.txt
```
ytimocin, line 456: This is not only going to make us skip the manifest-related resources. I think it is going to make us skip the deletion of a lot of other Radius resources too, right?

.github/workflows/long-running-azure.yaml
```diff
@@ -448,6 +449,17 @@ jobs:
             echo "Manifests not registered after 3 minutes."
             exit 1
           fi
+      - name: Generate skip-resource-list
```
ytimocin, line 452: I think we could use a better naming here like: **Create a list of resources not to be deleted**. I know it is long but what exactly is the **skip-resource-list**? What are we skipping the resources from?

.github/workflows/long-running-azure.yaml
```diff
@@ -448,6 +449,17 @@ jobs:
...
+        if: env.SKIP_BUILD != 'true'
+        run: |
+          echo "Generating skip-resource-list..."
+          kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" > skip-resource-list.txt
+      - name: Save skip-resource-list
```
ytimocin, line 457: I believe this also needs a better naming for future contributors.

.github/workflows/long-running-azure.yaml
```diff
@@ -553,10 +565,23 @@ jobs:
...
+          path: skip-resource-list.txt
+          key: skip-resource-list-file      π
       - name: Clean up cluster
         if: always()
         timeout-minutes: 60
-        run: ./.github/scripts/cleanup-long-running-cluster.sh
+        run: |
+          # Report if skip-resource-list.txt is missing or empty
+          if [ ! -s skip-resource-list.txt ]; then
+            echo "skip-resource-list.txt does not exist or is empty. Proceeding without skip list"
```
ytimocin, line 580:
```suggestion
            echo "skip-resource-list.txt does not exist or is empty. Proceeding without skip list..."
```

ytimocin (changes requested):
I believe we could use. better naming and we should add a note of why we need some resources to not be deleted.

.github/workflows/long-running-azure.yaml
lakshmimsft: skip-resources-list.txt is generated ONLY when new build is deployed. it will have the only necessary resources.

.github/workflows/long-running-azure.yaml
lakshmimsft: updated

.github/workflows/long-running-azure.yaml
ytimocin: It is going to get all the resources of type `resources.ucp.dev` right? Would that include resources other than resource providers?

lakshmimsft: [skip-delete-resource-list.txt](https://github.com/user-attachments/files/18562231/skip-delete-resource-list.txt)
The entries for these resource-providers include entries for all resourcetypes and their associated entries including, apiversions supported, locations and their operation statuses. attaching an example from local build. All these entries are needed to be preserved across runs before the next build/deploy.

ytimocin (approved):
Let's test this out but `kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" > skip-delete-resources-list.txt` will get all Radius resources to the `don't delete` list. Correct me if I am wrong.

PR #8391, opened 2025-01-31
remove validation of resource types against the static list in recipes codebase

nithyatsu:
# Description

Since we now have APIs that lets us query the valid resourcetypes and the resource types can include UDTs, we refactored the code to remove the validation of resource types against the static list of redaius resource types.

Also, After this PR, we covered the tests for these tasks. 

- Manually test registration of Recipe for a nonexistent resource type
- Manually test Recipe webhook for a nonexistent resource type. Operation/processing might fail but creation should happen

## Type of change

Fixes: https://dev.azure.com/azure-octo/Incubations/_workitems/edit/13653/

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable


pkg/corerp/backend/controller/createorupdateresource.go
```diff
@@ -60,16 +65,66 @@ func getDataModel(id resources.ID) (v1.DataModelInterface, error) {
 	}
 }
 
+func isPortableResource(resourceTypeResourceObj *database.Object) (bool, error) {
```
kachawla, line 68: Couple of callouts/questions about this approach -

1. Will this work for built in portable types in their current state before they are migrated to the UDT schema (with capabilites/supportsRecipe property)?
2. Is the output of this function only used for currently implemented built in portable resources or for UDT resources as well?
3. If this is expected to work for UDTs as well, then checking if recipes are supported isn't reliable since user can opt out of recipes.

pkg/corerp/backend/controller/createorupdateresource.go
```diff
@@ -60,16 +65,66 @@ func getDataModel(id resources.ID) (v1.DataModelInterface, error) {
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 // Run checks if the resource exists, renders the resource, deploys the resource, applies the
 // deployment output to the resource, deletes any resources that are no longer needed, and saves the resource.
 func (c *CreateOrUpdateResource) Run(ctx context.Context, request *ctrl.Request) (ctrl.Result, error) {
+	// This code is general and we might be processing an async job for a resource or a scope, so using the general Parse function.
```
kachawla, line 98: I think there is room for abstraction/refactoring both in this controller and deploymentprocessor. The code is getting too verbose/difficult to read.

kachawla (commented):
Could you please update the PR description to include manual testing that's mentioned in the linked issue to validate this change?

```
- Manually test registration of Recipe for a nonexistent resource type
- Manually test Recipe webhook for a nonexistent resource type. Operation/processing might fail but creation should happen
```

[Discussed this offline, there isn't much that can be tested here since the behavior currently is to allow registeration of recipe for a nonexistent resoruce type. So as long as existing funcitonal tests pass, this should be good.]

pkg/corerp/backend/controller/createorupdateresource.go
lakshmimsft: Note: portable types are currently registered with the UDT schema and with supportsRecipies capability.

lakshmimsft, line 68: Question: So, we see this function returns resources that supportRecipes. Should we rename the function to reflect that??
Also where is this being used. 1. We use it in recipe_webhook.go to confirm if the resource runs with recipes 2. we refer to it to see if the resource can have AppId = "".
I don't know is the term isPortableResource is conveying the right meaning.

kachawla: @lakshmimsft aren't we still relying on typespec to generate schema for portable resources: https://github.com/radius-project/radius/blob/main/typespec/Applications.Datastores/mongoDatabases.tsp? UDT based spec for portable resources is minimal, which I'm assuming was done purely for routing?

nithyatsu: I thought portable resources are portable because of recipes and hence the function name. The name existed earlier too, but was a hardcoded listof built in types that supported recipes. I agree its a good idea to discuss. The idea of AppID= "" is because it is an environment resource (again prod env can have its recipe of mongo, test env has its recipe of mongo ). The logic and naming were already existing. The aspect I changed was to remove hardcoding of "portable" resource list and detect it based on manifest.

pkg/corerp/backend/deployment/deploymentprocessor_test.go
```diff
@@ -299,80 +299,6 @@ func Test_Render(t *testing.T) {
...
-	t.Run("verify render success", func(t *testing.T) {
-		mocks := setup(t)
-		dp := deploymentProcessor{mocks.model, mocks.databaseClient, nil, nil}
-
-		testResource := getTestResource()
-		testRendererOutput := getTestRendererOutput()
-		resourceID := getTestResourceID(testResource.ID)
-
-		depId1, _ := resources.ParseResource("/subscriptions/test-subscription/resourceGroups/test-resource-group/providers/Applications.Datastores/mongoDatabases/test-mongo")
-		requiredResources := []resources.ID{depId1}
```
nithyatsu, line 311: I removed this test case, since it used a mongo db for example. "verify render success lowercase resourcetype" and "verify render success uppercase resourcetype" capture render success using a container.

lakshmimsft: adding notes: the test works if we update ln 358 and update the mongo resource to include an ApplicationID. It makes sense overall to delete the test, however, since a mongo resource request will not go through this code path.

lakshmimsft (dismissed):

pkg/corerp/backend/deployment/deploymentprocessor.go
```diff
@@ -575,18 +574,13 @@ func (dp *deploymentProcessor) getResourceDataByID(ctx context.Context, resource
...
-		}
-		appID = &parsedID
-	} else if rp_pr.IsValidPortableResourceType(resourceID.TypeSegments()[0].Type) {
-		// Application id is optional for portable resource types
-		appID = nil
-	} else {
-		return ResourceData{}, fmt.Errorf("missing required application id for the resource %q", resourceID.String())
+	// This code path is used only by core application resources.
+	// Application ID is a required property for these resources and cannot be empty.
+	parsedID, err := resources.ParseResource(applicationID)
```
kachawla, line 579: I'm a bit confused.. `buildResourceDependency` is used for non-core resources as well: https://github.com/radius-project/radius/blob/0ac499f38d53497abe20a24602accfef7d33fabd/pkg/corerp/backend/deployment/deploymentprocessor.go#L516-L569

The function that we discussed is only called for core resources is `getApplicationAndEnvironmentForResourceID`
https://github.com/radius-project/radius/blob/0ac499f38d53497abe20a24602accfef7d33fabd/pkg/corerp/backend/deployment/deploymentprocessor.go#L216

pkg/corerp/backend/deployment/deploymentprocessor_test.go
```diff
@@ -299,80 +299,6 @@ func Test_Render(t *testing.T) {
...
 
-	t.Run("verify render success", func(t *testing.T) {
-		mocks := setup(t)
-		dp := deploymentProcessor{mocks.model, mocks.databaseClient, nil, nil}
-
-		testResource := getTestResource()
-		testRendererOutput := getTestRendererOutput()
-		resourceID := getTestResourceID(testResource.ID)
-
-		depId1, _ := resources.ParseResource("/subscriptions/test-subscription/resourceGroups/test-resource-group/providers/Applications.Datastores/mongoDatabases/test-mongo")
```
kachawla, line 310: Mongo resource is a dependency here, so this should continue to be a valid scenario

pkg/corerp/backend/deployment/deploymentprocessor_test.go
```diff
@@ -584,27 +584,6 @@ func Test_Render(t *testing.T) {
 		require.Equal(t, "application ID \"invalid-app-id\" for the resource \"/subscriptions/test-subscription/resourceGroups/test-resource-group/providers/Applications.Core/containers/test-resource\" is not a valid id. Error: 'invalid-app-id' is not a valid resource id", err.(*v1.ErrClientRP).Message)
 	})
```
nithyatsu, line 586: This test actually captures the previous logic, where we had a check for application ID for the resource that is getting rendered. I have removed this.

pkg/corerp/backend/deployment/deploymentprocessor.go
```diff
@@ -575,17 +574,16 @@ func (dp *deploymentProcessor) getResourceDataByID(ctx context.Context, resource
 
 func (dp *deploymentProcessor) buildResourceDependency(resourceID resources.ID, applicationID string, resource v1.DataModelInterface, outputResources []rpv1.OutputResource, computedValues map[string]any, secretValues map[string]rpv1.SecretValueReference, recipeData portableresources.RecipeData) (ResourceData, error) {
 	var appID *resources.ID
+	// Application id is mandatory for core resource types and is a required field.
```
kachawla, line 577: minor update since some resources under Applications.Core namespace do not require application id like SecretStore and Extender.

```suggestion
	// Application id is mandatory for some of the core resource types and is a required field.
```

pkg/corerp/backend/deployment/deploymentprocessor_test.go
kachawla: Do we want to keep this test with the updated error it would return from https://github.com/radius-project/radius/blob/main/pkg/corerp/backend/deployment/deploymentprocessor.go#L228? Just to prevent against any unintended changes in downstream code.

nithyatsu: added the change we discussed

kachawla (approved):

PR #8818, opened 2025-03-13
Update Terraform Postgres Recipe Test

### LINKED ISSUE #8786: Scheduled long running test failed - Run ID: 13773700990
## Bug information 

This issue is automatically generated if the scheduled long running test fails. The Radius long running test operates on a schedule of every 2 hours everyday. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](https://github.com/radius-project/radius/actions/runs/13773700990).

[AB#14917](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/14917)

kachawla:
# Description

* <s>Update resource name to reset generated TF state name in long running tests.</s> - Updated in https://github.com/radius-project/radius/pull/8835

* <s>Remove skipping of resource deletion, there is no reason why we shouldn't be deleting resources for this test.</s> - Updated in https://github.com/radius-project/radius/pull/8821 instead along with server side bug fix.

* Remove manual verification and deletion of Kubernetes secret that stores the TF state file. This is only needed in one Terraform test to ensure that the delete can go through if the state is deleted outside of Radius, which is already covered in another test.

## Type of change

Attempts to fix: https://github.com/radius-project/radius/issues/8786

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable

### DIFF
--- a/test/functional-portable/corerp/noncloud/resources/recipe_terraform_test.go
+++ b/test/functional-portable/corerp/noncloud/resources/recipe_terraform_test.go
@@ -139,11 +139,14 @@ func Test_TerraformRecipe_KubernetesRedis(t *testing.T) {
 //
 // - Create an extender resource using a Terraform recipe that deploys Postgres on Kubernetes.
 // - The recipe deployment creates a Kubernetes deployment and a Kubernetes service and a postgres db.
+//
+// The purpose of this test is to validate that the Terraform provider config referencing secret store
+// resource is passed correctly during recipe deployment.
 func Test_TerraformRecipe_KubernetesPostgres(t *testing.T) {
 	template := "testdata/corerp-resources-terraform-postgres.bicep"
 	appName := "corerp-resources-terraform-pg-app"
 	envName := "corerp-resources-terraform-pg-env"
-	extenderName := "pgs-resources-terraform-pgsapp"
+	extenderName := "corerp-resources-terraform-pg"
 	secretName := "pgs-secretstore"
 	secretResourceName := appName + "/" + secretName
 	userName := "postgres"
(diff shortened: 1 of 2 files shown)


test/functional-portable/corerp/noncloud/resources/recipe_terraform_test.go
```diff
@@ -196,32 +199,6 @@ func Test_TerraformRecipe_KubernetesPostgres(t *testing.T) {
 				},
 			},
 			SkipObjectValidation: true,
-			SkipResourceDeletion: true,
```
kachawla, line 199: Skipping this was hiding a bug on the delete path. Handling that here now https://github.com/radius-project/radius/pull/8821

kachawla (commented):
Skipping resource deletion in this test was causing a bug to go unnoticed, handling it here now: https://github.com/radius-project/radius/pull/8821

test/functional-portable/corerp/noncloud/resources/recipe_terraform_test.go
nicolejms: Since you're removing the verification step, is there a 'default' verification for the tests or are we just looking for no errors/exceptions?

kachawla: This change, `SkipResourceDeletion: true`, is actually adding resource deletion, which should have been implemented from the beginning. Without this, we lacked test coverage for the delete path, and the delete functionality never worked from the start of the providers feature implementation.

The removal of the post-verify step is because it was redundant. It was likely copied from another test without realizing that it wasn't needed. The post-verify step was intended to manually delete the Kubernetes secret that stores the TF state file and validate that the recipe deletion still works (which, in fact, this test wasn’t addressing since deletion was being skipped). We already cover this scenario in this test: https://github.com/radius-project/radius/blob/main/test/functional-portable/corerp/noncloud/resources/recipe_terraform_test.go#L103. Duplicating this in all tests would be a waste of resources and prevent us from testing recipe deletion from the user's perspective, where the state file would typically exist, except in rare scenarios where it's deleted manually outside of Radius.

kachawla: For `SkipObjectValidation: true` - the test frame still by default validates that Radius resources are created. I don't have context on why this flag was set to skipped on this test, it basically skips validation of kubernetes resources. But it looks like it was merged in half baked state because it defines K8sObjects but skips it's validation. I don't want to change too many things at once and handle removing this flag after fixing rest of the problems with this test + long running tests.

kachawla:
Recipe deletion is failing now with connection issues:

```
"code": "RecipeDeletionFailed",
        	            	    "message": "terraform destroy failure: exit status 1\n\nError: Error connecting to PostgreSQL server  (scheme: postgres): dial tcp [::1]:5432: connect: connection refused\n\n  with module.defaultpostgres.postgresql_database.pg_db_test,\n  on .terraform/modules/defaultpostgres/main.tf line 77, in resource \"postgresql_database\" \"pg_db_test\":\n  77: resource postgresql_database \"pg_db_test\" {\n\n"
        	            	  }
```

Since this path has never been tested before and requires more investigation, I have separated out resource name update changes in another PR to see if that at least unblocks long running tests. https://github.com/radius-project/radius/pull/8835

kachawla:
An update here - my notes on continued investigation on this are below. TLDR; if we can replace this test with something lighter weight, that'll be much better, since we don't really need to test connection to database to test provider config.

https://github.com/radius-project/radius/issues/8786#issuecomment-2727542641
https://github.com/radius-project/radius/issues/8786#issuecomment-2731329683

I have been looking into replacing this with a recipe that creates an Azure resource group, but given that our long running tests use service principal but functional tests use workload identity, we can't really use a custom azure provider for test. I might send a PR to validate if we can use Azure provider for workload identity to achieve the purpose of this and then replace it once long running tests are migrated: https://github.com/radius-project/radius/issues/8889. Closing this PR for now, will likely keep experimenting with it.

PR #8999, opened 2025-03-28
Update Deployment Engine release process

willdavsmith:
# Description

* Remove Deployment Engine from release process

## Type of change

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable


sk593 (approved):

.github/scripts/release-get-version.sh
```diff
@@ -77,7 +77,20 @@ if [[ -z "$RELEASE_BRANCH_NAME" ]]; then
...
+# REL_CHANNEL is the release channel (e.g. 0.1, 0.1.0-rc1)
+REL_CHANNEL=""
+
+# If the release version is a release candidate (e.g. 0.1.0-rc1), use the full version as the channel
+# Otherwise, use the major and minor version (e.g. 0.1)
+if [[ "$RELEASE_VERSION" == *"-rc"* ]]; then
+  REL_CHANNEL=$RELEASE_VERSION
+else
+  REL_CHANNEL=$(echo "$RELEASE_VERSION" | cut -d '.' -f 1,2)
+fi
```
ytimocin, line 89: Can this be edge or latest by any chance? I guess my questions is if this can be anything other than 0.1.0-rc1 or 0.1.0.

ytimocin: I see down below that this is just for echoing the value, right?

.github/workflows/release.yaml
```diff
@@ -226,3 +216,30 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: rad-ci-bot
+          password: ${{ secrets.GH_RAD_CI_BOT_PAT }}
+      - name: Copy Deployment Engine images from ACR to GHCR
+        if: success() && steps.release-branch-exists.outputs.result == 'false'
+        run: |
+          TAG=${{ steps.get-version.outputs.release-channel }}
+          docker buildx imagetools create \
```
ytimocin, line 243: What does `buildx imagetools create` exactly do? The step title says `Copy DE images from ACR to GHCR` which sounds kind of like an easy move of images.

.github/scripts/release-get-version.sh
willdavsmith: this value comes directly from the `versions.yaml`, so it can't be latest. it will always be `0.1.0` or `0.1.0-rc1`. at the end we're echoing the value out to GITHUB_OUTPUT so that we can use it in the rest of the workflow

.github/workflows/release.yaml
willdavsmith: Since these images are multi-arch, we can't use plain `docker pull` and `docker push`. I found this command online, it seems like exactly what we need here

https://stackoverflow.com/questions/68567983/how-to-copy-multi-arch-docker-images-to-a-different-container-registry
https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/

Issue #8901, opened 2025-03-18
Put tool and framework versions in one place


brooke-hamilton:
### Area for Improvement

We depend on multiple tools and platforms and each one of them has a version number that must be tracked and updated in multiple places. For example, Kubernetes, Kind, k3d, Helm, and gotestsum. We have version numbers set in Make files, GitHub workflows, Bicep files, and Go source code.

### Desired behavior

This issue does not propose a specific solution. The outcome should be that for each dependency there is one place to update the version number. 

### rad Version

edge

### Would you like to support us?

- [ ] Yes, I would like to support you

brooke-hamilton:
The solution should include setting a Bicep version instead of taking latest.

DariuszPorowski:
@brooke-hamilton as we discussed, I am happy to prepare PoC with [task](https://taskfile.dev/) as a single source of truth for:
- local dev (Windows, macOS, Linux)
- GH workflows
- Dev Container / Codespaces / WSL

PR #9054, opened 2025-04-01
Fix Flux functional tests in long-running tests workflow

willdavsmith:
# Description

* Flux functional tests were not running successfully on the long-running test cluster because of setup issues. This PR adds Gitea and Flux installs to the long-running test workflow and adds a port-forwarding step in the runner to enable access to the Gitea server.

Manual LRT run: https://github.com/radius-project/radius/actions/runs/14202310846

## Type of change

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable


.github/actions/install-gitea/gitea-config.yaml
```diff
@@ -6,10 +6,3 @@ redis:
   enabled: true
 postgresql:
   enabled: true
-service:
-  http:
-    type: NodePort
-    nodePort: 30080
```
willdavsmith, line 12: we can remove these now because we're accessing the gitea server via port-forward instead of nodeport

sk593 (dismissed):

.github/workflows/long-running-azure.yaml
```diff
@@ -521,6 +535,33 @@ jobs:
         if: failure()
         run: |
           echo ":x: Failed to install Radius for functional test." >> $GITHUB_STEP_SUMMARY
+      - name: Install Flux Source Controller
+        run: |
+          curl -s https://fluxcd.io/install.sh | FLUX_VERSION=2.5.1 sudo bash
+          flux install --namespace=flux-system --version=v2.5.1 --components=source-controller --network-policy=false
```
nithyatsu, line 541: is it ok for the version here to be hardcoded?

willdavsmith: I think so - it's probably best to lock the version instead of using `latest` in case something breaks

brooke-hamilton: We should move this to an action because it is repeated in the non-cloud tests, but we can do that as part of the planned work on the long running tests. We also have other planned work to set version numbers in one place: [#8901](https://github.com/radius-project/radius/issues/8901)

brooke-hamilton (approved):
One comment about the version number being set in the action vs. the workflow, but that's a minor issue and I'm fine with others' opinions on it.

nithyatsu (approved):

PR #9197, opened 2025-04-18
Add UserAssigned Managed Identity

shivg7795:
Add support for ACI-Ngroups managed identity. 
Changes:
Update managed identity data model 
Update environment_conversion to map from API to DM and vice-versa
Update renderer to read and convert array of identity urls to a map of user-assigned managed identity
Update conversion tests to validate identity urls mapping from template to radius API model

### DIFF
--- a/pkg/to/value.go
+++ b/pkg/to/value.go
@@ -34,8 +34,6 @@ func StringArray(s []*string) []string {
 	for _, ptr := range s {
 		if ptr != nil {
 			values = append(values, *ptr)
-		} else {
-			values = append(values, "") // or handle nil however you'd like
 		}
 	}
 	return values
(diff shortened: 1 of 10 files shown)


pkg/to/value.go
```diff
@@ -28,6 +28,19 @@ func String(s *string) string {
 	return ""
 }
 
+// Map array of string pointers to an array of strings. If a value in the the array is nil then append empty string.
+func StringArray(s []*string) []string {
+	var values []string
+	for _, ptr := range s {
+		if ptr != nil {
+			values = append(values, *ptr)
+		} else {
```
sk593, line 37: This is an edge case, but I don't think we'd want to append empty strings if a value is nil. We should only be adding to the array if we get a managed identity string

sk593 (approved):
Looks great, thanks!!

PR #9219, opened 2025-04-22
Disable database control plane component by default

willdavsmith:
# Description

* Disable database control plane component by default

You can re-enable it by passing `--set database.enabled=true` in `rad install kubernetes`

## Type of change

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable

### DIFF
--- a/deploy/Chart/values.yaml
+++ b/deploy/Chart/values.yaml
@@ -114,7 +114,7 @@ dashboard:
       memory: "300Mi"
 
 database:
-  enabled: false
+  enabled: false # Enable the Postgres database install
   postgres_user: "radius"
   image: ghcr.io/radius-project/mirror/postgres
   tag: latest
--- a/deploy/Chart/values.yaml
+++ b/deploy/Chart/values.yaml
@@ -114,6 +114,7 @@ dashboard:
       memory: "300Mi"
 
 database:
+  enabled: false
   postgres_user: "radius"
   image: ghcr.io/radius-project/mirror/postgres
   tag: latest
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -469,6 +469,7 @@ jobs:
           rad install kubernetes --reinstall \
             --chart ${{ env.RADIUS_CHART_LOCATION }} \
             --set global.azureWorkloadIdentity.enabled=true \
+            --set database.enabled=true \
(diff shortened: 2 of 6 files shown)


ytimocin (commented):
LGTM

ytimocin (dismissed):

kachawla (commented):
@willdavsmith  can you add this in the `rad install kubernetes` reference docs? https://docs.radapp.io/reference/cli/rad_install_kubernetes/

deploy/Chart/values.yaml
```diff
@@ -114,6 +114,7 @@ dashboard:
       memory: "300Mi"
 
 database:
+  enabled: false
```
kachawla, line 117: Let's add a comment with some context.

kachawla (commented):
Can we add a functional test that validates the non default option (database enable true)?

kachawla (commented):
Do we need to manually clean it up from LRT cluster?

willdavsmith:
> @willdavsmith can you add this in the `rad install kubernetes` reference docs? https://docs.radapp.io/reference/cli/rad_install_kubernetes/

https://github.com/radius-project/docs/pull/1419

willdavsmith:
> Can we add a functional test that validates the non default option (database enable true)?

What would this look like? AFAIK functionality for the database isn't in yet, so the only thing to test is that the database gets installed or not, but to do this we'd have to create an entire new step in our functional tests (since the rad installs have to be separated). Based on this I'm not sure this is worth it since this `--set component.enabled=true` is a pattern we use in lots of places. What do you think?

willdavsmith:
> Do we need to manually clean it up from LRT cluster?

A `rad install kubernetes --reinstall` should uninstall it I think.

kachawla:
> > Can we add a functional test that validates the non default option (database enable true)?
> 
> What would this look like? AFAIK functionality for the database isn't in yet, so the only thing to test is that the database gets installed or not, but to do this we'd have to create an entire new step in our functional tests (since the rad installs have to be separated). Based on this I'm not sure this is worth it since this `--set component.enabled=true` is a pattern we use in lots of places. What do you think?

At this point, the goal of the test would be to make sure we don't introduce a regression on postgres installation path through other changes. Once the full postgres as datastore for Radius support is implemented, we will extend this test to validate end to end support.

kachawla:
> > > Can we add a functional test that validates the non default option (database enable true)?
> > 
> > 
> > What would this look like? AFAIK functionality for the database isn't in yet, so the only thing to test is that the database gets installed or not, but to do this we'd have to create an entire new step in our functional tests (since the rad installs have to be separated). Based on this I'm not sure this is worth it since this `--set component.enabled=true` is a pattern we use in lots of places. What do you think?
> 
> At this point, the goal of the test would be to make sure we don't introduce a regression on postgres installation path through other changes. Once the full postgres as datastore for Radius support is implemented, we will extend this test to validate end to end support.

Discussed and decided to add it in the long running workflow.

deploy/Chart/values.yaml
```diff
@@ -114,6 +114,7 @@ dashboard:
       memory: "300Mi"
 
 database:
+  enabled: false # Enable the Postgres database install
```
ytimocin, line 117: Do we need the comment?

willdavsmith: @kachawla asked me to add one here: https://github.com/radius-project/radius/pull/9219#discussion_r2055084287

ytimocin (approved):

PR #10041, opened 2025-07-22
Add validation for reserved words. (UDT)

lakshmimsft:
This pull request introduces updates to schema validation and error handling, and updates to associated tests
1. Add validation for reserved words:
`application` property must be a string
`environment` property must be a string
`environment` property must be included in schema
`connections` must be a map object
`status` cannot be used as a property
`recipe` cannot be used as a property
2. Updated to return all errors for all schemas in a document when registering types.
3. Added restriction to not allow: additionalProperties: true

## Type of change
- This pull request fixes a bug in Radius and has an approved issue (#6688 ).

Fixes: Part of #6688 

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable

### DIFF
--- a/pkg/schema/validator.go
+++ b/pkg/schema/validator.go
@@ -182,34 +182,32 @@ func (v *Validator) validateRadiusConstraints(schema *openapi3.Schema) error {
 	}
 
 	// Also validate additionalProperties if present
-	if schema.AdditionalProperties.Has != nil {
-		if addPropSchema := schema.AdditionalProperties.Schema; addPropSchema != nil {
-			if addPropSchema.Ref != "" {
-				// The $ref validation is already handled by checkRefUsage above
-			} else if addPropSchema.Value != nil {
-				if err := v.validateRadiusConstraints(addPropSchema.Value); err != nil {
-					// Add context to error
-					if valErrs, ok := err.(*ValidationErrors); ok {
-						for _, ve := range valErrs.Errors {
-							// Clone the error to avoid modifying the original
-							contextualErr := &ValidationError{
-								Type:    ve.Type,
-								Field:   "additionalProperties." + ve.Field,
-								Message: ve.Message,
-							}
-							errors.Add(contextualErr)
-						}
-					} else if valErr, ok := err.(*ValidationError); ok {
+	if addPropSchema := schema.AdditionalProperties.Schema; addPropSchema != nil {
+		if addPropSchema.Ref != "" {
+			// The $ref validation is already handled by checkRefUsage above
+		} else if addPropSchema.Value != nil {
+			if err := v.validateRadiusConstraints(addPropSchema.Value); err != nil {
+				// Add context to error
+				if valErrs, ok := err.(*ValidationErrors); ok {
+					for _, ve := range valErrs.Errors {
 						// Clone the error to avoid modifying the original
 						contextualErr := &ValidationError{
-							Type:    valErr.Type,
-							Field:   "additionalProperties." + valErr.Field,
-							Message: valErr.Message,
+							Type:    ve.Type,
+							Field:   "additionalProperties." + ve.Field,
+							Message: ve.Message,
 						}
 						errors.Add(contextualErr)
-					} else {
-						errors.Add(NewSchemaError("additionalProperties", err.Error()))
 					}
+				} else if valErr, ok := err.(*ValidationError); ok {
+					// Clone the error to avoid modifying the original
+					contextualErr := &ValidationError{
+						Type:    valErr.Type,
+						Field:   "additionalProperties." + valErr.Field,
+						Message: valErr.Message,
+					}
+					errors.Add(contextualErr)
+				} else {
+					errors.Add(NewSchemaError("additionalProperties", err.Error()))
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
 	return nil
 }
+
+// checkReservedProperties validates reserved property constraints
+func (v *Validator) checkReservedProperties(schema *openapi3.Schema) error {
+	if schema.Properties == nil {
+		return nil
+	}
+
+	var errors ValidationErrors
```
ytimocin, line 477: Just trying to understand why we have both **ValidationErrors** and **ValidationError**. Could **ValidationErrors** be an array of **ValidationError**. Not sure if this is already on main which is also okay for me.

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+
+	for propName, propRef := range schema.Properties {
+		// Check for restricted property names
+		if propName == "status" || propName == "recipe" {
+			err := NewConstraintError(propName, fmt.Sprintf("property '%s' is reserved and cannot be used", propName))
```
nithyatsu, line 482: What if the user defines status as a map[string]any conforming to radius's internal way? Should we allow that since it woudnt break anything?

nithyatsu: I think now, we cant do that since we wouldnt allow any correct?

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+			errors.Add(err)
+		}
+
+		// Check specific property type constraints
+		if propName == "application" || propName == "environment" {
+			if propRef.Value != nil {
```
nithyatsu, line 488: use constants instead of "application" "environment" "connections"

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+				}
+			}
+		}
+
+		if propName == "connections" {
+			if propRef.Value != nil {
+				// Check if it's an object type
+				if propRef.Value.Type != nil && !propRef.Value.Type.Is("object") {
+					err := NewConstraintError(propName, "property 'connections' must be a map object")
+					errors.Add(err)
```
nithyatsu, line 501: And it must contain a source as property and nothing else.  Might be easiest to just provide the snippet like below or link to testresources.yaml url for reference.  
```
connections:
              type: object
              additionalProperties:
                type: object
                properties:
                  source:
                    type: string
                    description: The resourceID of the source of the connection.
```

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+				// If it's an object, ensure it's map-like (must have additionalProperties)
+				if propRef.Value.Type != nil && propRef.Value.Type.Is("object") {
+					hasAdditionalProps := (propRef.Value.AdditionalProperties.Has != nil && *propRef.Value.AdditionalProperties.Has) ||
+						propRef.Value.AdditionalProperties.Schema != nil
+
+					if !hasAdditionalProps {
+						err := NewConstraintError(propName, "property 'connections' must be a map object (use additionalProperties)")
+						errors.Add(err)
+					}
+				}
```
nithyatsu, line 513: We have to check for. source property in the additionalProperties and no other property

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+			}
+		}
+	}
+
+	// Check that environment property is always included
+	if schema.Properties != nil {
```
nithyatsu, line 510: If Environment Property is not provided, but Application property is provided, we can get the environment from application property. I think that is the existing logic. So condition should be either appID or EnvID is provided.

pkg/schema/validator.go
```diff
@@ -356,15 +442,81 @@ func (v *Validator) checkObjectPropertyConstraints(schema *openapi3.Schema) erro
...
+		if _, hasEnv := schema.Properties["environment"]; !hasEnv {
+			err := NewConstraintError("environment", "property 'environment' must be included in schema")
```
nithyatsu: Should we say either environemnt OR application property must be included in schema and marked required? Do we validate "required" fields when we create the object?

pkg/schema/validator.go
lakshmimsft: I'm not for restriction of propertyname 'source' since this will possibly expand to other fields for different types of connections in the future. we're restricting to types allowed. let me know if we should discuss.

pkg/schema/validator.go
lakshmimsft: required fields are validated by openapi schema validation which will occur in the call. For Radius Constraints, we're making sure environment is included

pkg/schema/validator.go
lakshmimsft: pls see above comment

pkg/schema/validator.go
lakshmimsft: yes it is and it's used across packages.

pkg/schema/validator.go
lakshmimsft: updated

pkg/schema/validator.go
nithyatsu: downside is that today thats the only thing that works. But I understand your thinking that this might change very soon. I am ok with thinking about it if we recieve such a feedback. @zachcasper / @Reshrahim 's documentation on schema/ udt might be sufficient now for users to understand what to key in for conections to work.

pkg/schema/validator.go
nithyatsu: we would not "require" application right?

nithyatsu:
LGTM expect for a minor question.

nithyatsu (approved):

PR #10114, opened 2025-07-29
Adding fix to increse qps based on config

vishwahiremat:
# Description

- updating the code to use the config qps and burst when its defined in the config.
- FYI: https://github.com/radius-project/radius/pull/10031

## Type of change

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


pkg/kubeutil/config.go
```diff
@@ -88,10 +88,7 @@ func NewClientConfig(options *ConfigOptions) (*rest.Config, error) {
 	}
 
 	if options.QPS > 0.0 {
-		config.QPS = options.QPS
-	} else {
-		// Use environment variable override or default
-		qps := DefaultServerQPS
+		qps := options.QPS
```
ytimocin, line 91: Is this reverting the other changes? If so, can we do a commit revert?

vishwahiremat: no, its not reverting the changes, but fixing it, as the previous changes added for using the qps/burst changes were not getting used as they were added in the `else` block and it was never used.

ytimocin (approved):

sylvainsf (approved):
Looks great thanks for the cleaner fix!

pkg/kubeutil/config.go
lakshmimsft, line 91: Should line 92 be inside the ` if options.QPS > 0.0 { ` block? should we apply the env variable values if they exist regardless of options.QPS (etc) values > 0?

vishwahiremat: we could, but i tried to keep the functionality we had before, like we set the values only when `NewClientConfig` is called using the values.

PR #10147, opened 2025-08-01
Update TF install to shared binary approach

### LINKED ISSUE #10179: LRT/Non-Cloud Func tests: Terraform State Lock/409 Issues
We see an increase in terraform state-lock issues with messages similar to:

 "message": "terraform destroy failure: exit status 1\n\nError: Error acquiring the state lock\n\nError message: the state is already locked by another terraform client\nLock Info:\n  ID:        3d20d141-88c4-bb14-d677-88c5c9ec329a\n  Path:      \n  Operation: OperationTypeApply\n  Who:       rpuser@applications-rp-6f499d645b-vvmnl\n  Version:   1.12.2\n  Created:   2025-07-23 21:13:45.004096153 +0000 UTC\n  Info:      \n\n\nTerraform acquires a state lock to protect the state from being written\nby multiple users at the same time. Please resolve the issue above and try\nagain. For most commands, you can disable locking with the \"-lock=false\"\nflag, but this is not recommended.\n"

[AB#16708](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/16708)

lakshmimsft:
# Description
This pull request updates Terraform install to  a global shared binary approach, which alleviates race conditions and state lock errors caused by concurrent operations. Additionally, the changes include state lock timeout handling and error checking around Terraform binary availability.

## Type of change

Fixes: #10179 

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable

### DIFF
--- a/pkg/recipes/driver/terraform.go
+++ b/pkg/recipes/driver/terraform.go
@@ -98,11 +98,12 @@ func (d *terraformDriver) Execute(ctx context.Context, opts ExecuteOptions) (*re
 	}
 
 	tfState, err := d.terraformExecutor.Deploy(ctx, terraform.Options{
-		RootDir:        requestDirPath,
-		EnvConfig:      &opts.Configuration,
-		ResourceRecipe: &opts.Recipe,
-		EnvRecipe:      &opts.Definition,
-		Secrets:        opts.Secrets,
+		RootDir:          requestDirPath,
+		EnvConfig:        &opts.Configuration,
+		ResourceRecipe:   &opts.Recipe,
+		EnvRecipe:        &opts.Definition,
+		Secrets:          opts.Secrets,
+		StateLockTimeout: terraform.DefaultStateLockTimeout,
 	})
 
 	unsetError := unsetGitConfigForDirIfApplicable(secretStoreID, opts.Secrets, requestDirPath, opts.Definition.TemplatePath)
@@ -150,11 +151,12 @@ func (d *terraformDriver) Delete(ctx context.Context, opts DeleteOptions) error 
 	}
 
 	err = d.terraformExecutor.Delete(ctx, terraform.Options{
-		RootDir:        requestDirPath,
-		EnvConfig:      &opts.Configuration,
-		ResourceRecipe: &opts.Recipe,
-		EnvRecipe:      &opts.Definition,
-		Secrets:        opts.Secrets,
+		RootDir:          requestDirPath,
+		EnvConfig:        &opts.Configuration,
+		ResourceRecipe:   &opts.Recipe,
+		EnvRecipe:        &opts.Definition,
+		Secrets:          opts.Secrets,
+		StateLockTimeout: terraform.DefaultStateLockTimeout,
 	})
 
 	unsetError := unsetGitConfigForDirIfApplicable(secretStoreID, opts.Secrets, requestDirPath, opts.Definition.TemplatePath)
--- a/pkg/recipes/terraform/config/backends/kubernetes.go
+++ b/pkg/recipes/terraform/config/backends/kubernetes.go
@@ -108,7 +108,7 @@ func generateSecretSuffix(resourceRecipe *recipes.ResourceMetadata) (string, err
 	}
 
 	inputString := strings.ToLower(fmt.Sprintf("%s-%s-%s", parsedEnvID.Name(), parsedAppID.Name(), parsedResourceID.String()))
-	
+
 	hasher := sha1.New()
 	_, err = hasher.Write([]byte(inputString))
 	if err != nil {
@@ -116,10 +116,6 @@ func generateSecretSuffix(resourceRecipe *recipes.ResourceMetadata) (string, err
 	}
 	hash := hasher.Sum(nil)
 	suffix := fmt.Sprintf("%x", hash)
-
-	// Debug logging to help investigate state lock collisions
-	fmt.Printf("DEBUG STATE KEY: ResourceID=%s, EnvID=%s, AppID=%s, Input=%s, Suffix=%s\n", 
-		resourceRecipe.ResourceID, resourceRecipe.EnvironmentID, resourceRecipe.ApplicationID, inputString, suffix)
 
 	return suffix, nil
 }
--- a/pkg/recipes/terraform/config/backends/kubernetes.go
+++ b/pkg/recipes/terraform/config/backends/kubernetes.go
@@ -107,14 +107,21 @@ func generateSecretSuffix(resourceRecipe *recipes.ResourceMetadata) (string, err
 		return "", err
 	}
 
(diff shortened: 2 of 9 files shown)


pkg/recipes/driver/terraform.go
```diff
@@ -98,11 +98,12 @@ func (d *terraformDriver) Execute(ctx context.Context, opts ExecuteOptions) (*re
...
-		EnvConfig:      &opts.Configuration,
-		ResourceRecipe: &opts.Recipe,
-		EnvRecipe:      &opts.Definition,
-		Secrets:        opts.Secrets,
+		RootDir:          requestDirPath,
+		EnvConfig:        &opts.Configuration,
+		ResourceRecipe:   &opts.Recipe,
+		EnvRecipe:        &opts.Definition,
+		Secrets:          opts.Secrets,
+		StateLockTimeout: terraform.DefaultStateLockTimeout,
```
vishwahiremat, line 106: what happens if the deployment takes longer than `StateLockTimeout` ?

pkg/recipes/terraform/execute.go
```diff
@@ -391,19 +379,28 @@ func initAndApply(ctx context.Context, tf *tfexec.Terraform) (*tfjson.State, err
...
 	}
 
 	// Load Terraform state to retrieve the outputs
 	logger.Info("Fetching Terraform state")
+
+	// Verify terraform binary is still accessible before state operation
+	if execPath := tf.ExecPath(); execPath != "" {
+		if _, err := os.Stat(execPath); err != nil {
+			logger.Info(fmt.Sprintf("ERROR: Terraform binary missing at %s during state fetch: %s", execPath, err.Error()))
+			return nil, fmt.Errorf("terraform binary disappeared at %s: %w", execPath, err)
```
vishwahiremat, line 395: i am not sure if we wanna add exec path in the error output.

pkg/recipes/terraform/config/backends/kubernetes.go
```diff
@@ -107,14 +107,17 @@ func generateSecretSuffix(resourceRecipe *recipes.ResourceMetadata) (string, err
...
-	_, err = hasher.Write([]byte(strings.ToLower(fmt.Sprintf("%s-%s-%s", parsedEnvID.Name(), parsedAppID.Name(), parsedResourceID.String()))))
+	_, err = hasher.Write([]byte(inputString))
 	if err != nil {
 		return "", err
 	}
 	hash := hasher.Sum(nil)
+	suffix := fmt.Sprintf("%x", hash)
 
-	return fmt.Sprintf("%x", hash), nil
+	return suffix, nil
```
ytimocin, line 120: thanks for the cleanup

pkg/recipes/terraform/types.go
```diff
@@ -34,6 +34,9 @@ import (
 const (
 	executionSubDir                = "deploy"
 	workingDirFileMode fs.FileMode = 0700
+
+	// DefaultStateLockTimeout is the default timeout for acquiring Terraform state locks
+	DefaultStateLockTimeout = "10m"
```
ytimocin, line 39: Is the app going to wait 10 minutes to acquire the lock? I think our worker will have already timed out by then because it uses 2 minutes in terms of timeout.

pkg/recipes/terraform/install.go
```diff
@@ -37,73 +38,117 @@ const (
...
-	// Create Terraform installation directory
-	installDir := filepath.Join(tfDir, installSubDir)
-	if err := os.MkdirAll(installDir, 0755); err != nil {
-		return nil, fmt.Errorf("failed to create directory for terraform installation for resource: %w", err)
+	return tf, nil
+}
+
+// ensureGlobalTerraformBinary ensures a global shared Terraform binary is available.
+// Uses mutex-based locking to prevent race conditions during concurrent access.
+func ensureGlobalTerraformBinary(ctx context.Context, installer *install.Installer, logger logr.Logger) (string, error) {
```
ytimocin, line 90: I believe that this function can be split into smaller pieces in terms of modularity. I see a few opportunities:

1. If `os.Stat(globalBinary)` && `os.Stat(globalMarker)` is true then we can just check to see if globalTerraformReady is true or false. Based on that we can call another function like `verifyBinaryWorks` which can be used in other places too.
2. Some `logger.Info`s could be `logger.Warn` if `Warn` exists for logger.
3. Download and Install Terraform could be split to another function.
4. We do `tfexec.NewTerraform` and `tf.Version` in a few places. We can split that to its own function.

My main concern with this function is that it does a lot of things and testing it would be difficult. We should simplify it as much as we can.

pkg/recipes/terraform/execute.go
```diff
@@ -375,8 +355,16 @@ func getTerraformConfig(ctx context.Context, workingDir string, options Options)
...
+func getStateLockTimeout(timeout string) string {
+	if timeout == "" {
+		return DefaultStateLockTimeout
+	}
+	return timeout
+}
+
 // initAndApply runs Terraform init and apply in the provided working directory.
-func initAndApply(ctx context.Context, tf *tfexec.Terraform) (*tfjson.State, error) {
+func initAndApply(ctx context.Context, tf *tfexec.Terraform, stateLockTimeout string) (*tfjson.State, error) {
```
ytimocin, line 367: Instead of passing stateLockTimeout, we can pass in `...tfexec.ApplyOption` (or an array of the same) because, in future, we can have more options and they are just going to float the signature of this function.

sylvainsf (approved):
Looks great!

pkg/recipes/terraform/types.go
lakshmimsft: No, it tries to acquire the lock immediately. It will wait for 10min if the lock is not available. https://developer.hashicorp.com/terraform/cli/commands/apply#lock-timeout-duration

pkg/recipes/driver/terraform.go
lakshmimsft: this is time taken to wait for lock to be available, so it can then run the deployment. https://developer.hashicorp.com/terraform/cli/commands/apply#lock-timeout-duration

pkg/recipes/terraform/install.go
lakshmimsft: I will address this in separate PR.

pkg/recipes/terraform/execute.go
lakshmimsft: I will address this in separate PR

pkg/recipes/terraform/execute.go
lakshmimsft: I think I might leave this as is for now. We can refactor when we need to.

PR #10318, opened 2025-09-03
Update make-build to skip bicep download if exists

brooke-hamilton:
# Description

The `make build` command downloads the Bicep binary to the `./dist` folder. This PR updates the logic of `make build` to skip the Bicep download if it already exists in the target folder, which prevents multiple downloads of the same binary each time `make build` is run, and speeds up the command.

## Type of change

Fixes: N/A

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable

### DIFF
--- a/build/install-bicep.sh
+++ b/build/install-bicep.sh
@@ -59,11 +59,17 @@ if [ "$ARCH" = "arm64" ]; then
   BICEP_ARCH="arm64"
 fi
 
-curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-$BICEP_ARCH
-if [ $? -ne 0 ]; then
-  echo "Failed to download Bicep CLI. Please check your internet connection or the URL."
-  exit 1
-fi
+# Check if bicep binary already exists in the target location
+if [ -f "$OUTPUT_DIR/bicep" ]; then
+  echo "Bicep CLI already exists at $OUTPUT_DIR/bicep, skipping download."
+else
+  echo "Downloading Bicep CLI..."
+  if ! curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-$BICEP_ARCH; then
+    echo "Failed to download Bicep CLI. Please check your internet connection or the URL."
+    exit 1
+  fi
 
-chmod +x bicep
-mv bicep "$OUTPUT_DIR"/bicep
+  chmod +x bicep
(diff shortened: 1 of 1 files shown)


build/install-bicep.sh
```diff
@@ -59,11 +59,17 @@ if [ "$ARCH" = "arm64" ]; then
...
 fi
 
-curl -Lo bicep https://github.com/Azure/bicep/releases/latest/download/bicep-linux-$BICEP_ARCH
-if [ $? -ne 0 ]; then
-  echo "Failed to download Bicep CLI. Please check your internet connection or the URL."
-  exit 1
-fi
+# Check if bicep binary already exists in the target location
+if [ -f "$OUTPUT_DIR/bicep" ]; then
+  echo "Bicep CLI already exists at $OUTPUT_DIR/bicep, skipping download."
```
nicolejms, line 64: Should we do a version check here?  How will it be updated when a new version comes out?

brooke-hamilton: I think we could do that, however I took these things into consideration when adding the simple check for whether the file exists:

- This download goes into the `./dist` folder which is excluded from git in the `.gitignore` file.
- Running `make clean` or a `git clean -fdx` command deletes the folder. 
- There is currently no version checking. The version that is downloaded is "latest".
- If this runs on a build runner, the file will not exist, and it will be downloaded (latest version).

On my dev machine, the `dist` folder does not stay around for long because at some point during the week I will have a need to run `make clean` or `git clean -fdx`. If teammates have very different practices then I'm OK with changing the approach. But downloading the binary every time we do a local build seems excessive.

We have issue #8901 that describes having tool version numbers in one place. As part of that issue we could add logic to check for the bicep version. We should be selecting a specific version throughout the system instead of always taking latest.

sylvainsf (approved):
This will save a ton of time and bandwidth!

build/install-bicep.sh
nicolejms: That makes sense. Since it's easy to rectify i'm ok with no version check.

nicolejms (approved):

PR #10326, opened 2025-09-04
Replace AWS::S3::Bucket with AWS::Logs::LogGroup for functional tests

willdavsmith:
# Description

* Replace AWS::S3::Bucket with AWS::Logs::LogGroup

## Type of change

Fixes: https://github.com/radius-project/radius/issues/5963

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


test/validation/aws.go
```diff
@@ -182,12 +182,12 @@ func GetResourceTypeName(ctx context.Context, resource *AWSResource) (string, er
 }
 
 // assertFieldsArePresent ensures that all fields in actual exist and are equivalent in expected
-func assertFieldsArePresent(t *testing.T, actual any, expected any) {
+func assertFieldsArePresent(t *testing.T, expected any, actual any) {
```
willdavsmith, line 184: I updated the signature to match the common order (expected, actual). there's only one call of this function and it was already calling it like this. there was a bug on line 189 that caused it to work, I guess 2 wrongs do make a right.

test/functional-portable/corerp/cloud/mechanics/aws_mechanics_test.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 func Test_AWSRedeployWithUpdatedResourceUpdatesResource(t *testing.T) {
+	t.Skip("Flaky tests: https://github.com/radius-project/radius/issues/5963")
```
sylvainsf: Is there a reason you want this test to skip every time? If it's deprecated we can remove it instead.

sylvainsf (commented):
There's a lot of great improvements in this PR but a few areas I'm hoping you can address. One area of caution is using AWS tags, new tags are a longer creation time that can have outliers instead of re-using tags for something like a global test cleanup (ie: tag all resources with the sha of the commit and they can be ensured cleanup at the end). 

This is particularly important if you're setting SkipResourceDeletion: true. I'm in favor of moving all test cleanup out to something atomic, cleaning up in each test encourages tests to have collisions etc and we just hope the cleanup is preventing it. 

There's a lot of test steps and bicep templates that are nearly identical except for one or two parameters such as RetentionInDays; you could get rid a a LOT of code by using a modular/paramterized model. Here's a simple example in go:

```go
var cases = []struct {
    name      string
    retention int
}{
    {"7days", 7},
    {"14days", 14},
}

for _, tc := range cases {
    t.Run(tc.name, func(t *testing.T) {
        // Use tc.retention as a parameter
    })
}
```

I also notice there's no retry/backoff logic, here's a simple modular approach:

```go
import "time"

func retryWithBackoff(action func() error, maxRetries int, initialDelay time.Duration) error {
    delay := initialDelay
    for i := 0; i < maxRetries; i++ {
        err := action()
        if err == nil {
            return nil
        }
        time.Sleep(delay)
        delay = delay * 2 // Exponential backoff
        if delay > 30*time.Second {
            delay = 30 * time.Second // Cap the delay
        }
    }
    return fmt.Errorf("all retries failed")
}

// Usage: retryWithBackoff(validateResource, 5, 1*time.Second)
```

nit: using t.Error() on env var checks fails the suite instantly, if that's intended great but t.Skip() might be better. I'm on the fence because I prefer failing fast.

sylvainsf:
Apologies I submitted that as a comment instead of a request for changes.

test/functional-portable/corerp/cloud/mechanics/testdata/aws-mechanics-redeploy-withcreateandwritepropertyupdate.step1.bicep
```diff
@@ -1,48 +0,0 @@
-extension aws
-
-param creationTimestamp string
-param dbSubnetGroupName string
-param dbName string
-
```
willdavsmith, line 6: these looked unused.

sylvainsf (dismissed):
Looks great, really clean parameterization.

sylvainsf (approved):
approved!

PR #10563, opened 2025-10-01
Replace Gitea with custom git http backend

willdavsmith:
# Description

* Replace gitea with custom git http backend

## Local testing

You can verify the behavior locally using these steps:

### 1. create KinD cluster with config to expose localhost:30080
```
cat <<'EOF' | kind create cluster --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraPortMappings:
    - containerPort: 30080
      hostPort: 30080
      protocol: TCP
EOF
```

### 2. Install Radius
```
rad init
```

### 3. Install Flux into the cluster
```
.github/actions/install-flux/install-flux.sh 2.6.4
```

### 4. Install local Git server into the cluster
```
.github/actions/install-git-http-backend/install-git-http-backend.sh testuser not-a-secret-password git-http-backend alpine/git:2.45.2
```

### 5. Port-forward Git server
```
kubectl port-forward -n git-http-backend svc/git-http 30080:3000 &
```
(note the PID so you can kill it later with `kill <PID>`)

### 6. Run Flux tests
```
GIT_HTTP_SERVER_URL="http://localhost:30080" GIT_HTTP_USERNAME="testuser" GIT_HTTP_EMAIL="testuser@radapp.io" GIT_HTTP_PASSWORD="not-a-secret-password" go test -timeout 180s -run ^Test_Flux_Basic$ github.com/radius-project/radius/test/functional-portable/kubernetes/noncloud
```

### 7. Cleanup
```
kind delete cluster
kill <PID_FROM_PORT_FORWARD>
```

## Type of change

Fixes: https://github.com/radius-project/radius/issues/10479

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


brooke-hamilton (commented):
My initial take is that this is a good approach. One thing we should do differently is to pull the container images from somewhere other than Docker hub so that we are not rate limited.

I will submit a more detailed review when I am able to run a test from a fork.

brooke-hamilton:
> My initial take is that this is a good approach. One thing we should do differently is to pull the container images from somewhere other than Docker hub so that we are not rate limited.
> 
> I will submit a more detailed review when I am able to run a test from a fork.

According to [this GitHub blog post](https://github.blog/enterprise-software/ci-cd/when-to-choose-github-hosted-runners-or-self-hosted-runners-with-github-actions/#cost-management), GitHub has an agreement with Docker Hub to allow unlimited pulls. I think we did not have that experience at one time, but maybe this agreement was implemented after we had the issue.

.github/workflows/functional-test-noncloud.yaml
```diff
@@ -374,32 +374,74 @@ jobs:
...
+            echo "Using default Git HTTP password from workflow env"
+            password="${DEFAULT_PASSWORD}"
+          else
+            echo "Using Git HTTP password from repository secret"
+          fi
+          echo "::add-mask::${password}"
+          echo "GIT_HTTP_PASSWORD=${password}" >> "$GITHUB_ENV"
+        env:
+          SECRET_PASSWORD: ${{ secrets.GITEA_PASSWORD }}
+          DEFAULT_PASSWORD: ${{ env.GIT_HTTP_PASSWORD }}
```
brooke-hamilton: Let's use the GIT_HTTP_PASSWORD variable and remove the GITEA_PASSWORD, which requires this logic that checks for the existence of the secret. This git repo lives for a short time in a protected runner and does not store any confidential information. I think the default GIT_HTTP_PASSWORD is fine.

.github/workflows/long-running-azure.yaml
```diff
@@ -554,27 +554,23 @@ jobs:
...
-          
-      #     # Test the connection to ensure port forwarding is working
-      #     curl -s http://localhost:30080 > /dev/null || (echo "Port forwarding failed" && exit 1)
-          
-      #     echo "Port forwarding established successfully"
+      - name: Install Git HTTP backend
+        uses: ./.github/actions/install-git-http-backend
+        with:
+          git-username: ${{ env.GIT_HTTP_USERNAME }}
+          git-password: ${{ secrets.GITEA_PASSWORD }}
```
brooke-hamilton: Replace with GIT_HTTP_PASSWORD

.github/workflows/functional-test-noncloud.yaml
```diff
@@ -374,32 +374,74 @@ jobs:
...
+
+      - name: Install Git HTTP backend
+        if: matrix.name == 'kubernetes-noncloud'
+        uses: ./.github/actions/install-git-http-backend
+        with:
+          git-username: ${{ env.GIT_HTTP_USERNAME }}
+          git-password: ${{ env.GIT_HTTP_PASSWORD }}
+
+      - name: Port-forward to Git server
```
brooke-hamilton, line 401: There is also a Port-forward to Git server step in `long-running-azure.yaml` that does the same thing with different code. Should this be included into the reusable action? Or if not, put it in a script and reference it from the other workflow.

.github/workflows/functional-test-noncloud.yaml
```diff
@@ -462,6 +504,43 @@ jobs:
           TEMP_CERT_DIR: ${{ steps.create-local-registry.outputs.temp-cert-dir }}
           SSL_CERT_FILE: ${{ steps.create-local-registry.outputs.temp-cert-dir }}/certs/${{ env.LOCAL_REGISTRY_SERVER }}/client.crt
 
+      - name: Verify Git server port-forward
```
brooke-hamilton: Why is this step here instead of after the port-forward step?
Should this logic be included in the reusable action? 
Is this logic necessary? Are there circumstances where the port-forwarding would fail due to Radius code?
Note: this type of logic is not in the long running test workflow.

brooke-hamilton (commented):
🚀 This is awesome 🚀 

Minor comments below:

The [create-kind-cluster](https://github.com/willdavsmith/radius/blob/willdavsmith/re-enable-gitea/.github/actions/create-kind-cluster/action.yaml#L48) action references the gitea port. Does this need to be updated?

.github/workflows/functional-test-noncloud.yaml
willdavsmith: removed

.github/workflows/functional-test-noncloud.yaml
willdavsmith: removed that one

.github/actions/create-kind-cluster/action.yaml
```diff
@@ -45,7 +45,7 @@ runs:
       run: |
         # https://kind.sigs.k8s.io/docs/user/local-registry/
         # Create the cluster with necessary configurations
-        # to expose 30080 port for Gitea
+        # to expose 30080 port for for the local Git server
```
brooke-hamilton: Typo
```suggestion
        # to expose 30080 port for the local Git server
```

.github/actions/create-kind-cluster/action.yaml
```diff
@@ -62,7 +62,7 @@ runs:
       shell: bash
       run: |
         # Create the cluster with necessary configurations
-        # to expose 30080 port for Gitea
+        # to expose 30080 port for for the local Git server
```
brooke-hamilton: Typo

```suggestion
        # to expose 30080 port for the local Git server
```

brooke-hamilton (changes requested):
I was able to go through the test steps, but I get errors when running the tests.

.github/actions/install-git-http-backend/git-http-backend-entrypoint.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+# Bootstrap script for the alpine/git container. Installs nginx/fcgiwrap, renders configuration from
+# templates, and launches the Git HTTP backend suitable for Radius functional tests.
+set -eu
+
+# Ensure nginx + fcgiwrap + git HTTP backend bits are present (alpine/git is minimal by default).
+apk add --no-cache nginx fcgiwrap spawn-fcgi apache2-utils gettext git-daemon
```
ytimocin, line 7: Should we check for the existence of the `apk` command? Just to be safe?

.github/actions/install-git-http-backend/git-http-backend-wrapper.sh
ytimocin, line 1: What is the difference between the wrapper.sh and entrypoint.sh?

ytimocin: Could we have merged the two together? Why do we need separate ones?

.github/actions/install-git-http-backend/git-http-backend-entrypoint.sh
willdavsmith: I don't it's necessary IMO - this should only ever be running in an `alpine/git` container, and if it isn't, `apk` will fail immediately anyway

.github/actions/install-git-http-backend/git-http-backend-wrapper.sh
willdavsmith: `entrypoint.sh`: needs root access to the container to install the right packages, configure git, start nginx server
`wrapper.sh`: runs on every nginx request

they do different things with different auth levels, so I don't think they should be merged together.

brooke-hamilton (dismissed):

brooke-hamilton (approved):

PR #10771, opened 2025-11-05
build: bump autorest and typespec

### LINKED ISSUE #8902: Generated clients_new should use "@autorest/go@4.0.0-preview.69" version like others
### Area for Improvement

1. `clients_new` is using an older version of autorest/go compared to other auto-generated apis: https://github.com/radius-project/radius/blob/main/pkg/cli/clients_new/README.md?plain=1#L32
2. We should update it to "@autorest/go@4.0.0-preview.69" like others: https://github.com/search?q=repo%3Aradius-project%2Fradius+%22%40autorest%2Fgo%404.0.0-preview.69%22&type=code

### Observed behavior

One of our generated APIs is using an older version of **autorest/go**

### Desired behavior

We should use the same version as the others

### Proposed Fix

1. Update the version in this line to 69: https://github.com/radius-project/radius/blob/main/pkg/cli/clients_new/README.md?plain=1#L32
2. Run `make generate`
3. Fix all errors (if there is any)

### rad Version

edge

### Operating system

macOS Sequoia 15.3.2

### Additional context

_No response_

### LINKED ISSUE #9497: Update Typespec npm packages to latest and resolve breaking changes
### Area for Improvement

The current versions of `@azure-tools/typespec-autorest` and `@azure-tools/typespec-azure-core` (version ~0.56.0 as of today) have a dev dependency on `@typespec/compiler@^1.0.0`.  
Also, `@typespec/http` and `@typespec/openapi` have released version 1. 

These package updates have breaking changes in our TypeSpec files. We need to update to the latest versions of all packages in `/typespec/package.json` and resolve the breaking changes in our TypeSpec files.

### Observed behavior

_No response_

### Desired behavior

Radius uses the latest versions of npm packages.

### Proposed Fix

_No response_

### rad Version

edge

### Operating system

_No response_

### Additional context

_No response_

DariuszPorowski:
# Description

- bump autorest and related typespec to the latest
- pin autorest and oav tools to the static version in every place that occurs
- remove deprecated `module-version` from autorest configs and replaced with `make generate-fix-version` as workaround - part of `make generate` | ref: https://github.com/Azure/autorest.go/blob/34605c2f2cc9c2df816e9e2072d89e2a31fec07f/packages/autorest.go/CHANGELOG.md?plain=1#L17
- fix operationId in RabbitMQ examples `RabbitMqQueues` -> `RabbitMQQueues` to mach 1:1 with .tsp config (required by updated typespec)
- update .tsp files with required updates by the latest typespec
- introduce `make generate-cleanup` command to remove all auto-generated .go files
- regenerate swaggers and .go files based on above updates
- `pkg/corerp/frontend/controller/applications/graph_util.go` updates to reflect changes from the updated models (struct tags are no longer emitted)

> [!TIP]
> For reviewers: any Go files started with `zz_` are auto generated. Not worth to spend time and review. All potential issues should be handled by tests.

## Type of change

close #8902, close #9497

> [!NOTE]
> Both work items are strongly related to each other and only a joint update seems to be a sensible solution.

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


.github/workflows/lint.yaml
```diff
@@ -57,9 +57,9 @@ jobs:
           npm ci
           popd
       - name: Install autorest
-        run: npm install -g autorest@3.7.1
+        run: npm install -g autorest@3.7.2
       - name: Install oav
-        run: npm install -g oav@3.3.5
+        run: npm install -g oav@4.0.2
```
ytimocin, line 62: Are these the most recent ones?

.github/workflows/publish-docs.yaml
```diff
@@ -57,16 +57,16 @@ jobs:
       - name: Setup NodeJS
         uses: actions/setup-node@v6
         with:
-          node-version: "18"
+          node-version: "22"
```
ytimocin, line 60: Thanks for this!

ytimocin: We may have old node versions in a few other workflows too.

build/generate.mk
```diff
@@ -52,7 +52,7 @@ generate-node-installed:
 .PHONY: generate-autorest-installed
 generate-autorest-installed:
 	@echo "$(ARROW) Detecting autorest..."
-	@which autorest > /dev/null || { echo "run 'npm install -g autorest' to install autorest"; exit 1; }
+	@which autorest > /dev/null || { echo "run 'npm install -g autorest@3.7.2' to install autorest"; exit 1; }
```
ytimocin, line 55: This locking is great!

ytimocin: Should we also check for the installed version and say something like "You don't have 3.7.2 installed"?

build/generate.mk
```diff
@@ -73,10 +73,22 @@ generate-controller: generate-controller-gen-installed ## Generates the CRDs for
...
+generate-fix-version: ## Fixes the module version in generated files.
+	@echo "$(ARROW) Fixing module version..."
+	grep -rl --include \zz_*.go 'moduleVersion = "v0.1.0"' | xargs -r sed -i "s/moduleVersion = \"v0.1.0\"/moduleVersion = \"v$(AUTOREST_MODULE_VERSION)\"/"
+	@echo "$(ARROW) Done."
+
 .PHONY: generate-genericcliclient
 generate-genericcliclient: generate-node-installed generate-autorest-installed
 	@echo "$(AUTOREST_MODULE_VERSION) is module version"
-	autorest pkg/cli/clients_new/README.md --tag=2023-10-01-preview
+	autorest pkg/cli/clients_new/README.md --tag=2023-10-01-preview && rm pkg/cli/clients_new/generated/go.mod
```
ytimocin: Why do we need to run `rm pkg/cli/clients_new/generated/go.mod` command after running the autorest command?

pkg/cli/clients_new/README.md
```diff
@@ -29,8 +32,8 @@ The following configuration generates track2 go models and client.
 
 ```yaml $(tag) != ''
 version: 3.*.*
-use: "@autorest/go@4.0.0-preview.44"
-module-version: 0.0.1
+use: "@autorest/go@4.0.0-preview.74"
+module: "github.com/radius-project/radius/pkg/cli/clients_new/generated"
```
ytimocin, line 36: This bump is amazing! Thanks a lot for working on this!

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
```diff
@@ -2448,6 +2448,7 @@
     },
     "Azure.ResourceManager.CommonTypes.TrackedResourceUpdate": {
       "type": "object",
+      "title": "Tracked Resource",
```
ytimocin, line 2451: Did you add this or was it generated?

swagger/specification/applications/resource-manager/Applications.Messaging/preview/2023-10-01-preview/examples/RabbitMQQueues_CreateOrUpdate.json
```diff
@@ -1,5 +1,5 @@
 {
-  "operationId": "RabbitMqQueues_CreateOrUpdate",
+  "operationId": "RabbitMQQueues_CreateOrUpdate",
```
ytimocin, line 2: 🙏

typespec/Applications.Core/main.tsp
```diff
@@ -41,7 +39,5 @@ interface Operations extends Azure.ResourceManager.Operations {}
 @doc("Supported API versions for the Applications.Core resource provider.")
 enum Versions {
   @doc("2023-10-01-preview")
-  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
-  @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
```
ytimocin, line 45: We don't need them anymore or do they get automatically injected in some other code path?

typespec/package.json
```diff
@@ -3,14 +3,14 @@
...
-    "@typespec/rest": "~0.66.0",
-    "@typespec/versioning": "~0.66.0"
+    "@typespec/compiler": "~1.5.0",
+    "@azure-tools/typespec-autorest": "~0.61.1",
+    "@azure-tools/typespec-azure-core": "~0.61.0",
+    "@azure-tools/typespec-azure-resource-manager": "~0.61.1",
+    "@typespec/http": "~1.5.0",
+    "@typespec/openapi": "~1.5.0",
+    "@typespec/rest": "~0.75.0",
+    "@typespec/versioning": "~0.75.0"
```
ytimocin: 🙏

Are these the latest?

go.mod
ytimocin, line 1: We may need to run `go mod tidy`...

.github/workflows/lint.yaml
DariuszPorowski: Yes, on today's date (2025-11-05) both tools are the recent

autorest: https://www.npmjs.com/package/autorest?activeTab=versions
oav: https://www.npmjs.com/package/oav?activeTab=versions

.github/workflows/publish-docs.yaml
DariuszPorowski: yes, that true... but not related with the context of this PR. So, let's address it in another one.

build/generate.mk
DariuszPorowski: yes, we already running it in the others autorest cases, this was missing and the updated config produce go.mod now which is not necessary

swagger/specification/applications/resource-manager/Applications.Core/preview/2023-10-01-preview/openapi.json
DariuszPorowski: generated

build/generate.mk
```diff
@@ -73,10 +73,22 @@ generate-controller: generate-controller-gen-installed ## Generates the CRDs for
 	controller-gen object:headerFile=./boilerplate.go.txt paths=./pkg/controller/api/...
 	controller-gen crd paths=./pkg/controller/api/... output:crd:dir=./deploy/Chart/crds/radius
 
+.PHONY: generate-cleanup
```
brooke-hamilton, line 76: What is the reason for adding the `generate-cleanup` target?

build/generate.mk
```diff
@@ -73,10 +73,22 @@ generate-controller: generate-controller-gen-installed ## Generates the CRDs for
...
+generate-cleanup: ## Deletes all generated code.
+	@echo "$(ARROW) Deleting generated code..."
+	find . -type f -name 'zz_*.go' ! -name 'zz_*.deepcopy.go' -delete
+	@echo "$(ARROW) Done."
+
+.PHONY: generate-fix-version
+generate-fix-version: ## Fixes the module version in generated files.
```
brooke-hamilton: What is the root cause of the module version being wrong in the generated files? Is there a related issue in whatever repo of whatever tool is causing this?

build/generate.mk
brooke-hamilton: Is there a way to instruct the generator to not produce go.mod instead of deleting it?

typespec/Applications.Core/main.tsp
DariuszPorowski: It's not required anymore, ref: https://github.com/Azure/typespec-azure/pull/3268

brooke-hamilton (commented):
🤖

go.mod
DariuszPorowski: already did, its part of this PR

image

typespec/package.json
DariuszPorowski: Yes, on today's date (2025-11-05) all are the latest

build/generate.mk
DariuszPorowski: autorest does not cleanup the output directory by default, so the new generated files not always match the old one (like in this case - you can observe a few deleted files because they are no longer part of auto-generation)

build/generate.mk
DariuszPorowski: I put the note in the PR description about `module-version` ref: https://github.com/Azure/autorest.go/blob/34605c2f2cc9c2df816e9e2072d89e2a31fec07f/packages/autorest.go/CHANGELOG.md?plain=1#L17

brooke-hamilton (commented):
`make generate` still produces a lot of warnings. Do we want to address that in a separate PR? (This one already has a lot of changes.) 

Because when there are a lot of warnings, there are effectively no warnings. 

Here's an example:
```
resourcegroups.tsp:45:3 - warning @typespec/http/metadata-ignored: path property will be ignored as it is inside of a @body property. Use @bodyRoot instead if wanting to mix.
> 45 |   name: ResourceNameString;
     |   ^^^^
```

build/generate.mk
DariuszPorowski: not possible in the latest autorest update as I know

build/generate.mk
DariuszPorowski: In short, the behavior has changed in the latest autorest/go update.

DariuszPorowski:
> @brooke-hamilton Is there a way to instruct the generator to not produce go.mod instead of deleting it?

We already doing this in all the other autorest cases (by using rm), this one was missing and the latest update of the autores/go produces always go.mod

DariuszPorowski:
> `make generate` still produces a lot of warnings. Do we want to address that in a separate PR? (This one already has a lot of changes.)
> 
> Because when there are a lot of warnings, there are effectively no warnings.
> 
> Here's an example:
> 
> ```
> resourcegroups.tsp:45:3 - warning @typespec/http/metadata-ignored: path property will be ignored as it is inside of a @body property. Use @bodyRoot instead if wanting to mix.
> > 45 |   name: ResourceNameString;
>      |   ^^^^
> ```

I will give a shot and let you know what the best way is (split or include in this PR)

build/generate.mk
DariuszPorowski: In general, it's not bad idea, but this should be addressed in different PR that handle context of tools versioning and consistency @brooke-hamilton and I have this on the plate, so I do not want to introduce this change in this one.

DariuszPorowski:
> `make generate` still produces a lot of warnings. Do we want to address that in a separate PR? (This one already has a lot of changes.)
> 
> Because when there are a lot of warnings, there are effectively no warnings.
> 
> Here's an example:
> 
> ```
> resourcegroups.tsp:45:3 - warning @typespec/http/metadata-ignored: path property will be ignored as it is inside of a @body property. Use @bodyRoot instead if wanting to mix.
> > 45 |   name: ResourceNameString;
>      |   ^^^^
> ```

@brooke-hamilton currently warnings do not impact any code generation changes/issues... and patch is big enough to split "warning" case into separate PR, thoughts?

ytimocin (dismissed):

pkg/corerp/frontend/controller/applications/graph_util.go
ytimocin, line 1: This may need to be tested by running `rad app graph` to make sure the logic is not broken. cc/ @nithyatsu

pkg/corerp/frontend/controller/applications/graph_util.go
```diff
@@ -458,52 +458,48 @@ func outputResourceEntryFromID(id resources.ID) corerpv20231001preview.Applicati
...
-
-	raw, _, err := p.Get(&resource)
-	if err != nil {
-		// Not found, this is fine.
+	// Directly access the nested weakly-typed property bag instead of using jsonpointer on the struct.
+	// The previous implementation relied on jsonpointer with the path /properties/status/outputResources.
+	// When applied to the struct value, jsonpointer failed to locate the lowercase JSON field name and
+	// silently returned an empty list, causing OutputResources to be omitted in the application graph.
+	statusRaw, ok := resource.Properties["status"]
+	if !ok || statusRaw == nil {
```
lakshmimsft: Are changes in this file directly related to updated in autorest, typespec ver? Should these changes be in another PR??

DariuszPorowski, line 466: @lakshmimsft its part of this change. I put explanation in the PR description (see graph_util.go point - struct tags are no longer emitted by autorest)

lakshmimsft: aah.. i saw that a few mins later and had deleted my comment

lakshmimsft (dismissed):

build/generate.mk
brooke-hamilton: I think we need a comment in the code that explains what is happening. The link doesn't really explain what the issue is. It points to this text:

```
* The `module-version` switch has been removed.
  * For new modules, the `moduleVersion` constant will have an initial value of `v0.1.0`.
  * For existing modules, the value of `moduleVersion` is externally maintained.
```

But that doesn't explain why we have to replace the initial value with a different value. Without something in the code to explain why this is here, it will either be left forever, or some future person will have to puzzle through why we are replacing the specific value "v0.1.0".

brooke-hamilton (changes requested):

DariuszPorowski:
@brooke-hamilton `generate-fix-version` is to keep the current value. I can remove generate-fix-version workaround completely and keep default value from autorest which is `v0.1.0` and drop the current one what is in the codebase `v0.0.1`.
If `v0.0.1` is not important to keep, then I happy to remove it completely :) thoughts? :)

image

brooke-hamilton (approved):

PR #10810, opened 2025-11-13
Restores the skip list from cache (if it exists)

### LINKED ISSUE #10809: Scheduled long running test failed - Run ID: 19327855969
## Bug information 

This issue is automatically generated if the scheduled long running test fails. The Radius long running test operates on a schedule of every 2 hours everyday. It's important to understand that the test may fail due to workflow infrastructure issues, like network problems, rather than the flakiness of the test itself. For the further investigation, please visit [here](https://github.com/radius-project/radius/actions/runs/19327855969).

[AB#17699](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/17699)

nithyatsu:
# Description

If new types are added, the LRT creates them successfully when skip_build is false. It also makes a skip list when skip_build is false. But it does not use the skip list when skip_build is false, so the new types added never make it to the skip list, causing LRT to pass only when skip_build is false. 

This PR runs cleanup using that skip list (protecting built-in types) irrespective of skip_build, as long as a file is available.

## Type of change

Fixes: #10809 

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -518,24 +518,12 @@ jobs:
         if: env.SKIP_BUILD != 'true'
         run: |
           kubectl get resources.ucp.dev -n radius-system --no-headers -o custom-columns=":metadata.name" > skip-delete-resources-list.txt
-          echo "Created skip-delete-resources-list.txt with the following resources:"
-          cat skip-delete-resources-list.txt
       - name: Save list of resources not to be deleted
         if: env.SKIP_BUILD != 'true'
         uses: actions/cache/save@v4
         with:
           path: skip-delete-resources-list.txt
           key: skip-delete-resources-list-file
-      - name: Display skip-delete-resources-list contents
-        if: always()
-        run: |
-          if [ -f skip-delete-resources-list.txt ]; then
-            echo "Contents of skip-delete-resources-list.txt:"
-            cat skip-delete-resources-list.txt
-            echo "Total resources in skip list: $(wc -l < skip-delete-resources-list.txt)"
-          else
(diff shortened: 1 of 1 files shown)


brooke-hamilton (approved):
🚢

kachawla:
Is this skip list only used for built in core resources or any resources that we provide including RRTs?

kachawla (approved):

DariuszPorowski (approved):

lakshmimsft:
This doesn't seem right. The skip list is to be recreated newly for fresh builds (skip-build=false) and should have a clean slate. so it should not include/have older types

nithyatsu:
> Is this skip list only used for built in core resources or any resources that we provide including RRTs?

It is created upon fresh install before running any tests, so will include all built in types we install through manifests. In this case, the fresh install created the manifest which included recipepacks, but at the end of test did not use skip list while deleting, so the recipepack type got deleted, causing the next run (which is not a fresh isntall)  to fail, causing an alternating pattern of failure and success.

nithyatsu:
> This doesn't seem right. The skip list is to be recreated newly for fresh builds (skip-build=false) and should have a clean slate. so it should not include/have older types

Yes, it is only created in fresh install and does not include any old type. But with this change, the fresh install's cleanup uses the skip list while cleanup too. Am I missing some case?

PR #10916, opened 2025-12-09
ci(dependency-review): add workflow for scanning vulnerable dependencies

DariuszPorowski:
# Description

This pull request adds a new GitHub Actions workflow to improve the security of dependency management in the repository. The workflow automatically scans for known-vulnerable dependencies whenever a pull request modifies dependency manifest files, helping prevent the introduction of insecure packages.

## Type of change

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,4 +1,4 @@
-# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+# yaml-language-server: $schema=https://www.schemastore.org/github-workflow.json
 
 # Dependency Review Action
 #
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,36 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+# Dependency Review Action
+#
+# This Action will scan dependency manifest files that change as part of a Pull Request,
+# surfacing known-vulnerable versions of the packages declared or updated in the PR.
+# Once installed, if the workflow run is marked as required,
+# PRs introducing known-vulnerable packages will be blocked from merging.
+#
+# Source repository: https://github.com/actions/dependency-review-action
+---
+name: Dependency Review
+
+on:
+  pull_request:
+
+permissions: {}
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      pull-requests: write
+      checks: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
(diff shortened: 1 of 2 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run Dependency Review
+        uses: actions/dependency-review-action@3c4e3dcb1aa7874d2c16be7d79418e9b7efd6261 # v4.8.2
+        with:
+          comment-summary-in-pr: always
```
brooke-hamilton: The `comment-summary-in-pr` requires the workflow permission `pull-requests: write` according to the docs [here](https://github.com/actions/dependency-review-action?tab=readme-ov-file#configuration-options). 

How about using the `on-failure` option so that we don't see extra comment blocks in the PR that don't require action? In other words, we don't need to see this information unless there is an action we need to take.

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,36 @@
...
+      checks: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run Dependency Review
+        uses: actions/dependency-review-action@3c4e3dcb1aa7874d2c16be7d79418e9b7efd6261 # v4.8.2
+        with:
```
brooke-hamilton, line 36: Should we add the `fail-on-severity: critical` option?

brooke-hamilton (changes requested):
🥇 This great. I have two overall suggestions: 

1. Can you run this on your fork and show an example of the output in the PR comment? What are we going to see when this runs?
2. Can you add something to the contribution documentation that links to the docs for this action and explains what a developer should do when there is a failure?

Other questions:
 - Does this align with the other security scans and reporting in GitHub?
 - Do we need to configure branch protection rules in addition to implementing this workflow?
 - Should we add license checks (which is an option)?
 - Should we add OpenSSF checks (another option)?

.github/workflows/dependency-review.yml
DariuszPorowski: I want to make sure the PR will fail on any security issue, even low to make sure we are aware of it.

.github/workflows/dependency-review.yml
DariuszPorowski: we have `pull-requests: write` in permissions (line 25).
OK, with `on-failure`

brooke-hamilton (approved):

PR #11294, opened 2026-02-24
Update radius to setup default recipepack during rad init and env operations.

nithyatsu:
# Description

rad init 
- uses Radius.Core API 
- creates default pack in default resource group and injects it into default environment

rad env create --preview 
- adds default recipe pack when creating a new Radius.Core environment

rad deploy
- before deploying, scans the compiled ARM template for all Radius.Core/environments.
For each environment found, inspects existing recipe packs, if none, adds default pack.

## Type of change

Fixes: #11091 

**https://github.com/radius-project/resource-types-contrib/pull/100 and this PR must be merged together.**

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


pkg/cli/cmd/deploy/deploy.go
```diff
@@ -650,6 +662,157 @@ func (r *Runner) setupCloudProviders(properties any) {
...
+		if err != nil {
+			return "", fmt.Errorf("failed to create default recipe pack: %w", err)
+		}
+	}
+	return recipepack.DefaultRecipePackID(), nil
+}
+
+// findRadiusCoreEnvironmentResources walks the template's resources and returns
+// all Radius.Core/environments resources found (as mutable maps).
+func findRadiusCoreEnvironmentResources(template map[string]any) []map[string]any {
```
kachawla: We have an existing function for this https://github.com/kachawla/radius/blob/main/pkg/cli/bicep/resources.go#L107

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -650,6 +662,157 @@ func (r *Runner) setupCloudProviders(properties any) {
...
+		}
+	}
+
+	return nil
+}
+
+// setupRecipePackForEnvironment sets up recipe packs for a single Radius.Core/environments resource.
+// If the environment already has any recipe packs set (literal IDs or ARM expression references),
+// no changes are made. Otherwise, it fetches or creates the default recipe pack from
+// the default scope and injects their IDs into the template.
```
kachawla: nit: we only expect one default recipe pack, right?


```suggestion
// no changes are made. Otherwise, it fetches or creates the default recipe pack from
// the default scope and injects its ID into the template.
```

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -650,6 +662,157 @@ func (r *Runner) setupCloudProviders(properties any) {
...
+	if !ok {
+		properties = map[string]any{}
+		outerProps["properties"] = properties
+	}
+
+	// If the environment already has any recipe packs configured (literal IDs or
+	// Bicep expression references), leave it as-is — the user is managing packs explicitly.
+	if hasAnyRecipePacks(properties) {
+		return nil
+	}
```
kachawla, line 713: What do you think about moving this as a property on https://github.com/kachawla/radius/blob/main/pkg/cli/bicep/resources.go#L38? Just provider better readability and abstraction.

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -650,6 +662,157 @@ func (r *Runner) setupCloudProviders(properties any) {
...
+	if !ok {
+		return false
+	}
+	return len(packsArray) > 0
+}
+
+// getOrCreateDefaultRecipePack attempts to GET the default recipe pack from
+// the default scope. If it doesn't exist (404), it creates it with all core
+// resource type recipes. Returns the full resource ID.
+func getOrCreateDefaultRecipePack(ctx context.Context, client *v20250801preview.RecipePacksClient) (string, error) {
```
kachawla: Should we move recipe pack specific functionality like this to https://github.com/nithyatsu/radius/blob/e204475b2e5c24030995ac45fd9b4ce4f1441920/pkg/cli/recipepack/recipepack.go?

test/functional-portable/cli/noncloud/cli_test.go
```diff
@@ -732,57 +727,6 @@ func Test_RecipeCommands(t *testing.T) {
 	test.Test(t)
 }
 
-// This test creates an environment by directly calling the CreateEnvironment function to test dev recipes.
```
lakshmimsft, line 735: if we're deleting these tests, shouldn't we be adding new ones? the functionality still exists for existing dev-recipes.

pkg/cli/recipepack/recipepack.go
```diff
@@ -0,0 +1,129 @@
...
+// 4. rad group switch prod
+// 5. .rad deploy <template contains the environment>
+func EnsureDefaultResourceGroup(ctx context.Context, createOrUpdate ResourceGroupCreator) error {
+	return createOrUpdate(ctx, "local", DefaultResourceGroupName, &ucpv20231001.ResourceGroupResource{
+		Location: to.Ptr(v1.LocationGlobal),
+	})
+}
+
+// SingletonRecipePackDefinition defines a singleton recipe pack for a single resource type.
+type SingletonRecipePackDefinition struct {
```
kachawla: I thought we weren’t going to create singleton recipe packs. Is this and the other singleton references something that carried over from previous implementation?

pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -973,6 +974,261 @@ func Test_Run(t *testing.T) {
...
+		}
+
+		err = runner.setupRecipePack(context.Background(), template)
+		require.NoError(t, err)
+
+		// Verify that the default recipe pack was injected.
+		envRes := template["resources"].(map[string]any)["env"].(map[string]any)
+		outerProps := envRes["properties"].(map[string]any)
+		innerProps := outerProps["properties"].(map[string]any)
+		packs, ok := innerProps["recipePacks"].([]any)
```
kachawla: This code is repeated multiple times, could we abstract it out in a helper method?

pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -973,6 +974,261 @@ func Test_Run(t *testing.T) {
...
+				"env": map[string]any{
+					"type": "Radius.Core/environments@2025-08-01-preview",
+					"properties": map[string]any{
+						"name": "myenv",
+						"properties": map[string]any{
+							"recipePacks": []any{existingPackID},
+						},
+					},
+				},
+			},
```
kachawla, line 1064: Same here, this is repeated code. Could we create a template builder helper function?

pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -973,6 +974,261 @@ func Test_Run(t *testing.T) {
...
+			DefaultScopeClientFactory: defaultScopeFactory,
+			ConnectionFactory:         &connections.MockFactory{ApplicationsManagementClient: mockAppClient},
+			Output:                    &output.MockOutput{},
+		}
+
+		// Template with a Radius.Core/environments resource and no recipe packs.
+		template := map[string]any{
+			"resources": map[string]any{
+				"env": map[string]any{
+					"type": "Radius.Core/environments@2025-08-01-preview",
```
kachawla: Would be good to extract `Radius.Core/environments@2025-08-01-preview` into a constant, so that if the version changes, we won't have to update it in multiple places in this file.

pkg/cli/cmd/deploy/deploy_test.go
kachawla, line 1: Do we need a test for if CreateOrUpdateResourceGroup returns an error?

pkg/cli/cmd/deploy/deploy_test.go
```diff
@@ -973,6 +974,261 @@ func Test_Run(t *testing.T) {
...
+
+	t.Run("creates default pack when not found in default scope", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		mockAppClient := clients.NewMockApplicationsManagementClient(ctrl)
+		mockAppClient.EXPECT().
+			CreateOrUpdateResourceGroup(gomock.Any(), "local", "default", gomock.Any()).
+			Return(nil).
+			Times(1)
+
+		// Default scope factory returns 404 on GET (packs don't exist yet)
```
kachawla, line 1175: Should we add another test for any error (not just 404)?

pkg/cli/cmd/env/create/preview/create.go
```diff
@@ -141,19 +146,46 @@ func (r *Runner) Run(ctx context.Context) error {
...
+			return err
+		}
+		r.DefaultScopeClientFactory = defaultClientFactory
+	}
+
+	defaultPack := recipepack.NewDefaultRecipePackResource()
+	_, err = r.DefaultScopeClientFactory.NewRecipePacksClient().CreateOrUpdate(ctx, recipepack.DefaultRecipePackResourceName, defaultPack, nil)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
kachawla: @nithyatsu this function includes updates as well, so we should check for existing packs. Also, can a user not provide their own recipe packs as part of rad env create command?

pkg/cli/cmd/env/update/preview/update.go
```diff
@@ -86,7 +86,7 @@ rad env update myenv --clear-kubernetes
 	cmd.Flags().Bool(commonflags.ClearEnvAzureFlag, false, "Specify if azure provider needs to be cleared on env")
 	cmd.Flags().Bool(commonflags.ClearEnvAWSFlag, false, "Specify if aws provider needs to be cleared on env")
 	cmd.Flags().Bool(commonflags.ClearEnvKubernetesFlag, false, "Specify if kubernetes provider needs to be cleared on env (--preview)")
-	cmd.Flags().StringArrayP("recipe-packs", "", []string{}, "Specify recipe packs to be added to the environment (--preview)")
+	cmd.Flags().StringSliceP("recipe-packs", "", []string{}, "Specify recipe packs to replace the environment's recipe pack list (--preview). Accepts comma-separated values.")
```
kachawla, line 89: This is not only used for replace though.. it can still be used to add new recipe packs if the user uses default in the list as well, right?

pkg/cli/cmd/env/update/preview/update.go
```diff
@@ -274,11 +274,10 @@ func (r *Runner) Run(ctx context.Context) error {
...
 	}
 
-	// add recipe packs if any
+	// replace recipe packs if any are specified
 	if len(r.recipePacks) > 0 {
-		if env.Properties.RecipePacks == nil {
-			env.Properties.RecipePacks = []*string{}
-		}
+		// Create a new list to replace the existing recipe packs
+		newRecipePacks := []*string{}
```
kachawla, line 284: Should we add a prompt/warning for users mentioning that the existing list will be replaced? I don't think this is very intuitive
@nithyatsu @willtsai

pkg/cli/cmd/env/update/preview/update_test.go
kachawla, line 1: Do we have a test to explicitly validate that the list is replaced?

pkg/cli/cmd/radinit/environment.go
```diff
@@ -53,41 +55,76 @@ func (r *Runner) CreateEnvironment(ctx context.Context) error {
...
+
+	if r.Options.Environment.Namespace != "" {
+		providers.Kubernetes = &corerpv20250801.ProvidersKubernetes{
+			Namespace: to.Ptr(r.Options.Environment.Namespace),
+		}
+	}
+
 	if r.Options.CloudProviders.Azure != nil {
-		providerList = append(providerList, r.Options.CloudProviders.Azure)
+		providers.Azure = &corerpv20250801.ProvidersAzure{
```
kachawla, line 68: Just curious - why do these need updating?

pkg/cli/cmd/radinit/environment.go
```diff
@@ -53,41 +55,76 @@ func (r *Runner) CreateEnvironment(ctx context.Context) error {
...
-		return err
+	envProperties := corerpv20250801.EnvironmentProperties{
+		Providers: providers,
 	}
 
-	var recipes map[string]map[string]corerp.RecipePropertiesClassification
-	if r.Options.Recipes.DevRecipes {
-		// Note: To use custom registry for recipes, users need to manually configure
-		// their environment after initialization or use custom recipe definitions
-		recipes, err = r.DevRecipeClient.GetDevRecipes(ctx)
```
kachawla, line 73: I think we should make sure capture this in release highlights that default dev recipes are removed?
@nithyatsu @willtsai

pkg/cli/recipepack/recipepack.go
```diff
@@ -0,0 +1,129 @@
...
+	corerpv20250801 "github.com/radius-project/radius/pkg/corerp/api/v20250801preview"
+	"github.com/radius-project/radius/pkg/to"
+	ucpv20231001 "github.com/radius-project/radius/pkg/ucp/api/v20231001preview"
+	"github.com/radius-project/radius/pkg/version"
+)
+
+const (
+	// DefaultRecipePackResourceName is the name of the default recipe pack
+	// resource that contains recipes for all core resource types.
+	DefaultRecipePackResourceName = "default"
```
kachawla, line 34: @willtsai Do we want to restrict users from using `default` as the name for their recipe packs or we don't care if they override it by using the same name?

@nithyatsu will it cause any issues in the assumptions we have made in the implementation if allow it to be used?

pkg/cli/recipepack/recipepack.go
```diff
@@ -0,0 +1,129 @@
...
+func GetDefaultRecipePackDefinition() []SingletonRecipePackDefinition {
+	tag := version.Channel()
+	if version.IsEdgeChannel() {
+		tag = "latest"
+	}
+	return []SingletonRecipePackDefinition{
+		{
+			Name:           "containers",
+			ResourceType:   "Radius.Compute/containers",
+			RecipeLocation: "ghcr.io/radius-project/kube-recipes/containers:" + tag,
```
kachawla, line 128: I only see latest on ghcr. How are we planning to publish different versions?

test/functional-portable/cli/noncloud/cli_test.go
```diff
@@ -732,57 +727,6 @@ func Test_RecipeCommands(t *testing.T) {
...
-// After dev recipes are confirmed, the environment is deleted.
-func Test_DevRecipes(t *testing.T) {
```
kachawla, line 737: Should we remove rest of the code related to dev recipes?

test/functional-portable/cli/noncloud/testdata/corerp-recipe-pack-test.bicep
```diff
@@ -3,16 +3,13 @@ resource computeRecipePack 'Radius.Core/recipePacks@2025-08-01-preview' = {
...
+      'Radius.Networking/gateways': {
         recipeKind: 'terraform'
-        recipeLocation: 'https://github.com/project-radius/resource-types-contrib.git//recipes/compute/containers/kubernetes?ref=v0.48'
-        parameters: {
-          allowPlatformOptions: true
-        }
+        recipeLocation: 'https://github.com/project-radius/resource-types-contrib.git//recipes/networking/gateways/kubernetes?ref=v0.48'
       }
-      'Radius.Security/secrets': {
+      'Radius.Messaging/queues': {
```
kachawla: What's causing this to change?

test/functional-portable/cli/noncloud/cli_test.go
nithyatsu: I thought we would remove it along with removing Applications.Core/environments. Should we do it as part of this effort instead? similar to https://github.com/radius-project/radius/pull/11294#discussion_r2956670866

pkg/cli/recipepack/recipepack.go
nithyatsu: https://github.com/radius-project/resource-types-contrib/issues/99 is supposed to add aworkflow that should be run during release. This PR and current PR have to be merged together.

pkg/cli/recipepack/recipepack.go
nithyatsu: I have not enforced the check to make sure this pack wouldnt be edited by users. But we have to do it, since overwriting it can cause issues. Could I do that as a quick follow up PR  and also log a story to override the default recipe-pack with user created pack?

pkg/cli/cmd/radinit/environment.go
nithyatsu: We had the datamodel for providers change between old and new 2025 versions. For instance the older model had scope and now there is subscription and resource group.  Hence we had to change.

pkg/cli/cmd/env/update/preview/update.go
nithyatsu: I made this change after our discussion, where we talked about how to remove a recipe pack through update, and not beung able to do so currently causing challenges. For instance if we want core recipe pack + user pack, then update with user pack alone would remove the core pack by design. Also, the rad env update command is used to set scope too. In that case, the scope gets replaced too.
I agree it might be non-intuitive to users. I added a warning.

pkg/cli/cmd/env/update/preview/update.go
nithyatsu: No, that makes the design  to remove default pack once user brings in a pack tricky. For example the case where user wants to add  user pack for new types but wants default pack for core types.

nithyatsu: currently user cant provide recipe pack through env create. It can happen through env update and through rad deploy env.bicep.

pkg/cli/recipepack/recipepack.go
nithyatsu: That is most likely unused code. I will remove it.

pkg/cli/cmd/radinit/environment.go
```diff
@@ -53,41 +55,76 @@ func (r *Runner) CreateEnvironment(ctx context.Context) error {
...
+			r.DefaultScopeClientFactory = defaultClientFactory
+		}
 	}
 
-	err = client.CreateOrUpdateEnvironment(ctx, r.Options.Environment.Name, &corerp.EnvironmentResource{
+	defaultPack := recipepack.NewDefaultRecipePackResource()
+	_, err = r.DefaultScopeClientFactory.NewRecipePacksClient().CreateOrUpdate(ctx, recipepack.DefaultRecipePackResourceName, defaultPack, nil)
+	if err != nil {
+		return clierrors.MessageWithCause(err, "Failed to create default recipe pack.")
+	}
```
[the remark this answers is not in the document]
nithyatsu: The user will not be allowed to modify the default recipe pack.

pkg/cli/cmd/env/create/preview/create.go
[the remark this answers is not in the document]
nithyatsu: @kachawla Today the user cannot provide recipe pack through rad env create command. It is only through rad env update.
I understand your point that if we do a rad env create myenv, and myenv exists, we would just update myenv. I thought for this scenario, myenv would be overwritten with defaults (including default pack).  Instead If retrieve the env and check if recipe pack is present, and if so, leave the env untouched, it feels like we are treating just one attribute special. Please let me know.

pkg/cli/cmd/radinit/environment.go
```diff
@@ -53,41 +55,76 @@ func (r *Runner) CreateEnvironment(ctx context.Context) error {
...
-		recipes, err = r.DevRecipeClient.GetDevRecipes(ctx)
+	// Initialize the Radius.Core client factory if not already set
+	if r.RadiusCoreClientFactory == nil {
+		clientFactory, err := cmd.InitializeRadiusCoreClientFactory(ctx, r.Workspace, r.Workspace.Scope)
 		if err != nil {
-			return err
+			return clierrors.MessageWithCause(err, "Failed to initialize Radius Core client.")
 		}
+		r.RadiusCoreClientFactory = clientFactory
 	}
```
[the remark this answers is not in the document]
nithyatsu: Not removing functionalities as part of this PR.

pkg/cli/cmd/deploy/deploy.go
nithyatsu: I agree it helps readability and abstraction, but the challenge I have is that when there are multiple environment resources, we need a map environmentID-->hasRecipePacks. The property is per environment..

test/functional-portable/cli/noncloud/testdata/corerp-recipe-pack-test.bicep
nithyatsu: I had changed it because earlier we had a complicated logic to delete with adding recipe pack that had core type recipes. So the deploy would have failed with conflict error. But I  reverted this now, since we add default pack only when there are no recipe packs in the env.

test/functional-portable/cli/noncloud/cli_test.go
nithyatsu: reverted delete. I had deleted it earlier, since I had made some changes in test bicep to accomodate for the complex conflicting recipe pack logic (which we decided to get rid of).

test/functional-portable/cli/noncloud/cli_test.go
nithyatsu: Looks like the code was coming into picture only for rad init, so now it was dead code. I removed the  function + test.

pkg/cli/cmd/env/create/preview/create.go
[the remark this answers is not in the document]
kachawla: Let me sync up with you on the update overwrite behavior offline to underestand this better.

> Today the user cannot provide recipe pack through rad env create command. It is only through rad env update.

I didn't know about this, what's the motivation to not support adding a recipe pack on rad env create?

pkg/cli/cmd/deploy/deploy.go
kachawla: makes sense

pkg/cli/recipepack/recipepack.go
kachawla: yeah that sounds good @nithyatsu

kachawla (approved):

pkg/cli/cmd/env/create/preview/create.go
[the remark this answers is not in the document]
nithyatsu: the user experience for recipe packs was the minimal set that was necessary to get e2e functionality. 
so the recipe pack was provided through env update , similar to how we set a provider, or directly through rad deploy env.bicep. But it was more of a first iteration decision, than deciding against offering the ability. cc @willtsai for anything I missed.

PR #11506, opened 2026-03-25
Add .specify and eng/specs to migration plan directory structure

nicolejms:
Adds the .specify directory and eng/specs directory to the proposed directory structure in the migration plan.


nicolejms:
wrong version of the doc

PR #11564, opened 2026-04-06
fix(lrt): improve resource deletion logic

### LINKED ISSUE #8845: Revisit Resource Type Cleanup in Long Running Tests
### Steps to reproduce

Since this is specific to long running tests, it cannot be reproduced locally. Multiple frequent failures in long running tests, example: https://github.com/radius-project/radius/issues/8843.

### Observed behavior

Sometimes long running test workflow deletes core resource types (during cleanup it looks like), which causes it to fail at configure Radius workspace step with the following error:

```
Error: {
  "code": "BadRequest",
  "message": "resource type \"Applications.Core/environments\" not found",
  "target": "/planes/radius/local/resourcegroups/kind-radius/providers/Applications.Core/environments/kind-radius"
}
```

### Desired behavior

I'm not if we have discussed the approach towards why, what, and how resource types should be cleaned up in long running tests. We should review it to understand the rationale for current approach and where the gaps are to make it more resilient.

### Workaround

None I believe, it seems to have been resolving on its own after consistenly failing for a few hours.

### rad Version

N/A

### Operating system

_No response_

### Additional context

_No response_

DariuszPorowski:
# Description

Fixes the long-running test (LRT) cleanup script deleting resource provider infrastructure (resource types, API versions, locations) from `resources.ucp.dev`, which causes `resource type not found` errors in subsequent runs. This was triggered by GitHub Actions cache misses on `skip-delete-resources-list.txt` (see issue comments on #8843).

The cleanup script now uses Kubernetes label selectors (`ucp.dev/resource-type`) to unconditionally preserve resource provider infrastructure entries, regardless of whether the skip-file is available.

**Key changes:**
- Query `resources.ucp.dev` entries with resource provider labels before the deletion loop and skip them unconditionally
- Broaden scope-skip from `scope.local.*`/`scope.aws.*` to `scope.*` to protect all scope entries
- Harden shell with `set -euo pipefail` and use `grep -qFx` for exact matching
- Simplify the pre-test cleanup step in the workflow to always pass the skip file path (the script handles missing files gracefully)

## Type of change

Fixes: #8845

## Contributor checklist

    - [x] Not applicable 
    - [x] Not applicable 
    - [x] Not applicable 
    - [x] Not applicable 
    - [x] Not applicable

### DIFF
--- a/.github/scripts/cleanup-long-running-cluster.sh
+++ b/.github/scripts/cleanup-long-running-cluster.sh
@@ -16,11 +16,11 @@
 # limitations under the License.
 # ------------------------------------------------------------
 
-set -e
+set -euo pipefail
 
 SKIP_RESOURCE_FILE="${1:-}"
 echo "cleaning up long-running cluster on Azure"
-echo "Using skip-resource-list from: $SKIP_RESOURCE_FILE"
+echo "Using skip-resource-list from: ${SKIP_RESOURCE_FILE}"
 
 # Delete all test resources in queuemessages.
 if kubectl get crd queuemessages.ucp.dev >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


.github/scripts/cleanup-long-running-cluster.sh
```diff
@@ -16,35 +16,71 @@
...
-    echo "delete all resources in resources.ucp.dev"
+    echo "delete test resources in resources.ucp.dev"
+
+    # Build a list of resource provider infrastructure entries that must always be
+    # preserved, regardless of whether the skip-delete-resources-list.txt exists.
+    # These entries are identified by their ucp.dev/resource-type label.
+    RP_INFRA_FILE="$(mktemp)"
+    if ! kubectl get resources.ucp.dev -n radius-system --no-headers \
+        -o custom-columns=":metadata.name" \
+        -l "ucp.dev/resource-type in (system.resources_resourceproviders,system.resources_resourceproviders_resourcetypes,system.resources_resourceproviders_resourcetypes_apiversions,system.resources_resourceproviders_locations)" \
```
lakshmimsft, line 44: I believe this will not include all types that are registered at startup located in https://github.com/radius-project/radius/tree/main/deploy/manifest/built-in-providers/self-hosted or dev folders. it will only include the meta types.

PR #11538, opened 2026-03-30
Re-enable and fix `Test_AWSRedeployWithUpdatedResourceUpdatesResource`

willdavsmith:
NOTE: The cloud functional test runs will fail in this PR because they're using the versions from main. Linked below is a manual cloud functional test run on this branch.

# Description

* Re-enable and fix Test_AWSRedeployWithUpdatedResourceUpdatesResource
* Manual cloud functional test run: https://github.com/radius-project/radius/actions/runs/23871332220

## Background

# Fix AWS functional test failure (`Test_AWSRedeployWithUpdatedResourceUpdatesResource`)

## Problem

The AWS mechanics test that creates a CloudWatch LogGroup and then updates its `RetentionInDays` property started failing around March 23, 2026 with:

> "The security token included in the request is invalid (Service: CloudWatchLogs, Status Code: 400)"

The CREATE step succeeded, but the UPDATE step always failed.

## Root Cause

Around March 23, AWS updated the CloudControl `GetResource` response for `AWS::Logs::LogGroup` to include new properties (`DeletionProtectionEnabled`, `FieldIndexPolicies`, etc.) that were added to the LogGroup API in late 2025. Our `GeneratePatch` function was generating `"remove"` operations in the JSON Patch for any property present in the AWS current state but absent from the user's desired state — unless it was a read-only or create-only property. These new updatable properties triggered `"remove"` operations in the patch, which caused CloudControl's async handler to fail.

This was a latent bug in our patch generation logic. The trigger was an AWS-side change to the `GetResource` response shape.

### Bug fix: Preserve unspecified properties during update (`pkg/aws/operations/operations.go`)

Changed `GeneratePatch` to preserve all properties from the current state that the user didn't explicitly specify, not just read-only/create-only ones. This prevents generating `"remove"` operations for properties the user doesn't mention, matching the standard behavior of Terraform and CloudFormation where unmentioned properties are left unchanged.

### Credential improvement: IRSA role chaining (`pkg/ucp/aws/ucpcredentialprovider.go`)

The IRSA credential path now performs a two-step credential retrieval: `AssumeRoleWithWebIdentity` followed by `AssumeRole` on the same role. This produces standard IAM session credentials instead of web identity federation tokens, which are more broadly compatible with AWS services. While this wasn't the root cause of the test failure, it is a robustness improvement.

### Regional STS endpoint support

[...]

PR #11569, opened 2026-04-07
Fix AWS UpdateResource by preserving unspecified properties in patch

willdavsmith:
NOTE: The actions on this run will fail.

Passing manual run: https://github.com/radius-project/radius/actions/runs/24103522599

## Problem

The AWS mechanics test (`Test_AWSRedeployWithUpdatedResourceUpdatesResource`) was failing with:

> "The security token included in the request is invalid (Service: CloudWatchLogs, Status Code: 400)"

The CREATE step succeeded, but the UPDATE step always failed. The test was skipped in PR #11500 to unblock the 0.56.0 release.

## Root Cause

Around March 23, AWS updated the CloudControl `GetResource` response for `AWS::Logs::LogGroup` to include new properties (`DeletionProtectionEnabled`, `FieldIndexPolicies`) that were added to the LogGroup API in late 2025. Our `GeneratePatch` function was generating `"remove"` operations in the JSON Patch for any property present in the AWS current state but absent from the user's desired state — unless it was a read-only or create-only property. These new updatable properties triggered `"remove"` operations in the patch, which caused CloudControl's async handler to fail with a misleading "invalid security token" error.

This was a latent bug in our patch generation logic. The trigger was an AWS-side change to the `GetResource` response shape.

## Fix

Changed `GeneratePatch` to preserve all properties from the current state that the user didn't explicitly specify, not just read-only/create-only ones. This prevents generating `"remove"` operations for properties the user doesn't mention, matching the standard behavior of Terraform and CloudFormation where unmentioned properties are left unchanged.

Also re-enables the previously skipped AWS mechanics test.

## Splitting note

This PR was split from #11538. The IRSA credential improvements (role chaining, regional STS endpoint) will be submitted as a separate follow-up PR.

## Contributor checklist

    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable 
    - [ ] Yes 
    - [x] Not applicable


kachawla (approved):
Looks good. One question - do we still have a clear way for a user to intentionally remove/reset a property? Might be worth adding a test for that so the new contract is explicit

PR #11870, opened 2026-05-13
Modify rad install to create default resource group and environment

### LINKED ISSUE #11869: `rad install kubernetes` should create a default resource group and environment (match `rad init` behavior)

The `rad install kubernetes` command should automatically create a default resource group and default environment, mirroring the behavior of `rad init`. Currently, users need to manually set up these resources after installation, adding unnecessary steps and inconsistencies between commands.

### Acceptance criteria

- `rad install kubernetes` creates a default resource group if one does not exist.
- `rad install kubernetes` creates a default environment if one does not exist.
- User experience for environment/resource setup is consistent with `rad init`.
- Documentation is updated to reflect this behavior.

### Additional context

This change will streamline initial setup for new users installing via the Kubernetes provider, making the CLI experience more predictable and eliminating extra manual setup steps performed after install.

zachcasper:
# Description

`rad install kubernetes` previously only installed the Radius Helm chart, leaving the cluster without a usable resource group or environment. Users then had to run additional commands (or use `rad init`) before they could deploy anything. This PR brings `rad install kubernetes` in line with `rad init` by automatically creating the `default` resource group and a `default` environment (in the `default` Kubernetes namespace) immediately after a successful Helm install.

**Behavior:**

- After `Helm.InstallRadius` succeeds, the runner builds an in-process workspace targeting the just-installed control plane and calls `CreateOrUpdateResourceGroup(ctx, "local", "default", …)` and `CreateOrUpdateEnvironment(ctx, "default", …)` with `KubernetesCompute{Namespace: "default"}`.
- `CreateOrUpdate` semantics make the calls idempotent — re-running with `--reinstall` will not corrupt user-edited groups/environments.
- The early-return path for an existing install (without `--reinstall`) is unchanged: nothing is touched.
- Kube context resolution mirrors what Helm itself targets: `--kubecontext` if set, otherwise the active context from kubeconfig. If kubeconfig has no active context, a clear error tells the user to pass `--kubecontext`.
- If the post-install group/environment creation fails (e.g. transient UCP error), the error message tells the user how to recover (`rad group create default` / `rad env create default`) so they don't have to reinstall.
- No client-side workspace is persisted to `~/.rad/config.yaml`; the workspace used here is in-process only.

**Files changed:**

- `pkg/cli/cmd/install/kubernetes/kubernetes.go`: adds `ConnectionFactory` + `KubernetesInterface` to `Runner`, the post-install `createDefaultGroupAndEnvironment` step, `resolveKubeContext` helper, default constants, updated `Long` description.
- `pkg/cli/cmd/install/kubernetes/kubernetes_test.go`: updates all successful-install tests with the new mock expectations, adds a test for the kubeconfig-fallback path and a test for the empty-current-context failure path. Mock expectations assert `Location == LocationGlobal` and `KubernetesCompute.Namespace == "default"` on the env payload.

## Type of change

Fixes: #11869

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes — `rad install kubernetes` user-facing behavior changes; docs should call out the new default group/env and the new failure modes.
    - [ ] Not applicable


pkg/cli/cmd/install/kubernetes/kubernetes.go
```diff
@@ -43,6 +57,12 @@ By default 'rad install kubernetes' will install Radius with the version matchin
 
 Radius will be installed in the 'radius-system' namespace. For more information visit https://docs.radapp.io/concepts/technical/architecture/
 
+After installing the control plane, this command also creates a default resource group named 'default' and
+a default environment named 'default' (using the 'default' Kubernetes namespace) so that the cluster is
+immediately ready to deploy applications. Re-running the command on an existing installation without
+'--reinstall' will leave existing resources unchanged. Using '--reinstall' may recreate or overwrite the
```
willdavsmith: nit: "may" or "will" overwrite the default rg and env?

pkg/cli/cmd/install/kubernetes/kubernetes.go
```diff
@@ -196,5 +221,63 @@ func (r *Runner) Run(ctx context.Context) error {
...
 
+	// Helm is configured with Wait=true for the Radius chart (see PopulateDefaultClusterOptions),
+	// so by the time InstallRadius returns the control plane pods are Ready and the UCP API is
+	// reachable. We can therefore call the management client immediately without an extra wait.
+	return r.createDefaultGroupAndEnvironment(ctx)
+}
+
+// createDefaultGroupAndEnvironment ensures that a resource group named "default" and an environment named
+// "default" exist on the cluster Radius was just installed on. Existing resources with the same names are
+// left unchanged (CreateOrUpdate semantics), so re-running install is safe.
```
willdavsmith: this seems to be inconsistent with the above. should we re-create env and rg in the case that it already exists?

pkg/cli/cmd/install/kubernetes/kubernetes.go
```diff
@@ -196,5 +221,63 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		return clierrors.MessageWithCause(err, "Failed to create the default environment. Radius was installed successfully; you can retry with 'rad env create default'.")
+	}
+
 	return nil
 }
+
+// resolveKubeContext returns the kube context the install targeted. If --kubecontext was passed, that value
+// is used. Otherwise the active kube context from the user's kubeconfig is used (matching what the Helm
+// install actually targeted).
+func (r *Runner) resolveKubeContext() (string, error) {
```
willdavsmith: why do we need this function? I assume that we have to do this type of operation elsewhere in the CLI. How do we handle it there?

pkg/cli/cmd/install/kubernetes/kubernetes_test.go
```diff
@@ -61,16 +68,70 @@ func Test_Validate(t *testing.T) {
...
+	// needed for the post-install "create default resource group + environment" step, and returns the
+	// trailing log writes the helper appends so tests can assemble the full expected output slice.
+	//
+	// When expectGetKubeContext is true the kubernetes interface mock will expect a single call to
+	// GetKubeContext() (the kubeconfig-fallback path triggered when --kubecontext is not set).
+	expectDefaultGroupAndEnvCreation := func(t *testing.T, ctrl *gomock.Controller, expectGetKubeContext bool) (connections.Factory, cli_kubernetes.Interface, []any) {
+		t.Helper()
+		mgmtMock := clients.NewMockApplicationsManagementClient(ctrl)
+		mgmtMock.EXPECT().
+			CreateOrUpdateResourceGroup(gomock.Any(), "local", "default", gomock.Any()).
```
willdavsmith, line 85: since these are mocks could we add a couple of tests to simulate failed rg and env creation?

willdavsmith (commented):
looks good, left a couple of comments

pkg/cli/cmd/install/kubernetes/kubernetes.go
zachcasper: Removed the overwriting functionality. Clarified that user customizations are never overwritten.

pkg/cli/cmd/install/kubernetes/kubernetes.go
zachcasper: Modified so that each resource is created only when it is missing.

pkg/cli/cmd/install/kubernetes/kubernetes.go
zachcasper: Good catch. I removed resolveKubeContext() and the post-install workspace now puts r.KubeContext directly into the connection map. Also dropped the kubernetesInterface and GetKubeContext mocks the tests.

pkg/cli/cmd/install/kubernetes/kubernetes_test.go
zachcasper: I added two failure-path tests in Test_Run:

Failure: default resource group creation fails — mocks GetResourceGroup → 404, then CreateOrUpdateResourceGroup → error. 

Failure: default environment creation fails — mocks the full happy-path through resource-group creation, then GetEnvironment → 404 and CreateOrUpdateEnvironment → error.

willdavsmith (approved):

Issue #7853, opened 2024-08-28
Remove tests: 'Test_Storage` and `Test_PersistentVolume`


willdavsmith:
### Area for Improvement

We will be disabling these tests due to an Azure `PolicyViolation`. Here's an example error message:

```
[...]

PR #11904, opened 2026-05-14
Run Azure functional tests locally against OS-process Radius

sylvainsf:
## Summary

Lets the Radius functional test suites — both **corerp-noncloud** and
**corerp-cloud (Azure)** — run against an OS-process Radius debug stack
(`make debug-start`) on an arbitrary k3d/kind cluster, instead of being
hard-wired to a `kind-radius` cluster and workspace. Along the way:

- Stabilizes several `make debug-*` targets that didn't reliably bring up
  their dependencies on k3d.
- Fixes a latent postgres pagination bug surfaced by the long-lived local
  debug stack (silently dropped rows when the postgres session timezone
  wasn't UTC).
- Adds a contributor doc capturing the bugs hit, why CI didn't catch each
  one, and remaining follow-ups.

CI behavior is unchanged — every change is gated behind an env var,
config flag, or local-only path; CI's workload-identity / `kind-radius`
flow continues exactly as before.

  - `GetSecretSuffix` derives the resource group from the active workspace
    (`cli.LoadConfig` → `GetWorkspace` → `ParseScope` →
    `FindScope("resourcegroups")`) instead of hardcoding `kind-radius`.
    Falls back to `default` when nothing is configured.
  - `Test_ApplicationGraph` PostStepVerify substitutes the fixture's
    `kind-radius` resource group with the active workspace's resource
    group before unmarshalling.
- `test/functional-portable/corerp/cloud/resources/recipe_terraform_test.go`
  - Derives the resource ID from the active workspace scope so it works
    against any RG (CI's `kind-radius` and local debug's `default`).
- `test/functional-portable/corerp/cloud/resources/extender_test.go`,
  `test/rp/rptest.go`, `test/ucp/ucptest.go`, `test/validation/shared.go`,
  — incidental cleanups required to run the suites against a
  non-`kind-radius` workspace; skip AWS-only tests cleanly when AWS env
  vars are unset; skip private-git redis test when `GH_TOKEN` is unset.

### Azure-cloud functional tests against a local OS-process stack

- `build/scripts/azure-local-testenv.sh` (new) — orchestrator with
  `setup`/`run`/`teardown`/`all` sub-commands. `run` and `all` accept
  passthrough `go test` flags (e.g. `-run`, `-v`).
  - **Auto-recovery**: `run` rebuilds state from the newest
    `radlocal-${USER}-*` resource group when the state file is missing
    (e.g. after `make debug-stop`), and re-applies the Azure scope on
    the default rad environment that `debug-start` wipes.
  - **Orphan GC**: `teardown --all-orphans` deletes every
    `radlocal-${USER}-*` RG and stops the `tf-module-server` port-forward.
- `pkg/recipes/terraform/config/providers/azure.go` — Terraform Azure
  provider falls back to `use_cli = true` when no Azure credential is
  registered with UCP (404), so the host RP's `az login` session
  authenticates. CI workload-identity path is unchanged.
- `build/scripts/ensure-encryption-key.sh` (new) — generates a stable
  encryption key for the local stack.

### Local OS-process debug stack stability

- `build/scripts/start-radius.sh`
  - Exports `TERRAFORM_TEST_GLOBAL_DIR` so the RP no longer tries to
    write to a read-only `/terraform` on the host.
  - Exports `RADIUS_INSECURE_LOOPBACK_REGISTRIES=true` so dlv-launched
    children inherit it and pull recipes over plain HTTP from the local
    `registry:2` on 127.0.0.1:5000.
  - Drops the grep-for-manifests UCP-readiness check — dlv buffers child
    stdout, so the "Successfully registered manifests" line can sit
    unflushed for minutes even though UCP is ready. A 200 from
    `/apis/api.ucp.dev/v1alpha3` is sufficient.
- `pkg/rp/util/registry.go` — new `shouldForcePlainHTTP` gated by
  `RADIUS_INSECURE_LOOPBACK_REGISTRIES`. When set, `ReadFromRegistry`
  forces `PlainHTTP=true` for recipe pulls whose `templatePath` host is
  `localhost` or a loopback IP. Env var unset = behavior unchanged.
- `pkg/cli/config.go` — in `getConfig`, if no path is passed and
  `RAD_CONFIG_FILE` is set in the environment, honour it. Lets
  `build/test.mk` route the functional tests' in-process
  `cli.LoadConfig("")` calls at the project-local debug rad config
  instead of `~/.rad/config.yaml`.
- `build/test.mk`
  - Auto-detects a live OS-process Radius controller via
    `debug_files/logs/controller.pid` and exports
    `RADIUS_SKIP_FLUX_TESTS=1` (the in-cluster Flux source-controller URL
    is unreachable from a host process).
  - Auto-detects a live OS-process UCP via `debug_files/logs/ucp.pid` and
    exports `RAD_CONFIG_FILE=build/configs/rad-debug-config.yaml`.
- `build/debug.mk`
  - `debug-install-contour`: drop Helm `--wait` (it doesn't behave for
    LoadBalancer Services on k3d) and instead do explicit
    `kubectl wait --for=condition=Available` + `kubectl rollout status`,
    so the target only returns once Contour is actually serving.
  - `debug-install-tf-module-server`: deploy the in-cluster nginx test
    module server and port-forward it to `localhost:8999`; add a `curl`
    readiness probe so subsequent recipe pulls don't race the pod
    becoming Ready.
  - Add `yq` to `debug-check-prereqs`.
- `build/recipes.mk`, `.github/scripts/publish-recipes.sh`,
  `build/scripts/mirror-test-images.sh` (new) — companion glue for
  running the suite locally with mirrored images and locally published
  recipes (the publish script learns `PLAIN_HTTP` for `localhost:5000`).
- `test/functional-portable/kubernetes/noncloud/flux_test.go` —
  `testFluxIntegration` honours `RADIUS_SKIP_FLUX_TESTS` and `t.Skip()`s
  with a clear reason.

### Postgres pagination fix (bug found by the local stack)

- `pkg/components/database/postgres/postgresclient.go`
  - List query compared the resumed pagination cursor against the
    `TIMESTAMPTZ created_at` column using `$5::TIMESTAMP`, which strips
    the timezone. In a non-UTC session timezone postgres reinterpreted
    the naive value in the session's local zone, shifting the comparison
    boundary by the local UTC offset and silently dropping rows from
    page N+1. Fix: cast to `TIMESTAMPTZ`.
  - Symptom: `Applications.Core/containers` LIST returned 10/12 results
    in a local debug stack (33 containers total, page size 10, host TZ
    `America/Los_Angeles`), causing `getGraph` to receive an empty
    resource slice and `Test_ApplicationGraph` to fail.
  - CI never tripped this because the postgres container runs with
    `TZ=UTC` (zero drift on the cast) and per-job clusters rarely
    accumulated more than `top` rows of any one resource type.
- `pkg/components/database/postgres/postgresclient_test.go` —
  regression test that forces `timezone=America/Los_Angeles` on the pool
  and round-trips 5 saved rows through paginated Query calls with
  `MaxQueryItemCount=2`.

### Misc

  `pkg/recipes/engine/engine.go` — small adjustments surfaced while
  running the suites end-to-end (HTTP body re-read idempotency, engine
  double-call guard, env-not-found 404 → no-op on delete).
- `pkg/corerp/frontend/controller/applications/testbicep_scan_test.go`
  (new) — small scan test added during investigation.
- `.gitignore` — ignore local debug artifacts and the local-only
  `bicepconfig.json` override that `make debug-publish-bicep-types`
  writes.

### Documentation

- `docs/contributing/contributing-code/contributing-code-debugging/radius-os-processes-debugging.md`
  - Documents running the Azure cloud suite against the local OS-process
    stack via `make debug-start` + `azure-local-testenv.sh`.
  - Documents the Flux-test limitation under the OS-process flow.
- `docs/contributing/contributing-code/contributing-code-debugging/noncloud-test-learnings.md`
  (new) — long-form notes from stabilizing the noncloud suite against
  the local stack: bugs hit, why CI didn't catch each one, and remaining
  follow-ups (Test_Run_Logger timeout, mechanics-test deadlock,
  dynamicrp type-loader mismatch, namespace audit, recipe-package split
  for parallelism).

## How to use locally

```sh
make debug-start
make debug-install-contour

# corerp-noncloud

# corerp-cloud (Azure) against the same local stack, using host az login
build/scripts/azure-local-testenv.sh all
```

## How tested

Full local runs against an OS-process debug stack on macOS arm64 + k3d:

- `corerp/noncloud` suite: 48 pass, 2 skip, 0 fail in 2m24s wall
  (gotestsum JSONL collected via `GOTESTSUM_JSONFILE_DIR=/tmp/timings`).
- `corerp/cloud/...` Azure suite via `azure-local-testenv.sh run -v`:
  | Test | Status |
  | --- | --- |
  | `Test_AzureConnections` | PASS |
  | `Test_ACI` | PASS |
  | `Test_TerraformRecipe_AzureResourceGroup` | PASS |
  | AWS-only tests (`Test_AWS_*`, `Test_Extender_RecipeAWS_LogGroup`, `Test_AWSRedeploy*`) | SKIP (no AWS creds) |
  | `Test_TerraformPrivateGitModule_KubernetesRedis` | SKIP (no `GH_TOKEN`) |
  | `Test_Storage`, `Test_PersistentVolume` | SKIP (issue #7853, pre-existing) |
- Unit tests for the changed postgres and terraform-provider packages
  pass (`go test ./pkg/components/database/postgres/... ./pkg/recipes/terraform/config/providers/... -count=1`).

## Backwards compatibility

- CI workload-identity / service-principal credentials registered with
  `rad credential register azure …` continue to work exactly as before.
  The `use_cli = true` Terraform fallback only activates when UCP returns
  404 for the credential lookup (i.e. no credential ever registered).
- `RADIUS_INSECURE_LOOPBACK_REGISTRIES`, `RADIUS_SKIP_FLUX_TESTS`, and
  `RAD_CONFIG_FILE` are opt-in env vars; behavior is unchanged when they
  are unset.
- Postgres pagination fix is a strict no-op on UTC-session databases
  (which is what CI and production use); only non-UTC sessions see the
  corrected behavior.

## Type of change

- [x] This pull request is a minor refactor, code cleanup, test improvement, or other maintenance task and doesn't change the functionality of Radius (issue link optional).


build/debug.mk
```diff
@@ -238,10 +322,256 @@ debug-start: debug-setup debug-build-all ## Start k3d cluster and all Radius com
...
+		exit 1; \
+	fi
+	@echo "Installing Flux source-controller v$(DEBUG_FLUX_VERSION) (matches local flux CLI)..."
+	@kubectl config use-context k3d-radius-debug >/dev/null
+	@for i in 1 2 3; do \
+		flux install --namespace=$(DEBUG_FLUX_NAMESPACE) --version=v$(DEBUG_FLUX_VERSION) \
+			--components=source-controller --network-policy=false && \
+		kubectl wait --for=condition=available deployment \
+			-l app.kubernetes.io/component=source-controller \
+			-n $(DEBUG_FLUX_NAMESPACE) --timeout=120s && break; \
```
lakshmimsft, line 460: this will continue with success mesaage after all 3 retries fail. Should we put a warning message that Flux tests will not work as Flux source-controller failed to install after 3 attempts or we do a hard fail?

pkg/recipes/engine/engine.go
```diff
@@ -138,6 +139,14 @@ func (e *engine) deleteCore(ctx context.Context, recipe recipes.ResourceMetadata
 	logger := ucplog.FromContextOrDiscard(ctx)
 	configuration, err := e.options.ConfigurationLoader.LoadConfiguration(ctx, recipe)
 	if err != nil {
+		// If the environment has already been deleted (e.g. test cleanup deleted
```
lakshmimsft, line 142: LoadConfiguration makes 2 calls to fetch environment and application. the 404 returned could be from application and this code will misinterpret it as env does not exist and vice versa. Is this change needed since I think ordering was addressed elsewhere in this PR ?

test/rp/rptest.go
```diff
@@ -508,7 +508,13 @@ func (ct RPTest) Test(t *testing.T) {
...
 		}
 
-		for _, resource := range step.RPResources.Resources {
+		// Delete resources in reverse declaration order so that the Environment
+		// (typically declared first) is deleted last. Otherwise the Environment
+		// can be removed while child resources (Applications/Containers/portable
+		// resources) are still mid-cascade, causing their recipe-driven cleanup
+		// to fail loading the (now-gone) environment configuration.
+		for i := len(step.RPResources.Resources) - 1; i >= 0; i-- {
+			resource := step.RPResources.Resources[i]
```
lakshmimsft, line 517: could we change this to,say, check type of resource instead.. this approach assumes tests have resources listed in a particular order which may/may not be correct now/in the future.

test/functional-portable/corerp/util.go
```diff
@@ -63,3 +94,19 @@ func GetSecretSuffix(resourceID, envName, appName string) (string, error) {
...
+func rewriteResourceGroup(resourceID, rg string) string {
+	parsed, err := resources.ParseResource(resourceID)
+	if err != nil {
+		return resourceID
+	}
+	current := parsed.FindScope(resources_radius.ScopeResourceGroups)
+	if current == "" || current == rg {
+		return resourceID
+	}
+	// Case-insensitive replacement of the resourcegroups segment value while preserving the
```
lakshmimsft, line 109: nit: comment is incorrect. strings.ReplaceAll() is a case-sensitive replacement...but since current is derived from resourceID, we should not have a case mismatch.

build/debug.mk
```diff
@@ -209,14 +273,34 @@ debug-build-rad: ## Build rad CLI with debug symbols + create drad alias
...
+		running=$$(echo "$$servers" | cut -d/ -f1); \
+		if [ "$$running" = "0" ]; then \
+			echo "k3d cluster 'radius-debug' exists but is stopped — starting it..."; \
+			k3d cluster start radius-debug; \
+		else \
+			echo "k3d cluster 'radius-debug' already running ($$servers servers)"; \
+		fi; \
 	else \
+		echo "Creating k3d cluster 'radius-debug'..."; \
 		k3d cluster create radius-debug --api-port 0.0.0.0:6443 --wait --timeout 60s; \
```
lakshmimsft, line 288: running this i hit an error with port already in use. should we consider updating this to get next available port if this is in use?

build/debug.mk
```diff
@@ -238,10 +322,256 @@ debug-start: debug-setup debug-build-all ## Start k3d cluster and all Radius com
...
+	echo "❌ tf-module-server port-forward not reachable on localhost:$(DEBUG_TF_MODULE_LOCAL_PORT)"; \
+	tail -20 $(DEBUG_DEV_ROOT)/logs/tf-module-server-pf.log 2>/dev/null || true; \
+	exit 1
+
+debug-stop-tf-module-server: ## Tear down the Terraform module server port-forward
+	@pkill -f "kubectl.*port-forward.*$(DEBUG_TF_MODULE_DEPLOYMENT).*$(DEBUG_TF_MODULE_LOCAL_PORT):" >/dev/null 2>&1 || true
+	@echo "✅ tf-module-server port-forward stopped"
+
+debug-publish-bicep-types: ## Build radius+testresources bicep extensions as local .tgz files and drop a closer-wins bicepconfig override next to the test templates
+	@command -v bicep >/dev/null 2>&1 || { echo "❌ bicep CLI not found on PATH (install via: az bicep install / brew install bicep)"; exit 1; }
```
lakshmimsft, line 527: debug-publish-bicep-types is the last step of debug-start, and it hard-requires a standalone bicep on PATH. But bicep is not listed in the doc's "Required Tools" and not checked by debug-check-prereqs (which does check yq). Result: a fresh setup builds the entirestack, then fails at the very end. 
Consider either we add bicep to debug-check-prereqs (fail fast, with the other tools) and to the doc prerequisites or have this step auto-install via az bicep install / rad bicep download.

docs/contributing/contributing-code/contributing-code-debugging/radius-os-processes-debugging.md
lakshmimsft, line 83: Running the doc end-to-end requires three tools/conditions not in this list:

  - yq — required and enforced by make debug-check-prereqs (debug.mk:174), but missing here.
  - bicep — required by debug-publish-bicep-types (debug.mk:527), but missing here.
  - PostgreSQL must be running (not just installed) before debug-start; the container/service has to be up or prereqs fail.

 Please add these.

PR #11907, opened 2026-05-15
Feature spec: Dynamic Bicep Extensions Served by the Radius Control Plane

zachcasper:
# Description

Add a design document proposing that the Radius control plane become the single source of truth for Bicep extensions, replacing the current dual-pipeline model (release-time ACR publishes + manual `rad bicep publish-extension`).

Under the proposal:

- A new control-plane API returns a Bicep extension artifact computed on demand from the currently-registered resource types. No new Ingress, OCI registry, DNS, or TLS provisioning is required to expose it; it reuses the auth path `rad` already uses.
- A new `rad resource-type update` CLI command fetches that artifact and writes it to `~/.rad/bicep-extension.tgz`. `rad deploy` and `rad run` refresh it as a side effect, closing the staleness window in the normal authoring → deploy loop.
- `rad init` becomes kubecontext-aware: it installs Radius if the current kubecontext has none, otherwise skips the install step. In both cases it writes `~/.rad/bicep-extension.tgz` and creates or merges `~/bicepconfig.json` with a single `extensions.radius` entry pointing at that file. Bicep's existing parent-directory `bicepconfig.json` resolution covers every `.bicep` file under `$HOME` automatically.
- The four split `radius*` aliases (`radius`, `radiusCompute`, `radiusData`, `radiusSecurity`) are consolidated into a single `radius` alias; legacy entries in an existing `~/bicepconfig.json` are removed by `rad init` as an intentional forcing function.
- `rad bicep publish-extension` is removed from the CLI. Previously-published artifacts on `biceptypes.azurecr.io` remain readable for users on older `rad` versions.

The document records the four other delivery models that were evaluated (long-running proxy daemon, Ingress + anonymous OCI endpoint, Kubernetes aggregation layer, and control-plane-pushes-to-customer-registry) and why each was rejected for v1.

No code changes; design document only.

## Type of change

- This pull request is a design document and only includes files in the eng/design-notes directory.

Fixes: #_TBD_

## Contributor checklist

    - [x] Not applicable
    - [x] Yes
    - [ ] Yes
    - [x] Not applicable _(pending review on this PR)_
    - [x] Not applicable
    - [x] Not applicable
    - [x] Not applicable _(follow-up docs PR required when implementation lands)_


eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
```diff
@@ -0,0 +1,301 @@
...
+
+Developer (first-time on a workstation):
+
+1. Install the `rad` CLI.
+2. Ensure `kubectl` is pointed at the cluster where the platform engineer installed Radius, then run `rad init`. Radius is already installed, so `rad init` skips the install step and just writes `~/.rad/bicep-extension.tgz` and merges `~/bicepconfig.json`.
+3. Open any `.bicep` file under `$HOME` in VS Code with the Bicep extension installed. Type-checking and property completion work immediately.
+
+Developer (steady state, type set changes server-side):
+
+1. Run `rad resource-type update` to refresh now, or just run the next `rad deploy` / `rad run`, which refresh `~/.rad/bicep-extension.tgz` as a side effect.
```
Reshrahim: I like the idea of having a separate command that imports the server-side type library to developer workstations, but I think it should be separate from the `rad resource-type` CRUDL command.

`rad resource-type update` implies that the developer is updating the type schema on the server. In an enterprise setup, platform engineers would typically use RBAC to restrict who can create or update resource type definitions.

I think it would be better to call this something like `rad extension update` or `rad extension import` implying what the command is actually doing - pulling the extension from the Radius control plane and updating the local extension configuration on the developer machine.

It also maps to the existing `extension radius` experience in application definition, so the mental model stays the same .

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Scenario 3: Platform engineer onboards a developer to a new installation in one step (P2)
+
+A new developer's setup is: install the `rad` CLI, then `rad init`. They never receive `.tgz` files, never log into a Bicep types registry, and never edit `bicepconfig.json`. The single `radius` alias suffices for both built-in and user-defined types.
+
+## Key dependencies and risks
+
+* **Dependency – Bicep CLI local-file extension references and parent-directory `bicepconfig.json` resolution.** The integration point is Bicep's existing support for a local-file `extensions` entry plus its existing parent-walk to find `bicepconfig.json`. Bicep's walk reaches `$HOME` for any `.bicep` source under `$HOME`, but does not search `~/.bicep/` or XDG paths. No change to the upstream Bicep CLI or Bicep VS Code extension is required. Risk: if Bicep changes its parent-walk semantics, every workstation breaks. Mitigation: this is documented Bicep behavior; pin Bicep version testing in CI.
```
Reshrahim, line 222: As I understand here - `bicepconfig.json` will still be maintained at the repo level and the extensions in `.rad/` will be refreshed by `rad init` or an explicit command to update extensions locally. We maintain the `bicepconfig.json` outside of `.rad/` because of Bicep's config resolution?

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
```diff
@@ -0,0 +1,301 @@
...
+* **Dependency – `rad` is installed on every workstation authoring Bicep against Radius.** Effectively already true (no `rad`-less deploy flow today), but becomes a hard prerequisite under this feature.
+* **Dependency – Authentication path `rad` already uses for control-plane calls.** The new extension API reuses the same auth. The Bicep CLI itself never authenticates to the control plane (it only sees the local file).
+* **Risk – Stale `~/.rad/bicep-extension.tgz` on a workstation that has not run `rad` since the control plane changed.** Mitigated by routine `rad` commands (`deploy`, `run`) refreshing as a side-effect and by `rad resource-type update` being a single command.
+* **Risk – `.bicep` files outside `$HOME` are not covered automatically.** Mitigated by documentation; advanced users place a `bicepconfig.json` somewhere in their source tree by hand.
+* **Risk – Devcontainer / Codespaces / remote-SSH workspaces have a different `$HOME`.** `rad init` must be re-run inside the container; the artifact is per-environment. Documented limitation.
```
Reshrahim, line 228: Would this be a risk if `bicepconfig.json` is maintained at per repo level ?

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: No, rad init will now write bicepconfig.json in the users' home directory. The contents will be:
```
{
  "extensions": {
    "radius": ".rad/bicep-extension.tgz"
  }
}
```
The downloaded Bicep extension will be ~/.rad/bicep-extension.tgz.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: If a is running in a dev container (or another workstation for that matter), they will not have the Bicep extension in their environment. This is obvious and a somewhat silly risk to be documented.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: Yes, I debated many options here. I'm open to suggestions. Here is what was considered:

# `rad resource-type update` — command-naming debate

## Options considered

| # | Candidate | Pros | Cons |
|---|---|---|---|
| 1 | **`rad bicep extension sync`** (original) | Most precise; explicitly names artifact + namespace. | Verbose. Only makes sense if more `rad bicep extension <verb>` siblings (`show`, `path`) are likely. |
| 2 | **`rad bicep sync`** | Shorter; discoverable next to other `rad bicep …` commands. | Still ties the verb to the artifact format. Hides that the catalog (not the file) is the real input. |
| 3 | **`rad sync`** | Reads cleanly; leaves room for future syncs. | Vague; future-proofing was speculative (YAGNI). No concrete second target identified. |
| 4 | **`rad resource-type sync`** | Names *what* is synced (the catalog), not the file format. Mirrors `rad resource-type create`. Leaves room for non-Bicep outputs (Terraform schema, LSP index) without renaming. | `sync` has weak prior art for schema/catalog caches — the ecosystem (`helm repo update`, `apt update`, `brew update`, `terraform init -upgrade`) overwhelmingly uses `update`. |
| 5 | **`rad resource-type refresh`** | Semantically precise; no collision. | Less familiar verb than `sync` or `update` in modern CLIs. |
| 6 | **`rad extension update`** | Short; uses the established `update` verb. | "extension" collides hard with Bicep's own vocabulary (`extension` keyword, `extensions` block in `bicepconfig.json`, "extension types"); ambiguous with a future `rad` plugin system. Names a file artifact rather than the catalog. |
| 7 | **`rad resource-type update`** (selected) | Names the catalog, not the artifact. Mirrors `rad resource-type create` (symmetric: platform engineer creates server-side, developer updates client-side). Uses `update` — the established verb in this ecosystem for "pull fresh catalog from server." Singular form matches existing CLI conventions. Preserves optionality for future outputs (Terraform, LSP) under the same noun. | Verb collides loosely with the idea of "modify a resource type" — mitigated because schema edits happen via `rad resource-type create -f manifest.yaml` in practice, not via an `update` subcommand. |

## Why option 7 won

1. **Catalog, not artifact.** The thing being kept fresh is the resource-type catalog. The `.tgz` is one current serialization; in the future the same command can emit other formats.
2. **Symmetric with `rad resource-type create`.** Platform engineer creates types; developer updates the local mirror. Same noun on both sides.
3. **`update` is the dominant verb in the ecosystem.** `helm repo update`, `apt update`, `brew update`, `terraform init -upgrade`. `sync` is mostly used for filesystem mirrors (`rclone sync`, `aws s3 sync`) and fork-to-upstream (`gh repo sync`), neither of which matches our shape.
4. **Avoids the overloaded word "extension."** Reserves that term for Bicep's own meaning and for a future `rad` plugin namespace.
5. **Lets `rad init` absorb the first-run case.** Because `rad init` became kubecontext-aware in the same change, `rad resource-type update` is only needed for explicit refresh — the everyday developer never has to type it, which lowers the cost of choosing a slightly longer name.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
Reshrahim: Oh I was thinking about this from Repo Radius perspective. How this would be created and maintained for repositories with radius installed? May be some clarity around that would help.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: I changed the command to `rad resource-type sync`. You're right that update should be reserved for CRUDL control plane operations. Sync infers updating the local environment based on the control plane.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: Please suggest what clarity should be added, if any. If Radius was running in a CD runner, all of the above still applies. The user's home directory would have the bicepconfig.json and any Git repository should be cloned in a directory in the user's home directory, as is typical.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
```diff
@@ -0,0 +1,301 @@
...
+
+> As a developer, the type set I'm allowed to use is whatever my platform engineer told me to put in my `bicepconfig.json` — and I have no way to confirm it matches what the control plane will actually accept until I try to deploy. Onboarding to a new Radius installation is a docs-following exercise: install `rad`, then log into some company-specific OCI registry, then paste a snippet into `bicepconfig.json` for each of half a dozen aliases. When the platform team changes a type, nobody tells me, and my editor stops being a reliable source of truth.
+
+## Desired user experience outcome
```
nicolejms, line 118: To clarify - we will overwrite all files in this directory with current (without comparing files) so any changes that are breaking show up immediately and not during deploy time.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
```diff
@@ -0,0 +1,301 @@
...
+
+This feature makes the Radius control plane the single source of truth for Bicep extensions. A new `rad resource-type sync` command (also invoked implicitly by `rad init`, `rad deploy`, and `rad run`) fetches the control plane's currently-registered types, packages them into a Bicep extension, and writes it to `~/.rad/bicep-extension.tgz`. `rad init` becomes Kubernetes context aware: if no Radius control plane is detected in the current Kubernetes context, it installs Radius (today's behavior); if one is already installed, it skips the install step. In both cases it creates or merges `~/bicepconfig.json` with a single `radius` extension entry pointing at that file. Bicep's existing parent-directory resolution then picks the configuration up for any `.bicep` file under the home directory that does not have a closer `bicepconfig.json` of its own — no per-repository setup, no external OCI registry, and no separate publish step is required. Repositories that already ship their own `bicepconfig.json` are unaffected by the home-directory file (see edge cases below); the expected migration is for those repos to add the same `radius` entry or remove their local config.
```
nicolejms, line 9: should we use 'update' for consistency instead of 'sync'?

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
Reshrahim: Today we maintain the `bicepconfig.json` at each repository level (docs, radius) to run Radius CI tests and in the case of Repo Radius too, we might maintain this config in the `~/.radius` folder of a repository to compile and deploy the `app.bicep` with custom resource ty[es

With this proposal, there is no need to have this file at repo level and the workflows can automatically sync the `bicepconfig.json` before deploy. Is that correct ?

zachcasper: Sounds correct.

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: There were a number of verbs considered (see [this comment](https://github.com/radius-project/radius/pull/11907/#discussion_r3249654149)). We landed on the verb `sync` meaning to download information from the control plane to the local workstation. This is distinct from `update` which is a control plane CRUDL operation updating the control plane (e.g., the other direction).

eng/design-notes/bicep/2026-05-dynamic-bicep-extensions.md
zachcasper: Two files are in play:

`~/bicepconfig.json`: Never overwritten, the `radius` extension is upserted.

`~/.rad/bicep-extension.tgz`: Always overwritten with latest from control plane

This happens every time a user runs rad init, rad deploy, rad run, or rad resource-types sync.

PR #11911, opened 2026-05-15
Automate default resource type registration from resource-types-contrib

kachawla:
## Overview

Implement automated sync of default resource type manifests from `resource-types-contrib` into the `radius` repo, as described in the design: https://github.com/radius-project/radius/pull/11610/changes.

This eliminates manual cross-repo file editing and schema drift by making `resource-types-contrib` the single source of truth for resource type manifests that ship as defaults in Radius.

## How it works

There are two pieces: a **manual Makefile target** that copies manifests, and a **CI check** that verifies the copies are up to date.

### 1. Manual sync (developer runs locally)

When a developer wants to update the default resource types - either to pick up schema changes from `resource-types-contrib` or to add/remove a default type - they run:

```bash
make update-resource-types   # bumps go.mod to latest resource-types-contrib, then copies manifests
```

`make sync-resource-types` (copy only, no version bump) is useful when:
- A new entry was added to `defaults.yaml` without changing the `resource-types-contrib` version
- A previously-copied manifest was accidentally hand-edited and needs to be restored
- CI reports drift and the developer needs to regenerate the copies

```bash
make sync-resource-types     # copies manifests from the version already pinned in go.mod
```

This reads `deploy/manifest/defaults.yaml` to determine which resource types to copy, resolves each entry to a file in the `resource-types-contrib` Go module cache, and copies it into `deploy/manifest/built-in-providers/{dev,self-hosted}/`. It also removes any stale previously-copied files that are no longer in `defaults.yaml`. The developer then commits the resulting changes.

**There is no automated workflow that does the copying** - it is always a deliberate local action followed by a PR.

### 2. CI drift detection (runs automatically on PRs)

The `verify-resource-types.yaml` workflow runs on every PR that touches `go.mod`, `go.sum`, `defaults.yaml`, the copied manifests, or the sync logic. It:

1. Re-runs `make sync-resource-types` in CI (copy only, no version bump).
2. Checks `git diff` on the `built-in-providers/` directories.
   - **No diff** → the committed copies match the pinned version → check **passes**.
   - **Diff found** → the committed copies have drifted → check **fails** with a message telling the developer to run `make sync-resource-types` and commit.

This catches cases where someone bumps `go.mod` but forgets to re-sync, manually edits a copied file, or adds an entry to `defaults.yaml` without re-running the copy.

## Key design points

- **Version pinning via `go.mod`**: `resource-types-contrib` is added as a Go module dependency. A blank import in `pkg/resourcetypescontrib/import.go` keeps `go mod tidy` from removing it. The version pinned in `go.mod` is the single source of truth for which version of manifests are used.
- **No runtime changes**: The existing `RegisterDirectory` startup path is reused as-is. Copied manifests have no `location` field, so UCP routes via `DefaultDownstreamEndpoint` (dynamic-rp).
- **Per-type file layout**: Previously, `built-in-providers/` contained one YAML file per namespace (e.g., `radius_compute.yaml` with `containers`, `routes`, and `persistentVolumes` all in one file). Now each resource type has its own file (e.g., `containers.yaml`, `routes.yaml`). This matches the per-type file layout in `resource-types-contrib`, makes it straightforward to map `defaults.yaml` entries to files, and keeps diffs scoped to the individual type that changed.

### New files
- **`deploy/manifest/defaults.yaml`** - Lists which resource types from `resource-types-contrib` ship as defaults.
- **`build/resource-types.mk`** - Makefile targets:
  - `make update-resource-types` - Bumps `go.mod` to the latest `resource-types-contrib` version and syncs manifest files.
  - `make sync-resource-types` - Copies manifest files from the version pinned in `go.mod` (no version bump). Used by CI for drift detection.
- **`.github/workflows/verify-resource-types.yaml`** - CI drift detection: runs `make sync-resource-types` and fails if committed copies diverge from the pinned version.
- **`pkg/resourcetypescontrib/import.go`** - Blank import to retain the `resource-types-contrib` dependency in `go.mod`.
- **Per-type manifest files** in both `dev/` and `self-hosted/`: `containers.yaml`, `persistentVolumes.yaml`, `routes.yaml`, `mySqlDatabases.yaml`, `secrets.yaml`

### Removed files
- `radius_compute.yaml` (dev + self-hosted) - Replaced by per-type files: `containers.yaml`, `persistentVolumes.yaml`, `routes.yaml`
- `radius_data.yaml` (dev + self-hosted) - Replaced by per-type file: `mySqlDatabases.yaml`
- `radius_security.yaml` (dev + self-hosted) - Replaced by per-type file: `secrets.yaml`

### Modified files
- **`Makefile`** - Includes `build/resource-types.mk`
- **`go.mod`** - Adds `resource-types-contrib` dependency (pinned to `v0.0.0-20260515162318-067b29d71eb8`)

## Dependencies

- Depends on https://github.com/radius-project/resource-types-contrib/pull/158 (adds `go.mod` and `doc.go` to `resource-types-contrib`) - **merged**

## Test plan

- `make sync-resource-types` copies exactly the files listed in `defaults.yaml`
- Running twice produces no diff (idempotent)
- Adding a missing path to `defaults.yaml` causes the script to fail clearly
- Existing `Test_ResourceProvider_RegisterManifests` continues to pass
- CI drift detection catches manual edits or missed syncs

### DIFF
--- a/.github/workflows/verify-resource-types.yaml
+++ b/.github/workflows/verify-resource-types.yaml
@@ -52,8 +52,7 @@ on:
       - 'Makefile'
       - 'build/resource-types.mk'
 
-permissions:
-  contents: read
+permissions: {}
 
 concurrency:
   group: verify-resource-types-${{ github.event.pull_request.number || github.sha }}
@@ -63,6 +62,8 @@ jobs:
   verify-resource-types:
     name: Verify resource type copies are in sync
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/verify-resource-types.yaml
+++ b/.github/workflows/verify-resource-types.yaml
@@ -1,0 +1,98 @@
+# ------------------------------------------------------------
+# Copyright 2023 The Radius Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------
+
+# This workflow verifies that the resource type manifest files committed in
+# deploy/manifest/built-in-providers/ match the versions in the
+# resource-types-contrib module pinned in go.mod.
+#
+# It runs "make sync-resource-types", which re-copies the manifest files from
+# the Go module cache, and then checks for any diff.
+#
+# - If there is no diff, the committed copies are in sync and the check passes.
+# - If there is a diff, the committed copies have drifted from the pinned
+#   version and the check fails. The developer must run
+#   "make sync-resource-types" locally and commit the updated files.
+#
+# This catches:
+#   - A go.mod bump without re-running "make sync-resource-types"
+#   - Manual edits to copied manifest files that drift from upstream
+#   - New entries in defaults.yaml without re-syncing
+#
+# Similar in spirit to the "go mod tidy" check, ensuring generated/copied
+# artifacts are consistent with their source.
+
+name: Verify resource type manifests
+
+on:
+  pull_request:
+    branches:
+      - main
+      - release/*
+    paths:
+      # Trigger when the default list, the copied manifests, the dependency
+      # version (go.mod/go.sum), or the sync logic (Makefile, resource-types.mk)
+      # changes.
+      - 'go.mod'
+      - 'go.sum'
+      - 'deploy/manifest/defaults.yaml'
+      - 'deploy/manifest/built-in-providers/**'
+      - 'Makefile'
+      - 'build/resource-types.mk'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: verify-resource-types-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  verify-resource-types:
+    name: Verify resource type copies are in sync
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: go.mod
+
+      - name: Install yq
+        run: |
+          mkdir -p "${RUNNER_TEMP}/bin"
+          GOBIN="${RUNNER_TEMP}/bin" go install github.com/mikefarah/yq/v4@v4.44.3
+          echo "${RUNNER_TEMP}/bin" >> "${GITHUB_PATH}"
+
+      # Re-run the copy step (without bumping the version) to regenerate the
+      # manifest files from the module version pinned in go.mod.
+      - name: Sync resource types from pinned version
+        run: make sync-resource-types
+
+      # If the regenerated files differ from what's committed, the PR has drift.
+      - name: Check for drift
+        run: |
+          if ! git diff --exit-code deploy/manifest/built-in-providers/; then
+            echo ""
+            echo "ERROR: Resource type manifest copies are out of sync with the pinned"
+            echo "resource-types-contrib version in deploy/manifest/defaults.yaml."
+            echo ""
+            echo "Run 'make sync-resource-types' and commit the updated files."
+            exit 1
+          fi
+          echo "Resource type manifest copies are in sync."
--- a/build/resource-types.mk
+++ b/build/resource-types.mk
@@ -1,0 +1,142 @@
+# ------------------------------------------------------------
+# Copyright 2023 The Radius Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------
+
+# resource-types.mk provides targets for synchronizing default resource type
+# manifests from the resource-types-contrib repository.
+#
+# resource-types-contrib is added to go.mod as a Go module dependency. It
+# contains no executable Go code. We depend on it purely to leverage Go's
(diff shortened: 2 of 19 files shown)


deploy/manifest/defaults.yaml
```diff
@@ -0,0 +1,20 @@
...
+#   1. Merge the manifest in resource-types-contrib.
+#   2. Add the entry below.
+#   3. Run: make update-resource-types
+#   4. Commit the updated files (defaults.yaml, go.mod, go.sum, and copied YAMLs).
+defaultRegistration:
+  - Radius.Compute/containers
+  - Radius.Compute/persistentVolumes
+  - Radius.Compute/routes
+  - Radius.Data/mySqlDatabases
+  - Radius.Security/secrets
```
nithyatsu, line 20: I just checked we have only these today for constructing the recipepack. We might might have to remove mySqlDatabases too, and bring it in with the recipepack automation changes. 
	return []CoreTypesRecipeInfo{
		{
			ResourceType:   "Radius.Compute/containers",
			RecipeLocation: "ghcr.io/radius-project/kube-recipes/containers:" + tag,
		},
		{
			ResourceType:   "Radius.Compute/persistentVolumes",
			RecipeLocation: "ghcr.io/radius-project/kube-recipes/persistentvolumes:" + tag,
		},
		{
			ResourceType:   "Radius.Compute/routes",
			RecipeLocation: "ghcr.io/radius-project/kube-recipes/routes:" + tag,
		},
		{
			ResourceType:   "Radius.Security/secrets",
			RecipeLocation: "ghcr.io/radius-project/kube-recipes/secrets:" + tag,
		},
	}

pkg/resourcetypescontrib/import.go
```diff
@@ -0,0 +1,30 @@
...
+// Package resourcetypescontrib exists solely to keep the
+// github.com/radius-project/resource-types-contrib dependency in go.mod.
+//
+// resource-types-contrib is a Go module containing YAML manifest files (no
+// executable Go code). The blank import below prevents "go mod tidy" from
+// removing the dependency. The actual manifest files are copied from the
+// module cache into deploy/manifest/built-in-providers/ by
+// "make sync-resource-types".
+package resourcetypescontrib
+
```
nithyatsu, line 24: 👍

pkg/ucp/integrationtests/resourceproviders/resourceproviders_test.go
```diff
@@ -161,6 +161,34 @@ func Test_ResourceProvider_RegisterManifests(t *testing.T) {
 	deleteManifestResourceProvider(server)
 }
 
+// Test_ResourceProvider_RegisterManifests_NoLocation verifies that manifests
+// without a "location" field are registered successfully at startup. This is
+// the code path used by resource type manifests copied from
+// resource-types-contrib, which omit location so that UCP routes requests via
+// DefaultDownstreamEndpoint (dynamic-rp).
+func Test_ResourceProvider_RegisterManifests_NoLocation(t *testing.T) {
```
nithyatsu, line 169: I am not sure I understand this change. Were we just missing a testcase? Is it the same location that is set to global by default?

build/resource-types.mk
```diff
@@ -0,0 +1,137 @@
...
+		exit 1; \
+	fi && \
+	echo "  Module directory: $$MODULE_DIR" && \
+	for entry in $$(yq '.defaultRegistration[]' $(DEFAULTS_YAML)); do \
+		rel_path=$$(echo "$$entry" | sed 's/^Radius\.//') && \
+		type_name=$$(echo "$$rel_path" | cut -d'/' -f2) && \
+		src_path="$$MODULE_DIR/$$rel_path/$$type_name.yaml" && \
+		if [ ! -f "$$src_path" ]; then \
+			echo "ERROR: File not found: $$src_path (from entry '$$entry')"; \
+			echo "       Verify the entry in $(DEFAULTS_YAML) and the resource-types-contrib version."; \
```
nithyatsu, line 101: The resource-types-c0ntrib version is always latest (for now) correct?

nithyatsu (dismissed):

.github/workflows/verify-resource-types.yaml
```diff
@@ -0,0 +1,98 @@
...
+#
+# This catches:
+#   - A go.mod bump without re-running "make sync-resource-types"
+#   - Manual edits to copied manifest files that drift from upstream
+#   - New entries in defaults.yaml without re-syncing
+#
+# Similar in spirit to the "go mod tidy" check, ensuring generated/copied
+# artifacts are consistent with their source.
+
+name: Verify resource type manifests
```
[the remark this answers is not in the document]
kachawla: We have other workflows in the repo use human-readable names ('Check code is up-to-date' in lint.yaml, 'Build and Test' in build.yaml). Keeping name more verbose and readable.

deploy/manifest/defaults.yaml
kachawla: thanks for catching and fixing this. It was a miss when I copied over mysql manifest manually

pkg/ucp/integrationtests/resourceproviders/resourceproviders_test.go
kachawla: This is the routing location for UCP to route the request to dynamic rp vs app rp. Before this change all manifests in built-in-providers/ had an explicit location field (global: 'http://localhost:8082' for example). The manifests in resource-types-contrib do not include location and we intentionally don't add it on top when copying them over. Because when there's no location, UCP falls back to DefaultDownstreamEndpoint which points to dynamic-rp, since this is the endpoint where we want all RRTs to be routed to adding it is redundant. 

The behavior is the same, but the mechanism is different (fallback routing vs explicit address), so the test was added to verify it works.

build/resource-types.mk
kachawla: go.mod pins a specific version ([currently v0.0.0-20260515162318-067b29d71eb8](https://github.com/radius-project/radius/blob/755bd82c1989a0b933380a91725e615bee27a934/go.mod#L84)). `make update-resource-types` uses @latest  in the go get command to resolve the newest available version at that point in time, then pins it in. But after that it's pinned. `make sync-resource-types` (used by CI) always uses whatever version is in go.mod, not latest.

.github/workflows/verify-resource-types.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - 'go.mod'
+      - 'go.sum'
+      - 'deploy/manifest/defaults.yaml'
+      - 'deploy/manifest/built-in-providers/**'
+      - 'Makefile'
+      - 'build/resource-types.mk'
+
+permissions:
+  contents: read
+
```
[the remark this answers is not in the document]
kachawla: Fixed - moved permissions to job level with permissions: {} at workflow level.

pkg/ucp/integrationtests/resourceproviders/resourceproviders_test.go
```diff
@@ -161,6 +161,34 @@ func Test_ResourceProvider_RegisterManifests(t *testing.T) {
...
+	}))
+	defer server.Close()
+
+	createRadiusPlane(server)
+
+	noLocationNamespace := "Radius.Compute"
+	noLocationResourceProviderURL := "/planes/radius/local/providers/System.Resources/resourceproviders/" + noLocationNamespace + radiusAPIVersion
+	noLocationResourceTypeURL := "/planes/radius/local/providers/System.Resources/resourceproviders/" + noLocationNamespace + "/resourcetypes/containers" + radiusAPIVersion
+
+	require.EventuallyWithTf(t, func(collect *assert.CollectT) {
```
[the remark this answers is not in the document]
kachawla: Updated

nithyatsu (approved):

PR #11915, opened 2026-05-15
Integrate contrib types into unified Bicep extension

kachawla:
## Overview

Today, resource types from `resource-types-contrib` (e.g. `Radius.Compute/containers`, `Radius.Security/secrets`) are published as separate Bicep extensions (`radiusCompute`, `radiusData`, `radiusSecurity`), requiring users to configure multiple entries in `bicepconfig.json` and use different `extension` directives depending on the namespace. This PR is part of a series that consolidates them into the existing `radius` Bicep extension so users have a single `extension radius` for all Radius-authored types.

Specifically, this PR wires up the build pipeline so that contrib resource type manifests (maintained via [`make update-resource-types`](https://github.com/radius-project/radius/pull/11911)) are included in the unified `radius` extension alongside the existing TypeSpec/Swagger-generated types.

## What this PR does

1. **`generate-bicep-types-contrib` Makefile target** -- reads `deploy/manifest/defaults.yaml` to discover which contrib namespaces to include and passes all per-type manifests for each namespace to `manifest-to-bicep generate` for merging into `types.json` + `index.json` + `index.md`. Creates empty `docs/` directories so the [doc generation script](https://github.com/radius-project/docs/blob/main/.github/scripts/generate_resource_references.py) doesn't crash (contrib resource type docs tracked in [#11918](https://github.com/radius-project/radius/issues/11918)).
2. **`publish-bicep-extension` Makefile target** -- wraps `bicep publish-extension` for local testability.
3. **Shared `index-builder.ts` module** -- extracts `buildTypeIndex()` from `generate.ts` into a shared module so both the autorest pipeline and the new `rebuild-index` CLI can use it without code duplication.
4. **`rebuild-index.ts` CLI wrapper** -- standalone entry point that rebuilds the unified `index.json`/`index.md` after contrib `types.json` files are added to the generated tree.
5. **Extended `generate-bicep-types` target** -- now calls `generate-bicep-types-contrib` then `rebuild-index` after the existing autorest step, so the full pipeline produces one unified extension.
6. **Removed contrib extension references** -- `radiusCompute`, `radiusData`, `radiusSecurity` removed from `bicepconfig.json`, `install-bicep.sh`, and `pkg/cli/setup` since these namespaces are now part of the single `radius` extension.
[...]

PR #11916, opened 2026-05-15
Add workflow to sync contrib resource types and publish Bicep extensions

### LINKED ISSUE #11487: Automate publishing extensions

kachawla:
## Overview

Today the `radius` Bicep extension is published to `biceptypes.azurecr.io` via the existing [`build-and-push-bicep-types`](https://github.com/radius-project/radius/blob/main/.github/workflows/build.yaml#L356) job in `build.yaml`, which dispatches to the `radius-publisher` pipeline on every push to `main` and on version tag pushes. With [#11915](https://github.com/radius-project/radius/pull/11915) updating `make generate-bicep-types` to include contrib types, the existing publish pipeline automatically produces the combined extension -- no new publish workflow is needed.

However, there is no automation to pull updated resource type manifests from `resource-types-contrib` into this repo. When someone merges a schema change or a new resource type in contrib, the manifest copies committed under `deploy/manifest/built-in-providers/` must be refreshed manually via `make update-resource-types` before the next publish picks them up.

This PR adds a workflow that closes that gap by automating the manifest sync.

## How it works

```
resource-types-contrib merges to main
  |
  +--> notify-radius.yaml (contrib repo, PR 4) fires repository_dispatch
         |
         +--> contrib-update-resource-types.yaml (this PR) receives dispatch
                |
                +--> Runs 'make update-resource-types' to refresh manifest copies
                +--> Opens/updates PR on bot/update-resource-types branch
                       |
                       +--> Human reviews and merges the PR
                              |
                              +--> Push to main triggers build.yaml's existing
                                   build-and-push-bicep-types job
                                     |
                                     +--> Dispatches to radius-publisher
                                     +--> radius-publisher runs make generate-bicep-types
                                          (now includes contrib) and publishes
                                          radius:latest to biceptypes.azurecr.io
```

### `contrib-update-resource-types.yaml`

Handles `repository_dispatch` events (type: `resource-types-contrib-updated`) from `resource-types-contrib`.

**Triggers:**
- `repository_dispatch` -- fired by the contrib repo's `notify-radius.yaml` workflow (PR 4)
- `workflow_dispatch` -- commented out for production, can be enabled during development

**Steps:**
1. Validates `contrib_ref` as a hex commit SHA (informational only -- the actual version fetched is determined by `make update-resource-types` which runs `go get ...@latest`)
2. Installs yq (required by `make update-resource-types` to parse `defaults.yaml`)
3. Runs `make update-resource-types` to bump `go.mod` to latest contrib and copy manifests
4. If changes are detected (using `git status --porcelain` to catch both modified and new untracked files), opens or updates a PR on the `bot/update-resource-types` branch
5. Merging that PR triggers the existing publish pipeline to republish `radius:latest`

**Security:**
- `contrib_ref` is validated against `^[a-f0-9]{7,40}$` and passed via environment variables (not inline `${{ }}` interpolation) to prevent shell and script injection
- Uses `GH_RAD_CI_BOT_PAT` for checkout and PR creation so the resulting push triggers CI checks (the default `GITHUB_TOKEN` cannot trigger workflows on pushes it creates)

**Note:** This workflow depends on `make update-resource-types` from [#11911](https://github.com/radius-project/radius/pull/11911). It includes a pre-flight check that fails fast with a descriptive error if the target is not yet available.

## Dependencies

- [Integrate contrib types into unified Bicep extension](https://github.com/radius-project/radius/pull/11915)
- [Automated default resource type registration](https://github.com/radius-project/radius/pull/11911) (provides `make update-resource-types`)
- Required secret: `GH_RAD_CI_BOT_PAT`

## Changes

- `.github/workflows/contrib-update-resource-types.yaml`: New workflow

## Part of

Unified Bicep extension publishing (PR 3/4). See [design doc](https://github.com/radius-project/radius/pull/11892).


nithyatsu (dismissed):

.github/workflows/contrib-update-resource-types.yaml
```diff
@@ -0,0 +1,209 @@
...
+          echo "ref=${ref}" >> "${GITHUB_OUTPUT}"
+          echo "Contrib ref: ${ref}"
+
+      - name: Checkout
+        # Uses GH_RAD_CI_BOT_PAT instead of the default GITHUB_TOKEN so that
+        # the bot can push the PR branch and the resulting PR triggers CI checks.
+        # The default GITHUB_TOKEN cannot trigger workflows on pushes it creates.
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          token: ${{ secrets.GH_RAD_CI_BOT_PAT }}
```
DariuszPorowski: issue: related to https://github.com/radius-project/resource-types-contrib/pull/160#discussion_r3251543652 we must eliminate PATs/fake accounts and use GH Apps

.github/workflows/contrib-update-resource-types.yaml
```diff
@@ -0,0 +1,209 @@
...
+          go-version-file: go.mod
+          cache-dependency-path: go.sum
+          cache: true
+
+      - name: Install yq
+        # Required by make update-resource-types / sync-resource-types to parse
+        # deploy/manifest/defaults.yaml.
+        run: |
+          mkdir -p "${RUNNER_TEMP}/bin"
+          GOBIN="${RUNNER_TEMP}/bin" go install github.com/mikefarah/yq/v4@v4.44.3
```
DariuszPorowski: suggestion: if the tool releases official binaries, then we should use them (with checksums validation if possible) instead of compiling the source on us.

.github/workflows/contrib-update-resource-types.yaml
```diff
@@ -0,0 +1,209 @@
...
+      - name: Open or update pull request
+        # Creates a new PR if none exists for the bot branch, or updates the
+        # existing one's title and body. Uses github-script because we need
+        # conditional create-or-update logic that gh pr create doesn't support.
+        if: steps.changes.outputs.changed == 'true'
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        env:
+          CONTRIB_REF: ${{ steps.contrib.outputs.ref }}
+        with:
+          github-token: ${{ secrets.GH_RAD_CI_BOT_PAT }}
```
DariuszPorowski: issue: same here what I mentioned above - use GH Apps instead of PATs/fake accounts

DariuszPorowski (changes requested):

.github/workflows/contrib-update-resource-types.yaml
kachawla: Good point, updated.

Also, broader question for you and @brooke-hamilton - do we have a convention for tool installation in workflows? Right now yq is installed separately in each workflow that needs it. Should we consolidate this into a composite action so the version/checksum management lives in one place, or is the preference to keep installing tools directly in each workflow?

.github/workflows/contrib-update-resource-types.yaml
kachawla: We use this PAT in rest of the workflows in the repo as well. Given that we have to plan migrate all of the workflows to GH app in the near future, can we include this in the same effort instead of blocking this PR?

.github/workflows/contrib-update-resource-types.yaml
DariuszPorowski: We do not have, already talked about that with @brooke-hamilton, we have some ideas for standardization but its P3 for now. So, keep in-line for now and we will change in the future, thx!

.github/workflows/contrib-update-resource-types.yaml
kachawla: Updated

.github/workflows/contrib-update-resource-types.yaml
kachawla: Updated. Thanks

.github/workflows/contrib-update-resource-types.yaml
kachawla: Updated

DariuszPorowski (approved):

PR #11931, opened 2026-05-18
Wrap last column to terminal width in CLI table formatter

zachcasper:
# Description

Fixes the misaligned/ugly CLI table output when a cell in the last column (typically `DESCRIPTION`) is longer than the terminal width. The previous `text/tabwriter`-based renderer let long lines overflow and wrap at the terminal boundary, breaking column alignment on continuation lines.

This PR replaces the table renderer with a custom implementation that:

- Detects terminal width via `golang.org/x/term`.
- Word-wraps content in the last column to fit the remaining width.
- Pads continuation lines so they sit under the correct column.
- Is rune-aware (UTF-8 safe) for both column padding and wrapping.
- Falls back to no wrapping when the writer is not a terminal (e.g., piped output), preserving scriptability.

## Example

**Before:**

```
$ rad resource-type show Radius.Security/secrets
TYPE                     NAMESPACE
Radius.Security/secrets  Radius.Security

DESCRIPTION:
...
API VERSION: 2025-08-01-preview

TOP-LEVEL PROPERTIES:

NAME         TYPE      REQUIRED  READ-ONLY  DESCRIPTION
application  string    false     false      (Optional) The Radius Application ID. `myApplication.id` for example.
data         object    true      false      (Required) Map of secret data. For example: `data: { username: { value: user1 } password: { value: pass }}`
environment  string    true      false      (Required) The Radius Environment ID. Typically set by the rad CLI. Typically value should be `environment`.
kind         string    false     false      (Optional) The kind of content of the secret. If not specified, generic is assumed. basicAuthentication, awsIRSA, and azureWorkloadIdentity should only be used for configuring authentication to OCI registries for storing Bicep templates. This will change in the future.

OBJECT PROPERTIES:

data

NAME      TYPE      REQUIRED  READ-ONLY  DESCRIPTION
encoding  string    false     false      (Optional) Content encoding of the value. If not specified, `string` is assumed.
value     string    true      false      (Required) The string value of the secret unless encoding is set to 'base64'.
```

**After:**

```
$ rad resource-type show Radius.Security/secrets
TYPE                     NAMESPACE
Radius.Security/secrets  Radius.Security

DESCRIPTION:
...
API VERSION: 2025-08-01-preview

TOP-LEVEL PROPERTIES:

NAME         TYPE      REQUIRED  READ-ONLY  DESCRIPTION
application  string    false     false      (Optional) The Radius Application ID. `myApplication.id` for example.
data         object    true      false      (Required) Map of secret data. For example: `data: { username: { value: user1 } password: { value: pass }}`
environment  string    true      false      (Required) The Radius Environment ID. Typically set by the rad CLI. Typically value should be `environment`.
kind         string    false     false      (Optional) The kind of content of the secret. If not specified, generic is assumed. basicAuthentication, awsIRSA,
                                            and azureWorkloadIdentity should only be used for configuring authentication to OCI registries for storing Bicep
                                            templates. This will change in the future.

OBJECT PROPERTIES:

data

NAME      TYPE      REQUIRED  READ-ONLY  DESCRIPTION
encoding  string    false     false      (Optional) Content encoding of the value. If not specified, `string` is assumed.
value     string    true      false      (Required) The string value of the secret unless encoding is set to 'base64'.
```

## Type of change

Fixes: #9756

## Contributor checklist

    - [x] Not applicable
    - [x] Not applicable
    - [x] Not applicable
    - [x] Not applicable
    - [x] Not applicable
    - [x] Not applicable

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -119,7 +119,10 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
-require golang.org/x/term v0.43.0
+require (
+	github.com/mattn/go-runewidth v0.0.23
+	golang.org/x/term v0.43.0
+)
 
 require (
 	cel.dev/expr v0.25.1 // indirect
@@ -253,7 +256,6 @@ require (
 	github.com/mailru/easyjson v0.9.0 // indirect
 	github.com/mattn/go-colorable v0.1.14 // indirect
 	github.com/mattn/go-localereader v0.0.1 // indirect
-	github.com/mattn/go-runewidth v0.0.23 // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -119,6 +119,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require golang.org/x/term v0.43.0
+
 require (
 	cel.dev/expr v0.25.1 // indirect
 	cloud.google.com/go v0.123.0 // indirect
@@ -319,7 +321,6 @@ require (
 	golang.org/x/net v0.54.0 // indirect
 	golang.org/x/oauth2 v0.36.0 // indirect
 	golang.org/x/sys v0.44.0 // indirect
-	golang.org/x/term v0.43.0 // indirect
 	golang.org/x/time v0.15.0 // indirect
 	golang.org/x/tools v0.44.0 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
--- a/pkg/cli/output/table.go
+++ b/pkg/cli/output/table.go
@@ -23,8 +23,8 @@ import (
 	"os"
 	"strings"
 	"unicode"
-	"unicode/utf8"
 
+	"github.com/mattn/go-runewidth"
 	"golang.org/x/term"
 	"k8s.io/client-go/util/jsonpath"
 )
@@ -120,14 +120,14 @@ func (f *TableFormatter) formatWithWidth(obj any, writer io.Writer, options Form
 	numCols := len(headings)
 	slots := make([]int, numCols)
 	for i, h := range headings {
-		if w := utf8.RuneCountInString(h) + TablePadSize; w > slots[i] {
+		if w := runewidth.StringWidth(h) + TablePadSize; w > slots[i] {
 			slots[i] = w
 		}
 	}
 	for _, cells := range cellLines {
 		for i, lines := range cells {
 			for _, line := range lines {
-				if w := utf8.RuneCountInString(line) + TablePadSize; w > slots[i] {
+				if w := runewidth.StringWidth(line) + TablePadSize; w > slots[i] {
 					slots[i] = w
 				}
 			}
@@ -221,28 +221,35 @@ func (f *TableFormatter) formatWithWidth(obj any, writer io.Writer, options Form
 	return nil
 }
 
-// padRight returns s padded with trailing spaces so that the resulting string is at least
-// width runes long. If s is already wider than width, s is returned unchanged.
+// padRight returns s padded with trailing spaces so that the resulting string occupies
+// at least width terminal columns. Width is measured in display columns (so wide
+// characters like CJK or emoji count as 2 and combining marks count as 0), so the next
+// column starts at the correct visual position. If s is already wider than width, s is
+// returned unchanged.
 func padRight(s string, width int) string {
-	n := utf8.RuneCountInString(s)
+	n := runewidth.StringWidth(s)
 	if n >= width {
 		return s
 	}
 	return s + strings.Repeat(" ", width-n)
 }
 
-// wordWrap wraps s into lines no wider than width runes, breaking on whitespace where
-// possible. Words longer than width are broken mid-word on rune boundaries (never inside
-// a UTF-8 sequence). Internal whitespace runs are preserved within a line; trailing
-// whitespace at a wrap boundary is dropped. An empty input returns a single empty line.
+// wordWrap wraps s into lines whose display width is no more than width terminal
+// columns, breaking on whitespace where possible. Width is measured in display
+// columns (wide characters count as 2, combining marks as 0), so the result fits
+// the terminal even when the content contains CJK characters or emoji. Words wider
+// than width are broken at rune boundaries (never inside a UTF-8 sequence). Internal
+// whitespace runs are preserved within a line; trailing whitespace at a wrap boundary
+// is dropped. An empty input returns a single empty line.
 func wordWrap(s string, width int) []string {
-	if width <= 0 || utf8.RuneCountInString(s) <= width {
+	if width <= 0 || runewidth.StringWidth(s) <= width {
 		return []string{s}
 	}
 
 	runes := []rune(s)
 	var lines []string
 	var current []rune
+	currentWidth := 0
 	i := 0
 	for i < len(runes) {
 		// Collect the next non-whitespace word.
@@ -251,6 +258,7 @@ func wordWrap(s string, width int) []string {
 			j++
 		}
 		word := runes[i:j]
+		wordWidth := runewidth.StringWidth(string(word))
 
 		// Collect the whitespace that follows the word.
 		k := j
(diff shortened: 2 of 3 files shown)


pkg/cli/output/table.go
```diff
@@ -57,79 +85,223 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
 }
 
+// padRight returns s padded with trailing spaces so that the resulting string is at least
+// width runes long. If s is already wider than width, s is returned unchanged.
+func padRight(s string, width int) string {
+	n := utf8.RuneCountInString(s)
+	if n >= width {
+		return s
+	}
+	return s + strings.Repeat(" ", width-n)
```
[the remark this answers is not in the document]
zachcasper: Good catch. Switched `padRight` and the slot/width computation to use `runewidth.StringWidth`, which accounts for wide characters (CJK, emoji = 2 columns) and combining marks (0 columns). Fixed in cb5bdddc.

pkg/cli/output/table.go
```diff
@@ -57,79 +85,223 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
+}
+
+// wordWrap wraps s into lines no wider than width runes, breaking on whitespace where
+// possible. Words longer than width are broken mid-word on rune boundaries (never inside
+// a UTF-8 sequence). Internal whitespace runs are preserved within a line; trailing
+// whitespace at a wrap boundary is dropped. An empty input returns a single empty line.
+func wordWrap(s string, width int) []string {
+	if width <= 0 || utf8.RuneCountInString(s) <= width {
+		return []string{s}
```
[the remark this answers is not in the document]
zachcasper: Addressed in cb5bdddc. `wordWrap` now tracks display width via `runewidth.StringWidth`/`runewidth.RuneWidth`, and the long-word splitter walks rune boundaries while accumulating display columns so wide characters cannot push a line past the terminal width. Added `Test_Table_WrapsWideCharactersByDisplayWidth` (CJK) to enforce that no rendered line exceeds the terminal width in display columns.

pkg/cli/output/table.go
```diff
@@ -35,12 +38,37 @@ const (
 	TableFlags          = 0
 )
 
+// terminalWidthForWriter is overridable for testing.
+var terminalWidthForWriter = func(w io.Writer) int {
```
brooke-hamilton, line 42: Two concerns in this area:

1. **Dead exported constants above.** `TableTabSize`, `TablePadCharacter`, and `TableFlags` were only used to configure `text/tabwriter`. After this change they have no callers in the workspace. Because they are exported, leaving them dangling pollutes the package API. Please remove them, or add a doc comment explaining they are retained for backwards compatibility.
2. **`terminalWidthForWriter` as a mutable package-level `var`.** It's a test seam, but none of the new tests actually swap it (they all call `formatWithWidth` directly), so the seam is currently unused. Either drop it and test `Format` end-to-end with an `*os.File` pipe, or keep it and add a test that overrides it with `t.Cleanup` restoring the original. Also: width is only detected when the writer's concrete type is `*os.File`. Anything wrapping stdout (e.g., a color-enabling writer, `bufio.Writer`, a tee) silently disables wrapping. Please confirm the CLI's real stdout path is a raw `*os.File`; otherwise users on real terminals won't get the new behavior.

pkg/cli/output/table.go
```diff
@@ -57,79 +85,245 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
+	}
 
-		renderedRows = append(renderedRows, currentRows...)
+	// The last column is rendered without trailing padding, so its effective content width
+	// is its slot width minus the pad we added when computing the slot. When we know the
+	// terminal width and the table would overflow, shrink the last column so its content
+	// fits in the remaining space, and wrap longer cells onto continuation lines.
+	lastContentWidth := slots[numCols-1] - TablePadSize
+	if terminalWidth > 0 {
+		nonLastSum := 0
```
brooke-hamilton, line 148: **(High priority — flagged by both reviewers.)** When `nonLastSum + TableColumnMinWidth > terminalWidth`, the last column is forced to `TableColumnMinWidth` and the rendered line will exceed the terminal width — the very thing this PR fixes. For example, if non-last columns consume 75 cells of an 80-cell terminal, `available` becomes 5, this clamp raises it to 10, and every line ends up 85 cells wide.

Consider `available = max(1, terminalWidth-nonLastSum)` so wrapping still occurs in narrow terminals, or — if you intentionally prefer a readable minimum at the cost of overflow — add a comment documenting that fixed-column-driven overflow can still occur. Either way, please add a regression test where non-last columns leave fewer than `TableColumnMinWidth` cells for the last column.

pkg/cli/output/table.go
```diff
@@ -57,79 +85,245 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
+			nonLastSum += slots[i]
+		}
+		available := terminalWidth - nonLastSum
+		if available < TableColumnMinWidth {
+			available = TableColumnMinWidth
+		}
+		if available < lastContentWidth {
+			lastContentWidth = available
+		}
 	}
```
brooke-hamilton, line 160: Two minor cleanups:

1. `if numCols > 0` is dead: `numCols == len(options.Columns)` and the function returns early at the top when `len(options.Columns) == 0`. Remove the guard.
2. `cells[lastIdx] = wrapped` mutates the caller's `cells` slice. It's harmless today because each row is iterated once, but it's a non-obvious side effect on a closure parameter. Consider assigning into a local `colCells` variable, or document the contract on the closure.

pkg/cli/output/table.go
```diff
@@ -57,79 +85,245 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
 
+	for _, cells := range cellLines {
+		if err := writeRow(cells); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
```
brooke-hamilton, line 223: When `width <= 0`, this returns `[]string{s}` without splitting on newlines. That is correct (callers have already split on `\n`), but the public contract isn't obvious from the signature — please add a one-line comment to that effect.

pkg/cli/output/table.go
```diff
@@ -57,79 +85,245 @@ func (f *TableFormatter) Format(obj any, writer io.Writer, options FormatterOpti
...
+
+		// Collect the whitespace that follows the word.
+		k := j
+		for k < len(runes) && unicode.IsSpace(runes[k]) {
+			k++
+		}
+		space := runes[j:k]
+		spaceWidth := runewidth.StringWidth(string(space))
+		i = k
+
```
brooke-hamilton, line 271: Long-word splitter edge case: if `width == 1` and a wide rune (display width 2) appears, the `ci > chunkStart` guard prevents emitting an empty chunk, so a one-rune chunk of display width 2 will be produced — exceeding `width`. This is unreachable today because `available` is clamped to `TableColumnMinWidth == 10`, but worth either an assertion comment or a `wordWrap` test with `width == 1` to document the contract.

Also: `ci` here is a rune index because `word` is `[]rune`, but the surrounding code uses `[]rune(s)` conversions elsewhere — a one-line comment clarifying that `ci` is a rune index (not a byte index) would help future readers.

pkg/cli/output/table_test.go
```diff
@@ -145,6 +147,172 @@ smoothness
...
+}
+
+var wrappableInputOptions = FormatterOptions{
+	Columns: []Column{
+		{Heading: "NAME", JSONPath: "{ .Name }"},
+		{Heading: "DESCRIPTION", JSONPath: "{ .Description }"},
+	},
+}
+
+func Test_Table_WrapsLastColumnToTerminalWidth(t *testing.T) {
```
brooke-hamilton, line 162: None of the new tests (`Test_Table_WrapsLastColumnToTerminalWidth`, `Test_Table_NoWrapWhenTerminalWidthUnknown`, `Test_Table_WrapsLongUnbreakableWord`, `Test_Table_AlignsAndWrapsUnicode`, `Test_Table_PreservesInternalWhitespace`, `Test_Table_WrapsWideCharactersByDisplayWidth`) call `t.Parallel()`. Per the repo's Go instructions, tests without shared state should run in parallel. All of these are independent and safe to parallelize.

Additionally, they share an identical setup pattern (build `[]wrappableInput`, create formatter + buffer, call `formatWithWidth`, compare to expected string). Please consolidate into a single table-driven `Test_Table_Wrapping(t *testing.T)` with sub-tests using `t.Run` + `t.Parallel()`. This eliminates duplication and makes adding cases trivial.

brooke-hamilton (commented):
Automated review consolidating findings from two reviewer passes. Two reviewers independently flagged the `TableColumnMinWidth` clamp as the primary functional bug; other comments cover API hygiene, test coverage gaps, and small dead-code/style items.

**Top priorities**
1. The `available < TableColumnMinWidth` clamp can still overflow the terminal — the very condition this PR is meant to fix.
2. `Format` (the production entry point) is not exercised by any new test; all new tests bypass it via `formatWithWidth`.
3. New tests do not use `t.Parallel()` and could be consolidated into a single table-driven test.

**PR title:** Accurate and descriptive — no change suggested.

**Contributor doc impact:** None. Internal renderer change only; no CLI surface, build, or workflow changes.

pkg/cli/output/table_test.go
```diff
@@ -145,6 +147,172 @@ smoothness
...
+	expected := "NAME      DESCRIPTION\n" +
+		"size      The size of the PostgreSQL database, non production environments can\n" +
+		"          be (S)mall or (M)edium, production environments can be or (S)mall,\n" +
+		"          (M)edium, (L)arge, or (XL)arge\n" +
+		"host      The host name of the database server\n"
+
+	require.Equal(t, expected, buffer.String())
+}
+
+func Test_Table_NoWrapWhenTerminalWidthUnknown(t *testing.T) {
```
brooke-hamilton, line 192: Every new test bypasses `Format` and calls the unexported `formatWithWidth`. Please add at least one test that drives `Format` with a non-`*os.File` writer (a `bytes.Buffer`) to verify the "unknown width → no wrap" path through the production entry point. This is also the simplest way to lift the `terminalWidthForWriter` coverage hole flagged by Codecov.

pkg/cli/output/table_test.go
```diff
@@ -145,6 +147,172 @@ smoothness
...
+	expected := "NAME      DESCRIPTION\n" +
+		"row       a   b   c\n"
+	require.Equal(t, expected, buffer.String())
+}
+
+// Test_Table_WrapsWideCharactersByDisplayWidth verifies that columns and wrapping use
+// terminal display width rather than rune count: wide characters (e.g. CJK ideographs)
+// occupy two columns, so the column slot must grow accordingly and wrapping must occur
+// before content exceeds the terminal width.
+func Test_Table_WrapsWideCharactersByDisplayWidth(t *testing.T) {
```
brooke-hamilton, line 286: Missing edge-case coverage worth adding:

- Single-column table (`numCols == 1`) — exercises the `lastContentWidth` path with `nonLastSum == 0`.
- Cell containing literal `\n` combined with wrapping — verifies the doc claim that embedded newlines still expand into multiple rows even when `terminalWidth == 0`.
- Empty rows / empty last cell — verifies no spurious blank continuation lines.
- Heading longer than the available last-column width — current code will wrap the heading; either assert this behavior or add a special case so headings are never wrapped.
- Narrow terminal where non-last columns leave fewer than `TableColumnMinWidth` cells for the last column — regression test for the overflow issue flagged on `table.go`.

go.mod
```diff
@@ -119,6 +119,11 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require (
+	github.com/mattn/go-runewidth v0.0.23
+	golang.org/x/term v0.43.0
+)
```
brooke-hamilton, line 125: `golang.org/x/term v0.43.0` and `github.com/mattn/go-runewidth v0.0.23` are correctly promoted to direct dependencies and removed from the indirect block. The diff stat shows no `go.sum` change — please confirm `go mod tidy` was run and `go.sum` is consistent before merge.

PR #11932, opened 2026-05-18
Long-running Azure test reliability and test-code ref override

brooke-hamilton:
# Description

Improvements to the long-running Azure test workflow to prevent timeouts and to allow test code to be patched without cutting a product patch release.

Three independent fixes:

1. **Optional `test_code_ref` override for `checkout-release-codebase.sh` and the long-running Azure workflow.** When set (via positional arg, `TEST_CODE_REF` env var, or workflow_dispatch input), the script clones the supplied git ref into `current_release/` instead of the tag matching the installed CLI. The product under test is still the installed release; only the on-disk test/infrastructure code changes. This lets us iterate on test fixes without releasing a new patch.
2. **Pre-restore the UDT `usertypealpha-recipe.bicep` types in the workflow.** Restoring this lazily later in the run was racing with Azure CLI token expiry during the long-running tests.
3. **Robust namespace and DeploymentTemplate teardown in the Kubernetes/Flux noncloud tests.**
   - `deleteNamespace` now waits for normal deletion with `assert.Eventually` and falls back to clearing namespace finalizers (test-only escape hatch) when a namespace is stuck `Terminating`.
   - `testFluxIntegration` now deletes `DeploymentTemplate` resources first and waits for the Radius finalizer to drain before tearing down namespaces, so namespaces are not stuck on `radapp.io/deployment-template-finalizer` and `radius-rp` does not recreate application namespaces while their backing Applications.Core resources still exist.
   - Teardown uses `assert.*` instead of `require.*` so one stuck resource does not skip cleanup of the rest.

## Type of change

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable

### DIFF
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -42,6 +42,17 @@ name: Long-running test on Azure
 on:
   # Enable manual trigger for testing.
   workflow_dispatch:
+    inputs:
+      test_code_ref:
+        description: |
+          Optional git ref (branch, tag, or SHA) in radius-project/radius to use
+          for test code. Leave blank to use the tag matching the installed
+          release CLI (default behavior). The product under test is always the
+          installed release; only the on-disk test/infrastructure code is taken
+          from this ref.
+        required: false
+        default: ''
+        type: string
   schedule:
     # Run once per day at 3:00 AM PDT (10:00 AM UTC)
     - cron: 0 10 * * *
@@ -161,6 +172,9 @@ jobs:
 
       - name: Checkout release codebase
         id: checkout-release-codebase
+        env:
+          # Optional override; empty string means "use the release tag matching the installed CLI".
+          TEST_CODE_REF: ${{ inputs.test_code_ref }}
         run: |
           export PATH=$GITHUB_WORKSPACE/bin:$PATH
           ./.github/scripts/checkout-release-codebase.sh
@@ -355,6 +369,8 @@ jobs:
           bicep restore "./${RELEASE_DIR}/test/functional-portable/corerp/cloud/resources/testdata/corerp-azure-connection-database-service.bicep" --force
           # Restore AWS Bicep types
           bicep restore "./${RELEASE_DIR}/test/functional-portable/corerp/cloud/resources/testdata/aws-s3-bucket.bicep" --force
+          # Restore UDT testresources Bicep types - required to avoid Azure CLI token expiry during the long-running tests
+          bicep restore "./${RELEASE_DIR}/test/functional-portable/dynamicrp/noncloud/resources/testdata/usertypealpha-recipe.bicep" --force
         env:
           RELEASE_DIR: ${{ steps.checkout-release-codebase.outputs.release-dir }}
 
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -45,10 +45,10 @@ on:
     inputs:
       test_code_ref:
         description: |
-          Optional git  branch in radius-project/radius for the test code to use
-          during the run. Leave blank to use the tag matching the release
+          Optional git branch or tag in radius-project/radius for the test code
+          to use during the run. Leave blank to use the tag matching the release
           (default behavior). This setting does not affect the code being tested,
-          which is always the installed release. This setting allows testing 
+          which is always the installed release. This setting allows testing
           updated test code, not updated product code.
         required: false
         default: ''
--- a/.github/workflows/long-running-azure.yaml
+++ b/.github/workflows/long-running-azure.yaml
@@ -45,11 +45,11 @@ on:
     inputs:
       test_code_ref:
         description: |
-          Optional git ref (branch, tag, or SHA) in radius-project/radius to use
-          for test code. Leave blank to use the tag matching the installed
-          release CLI (default behavior). The product under test is always the
-          installed release; only the on-disk test/infrastructure code is taken
-          from this ref.
+          Optional git  branch in radius-project/radius for the test code to use
+          during the run. Leave blank to use the tag matching the release
+          (default behavior). This setting does not affect the code being tested,
+          which is always the installed release. This setting allows testing 
+          updated test code, not updated product code.
         required: false
         default: ''
         type: string
@@ -369,7 +369,9 @@ jobs:
           bicep restore "./${RELEASE_DIR}/test/functional-portable/corerp/cloud/resources/testdata/corerp-azure-connection-database-service.bicep" --force
           # Restore AWS Bicep types
           bicep restore "./${RELEASE_DIR}/test/functional-portable/corerp/cloud/resources/testdata/aws-s3-bucket.bicep" --force
-          # Restore UDT testresources Bicep types - required to avoid Azure CLI token expiry during the long-running tests
+          # Pre-restore UDT testresources Bicep types so the first deploy in the run does not have to fetch them.
+          # Without pre-restore, the lazy restore happens mid-test and has occasionally exceeded the Azure CLI
+          # token lifetime during long-running tests.
           bicep restore "./${RELEASE_DIR}/test/functional-portable/dynamicrp/noncloud/resources/testdata/usertypealpha-recipe.bicep" --force
         env:
           RELEASE_DIR: ${{ steps.checkout-release-codebase.outputs.release-dir }}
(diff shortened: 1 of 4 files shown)


test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -404,14 +405,52 @@ func deleteNamespace(ctx context.Context, t *testing.T, namespace string, opts r
...
+	// below instead of failing the test outright.
+	if assert.Eventually(t, func() bool {
+		ns := &corev1.Namespace{}
+		getErr := opts.Client.Get(ctx, types.NamespacedName{Name: namespace}, ns)
+		if apierrors.IsNotFound(getErr) {
+			return true
+		}
+		// Tolerate transient errors and keep polling.
+		return false
+	}, time.Minute*2, time.Second*5, "waiting for namespace %s to be deleted", namespace) {
```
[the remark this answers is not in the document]
brooke-hamilton: Fixed

test/functional-portable/kubernetes/noncloud/flux_test.go
```diff
@@ -307,13 +317,8 @@ func testFluxIntegration(t *testing.T, testName string, steps []GitOpsTestStep,
...
 
 			deploymentTemplate, err := waitForDeploymentTemplateToBeReadyWithGeneration(t, ctx, types.NamespacedName{Name: name, Namespace: namespace}, stepNumber, opts.Client)
-			defer func() {
-				err := opts.Client.Delete(ctx, deploymentTemplate)
-				if controller_runtime.IgnoreNotFound(err) != nil {
-					t.Logf("Error deleting deployment template: %v", err)
-				}
-			}()
 			require.NoError(t, err)
+			deploymentTemplates = append(deploymentTemplates, deploymentTemplate)
```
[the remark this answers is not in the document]
brooke-hamilton: Fixed

.github/workflows/long-running-azure.yaml
```diff
@@ -161,6 +172,9 @@ jobs:
 
       - name: Checkout release codebase
         id: checkout-release-codebase
+        env:
+          # Optional override; empty string means "use the release tag matching the installed CLI".
+          TEST_CODE_REF: ${{ inputs.test_code_ref }}
```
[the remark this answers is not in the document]
brooke-hamilton: Thanks — flagging this so it isn't lost, but I don't think we should restrict the override in this PR.

`workflow_dispatch` is already gated to users with write/maintain access on this repo (and the job has `if: github.repository == vars.RADIUS_REPOSITORY`). Anyone with the ability to dispatch this workflow can already push a branch and run it under the existing path, so restricting `test_code_ref` to "trusted/protected refs" doesn't reduce the attack surface — it just defeats the feature's purpose, which is exactly to iterate on test-code fixes against the installed release without cutting a patch.

If we want defense-in-depth, the standard mitigation is to wrap the job in `environment: <name>` with required reviewers so each manual dispatch needs an approval. I'd prefer to do that as a follow-up if/when we decide it's warranted, rather than block this PR.

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -404,14 +404,50 @@ func deleteNamespace(ctx context.Context, t *testing.T, namespace string, opts r
 	err := opts.K8sClient.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{})
 	if !apierrors.IsNotFound(err) {
 		require.NoError(t, err, "failed to delete namespace %s", namespace)
-		t.Logf("Namespace %s deleted successfully", namespace)
+		t.Logf("Namespace %s delete requested", namespace)
```
willdavsmith: excellent

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
```diff
@@ -404,14 +404,50 @@ func deleteNamespace(ctx context.Context, t *testing.T, namespace string, opts r
...
+	ns, err := opts.K8sClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return
+	}
+	require.NoError(t, err)
+
+	if len(ns.Spec.Finalizers) > 0 {
+		t.Logf("Namespace %s stuck terminating with finalizers: %v; clearing them", namespace, ns.Spec.Finalizers)
+		ns.Spec.Finalizers = nil
+		_, err = opts.K8sClient.CoreV1().Namespaces().Finalize(ctx, ns, metav1.UpdateOptions{})
```
willdavsmith: I think having a finalizer is a symptom that there are still hanging resources. IMO we should never force delete finalizers in this case, since it can result in the next run having unexpected behavior (existing but untracked resources).

test/functional-portable/kubernetes/noncloud/flux_test.go
```diff
@@ -237,14 +241,60 @@ func testFluxIntegration(t *testing.T, testName string, steps []GitOpsTestStep,
...
 	}()
 	require.NoError(t, err)
 
 	// Wait for the GitRepository to be ready.
 	_, err = waitForGitRepositoryReady(t, ctx, types.NamespacedName{Name: gitRepoName, Namespace: fluxSystemNamespace}, opts.Client, fluxGitRepository.ResourceVersion)
 	require.NoError(t, err)
 
+	// Track DeploymentTemplates created across steps so we can delete them and wait for
+	// the Radius finalizer to drain before tearing down namespaces.
+	var deploymentTemplates []*radappiov1alpha3.DeploymentTemplate
```
willdavsmith, line 256: where is this slice being populated?

test/functional-portable/kubernetes/noncloud/flux_test.go
```diff
@@ -237,14 +241,60 @@ func testFluxIntegration(t *testing.T, testName string, steps []GitOpsTestStep,
...
+					current := &radappiov1alpha3.DeploymentTemplate{}
+					getErr := opts.Client.Get(ctx, types.NamespacedName{Name: dt.Name, Namespace: dt.Namespace}, current)
+					if !apierrors.IsNotFound(getErr) {
+						return false
+					}
+				}
+				return true
+			}, time.Minute*5, time.Second*5, "DeploymentTemplates were not deleted in time")
+		}
+
```
willdavsmith, line 291: overall I think this is great. we need better reliability from these tests and I think this is the right approach (add better cleanup in the tests)

test/functional-portable/kubernetes/noncloud/flux_test.go
```diff
@@ -237,14 +241,60 @@ func testFluxIntegration(t *testing.T, testName string, steps []GitOpsTestStep,
...
+		for _, namespace := range namespaces {
+			t.Logf("Deleting namespace: %s", namespace)
+			deleteNamespace(ctx, t, namespace, opts)
```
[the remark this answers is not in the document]
brooke-hamilton: Fixed

test/functional-portable/kubernetes/noncloud/flux_test.go
brooke-hamilton: Populated inside the per-step loop at [test/functional-portable/kubernetes/noncloud/flux_test.go#L359-L361](https://github.com/radius-project/radius/blob/brooke-hamilton/fix-lrt-release/test/functional-portable/kubernetes/noncloud/flux_test.go#L359-L361):

```go
deploymentTemplate, err := waitForDeploymentTemplateToBeReadyWithGeneration(t, ctx, types.NamespacedName{Name: name, Namespace: namespace}, stepNumber, opts.Client)
require.NoError(t, err)
deploymentTemplates = append(deploymentTemplates, deploymentTemplate)
```

For every step, after each `DeploymentTemplate` becomes Ready we append it to the slice the `t.Cleanup` closes over, so teardown deletes every DT created across all steps before tearing down namespaces.

test/functional-portable/kubernetes/noncloud/deploymenttemplate_test.go
brooke-hamilton: Removed

DariuszPorowski:
@brooke-hamilton @willdavsmith any progress on this fix?

brooke-hamilton:
Superseded by other changes.

PR #11933, opened 2026-05-18
Fix resource types missing from rad resource-type list with per-type manifest files

kachawla:
## Background

PR #11911 introduced automated synchronization of default resource type manifests from `resource-types-contrib`. As part of that change, the manifest file layout in `deploy/manifest/built-in-providers/` changed from one file per namespace (e.g., `radius_compute.yaml` containing `containers`, `routes`, and `persistentVolumes`) to one file per type (e.g., `containers.yaml`, `routes.yaml`, `persistentVolumes.yaml`). This matches the per-type layout in `resource-types-contrib` and keeps diffs scoped to the type that changed.

However, the initializer's `registerResourceProviderDirect` function was not designed for multiple files sharing the same namespace. It saves the Location and ResourceProviderSummary per namespace on each file, and each save was a full overwrite rather than a merge.

## Problem

When multiple manifest files share the same namespace (e.g., `containers.yaml` and `persistentVolumes.yaml` both under `Radius.Compute`), the initializer was overwriting the Location and ResourceProviderSummary on each file, causing earlier types to disappear from `rad resource-type list`. Only the last file's types (alphabetically) were visible.

The individual ResourceType records were saved correctly (one per type), but the Location and Summary - which UCP's routing layer and `rad resource-type list` read from - only contained the last file's types.

## Fix

`registerResourceProviderDirect` now reads the existing Location and Summary from the database before saving, merging in new types alongside existing ones. If no existing record is found (first file for this namespace), it behaves as before. Only `ErrNotFound` is treated as "no existing record"; other errors (database outage, decode failure) cause the initializer to fail fast.

The fix also preserves the existing location address when a later manifest for the same namespace does not specify one.

## Changes

- `pkg/ucp/initializer/service.go` - Read-merge-write for Location and ResourceProviderSummary, with ErrNotFound-only error handling and address preservation
- `pkg/ucp/integrationtests/resourceproviders/resourceproviders_test.go` - Updated `Test_ResourceProvider_RegisterManifests_NoLocation` to verify both types from two files sharing a namespace are registered and present in the location's resourceTypes map. Added `Test_ResourceProvider_DefaultsRegistered` that reads `defaults.yaml`, loads real manifests, and verifies all types are registered with correct location entries.
- `testdata/manifests-no-location/persistentVolumes.yaml` - Second test manifest sharing the `Radius.Compute` namespace
- `test/functional-portable/dynamicrp/noncloud/resources/default_containers_test.go` - End-to-end functional test deploying `Radius.Compute/containers` and `Radius.Compute/routes` (two types from the same namespace) using the default recipes
- `test/functional-portable/dynamicrp/noncloud/resources/testdata/default-containers-test.bicep` - Bicep template for the e2e test

## Testing

- **Integration test** (`Test_ResourceProvider_RegisterManifests_NoLocation`): Two test manifests sharing `Radius.Compute` namespace, verifies both types appear in the location's resourceTypes map.
- **Integration test** (`Test_ResourceProvider_DefaultsRegistered`): Reads `defaults.yaml`, loads manifests from `built-in-providers/dev/`, verifies all types are registered with correct location entries.
- **Functional test** (`Test_DefaultContainers_Deploy`): End-to-end deployment of `Radius.Compute/containers` and `Radius.Compute/routes` using default recipes.


pkg/ucp/initializer/service.go
```diff
@@ -202,8 +203,22 @@ func registerResourceProviderDirect(ctx context.Context, dbClient database.Clien
...
+
+	existingLocation, err := database.GetResource[datamodel.Location](ctx, dbClient, locationID)
+	if err != nil && !errors.Is(err, &database.ErrNotFound{}) {
+		return fmt.Errorf("failed to read existing location %s/%s: %w", rp.Namespace, locationName, err)
+	}
+	if err == nil {
+		for name, config := range existingLocation.Properties.ResourceTypes {
+			locationResourceTypes[name] = config
+		}
+	}
```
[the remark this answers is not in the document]
kachawla: Updated

pkg/ucp/initializer/service.go
```diff
@@ -75,21 +75,49 @@ func (w *Service) Run(ctx context.Context) error {
...
-			return fmt.Errorf("failed to register manifest %s: %w", filePath, err)
+		existing, ok := merged[rp.Namespace]
+		if !ok {
+			merged[rp.Namespace] = rp
+			continue
+		}
+
+		// Merge types from this file into the existing provider for this
+		// namespace. If a type appears in multiple files, the later file wins.
+		for typeName, resourceType := range rp.Types {
```
nithyatsu, line 110: is this a decision we took instead of flagging multiple definitions? if so, should we add a log to help with debugging.

pkg/ucp/initializer/service.go
```diff
@@ -239,6 +267,7 @@ func registerResourceProviderDirect(ctx context.Context, dbClient database.Clien
 
 	// 4. Save ResourceProviderSummary
 	summaryID := rootScope + "/providers/System.Resources/resourceProviderSummaries/" + rp.Namespace
+
```
nithyatsu: nit: extra line

pkg/ucp/initializer/service.go
kachawla: Good question. My original thought was that it shouldn't be blocking as we have PR reviews that should catch it. But on second thought, it should cause the startup to fail because it can go unnoticed otherwise. Updating to return an error.

Thanks for reviewing!

kachawla: Updated

pkg/ucp/initializer/service.go
kachawla: removed

nithyatsu (approved):

PR #11970, opened 2026-05-21
Make functional tests workspace-aware and stabilize local debug stack

sylvainsf:
## Summary

Lets the corerp-noncloud functional test suite run against an OS-process Radius
debug stack on an arbitrary k3d/kind cluster, instead of being hard-wired to a
`kind-radius` cluster + workspace. Also fixes a couple of `make debug-*` targets
that didn't reliably bring up dependencies on k3d.

End result: **152 / 154 corerp-noncloud tests pass locally** in ~2 minutes
against `make debug-start` + a k3d cluster. The 2 remaining failures are in
`Test_ApplicationGraph`, which exercises the deprecated
`Applications.Core/containers` resource type (superseded by `Radius.Compute/*`)
and reproduces outside the test as well — see "Known issue" below.

  - `GetSecretSuffix` now derives the resource group from the active workspace
    via `cli.LoadConfig("") → GetWorkspace → ParseScope → FindScope("resourcegroups")`,
    instead of hardcoding `kind-radius`. Falls back to `default` when nothing is
    configured.
  - Adds a small `rewriteResourceGroup` helper used by callers that hold ARM
    IDs templated for `kind-radius`.
  - `Test_ApplicationGraph` PostStepVerify: substitutes
    `/resourcegroups/kind-radius/` in the JSON fixture with the active
    workspace's resource group before `json.Unmarshal`, instead of forcing the
    test environment to use `kind-radius`.
- `test/rp/rptest.go`, `test/ucp/ucptest.go`, `test/validation/shared.go`,
  — incidental cleanups required to run the suite against a non-`kind-radius`
  workspace.

### `make debug-*` reliability
- `build/debug.mk`
  - `debug-install-contour`: drop Helm `--wait` (it doesn't work for
    LoadBalancer Services on k3d) and instead do an explicit
    `kubectl wait --for=condition=Available` + `kubectl rollout status` so the
    target only returns once Contour is actually serving.
  - `debug-install-tf-module-server`: add a port-forward + `curl` readiness
    probe (8999:80) so subsequent recipe pulls don't race the pod becoming
    Ready.
- `build/test.mk`, `build/recipes.mk`, `.github/scripts/publish-recipes.sh`,
  `build/scripts/mirror-test-images.sh` (new) — companion glue for running the
  suite locally with mirrored images and locally published recipes.

### Misc

  `pkg/recipes/engine/engine.go`, `pkg/rp/util/registry.go`
  — small adjustments surfaced while running the suite end-to-end.
- `pkg/corerp/frontend/controller/applications/testbicep_scan_test.go` (new)
  — small scan test added during investigation.
- `.gitignore` — ignore local debug artifacts.

## How I verified

```sh
make debug-start                   # OS-process ARP/UCP/controller/dynamic-rp
make debug-install-contour
# 152 / 154 PASS in ~124s
```

The 4 `TerraformRecipe*` tests that previously failed because of
`kind-radius`-bound secret suffixes now pass.

## Known issue (not in scope)

`Test_ApplicationGraph` still fails, but the failure reproduces independently
of the test:

```sh
rad resource show Applications.Core/containers http-back-ctnr-simple1   # Succeeded
rad resource list Applications.Core/containers                          # does NOT list it
rad app graph corerp-application-simple1                                # (empty)
```

`getGraph` returns an empty `Resources` array even though both containers are
deployed and `Succeeded`. The same flow against
`corerp-resources-container-secrets` works. This is consistent with a
listing/typing issue in the deprecated `Applications.Core/containers`
plane — `Radius.Compute/containers` is the supported replacement — so I am
**deliberately not fixing it here**.

## Why draft

Opening as draft to let CI exercise the diff and surface any regressions in
the touched `pkg/*` files (`unfold.go`, `updatefilter.go`, `engine.go`,
`registry.go`) before requesting review.

### DIFF
--- a/.github/scripts/publish-recipes.sh
+++ b/.github/scripts/publish-recipes.sh
@@ -65,5 +65,9 @@ for RECIPE in $(find "$DIRECTORY" -type f -name "*.bicep"); do
 
     echo "Publishing $RECIPE to $PUBLISH_REF"
     echo "- $PUBLISH_REF" >>$GITHUB_STEP_SUMMARY
-    rad bicep publish --file $RECIPE --target "br:$PUBLISH_REF"
+    PUBLISH_FLAGS=()
+    if [[ "${PLAIN_HTTP:-false}" == "true" ]]; then
+        PUBLISH_FLAGS+=(--plain-http)
+    fi
+    rad bicep publish --file "$RECIPE" --target "br:$PUBLISH_REF" "${PUBLISH_FLAGS[@]}"
 done
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,11 @@ debug_files/*
 drad
 *.lock
 
+# Local-only bicepconfig override written by `make debug-publish-bicep-types`.
+# Lives next to the .bicep templates so bicep's nearest-config search picks it up
+# over the tracked sibling at .../resources/bicepconfig.json. Removed by debug-stop.
+test/functional-portable/dynamicrp/noncloud/resources/testdata/bicepconfig.json
+
 # Demo app
 demo/*
 demo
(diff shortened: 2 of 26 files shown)


sylvainsf:
Closing in favor of a fresh PR from an upstream branch (sylvainsf/local-functional) — the fork-PR run kept hitting tj-actions/changed-files "Unable to locate the commit sha" because the base ref had moved on upstream main.

PR #11975, opened 2026-05-21
Make functional tests workspace-aware and stabilize local debug stack

sylvainsf:
## Summary

Lets the Radius functional test suites — both **corerp-noncloud** and
**corerp-cloud (Azure)** — run against an OS-process Radius debug stack
(`make debug-start`) on an arbitrary k3d/kind cluster, instead of being
hard-wired to a `kind-radius` cluster and workspace. Also stabilizes a few
`make debug-*` targets that didn't reliably bring up their dependencies
on k3d.

  - `GetSecretSuffix` derives the resource group from the active workspace
    (`cli.LoadConfig` → `GetWorkspace` → `ParseScope` →
    `FindScope("resourcegroups")`) instead of hardcoding `kind-radius`.
    Falls back to `default` when nothing is configured.
  - `Test_ApplicationGraph` PostStepVerify substitutes the fixture's
    `kind-radius` resource group with the active workspace's resource group
    before unmarshalling.
- `test/functional-portable/corerp/cloud/resources/recipe_terraform_test.go`
  - Derives the resource ID from the active workspace scope so it works
    against any RG (CI's `kind-radius` and local debug's `default`).
- `test/functional-portable/corerp/cloud/resources/extender_test.go`,
  `test/rp/rptest.go`, `test/ucp/ucptest.go`, `test/validation/shared.go`,
  — incidental cleanups required to run the suites against a
  non-`kind-radius` workspace; skip AWS-only tests cleanly when AWS env
  vars are unset; skip private-git redis test when `GH_TOKEN` is unset.

### Azure-cloud functional tests against a local OS-process stack

- `build/scripts/azure-local-testenv.sh` — new orchestrator with
  `setup`/`run`/`teardown`/`all` sub-commands. `run` and `all` accept
  passthrough `go test` flags (e.g. `-run`, `-v`).
  - Auto-recovery: `run` rebuilds state from the newest
    `radlocal-${USER}-*` resource group when the state file is missing
    (e.g. after `make debug-stop`), and re-applies the Azure scope on the
    default rad environment that `debug-start` wipes.
  - Orphan GC: `teardown --all-orphans` deletes every
    `radlocal-${USER}-*` RG and stops the `tf-module-server`
    port-forward.
- `pkg/recipes/terraform/config/providers/azure.go` — Terraform Azure
  provider falls back to `use_cli = true` when no Azure credential is
  registered with UCP (404), so the host RP's `az login` session
  authenticates. CI workload-identity path is unchanged.
- `build/scripts/start-radius.sh` — exports
  `TERRAFORM_TEST_GLOBAL_DIR` so the RP no longer tries to write to a
  read-only `/terraform`.
- `build/scripts/ensure-encryption-key.sh` (new) — generates a stable
  encryption key for the local stack.

### `make debug-*` reliability

- `build/debug.mk`
  - `debug-install-contour`: drop Helm `--wait` (it doesn't behave for
    LoadBalancer Services on k3d) and instead do explicit
    `kubectl wait --for=condition=Available` + `kubectl rollout status`,
    so the target only returns once Contour is actually serving.
  - `debug-install-tf-module-server`: deploy the in-cluster nginx test
    module server and port-forward it to `localhost:8999`; add a `curl`
    readiness probe so subsequent recipe pulls don't race the pod
    becoming Ready.
- `build/test.mk`, `build/recipes.mk`,
  `.github/scripts/publish-recipes.sh`,
  `build/scripts/mirror-test-images.sh` (new) — companion glue for
  running the suite locally with mirrored images and locally published
  recipes (the publish script learns `PLAIN_HTTP` for `localhost:5000`
  pushes).

### Misc

  `pkg/recipes/engine/engine.go` — small adjustments surfaced while
  running the suites end-to-end.
- `pkg/corerp/frontend/controller/applications/testbicep_scan_test.go`
  (new) — small scan test added during investigation.
- `.gitignore` — ignore local debug artifacts and the local-only
  `bicepconfig.json` override that `make debug-publish-bicep-types`
  writes.

### Documentation

- `docs/contributing/contributing-code/contributing-code-debugging/radius-os-processes-debugging.md`
  documents running the Azure cloud suite against the local OS-process
  stack via `make debug-start` + `azure-local-testenv.sh`.

## How to use locally

```sh
make debug-start
make debug-install-contour

# corerp-noncloud

# corerp-cloud (Azure) against the same local stack, using host az login
build/scripts/azure-local-testenv.sh all
```

### DIFF
--- a/.github/scripts/publish-recipes.sh
+++ b/.github/scripts/publish-recipes.sh
@@ -65,5 +65,9 @@ for RECIPE in $(find "$DIRECTORY" -type f -name "*.bicep"); do
 
     echo "Publishing $RECIPE to $PUBLISH_REF"
     echo "- $PUBLISH_REF" >>$GITHUB_STEP_SUMMARY
-    rad bicep publish --file $RECIPE --target "br:$PUBLISH_REF"
+    PUBLISH_FLAGS=()
+    if [[ "${PLAIN_HTTP:-false}" == "true" ]]; then
+        PUBLISH_FLAGS+=(--plain-http)
+    fi
+    rad bicep publish --file "$RECIPE" --target "br:$PUBLISH_REF" "${PUBLISH_FLAGS[@]}"
 done
--- a/.gitignore
+++ b/.gitignore
@@ -76,6 +76,11 @@ debug_files/*
 drad
 *.lock
 
+# Local-only bicepconfig override written by `make debug-publish-bicep-types`.
+# Lives next to the .bicep templates so bicep's nearest-config search picks it up
+# over the tracked sibling at .../resources/bicepconfig.json. Removed by debug-stop.
+test/functional-portable/dynamicrp/noncloud/resources/testdata/bicepconfig.json
+
 # Demo app
 demo/*
 demo
--- a/build/debug.mk
+++ b/build/debug.mk
@@ -209,14 +209,25 @@ debug-build-rad: ## Build rad CLI with debug symbols + create drad alias
 	@echo "💡 Use './drad' for debug-configured CLI (preserves 'rad' for your installed version)"
 
 debug-start: debug-setup debug-build-all ## Start k3d cluster and all Radius components as OS processes
-	@echo "Creating k3d cluster..."
-	@if k3d cluster list | grep -q "radius-debug"; then \
-		echo "k3d cluster 'radius-debug' already exists"; \
+	@echo "Ensuring k3d cluster 'radius-debug' is running..."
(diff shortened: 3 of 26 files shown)


sylvainsf:
Closing in favor of #11904, which contains every commit from this branch plus the postgres TIMESTAMPTZ pagination fix, the noncloud-test-learnings doc, additional debug-stack stabilization, and the CI workflow unblock. The PR description over there has been updated to a synthesis covering all of it.

PR #11994, opened 2026-05-26
Document Radius extensibility architecture document

brooke-hamilton:
## Summary

Adds a code-grounded architecture document for Radius extensibility covering:

- resource type registration through `rad resource-type create` and the UCP initializer
- recipe registration through `Applications.Core/environments` and `Radius.Core/recipePacks`
- deployment and delete invocation through UCP routing, dynamic-rp, the recipe engine, and drivers
- invariants and change-safety guidance for resource types, recipes, drivers, and capabilities

Also updates the architecture README so the new extensibility document appears in the start-here list and reading order.

## Validation

- Reviewed the document against the referenced Go implementation paths
- VS Code diagnostics for `docs/architecture/extensibility.md`: no errors found


brooke-hamilton:
@copilot resolve the merge conflicts in this pull request

docs/architecture/extensibility.md
```diff
@@ -0,0 +1,495 @@
...
+  UCP->>DB: persist records
+
+  Note over Dev,Init: Path B — startup seed
+  Init->>Init: read manifest directory, merge by namespace
+  Init->>DB: write ResourceProvider / Type / APIVersion / Location directly
+```
+
+## Recipe Registration
+
+`rad recipe register` stores recipes as entries on an
```
lakshmimsft, line 185: this has now changed to use recipe packs. configuring and deploying a recipepack resource and associating it with an environment is what will now tie recipes to environments. 
oh nvm, i see this is addressed later.
i'm wondering if we should remove references to Applications.Core/* or leave them in for now.

docs/architecture/extensibility.md
```diff
@@ -0,0 +1,495 @@
...
+1. Fetches the target environment.
+2. Builds either a `BicepRecipeProperties` or `TerraformRecipeProperties`
+   value depending on the `--template-kind` flag, populating
+   `TemplateKind`, `TemplatePath`, optional `TemplateVersion` /
+   `PlainHTTP`, and any `Parameters`.
+3. Inserts the recipe under
+   `envResource.Properties.Recipes[<resourceType>][<recipeName>]`.
+4. Calls `CreateOrUpdateEnvironment` to persist.
+
+The resulting `Applications.Core/environments` shape:
```
lakshmimsft: maybe we should give example of Radius.Core/environments with recipe pack

brooke-hamilton: Good idea. Added two YAML examples: a `Radius.Core/environments` resource that references a recipe pack by ID (with `recipeParameters` overrides), and the referenced `Radius.Core/recipePacks` resource showing recipe definitions keyed by resource type.

docs/architecture/extensibility.md
```diff
@@ -0,0 +1,495 @@
...
+fetches the environment's recipe pack IDs, finds the first pack with a recipe
+whose key matches the resource type, and reconciles that recipe's parameters
+with environment-level `recipeParameters` for the same type.
+
+Recipe names are selected by the resource being deployed. For dynamic
+resources, omitting `properties.recipe` does not disable recipes;
+[DynamicResource.GetRecipe](../../pkg/dynamicrp/datamodel/dynamicresource.go)
+returns a recipe named `default` when the property is absent. That means an
+environment recipe named `default` is the implicit convention for resource
+types that should work without an explicit recipe block.
```
lakshmimsft, line 260: we should delete this example (Radius.Data/redis.. ~ln206-217)  and the paragraph above (ln231-236).. we've dropped multiple recipe names for types with Extensibility.. this seems to be a reference to older portable types (where this capability existed)

docs/architecture/extensibility.md
```diff
@@ -0,0 +1,495 @@
...
+  `properties.recipes[<type>]` entries store recipe kind, location,
+  parameters, and `plainHttp` for the `Radius.Core` environment path.
+- **Recipe engine** — selects a driver and runs the recipe to produce
+  resources, values, and secrets.
+
+## Resource Type Registration
+
+A resource type is registered by writing its manifest into UCP under the
+`System.Resources` provider. There are two paths to do this; both end at the
+same database records.
```
lakshmimsft: It would be worth clarifying, the paths are not interchangeable: Path A (rad resource-type create) is how custom / user-defined types are registered against a running cluster, while Path B (the UCP initializer) only runs at UCP pod startup and only registers the built-in/bundled dynamic types that ship with Radius.

docs/architecture/extensibility.md
```diff
@@ -0,0 +1,495 @@
...
+- Dynamic resources default to recipe name `default` when `properties.recipe`
+  is omitted.
+- The driver lookup key is the recipe's `templateKind` for
+  `Applications.Core` recipes and the mapped `recipeKind` for
+  `Radius.Core` recipe packs. Adding a new driver means registering it in the
+  engine's `Drivers` map.
+- Dynamic recipe output only becomes user-visible resource properties when
+  those properties exist in the registered schema and are not basic Radius
+  properties such as `application` or `environment`.
+- Sensitive input fields marked by schema annotations are encrypted at rest and
```
lakshmimsft: This is not accurate. ln 355 above is more accurate.  the backend persists a redacted copy with sensitive fields set to nil — not an encrypted at rest

docs/architecture/extensibility.md
brooke-hamilton: Good catch — the `memorydb` second recipe implied per-type multiple named recipes, which is the legacy portable-resources model rather than the extensibility model. I removed that entry so the `Applications.Core` example now shows a single `default` recipe, and replaced the rest of this section's emphasis with the `Radius.Core/recipePacks` examples (see the reply above). I kept a brief `Applications.Core` description because the recipe map still exists in code for that path — happy to drop the `Applications.Core` path entirely if you'd prefer, since this overlaps with your earlier question about removing `Applications.Core/*` references.

docs/architecture/extensibility.md
brooke-hamilton: Fixed — you're right. Confirmed `schemautil.RedactFields` sets the sensitive paths to `nil`. The invariant now reads that the frontend encrypts the fields when the accepted resource is first stored, then the backend recipe controller decrypts an in-memory copy for execution and persists a redacted resource with those fields set to `nil`, so the durable record holds neither plaintext nor ciphertext. This matches the earlier paragraph you pointed to.

docs/architecture/extensibility.md
brooke-hamilton: Added two YAML examples: a `Radius.Core/environments` resource referencing a recipe pack by ID (with a `recipeParameters` override), and the referenced `Radius.Core/recipePacks` resource showing recipe definitions keyed by resource type (`recipeKind` / `recipeLocation` / `plainHttp`). The property casing matches the 2025-08-01-preview TypeSpec and examples.

docs/architecture/extensibility.md
brooke-hamilton: Clarified. The intro now states the two paths produce equivalent records but are not interchangeable: Path A (`rad resource-type create`) registers custom, user-defined types against a running cluster, while Path B (the UCP initializer) runs only at UCP pod startup and registers the built-in/bundled types that ship with Radius.

lakshmimsft (approved):

PR #11995, opened 2026-05-26
Add default Contour Gateway for route recipes

### LINKED ISSUE #12038: Radius.Compute/routes default recipe does not work after default Kubernetes install
## Bug

`Radius.Compute/routes` is available as a built-in resource type and is included in the default recipe pack, but it does not work end to end after a default `rad install kubernetes`.

The default recipe pack includes:

- `Radius.Compute/containers`
- `Radius.Compute/persistentVolumes`
- `Radius.Compute/routes`
- `Radius.Security/secrets`

The `Radius.Compute/routes` recipe renders Kubernetes Gateway API route resources such as `HTTPRoute`. However, a default Radius install currently installs Contour but does not create the Gateway API infrastructure that the route recipe needs to attach to. The install also does not grant the dynamic RP Gateway API permissions for route resources.

As a result, an application using `Radius.Compute/routes` can fail even though the resource type and default recipe are present.

## Expected behavior

After a default `rad install kubernetes`, users should be able to deploy a Radius app using the default `Radius.Compute/routes` recipe without manually creating ingress infrastructure.

If Contour remains installed by default, Radius should create or otherwise document/provide the default Gateway API resources required by the route recipe, for example:

- `GatewayClass/contour`
- `Gateway/radius` in `radius-system`
- Gateway API RBAC for the dynamic RP

## Actual behavior

The route resource type and default recipe registration exist, but the installed cluster is missing the Gateway API runtime prerequisites. The route recipe may fail to create or attach route resources, or the route may not receive traffic.

## Why this matters

This creates a confusing default experience:

- `Radius.Compute/routes` appears to be supported by default.
- The default recipe pack includes a routes recipe.
- A default Kubernetes install does not provide the Gateway infrastructure that recipe needs.

Users reasonably expect default recipes to work after a default install.

### Workaround 1: Create Gateway API infrastructure manually

Install or use a Gateway API controller, then create a Gateway that the route recipe can attach to. For Contour, create something like:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: contour
spec:
  controllerName: projectcontour.io/gateway-controller
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: radius
  namespace: radius-system
spec:
  gatewayClassName: contour
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
```

Then configure the route recipe to target that Gateway, if the recipe exposes parameters:

- `gateway_name: radius`
- `gateway_namespace: radius-system`

### Workaround 2: Grant dynamic RP Gateway API permissions

Grant the `radius-system/dynamic-rp` service account permissions for Gateway API resources such as:

- `gateways`
- `httproutes`
- `tlsroutes`
- `tcproutes`
- `udproutes`
- `referencegrants`

Then restart the dynamic RP deployment.

### Workaround 3: Register an explicit recipe pack

Create/register a recipe pack that points `Radius.Compute/routes` at an existing Gateway in the cluster. This can be used with Contour Gateway API, NGINX Gateway Fabric, or another Gateway API controller.

For example, use recipe parameters to point the routes recipe at the Gateway created by the platform team.

### Workaround 4: Avoid routes for now

For applications that do not need external ingress, omit `Radius.Compute/routes` and rely on internal Kubernetes services created by `Radius.Compute/containers`.

## Suggested fix

Make the default install and default routes recipe consistent. If Contour remains installed by default, Radius should create the shared Gateway API resources and grant dynamic RP Gateway API RBAC during install/upgrade, then clean up Radius-managed Gateway resources during uninstall.

Alternatively, remove `Radius.Compute/routes` from the default recipe pack until Radius install provides the prerequisites needed for the default recipe to work.

willdavsmith:
## Summary

This PR updates the Contour recipe-pack design and implementation direction to use Gateway API as the default Contour-backed route path.

The default model is now:

- `rad install kubernetes` continues installing Contour by default for now.
- When Contour install is enabled, Radius creates the shared Gateway API infrastructure: `GatewayClass/contour` and `Gateway/radius` in `radius-system`.
- The default `Radius.Compute/routes` recipe is registered with `gatewayName=radius` and `gatewayNamespace=radius-system` so it can attach route resources to the managed Gateway.
- The default `Radius.Compute/routes` recipe renders Gateway API route resources such as `HTTPRoute`.
- Applications do not define an application-level gateway resource in the default path.

The design still keeps removal of default Contour installation as a separate review decision.

## Scope

- Adds install/upgrade/uninstall handling for the managed default Contour Gateway API `Gateway`.
- Adds default Gateway parameters to the built-in `Radius.Compute/routes` recipe registration.
- Grants the dynamic RP Gateway API permissions for Gateway and route resources.
- Rewrites the design note to describe Contour Gateway API as the default recipe path.
- Documents HTTPProxy as an alternate compatibility recipe-pack option, not the default.

## Related work

- Issue: https://github.com/radius-project/radius/issues/11952
- Demo repo: https://github.com/willdavsmith/radius-nginx-demo
- Companion recipe PR: https://github.com/radius-project/resource-types-contrib/pull/179

## Validation

Local:

- `GOCACHE=/private/tmp/radius-go-build-cache go test ./pkg/cli/helm ./cmd/rad/cmd`
- `GOCACHE=/tmp/radius-nginx-demo-go-cache GOMODCACHE=/tmp/radius-nginx-demo-go-mod go test ./pkg/cli/recipepack`
- `GOCACHE=/tmp/radius-nginx-demo-go-cache GOMODCACHE=/tmp/radius-nginx-demo-go-mod go test ./pkg/cli/cmd/env/create/preview ./pkg/cli/cmd/deploy`

Demo e2e:

- Contour Gateway API recipes: https://github.com/willdavsmith/radius-nginx-demo/actions/runs/26665457465
- NGINX Gateway API recipes: https://github.com/willdavsmith/radius-nginx-demo/actions/runs/26665457417

## Notes for reviewers

The key design point is that Radius can keep the same application shape while moving provider-specific route rendering into recipes. Contour remains installed by default for now; the default application routing path moves to Gateway API so the same route recipe shape can work with Contour, NGINX Gateway Fabric, or another Gateway API controller.

Fixes: https://github.com/radius-project/radius/issues/12038


eng/design-notes/recipes/2026-05-contour-recipe-packs.md
```diff
@@ -0,0 +1,210 @@
...
+Radius already has a default recipe pack experience for development scenarios. `rad init --preview` creates a default recipe pack named `default` in `/planes/radius/local/resourceGroups/default` and links it to the created environment. `rad deploy` also creates or fetches that default recipe pack and injects it into environment resources that do not specify recipe packs.
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/11952
+
+### Goals
+
+- Keep Contour installed by default for now.
+- Create the shared Contour Gateway API `Gateway` during Radius install when Contour install is enabled.
```
willtsai: When Contour install is disabled (e.g. using `rad install kubernetes --skip-contour-install `), will the Contour installation and Contour Gateway API `Gateway` creation be skipped? That's the desired behavior, so it would help to make that clearer in this design doc.

willdavsmith: done

eng/design-notes/recipes/2026-05-contour-recipe-packs.md
```diff
@@ -0,0 +1,210 @@
...
+          containerName: 'web'
+          containerPort: 80
+        }
+      }
+    ]
+  }
+}
+```
+
+The default route recipe attaches HTTP and TLS routes to `Gateway/radius` in `radius-system`. Users who want a different Gateway API controller, such as NGINX Gateway Fabric, can select a different recipe pack or pass recipe parameters that target a different Gateway.
```
sk593, line 83: Just confirming, the routes recipe has support for TCP and UDP routes depending on the application template. The expectation is that this would not be supported with a Contour installation, correct (I don't think Contour natively supports these)? i.e. we'd use either HTTP or the TLS routing with the Gateway API for default Contour installations

eng/design-notes/recipes/2026-05-contour-recipe-packs.md
```diff
@@ -0,0 +1,210 @@
...
+
+The existing `default` recipe pack should use the Gateway API `Radius.Compute/routes` recipe. Contour installation and recipe selection are separate concerns:
+
+- Installing Contour adds the ingress controller and Gateway API support to the cluster.
+- Radius install creates the shared Contour `Gateway` only when Contour install is enabled.
+- The default route recipe renders application routes that attach to that Gateway.
+
+Today the default recipe pack follows the Radius version channel, including `latest` on the edge channel. A future hardening step should pin default recipes to the Radius release or another explicit artifact version so the default experience does not depend on floating recipe artifacts.
+
+If Radius later stops installing Contour by default, default Gateway creation and default route recipe selection should be revisited together.
```
sk593, line 136: Could use PM input here. We might not need to revisit the design it if the expectation is that users bring their own networking infrastructure. This would just need to be called out explicitly in documentation

pkg/cli/helm/contour_gateway.go
```diff
@@ -0,0 +1,234 @@
...
+				"namespace": DefaultContourGatewayNamespace,
+				"labels":    radiusManagedLabels(),
+			},
+			"spec": map[string]any{
+				"gatewayClassName": ContourGatewayClassName,
+				"listeners": []any{
+					map[string]any{
+						"name":     "http",
+						"protocol": "HTTP",
+						"port":     int64(80),
```
sk593, line 246: Should we be dynamically allocating ports? Wondering what the outcome is if the user already has these ports being used

eng/design-notes/recipes/2026-05-contour-recipe-packs.md
willdavsmith: The `routes` recipe can model HTTP/TLS/TCP/UDP, but the default Contour install should only support HTTP and TLS via the managed 80/443 listeners. TCP needs an explicit user-provided listener/port, and UDP is not supported by Contour’s Gateway API path today. I’ll clarify that in the design doc.

pkg/cli/helm/contour_gateway.go
willdavsmith: I think these ports are the standard for default gateway access - if users want better customization they could disable the install and bring their own gateway

eng/design-notes/recipes/2026-05-contour-recipe-packs.md
willdavsmith: I guess the decision for now is to keep the default route recipe as a simple gateway API route resource so that it's compatible with most BYO-gateways. this PR just introduces a default gateway installation as well

pkg/cli/helm/cluster.go
```diff
@@ -246,6 +246,11 @@ func (i *Impl) InstallRadius(ctx context.Context, clusterOptions ClusterOptions,
 		return fmt.Errorf("failed to apply Contour Helm chart, err: %w", err)
 	}
 
+	output.LogInfo("Configuring Radius Contour Gateway...")
+	if err := configureDefaultContourGateway(ctx, kubeContext); err != nil {
+		return fmt.Errorf("failed to configure Radius Contour Gateway, err: %w", err)
+	}
```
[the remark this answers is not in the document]
willdavsmith: Fixed in `4ba74d2e7`: the Contour install path now forces `Wait=true` for the Contour chart before configuring the managed Gateway API resources, so CRDs/resources are available before reconciliation runs.

pkg/cli/helm/contour_gateway.go
```diff
@@ -0,0 +1,234 @@
...
+	existing, err := resource.Get(ctx, DefaultContourGatewayName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		_, err = resource.Create(ctx, newContourGateway(), metav1.CreateOptions{})
+		return err
+	} else if err != nil {
+		return err
+	}
+
+	if !isRadiusManaged(existing) {
+		return fmt.Errorf("Gateway %q in namespace %q already exists and is not managed by Radius", DefaultContourGatewayName, DefaultContourGatewayNamespace)
```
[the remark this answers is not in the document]
willdavsmith: Fixed in `4ba74d2e7`: changed the new Gateway error string to start lowercase.

willtsai (dismissed):
The design and user scenarios look good to me. I'll leave it up to @sk593 to review the code changes.

sk593 (approved):

PR #12011, opened 2026-05-28
Resource type icons feature specification

brooke-hamilton:
# Description

Adds the feature specification and requirements checklist for the **Resource Type Icons** feature (spec 003). Resource type authors can attach an optional SVG icon to a resource type so visual graph renderings (e.g., dashboard) can distinguish types. Icons travel with the resource type manifest via the same registration path used by both contributed and built-in types, are addressed by content hash, and fall back to a default when absent.

This PR contains specification documents only — no implementation changes. Files added:

- `specs/003-resource-type-icons/spec.md` — full feature specification, including authoring, registration, and rendering flows.
- `specs/003-resource-type-icons/checklists/requirements.md` — requirements checklist.
- `.specify/feature.json` — feature tracking entry.

## Type of change

- This pull request is a design document and only includes files in the eng/design-notes directory.

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [x] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable


specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+
+On a fresh install, every `Applications.Core/*` type listed in FR-013 (`environments`, `applications`, `containers`, `gateways`, `secretStores`, `extenders`) renders with its own distinct icon — not the default, not the same as each other. These types drive a new user's first-run experience.
+
+**Why P2**: New users see built-ins before contributed types. The first graph they render must be meaningful.
+
+**Test**: Fresh install, deploy a sample app exercising each `Applications.Core/*` type, run an icon check on each node and confirm no node resolves to the default-icon hash.
+
+### Story 4 — Contributor guidance and CI safety net (P2)
+
+A pull request that adds a malformed, oversized, or non-SVG icon is rejected by CI with an actionable error. A pull request in the Radius product repository that removes or breaks the embedded product default icon is also rejected. Documentation tells contributors exactly where to put the file.
```
nicolejms: How do we ensure no malicious code in the svg file? Are there existing tools for this?

brooke-hamilton: [FR-005a](https://github.com/radius-project/radius/pull/12011/changes#diff-ea5dbf53d7e22a0842e088a6cf68e55da60665a7e8d1df600b8984cae898194dR164) defines validations to prevent scripts and links and other non-SVG elements embedded in the file. We should be able to rely on either an existing Go library module for this scanning or derive our own scanning logic from the known set of SVG extension capabilities.

brooke-hamilton: We could also run an SVG minifier to remove tool comments, but that would affect the hash of the SVG file (and if we change the minifier that would affect all existing SVGs) so we would have to carefully consider where in the tool chain we add such a thing.

brooke-hamilton:
To Do: Add a runtime check for SVG extensions and remove them.

specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+    Fetch --> Cache[Browser cache /<br/>CDN serves on repeat<br/>NFR-007, SC-007]
+    Inline --> Render[Sanitize SVG NFR-006<br/>and render on node]
+    Cache --> Render
+```
+
+Renderer-side behavior on a 404, network failure, or malformed bytes (e.g. drawing a fallback glyph) is the rendering surface's concern and is out of scope for this spec — see [Out of Scope](#out-of-scope).
+
+Non-rendering consumers (`rad app graph -o json`, scripting clients) take the endpoint path by default — they receive `iconHash` only.
+
+## User Scenarios *(mandatory)*
```
DariuszPorowski, line 101: How about reusing existing libraries with icons? i.e. Fluent UI, provides massive amount of icons - it means we do not need to maintain own.

specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+    subgraph RenderBox["Rendering surface"]
+        Renderer([e.g. dashboard])
+    end
+
+    Author -- "authors YAML + sibling SVG(s)" --> CLI
+    Renderer -- "fetch graph<br/>(iconHash per node)" --> GraphAPI
+    Renderer -- "fetch SVG bytes by hash" --> Endpoint
+```
+
+## Authoring flow
```
Reshrahim, line 36: What is the name of the property for icons in the resource type yaml definition? Is it a required or optional property and What would be the value ?

An example of the type definition would be helpful

specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+
+A contributor creates a sibling `icon.svg` file next to a resource type definition and runs the existing publish command. The icon flows through publish → control plane → graph → dashboard with no hand-editing of the definition file.
+
+**Why P1**: Authors will not adopt a contract that requires hand-editing embedded XML in YAML.
+
+**Test**: Add `icon.svg` to a type folder, publish, run an icon check using the new icon's hash.
+
+### Story 3 — Built-in types ship with custom icons (P2)
+
+On a fresh install, every `Applications.Core/*` type listed in FR-008 (`environments`, `applications`, `containers`, `gateways`, `secretStores`, `extenders`) renders with its own distinct icon — not the default, not the same as each other. These types drive a new user's first-run experience.
```
willtsai, line 121: assuming this applies to the new style `Radius.Core/*` built-in resource types too?

specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+    A[Author edits<br/>fooBars.yaml] --> B{Need different icons<br/>per type in this YAML?}
+    B -- "no, one icon for all" --> C[Create sibling<br/>icon.svg]
+    B -- "yes" --> D[Create one sibling<br/>icon.&lt;typeName&gt;.svg<br/>per type that needs<br/>a distinct icon]
+    B -- "no icon at all" --> E[Skip — type will<br/>resolve to default<br/>at registration time]
+    C --> F[Run rad resource-type create<br/>-f fooBars.yaml]
+    D --> F
+    E --> F
+```
+
+The YAML definition for resource type authoring is not affected by the icon feature.
```
lakshmimsft: icon.\<typename\>.svg is not differentiating multiple icons for the same type.

specs/003-resource-type-icons/spec.md
brooke-hamilton: The yaml definition edited by resource type authors would not change. The icon would be added as a sibling file with a predefined naming convention. However, the internal data storage schema that stores the type would have an additional property for the SVG file.

specs/003-resource-type-icons/spec.md
```diff
@@ -0,0 +1,292 @@
...
+
+A developer views an application graph. Every node displays a meaningful icon — even when the type's author provided none — because the build always falls back to the repository-wide default icon.
+
+**Why P1**: Without a guaranteed default, every consumer must invent its own default icon.
+
+**Test**: Register a type that declares no icon, deploy an app using it, run an icon check using the default icon's hash.
+
+### Story 2 — Authoring a custom icon (P1)
+
+A contributor creates a sibling `icon.svg` file next to a resource type definition and runs the existing publish command. The icon flows through publish → control plane → graph → dashboard with no hand-editing of the definition file.
```
Reshrahim: Would the Resource type definition contain a reference to the `icon.svg` file ?

specs/003-resource-type-icons/spec.md
brooke-hamilton: Customers could pull from any SVG icon library and I think what you are suggesting makes sense as an authoring experience, i.e., maybe we can provide this as a default.

specs/003-resource-type-icons/spec.md
brooke-hamilton: Yes - all types will have either a default icon or a predefined icon.

specs/003-resource-type-icons/spec.md
brooke-hamilton: Multple type icons would be defined by the convention `icon.<type-name>.svg`, and each type would have only one icon.

specs/003-resource-type-icons/spec.md
brooke-hamilton: No - the resource definition does not have a reference to the icon file because we will use a naming convention to identify the right file.

willtsai: I got a bit confused by the term "sibling" at first, but after half way through this spec I've understood what it means (I think). Help me validate my understanding: to add a custom icon for a given resource type(s), the icon(s) should be added to the same parent folder (hence, "sibling" SVG files) as the resource type(s) definition file and the SVG icon files must follow the naming conventions you've described. Then, the Radius tooling will be able to resolve the icon files to their respective resource types based on the SVG location and name. Correct?

zachcasper: We should not use magic file names. I suggest
```
rad resource-type create -f my-types.yaml \
  --icon <path> \
  --icon <typeName>=<path>
```
When the YAML only contains a single resource type, `--icon icon.svg` is sufficient. When there are multiple resource types, the user can specify the type name. For example:

```
rad resource-type create -f my-types.yaml \
  --icon containers=container.svg \
  --icon gateways=gateway.svg
```
There is an edge case to resolve where there are name collisions.

brooke-hamilton: Yes - and thank you I will clarify and simplify the language.

specs/003-resource-type-icons/spec.md
lakshmimsft: makes sense..i misread 'Need different icons per type in this YAML?' :)

specs/003-resource-type-icons/spec.md
brooke-hamilton: Done - removed the sibling svg icon implementation and replaced it with the CLI syntax you suggested.

brooke-hamilton: Done - sibling language is now removed.

nithyatsu (approved):

PR #12017, opened 2026-05-29
Improve line-number accuracy in radius-code-review skill

brooke-hamilton:
## Description

Improves the `radius-code-review` skill to eliminate incorrect line numbers, which is a frequent issue with the review scripts. An alternate approach is to remove the generated script altogether. 

## Changes

`.github/skills/radius-code-review/SKILL.md`:

- **Anchor-based comments**: Every review comment now records an *anchor* (a unique code snippet on the target line). The anchor — not the integer — is the source of truth.
- **New "Line-number accuracy (mandatory)" section**: Instructs the agent to never type a line number from memory and to look up each line via `grep -n`, and to ensure the cited line is on the `RIGHT` side of the diff.
- **Mandatory line-number verification (Step 3)**: Adds a step to print actual file content at every cited line and confirm it matches the comment's anchor before generating the script.
- **Runtime anchor resolution (Step 5)**: Adds a `line_for` shell helper that resolves line numbers from anchors at script build time and fails loudly if an anchor matches zero or more than one line, so a wrong/stale anchor aborts the script instead of mis-placing a comment.
- **Checklist updates**: Adds items requiring anchors and line-number verification.

## Type of change

- This pull request fixes a bug or improves an existing skill.


sylvainsf (approved):
lgtm

Issue #6934, opened 2023-12-07
Manage applications in multiple environments on separate Kubernetes clusters


rynowak:
## Overview of feature request

Radius should be able to manage applications across environments that are not on the same Kubernetes cluster as the control plane.

Examples:

- Radius is installed in Kubernetes cluster `control-plane`. It can manage applications on Kubernetes clusters `prod-westus` and `prod-eu-2` using environments that target those clusters (`env-prod-westus`, `env-prod-eu-2`).
- Radius is running on a VM. It can manage applications on Kubernetes clusters `prod-westus` and `prod-eu-2` using environments that target those clusters (`env-prod-westus`, `env-prod-eu-2`).

**Out of scope:** Defining an environment that spans multiple compute clusters. Each environment will map to a single Kubernetes cluster. A Kubernetes cluster can still contain multiple environments.

## Acceptance criteria

- Decouple Radius from the Kubernetes cluster where applications are running.
- Define environments in a way that identifies and holds credentials for a specific Kubernetes cluster.
- Manage applications on (potentially multiple) Kubernetes clusters. Use the environment to select the destination.

[AB#10707](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/10707)

sylvainsf:
## Status update: multi-cluster work is landing in two stages

The design and implementation for deploying to external clusters is underway in #12106 (`multi-cluster` branch), with the technical design in `eng/design-notes/environments/2026-06-multi-cluster.md`.

Heads up that this is being delivered in two stages, and **this issue will stay open until the second stage lands**:

- **v1 (in progress, #12106):** a single `ClusterAccessResolver` seam plus an injected-kubeconfig path (`RADIUS_TARGET_KUBECONFIG` / `KUBE_CONFIG_PATH`). This unblocks deploying recipes to an external EKS/AKS cluster via the Repo Radius `verify`/`deploy` workflows, but the target cluster is supplied as a mounted kubeconfig — **not** named on the environment resource.
- **v2 (planned):** the named-cluster UX this issue actually describes — `providers.aws.eksClusterName` / `providers.azure.aksClusterName` on the environment, with Radius deriving cluster credentials in-process from the registered cloud credential. This is the stage that lets users "manage applications in multiple environments on separate clusters" directly through Radius.

[...]

PR #12106, opened 2026-06-15
Multi-cluster deployment: design + v1 implementation

sylvainsf:
# Description

Adds the **Multi-Cluster Deployment** technical design under `eng/design-notes/`
and is the working branch for its **v1** implementation.

The design introduces a single `ClusterAccessResolver` seam that decides, per
recipe execution, which Kubernetes cluster to deploy to and how to authenticate,
returning an in-memory `*rest.Config` that both the Bicep and Terraform recipe
engines consume instead of assuming the in-cluster config.

**v1 (this branch)** aligns to the Repo Radius `verify`/`deploy` workflows in the
separate [`github-extension`](https://github.com/radius-project/github-extension)
consumer repo. Radius owns the integration contract those workflows rely on — the
`RADIUS_TARGET_KUBECONFIG` / `KUBE_CONFIG_PATH` env-var seam — and v1 honors an
injected kubeconfig.

**v2 (future)** delivers the feature spec's named-cluster UX
(`aws.eksClusterName` / `azure.aksClusterName`) and in-process cloud-derived
credential acquisition (EKS `DescribeCluster` + STS presign; AKS
`ListClusterUserCredentials` + Entra token) behind the same seam.

The design documents where it diverges from the product-authored
[feature spec](https://github.com/radius-project/radius/blob/main/eng/design-notes/environments/2026-05-external-kubernetes.md)
and records the EKS ~15-minute token-lifetime decision (ship with the limit).

> **Scope vs. #6934:** This PR (v1) is a partial step toward
> [#6934](https://github.com/radius-project/radius/issues/6934). It enables
> recipe execution against an external cluster via an injected kubeconfig, but it
> does **not** deliver the full "name a cluster on the environment resource" UX
> that #6934 describes. #6934 stays open until the v2 named-cluster work lands.
> Hence this PR uses "Part of" rather than "Fixes".

## Execution plan (v1, all on this branch)

- **A. Resolver seam + `localStrategy`** — new `pkg/recipes/kubernetes/clusteraccess`;
  move the duplicated `rest.InClusterConfig()` / local-kubeconfig logic out of the
  Terraform provider/backend builders into one place. No behavior change.
- **B. `injectedKubeconfigStrategy`** — honor `RADIUS_TARGET_KUBECONFIG`
  (precedence: injected → local).
- **C. Terraform engine integration** — route the executor through the resolver
  per execution; keep `KUBE_CONFIG_PATH`; backend stays local.
- **D. Bicep engine integration** — render a kubeconfig from the resolved
  `*rest.Config` and thread it to the `extension kubernetes` provider; no
  Deployment Engine ServiceAccount change.
- **E. Tests + acceptance** — unit tests for strategy selection/config assembly;
  acceptance gate is the Repo Radius `verify`/`deploy` workflows green against a
  build of this branch (EKS + AKS, Bicep + Terraform). Depends on the OIDC
  workflow work for test clusters/permissions.

Order: A → (B, C, D in parallel) → E. v1 deliverables land together on this
branch; the design can be revised if review changes things.

> Draft: opened for design review. Implementation commits will follow on this
> branch.

## Type of change

Part of: #6934

## Contributor checklist

    - [x] Not applicable
    - [x] Yes
    - [ ] Not applicable
    - [x] Not applicable
    - [x] Not applicable
    - [ ] Not applicable


brooke-hamilton (changes requested):
## Validation summary

I ran the multi-cluster E2E from this branch (`a397595`) in a local WSL setup, mirroring the CI `multicluster-noncloud` leg: two kind clusters (control-plane `kind` + workload `external`), images built/pushed from this branch, `rad install` with `global.targetCluster.enabled=true`, and the PR's own functional tests against a real injected `RADIUS_TARGET_KUBECONFIG`.

**Both in-PR Go code paths work** — but only after two fixes. The feature mechanism is sound; the packaging/test wiring is not, so the PR is not CI-green as-is.

| Test (PR code path) | Result |
|---|---|
| Unit tests (clusteraccess, tf provider, asyncworker, kubeutil) | ✅ Pass |
| Helm render wiring (volume/mount/env on all 3 deployments) | ✅ Correct |
| `Test_MultiCluster_LegacyContainer` (applications-rp direct render) | ❌→✅ Pass *only* after Finding 1 fix |
| `Test_MultiCluster_TerraformRecipe` (dynamic-rp tf provider) | ❌→✅ Pass *only* after Finding 1 **and** Finding 2 fix |
| `Test_MultiCluster_BicepContainer` (DE — not in this PR's Go) | ⏭ Not run (needs HTTPS registry; outside this PR's Go scope) |

---

### Finding 1 — BLOCKING chart bug: injected kubeconfig secret is unreadable by the RP process

`deploy/Chart/templates/_helpers.tpl:127` mounts the `target-kubeconfig` secret with `defaultMode: 0400` (owner-read-only, owned `root:root`) in the `radius.targetCluster.volume` helper. All three RP images are distroless **nonroot** (uid 65532), and the pod `securityContext` is `{}` (no `fsGroup`/`runAsUser`; the container only sets `allowPrivilegeEscalation: false`). So the process cannot read its own kubeconfig:

```
failed to load target kubeconfig from RADIUS_TARGET_KUBECONFIG=".../kubeconfig":
open /etc/radius/target-kubeconfig/kubeconfig: permission denied
```

Because the helper is shared by `de/deployment.yaml`, `rp/deployment.yaml`, and `dynamic-rp/deployment.yaml`, this breaks **every** deploy when `targetCluster.enabled=true`. I proved the root cause by patching the live `defaultMode` 0400 → 0444, after which the legacy container test passes. This should also fail this PR's own CI `multicluster-noncloud` leg.

**Suggested fix:** `defaultMode: 0444`, or add pod `securityContext.fsGroup` + `0440`.

---

### Finding 2 — Terraform recipe test can never pass in CI: app namespace is never created on the external cluster

After applying the Finding 1 fix, `Test_MultiCluster_TerraformRecipe` still fails with a raw Terraform error:

```
terraform apply failure: exit status 1
Error: namespaces "mcluster-resources-terraform-redis-app" not found
```

The redis Terraform module assumes the target namespace already exists. The direct-render path emits the namespace into its rendered output (so the legacy container test creates it on the external cluster and works), but the recipe path has no equivalent step.

The design doc explicitly makes namespace auto-creation a non-goal:
- line 134: "Auto-creating namespaces on the external cluster (the workflow does this)."
- line 627: "Target namespace missing → clear error; no auto-creation (the workflow creates the namespace)."

However, neither the test, the test helpers (`external_cluster_test.go`), nor the CI `multicluster-noncloud` setup (which only pre-creates `radius-system` + the `target-kubeconfig` secret on the external cluster) ever creates the per-test **app** namespace on the external cluster. As written, the test will always fail in CI.

Two sub-issues:
1. The test is missing a "simulate the provisioning workflow" step that pre-creates the app namespace on the external cluster.
2. The design promises a *clear* error when the target namespace is missing, but a raw Terraform "namespace not found" leaks through instead — that clear-error UX isn't implemented.

**Proof:** pre-creating the namespace on the external cluster (simulating "the workflow") makes the test pass — the redis Deployment/Service land on the external cluster, the tfstate secret stays on the control-plane `radius-system`, and nothing lands on the control plane. This confirms the PR's Terraform kubernetes-provider redirect itself is correct.

---

### Not run: `Test_MultiCluster_BicepContainer`

Exercises the deployment-engine Bicep path (not part of this PR's Go) and needs locally-published `Radius.Compute`/`Radius.Core` preview types, which require an HTTPS registry (`bicep publish-extension` rejects an insecure HTTP registry). Lower priority since both in-PR Go code paths are already validated.

---

### Bottom line

The core multi-cluster v1 mechanism — inject kubeconfig → recipes and directly-rendered resources land on the external cluster while state stays on the control plane — **genuinely works**. But as-is the PR is not mergeable/CI-green: Finding 1 is a blocking chart bug, and Finding 2 means the Terraform functional test cannot pass in CI. Both are straightforward to address. Happy to re-validate after the fixes.

*Validation only — no PR/repo files were modified; the only changes were runtime `kubectl patch`/namespace creates on throwaway local clusters.*

brooke-hamilton (dismissed):
## Re-validation: both findings resolved — approving

I re-ran the full multi-cluster E2E from the updated head (`4045274`) **without any of my previous manual workarounds**. Both issues from my earlier `CHANGES_REQUESTED` review are now fixed, and the multi-cluster mechanism works out-of-the-box from the chart.

### Setup
Mirrored the CI `multicluster-noncloud` leg locally: two kind clusters (control-plane + workload `external`), images rebuilt/pushed from this branch, `helm upgrade` with the in-tree chart and `global.targetCluster.enabled=true`, then a `rollout restart` to force a re-pull of the freshly built images. All three deployments now render `defaultMode: 0444` directly from the chart (verified `292` on the live volumes — no runtime patch).

### Results

| Test (PR code path) | Result |
|---|---|
| Unit tests (incl. new `pkg/kubeutil` helpers) | ✅ Pass |
| `Test_MultiCluster_LegacyContainer` (applications-rp direct render) | ✅ **Pass out-of-the-box** (30.4s) |
| `Test_MultiCluster_TerraformRecipe` (dynamic-rp tf provider) | ✅ **Pass out-of-the-box** (36.2s) |
| `Test_MultiCluster_BicepContainer` (DE Bicep path) | ⏭ Not run (outside this PR's Go; needs an HTTPS registry for local `Radius.Compute` types) |

### Finding 1 (was BLOCKING) — RESOLVED
`_helpers.tpl` now mounts the `target-kubeconfig` secret with `defaultMode: 0444` instead of `0400`. The non-root (distroless, uid 65532) RP/DE containers can now read the injected kubeconfig. The legacy container test passes with no `permission denied` and no manual patch.

### Finding 2 (terraform test could never pass in CI) — RESOLVED
The app namespace is now created on the **deployment-target (external)** cluster:
- `CreateAppScopedNamespace` (`updatefilter.go`) resolves its client via `kubeutil.DeploymentTargetRuntimeClient`, so the legacy app-scoped namespace lands on the external cluster.
- The new-type path (`createorupdateenvironment.go`) validates namespace existence against the target cluster and returns a clear `BadRequest` when missing (implements the design's "clear error" UX), defaulting an unset namespace via `TargetClusterDefaultNamespace`.
- The test harness (`rptest.go`) provisions the namespace on the external cluster too.

I deleted the external app namespace beforehand and confirmed the **product code** recreated `mcluster-resources-terraform-redis-app` on the external cluster during deploy. The redis Service landed on the external cluster, the Terraform state secret stayed on the control-plane `radius-system`, and nothing leaked onto the control plane.

### Note (non-blocking)
`Test_MultiCluster_BicepContainer` still wasn't exercised here — it drives the deployment-engine Bicep path (not part of this PR's Go) and needs locally-published `Radius.Compute`/`Radius.Core` preview types, which require an HTTPS registry. CI covers it; nothing to change in this PR.

Nice, thorough response to the review — LGTM.

*Validation only — no PR/repo files were modified; the only changes were on throwaway local clusters.*

eng/design-notes/environments/2026-06-multi-cluster.md
```diff
@@ -0,0 +1,943 @@
...
+### Detailed Design
+
+The design is presented as changes aligned with the product's layers. v1 changes
+are the ones required for the Repo Radius `deploy` workflow to work against
+Radius `main`; v2 changes deliver the feature spec's named-cluster UX behind the
+same seam.
+
+#### Change 1 (v1) — Honor an injected kubeconfig
+
+Make recipe execution honor `RADIUS_TARGET_KUBECONFIG`, a path to a kubeconfig
```
sylvainsf, line 292: Correct, that's what sets it in the app use case. It's set in tests for the purpose of testing the functionality, but the usage happens outside the radius codebase.

nithyatsu (approved):

brooke-hamilton (approved):

PR #11780, opened 2026-04-29
Add Radius.Core/terraformConfigs and bicepConfigs resources

sylvainsf:
# Description

Adds two new resources in the `Radius.Core` namespace — `terraformConfigs` and `bicepConfigs` — that environments reference by resource ID. Platform teams can now define private registry authentication, Terraform CLI configuration (`provider_installation`, `credentials`, env vars), and Bicep registry authentication once and share it across multiple environments.

This implements a scoped subset of [design-notes#107](https://github.com/radius-project/design-notes/pull/107). Out of scope: Terraform binary lifecycle (`rad terraform install`), backend state stores, installer pipeline. See [docs/architecture/terraform-bicep-config.md](docs/architecture/terraform-bicep-config.md) for the full design.

**What's included**

- New TypeSpec/OpenAPI resources `Radius.Core/terraformConfigs` and `Radius.Core/bicepConfigs` (regenerated SDK).
- New `terraformConfig?` and `bicepConfig?` properties on `Radius.Core/environments`, validated for existence at PUT time.
- Datamodels, bidirectional API↔datamodel converters, REST controllers wired in `pkg/corerp/setup/setup.go`.
- Loader bridge in `pkg/recipes/configloader/environment.go` that resolves the new config resources at recipe execution time and populates `RecipeConfig` consumed by the shared driver.
- New `ProviderInstallation` field on the shared `TerraformConfigProperties` datamodel; the Terraform driver writes a `.terraformrc` to the working dir and points Terraform at it via `TF_CLI_CONFIG_FILE` on both Deploy and Delete. Field is optional and only populated by the Radius.Core path — the legacy `Applications.Core` `recipeConfig` flow is unchanged.
- Unit tests for the `.terraformrc` writer, the env-var setter, and the Delete-path helper.
- Functional test (`Test_TerraformConfig_BicepConfig_Combined`) deploying both configs in a single environment and running a Terraform recipe end-to-end.
- Architecture doc: [docs/architecture/terraform-bicep-config.md](docs/architecture/terraform-bicep-config.md).

**Known follow-ups (filed separately)**
- `bicepConfig.registryAuthentication.{AzureWI,AwsIrsa}` are accepted by the schema but not yet wired in the Bicep driver (only `BasicAuth` flows through today).
- `referencedBy` on the config resources is exposed in the schema but not yet populated.
- Delete protection (409 when a config is referenced by an environment) is not yet implemented.

## Type of change

- This pull request adds or changes features of Radius and has an approved issue (issue link required).

[...]

PR #12181, opened 2026-06-19
Refactor recipe settings resources to match the feature spec and support Radius.Security/secrets

### LINKED ISSUE #12122: bicepConfigs / terraformConfigs reject Radius.Security/secrets for registry secrets; only Applications.Core/secretStores accepted
### Steps to reproduce

1. Create a Radius.Security/secrets resource containing registry credentials.
2. Create a Radius.Core/bicepConfig (or Radius.Core/terraformConfig) that references that secret for registryAuthentications/basicAuth.
3. Apply the configuration (or run the functional test that references the config).
4. Observe that the config is rejected or the secret reference is not accepted.

Example/test data that currently works (uses Applications.Core/secretStores):
- test/functional-portable/dynamicrp/noncloud/resources/testdata/bicepconfig-test.bicep
  https://github.com/radius-project/radius/blob/6ec9c74569057b24aaa4ff8bd5326df63647af91/test/functional-portable/dynamicrp/noncloud/resources/testdata/bicepconfig-test.bicep

### Observed behavior

Radius.Core/bicepConfigs and Radius.Core/terraformConfigs accept Applications.Core/secretStores for registryAuthentications/basicAuth secret references, but attempts to reference Radius.Security/secrets are rejected or ignored. Functional test data in the repository uses Applications.Core/secretStores (see link above) and succeeds.

### Desired behavior

Allow Radius.Security/secrets to be referenced by Radius.Core/bicepConfigs and Radius.Core/terraformConfigs as an equivalent secret resource. Additionally, Radius.Core/bicepConfigs and Radius.Core/terraformConfigs should be compatible with all the new extensibility architecture resources, including Radius.Core/applications, Radius.Core/environments, Radius.Security/secrets, etc.

### Workaround

Workaround: Use Applications.Core/secretStores for registry secrets (this is what the current functional tests use).

### rad Version

0.58.0

### Operating system

_No response_

### Additional context

_No response_

sylvainsf:
# Description

Fixes #12122. `Radius.Core/bicepConfigs` and `Radius.Core/terraformConfigs` rejected `Radius.Security/secrets` for registry/module credentials, accepting only `Applications.Core/secretStores`. Fixing that surfaced two deeper problems — a naming divergence from the feature spec and a secret-kind coupling in the Bicep auth path — which this PR also corrects.

## 1. Support `Radius.Security/secrets` in the recipe secret loader (the bug)

The recipe secret loader was hardcoded to the `Applications.Core/secretStores` `ListSecrets` API and never dispatched on the referenced resource type. It now parses the secret resource ID and routes accordingly:

- `Applications.Core/secretStores` keeps the existing `ListSecrets` path, behavior-identical.
- `Radius.Security/secrets` is read by locating the backing Kubernetes Secret its recipe materializes (via `status.outputResources`) and reading the values directly. This is required because a provisioned `Radius.Security/secrets` has its sensitive `data` redacted from the database once provisioning succeeds, so the plaintext only lives in the backing Kubernetes Secret.

The Kubernetes client provider is threaded into the loader at both wiring sites (dynamic-rp and the recipe controller config).

## 2. Refactor: rename `Configs` → `Settings` to match the spec

These resource types were introduced (in #11780) as `Radius.Core/terraformConfigs` / `bicepConfigs` based on the earlier branch work by Yetkin/Vishwa, but the authoritative feature spec authored by @zachcasper (`terraform-bicep-settings`) names them `Radius.Core/terraformSettings` / `bicepSettings`. The naming divergence originated here, not in the spec, and the types are not yet referenced on the docs site — so this renames them to match the spec before they ship more widely:

- TypeSpec models/segments and the environment `terraformSettings` / `bicepSettings` properties.
- Datamodel types, converters, conversions, the frontend validator package, setup routing, the built-in provider manifests, and the embedded OpenAPI metadata mapping.
- Regenerated clients, fakes, swagger, and bicep-types.

The legacy `recipeConfig` transport types (`datamodel.BicepConfigProperties`, `datamodel.TerraformConfigProperties`) and the `Applications.Core/environments` path are intentionally untouched — these are new-environment-only resources.

## 3. Refactor: select Bicep registry auth from the settings resource, not the secret kind

Per the spec, the secret `kind` is optional and the Bicep auth method comes from the `bicepSettings` resource. Previously the Bicep driver chose the registry auth client from the secret's kind, so a kind-less `Radius.Security/secrets` carrying only `username`/`password` failed with `"invalid type"`. The `bicepSettings.authenticationMethod` is now threaded through and used to select the auth client (BasicAuth, AzureWI, AwsIrsa). The `Applications.Core/secretStores` path leaves the method empty and continues to use the secret store's own type.

## Type of change

- This pull request fixes a bug in Radius and has an approved issue (#12122).
- This pull request adds or changes features of Radius and has an approved issue (#12122).

## Contributor checklist

- [x] Existing and new functional tests, unit tests, integration tests, and end-to-end tests pass locally where runnable (functional test requires a cluster — runs in `dynamicrp-noncloud` CI).
- [x] Code generated from TypeSpec is regenerated (`make generate`).
- [ ] Design documents updated (the spec lives in `radius-project/design-notes`; this PR aligns the implementation to it).

## Tests

- Unit tests for the type-dispatching loader and the `Radius.Security/secrets` backing-secret reader.
- `Test_TerraformSettings_SecuritySecret_Credentials` — a functional test that provisions a `Radius.Security/secrets` resource and references it from a `terraformSettings` resource for private registry credentials, exercising the new loader end-to-end.

Fixes #12122


test/functional-portable/dynamicrp/noncloud/resources/testdata/tfbicep-combined-test.bicep
```diff
@@ -54,7 +54,7 @@ resource registrySecret 'Applications.Core/secretStores@2023-10-01-preview' = {
   }
 }
 
-resource bicepConfig 'Radius.Core/bicepConfigs@2025-08-01-preview' = {
+resource bicepSettings 'Radius.Core/bicepSettings@2025-08-01-preview' = {
   name: 'tfbicep-combined-bicepconfig'
```
brooke-hamilton: This `bicepSettings` resource is named `tfbicep-combined-bicepconfig`, but the combined functional test expects `tfbicep-combined-bicepsettings` (`terraformsettings_bicepsettings_test.go:283`). `RPResourceSet` validation matches deployed resources by name, so this mismatch will fail the combined test in CI. Rename to `tfbicep-combined-bicepsettings`.

(Also flagged by the Copilot reviewer on this same line.)

pkg/recipes/configloader/security_secrets_test.go
```diff
@@ -0,0 +1,179 @@
...
+	_, err := loader.LoadSecrets(context.Background(), map[string][]string{
+		"/planes/radius/local/resourceGroups/rg/providers/Radius.Security/secrets/my-secret": nil,
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "kubernetes client is not configured")
+}
+
+// Test_readBackingSecret verifies that, given a resolved namespace/name, secret values are read from the
+// backing Kubernetes Secret and filtered by the requested keys.
+func Test_readBackingSecret(t *testing.T) {
```
brooke-hamilton, line 158: `Test_readBackingSecret` is named as if it covers the backing-secret reader, but it never calls `loadSecuritySecret` — it only exercises the fake clientset and `ParseResource` directly. The actual happy path of `loadSecuritySecret` (kind extraction, key filtering, and the missing-key error) has no unit coverage. Consider driving the test through `loadSecuritySecret` with a `secretsLoader` whose `KubernetesProvider` is backed by the fake clientset, so the production read path is exercised.

brooke-hamilton (changes requested):
🚀

DariuszPorowski (changes requested):
Does it mean, we need to update refs, docs, samples, etc in other repos due to Config - > Settings change? Looks like it's a breaking change that needs to be coordinated with TPMs.

nicolejms:
@sylvainsf can you improve code coverage numbers. they're pretty low -  Patch coverage is 43.45238%

sylvainsf:
> Does it mean, we need to update refs, docs, samples, etc in other repos due to Config - > Settings change? Looks like it's a breaking change that needs to be coordinated with TPMs.

No, this is for extensibility namespace, there are no docs for that currently at all. That's on the PM task list.

sylvainsf:
Superseded by #12303, which moves this branch off the personal fork to the upstream `sylvainsf/fix-12122-security-secrets-recipe-configs` branch (same commits, rebased on `main`, plus the review-comment fixes). Closing in favor of #12303.

PR #12078, opened 2026-06-09
Add Repo Radius feature specification

zachcasper:
This pull request introduces a comprehensive feature specification document for "Repo Radius," detailing a new approach to delivering Radius as a lightweight, on-demand engine that runs within GitHub Actions, removing the need for a pre-configured Kubernetes cluster. The document outlines goals, user personas, scenarios, dependencies, risks, and required investments for implementing Repo Radius, making Radius more accessible to developers without platform engineering resources.

**Repo Radius Feature Specification:**

*Overview and Motivation:*
- Proposes transforming Radius from a persistent control plane requiring Kubernetes expertise into an ephemeral, on-demand service running inside GitHub Actions, with state stored in GitHub-native storage. This aims to lower the barrier to entry for developers and enable a "try-before-you-buy" experience.

*Key Scenarios and User Experience:*
- Describes end-to-end developer workflows, including deploying applications, setting up environments, authorizing cloud access via OIDC, resuming deployments, and migrating to a self-hosted Radius.
- Details how the UX component interacts with Repo Radius through GitHub APIs, including environment setup, OIDC configuration, workflow dispatch, execution, and output handling.

*Technical Implementation and Dependencies:*
- Outlines technical requirements such as supporting deployment to external Kubernetes clusters, externalizing the Radius data store, integrating with cloud provider credentials via OIDC, and supporting a customizable Terraform backend. (F232f160# Description

## Type of change

- This pull request is a design document and only includes files in the eng/design-notes directory.

## Contributor checklist
Please verify that the PR meets the following requirements, where applicable:

- An overview of proposed schema changes is included in a linked GitHub issue.
    - [ ] Yes
    - [X] Not applicable
- A design document is added or updated under `eng/design-notes/` in this repository, if new APIs are being introduced.
    - [X] Yes
    - [ ] Not applicable
- The design document has been reviewed and approved by Radius maintainers/approvers.
    - [ ] Yes
    - [X] Not applicable
- A PR for [resource-types-contrib](https://github.com/radius-project/resource-types-contrib/) is created, if resource types or recipes are affected by the changes in this PR.
    - [ ] Yes
    - [X] Not applicable
- A PR for [dashboard](https://github.com/radius-project/dashboard/) is created, if the Radius Dashboard is affected by the changes in this PR.
    - [ ] Yes
    - [X] Not applicable
- A PR for the [documentation repository](https://github.com/radius-project/docs) is created, if the changes in this PR affect the documentation or any user facing updates are made.
    - [ ] Yes
[...]

Issue #8398, opened 2025-02-03
FEATURE: Postgres DB initialization


superbeeny:
### Overview of feature request

As part of the Postgres integration we need to initialize the database with users and schemas. 

We considered 3 options:
1. Add the db initialization and maintain a postgres docker image.
2. Initialize the database on server start.
3. Initialize the database using a configmap on helm install.

We discounted option 1 as we did not want to have to maintain our own docker images for postgres.
We discounted option 2 as we would have to handle race conditions as the server start between ucp, applications-rp, dynamic-dp is not orchestrations.

This leaves us with option 3 as the preferred option.
Also, easy to roll back and the database is ready before the servers start.

### Acceptance criteria

1. The database is initialized with users and tables on install and before the servers start.

### Additional context

Parent Issue: #8096 

### Would you like to support us?

- [x] Yes, I would like to support you

ytimocin:
Will the migration file be kept as a Secret or a ConfigMap?

ytimocin:
Will the migration script be stored as a **Secret**, or will it be in a **ConfigMap** where it's visible in plain text?

How will we ensure that the database initialization runs only once and does not get re-applied unintentionally during subsequent deployments?

What do we mean exactly by this: `Initialize the database using a configmap on helm install.`? Does it mean during `rad install kubernetes`?

nicolejms:
Current implementation and remaining work (reviewed against main `ec979015e67de6222b2b83c462de1ae70790c3e4`).

### Implemented
- The ConfigMap-based PostgreSQL initialization creates the provider databases, users and tables and is mounted into PostgreSQL's initialization directory.
- Chart tests cover the initialization resources.

### Remaining gaps
- Confirm the explicit acceptance condition that initialization finishes before Radius servers start serving successfully. PostgreSQL's entrypoint orders its own initialization, but independent provider deployments have no explicit database-ready startup gate.
- Clarify whether usable-service readiness is the intended ordering guarantee; this is an acceptance question, not missing database initialization code.

[...]

Issue #8096, opened 2024-11-26
External data store for Radius


zachcasper:
### Overview of feature request

Today, Radius stores its configuration and resource data in etc within the host Kubernetes cluster. This presents challenges for users backing up their Radius resource data and coordinating with Kubernetes cluster updates. This feature is to implement options for advanced Radius deployments which use an external data store provided by the user. The first supported external data store should be PostgreSQL.

### Acceptance criteria

- Getting started users can install Radius without needing an external database
- Advanced users can install Radius and provide an existing PostgreSQL database

### Would you like to support us?

- [ ] Yes, I would like to support you

[AB#13755](https://dev.azure.com/azure-octo/e61041b4-555f-47ae-95b2-4f8ab480ea57/_workitems/edit/13755)

superbeeny:
Adding some more context based on conversations in the last design meeting:

### Some of the why we need to change the current implementation of the radius datastore?

1. It is hard to debug.
2. It is hard to migrate.
3. Over time performance degrades to a point the cluster becomes unusable.

### Some features to maintain

1. No more steps required from the user during `rad init`
2. Maintain the simplicity of the "Getting Started" and other tutorials.
3. We don't want to maintain a lot of stuff or be responsible for items outside our expertise. This includes operations on how to maintain and configure Postgres.
4. Production should support "bring your own database", BYOD.

### Workback Plan:
**Goal**: Radius uses the database as a datastore.

1. Complete implementation of the helm chart install.
2. Implement seeding script to initialize the schema.
3. Generate username and password and store in secrets.
4. Update the developer documentation.
5. Update the User documentation.
6. Update the production deployment documentation with instructions to bring your own.
7. Switch Radius over to using the db as the datastore.
8. Profit!

@rynowak and @zachcasper thoughts?

superbeeny:
@kachawla and @zachcasper is there a preference on how to initialize the db? can do either through adding to the docker image or possibly `shelling out` from `rad init`. Thoughts?

kachawla:
> [@kachawla](https://github.com/kachawla) and [@zachcasper](https://github.com/zachcasper) is there a preference on how to initialize the db? can do either through adding to the docker image or possibly `shelling out` from `rad init`. Thoughts?

Sorry missed this earlier. Synced up offline on this.

nicolejms:
Current implementation and remaining work (reviewed against main `ec979015e67de6222b2b83c462de1ae70790c3e4`).

### Implemented
[...]

PR #12213, opened 2026-06-23
Repo Radius: externalize control-plane and Terraform state (rad startup / rad shutdown)

sylvainsf:
# Repo Radius: externalize control-plane and Terraform state (`rad startup` / `rad shutdown`)

## Description

Implements **Investment 2 of the Repo Radius feature spec — externalization of the Radius data
store**, scoped to the state-storage aspects only (control-plane PostgreSQL state and Terraform
recipe state). It adds two kind-agnostic commands, `rad startup` and `rad shutdown`, that back up
and restore all durable Radius state across an ephemeral control plane, plus the Helm and
Terraform-state plumbing they depend on.

Design note: `eng/design-notes/2026-06-repo-radius-state-storage.md`.

There is **no dedicated workspace kind**. The commands operate on the current workspace's
Kubernetes context like any other command and do not create or delete clusters or install Radius —
cluster lifecycle is the caller's responsibility.

## What's included

The change is organized as three logical commits:

1. **PostgreSQL enablement (closes the chart gaps for `database.enabled=true`)**
   - UCP, Applications RP, and Dynamic RP configmaps/deployments now switch to the `postgresql`
     provider and inject `POSTGRES_PASSWORD` when `database.enabled=true` (previously hardcoded to
     `apiserver`).
   - Adds an init-db ConfigMap (option 3 from #8398) that creates the per-RP databases, users, and
     tables before the servers start.
   - Fixes the `POSTGRES_DB` secret value and pins the postgres image to `16-alpine`.
   - Fixes the `databaseProvider` URL env-var substitution in `factory.go`.
   - Adds helm-unittest coverage and a unit test for the env-var helper.

2. **Terraform recipe state backup/restore (`pkg/cli/tfstate`)**
   - Terraform recipes store state in Kubernetes Secrets (`tfstate=true`), not in PostgreSQL, so
     that state is lost on teardown. This package exports and restores those Secrets, including the
     chunked `tfstate-{workspace}-{suffix}-{index}` Secrets for large state.

3. **`rad startup` / `rad shutdown` + end-to-end lifecycle test**
   - `pkg/cli/pgbackup`: control-plane PostgreSQL backup/restore via `kubectl exec pg_dump/psql`.
   - `pkg/cli/gitstate`: persists the state directory to a `radius-state` git orphan branch in an
     isolated worktree; the backup push fails loudly when a remote is configured (a failed push
     would otherwise be silent data loss) and tolerates the no-remote local/test case.
   - `rad shutdown` backs up both stores then commits and pushes; `rad startup` waits for the
     database then restores both stores.

## Testing

Unit tests cover the chart rendering, the Terraform-state round-trip (fake Kubernetes client), the
git orphan-branch worktree behaviour (real temporary repos), and the command runners
(hand-written fakes). All pass, along with the existing 85 Helm chart tests.

### End-to-end test dependency (please read)

The end-to-end lifecycle test lives at `test/functional-portable/statestore` and exercises the
full path that this work exists to protect: install → deploy a Terraform-backed resource →
`rad shutdown` → tear down → reinstall → `rad startup` → **deploy an update to the same resource**.
The update is the path that fails when Terraform state is lost.

**This test depends on the separate Repo Radius workflow code (in flight) that creates the
ephemeral cluster, installs Radius, and runs the deploy.** Because `rad startup` / `rad shutdown`
are intentionally kind-agnostic and do not manage cluster lifecycle, the test needs that workflow
to stand up the environment. Until it lands, the test drives the install/uninstall itself and is
**gated behind the `RADIUS_STATE_E2E` environment variable**, so it does not run in the normal
functional suite. Once the shared cluster-create + deploy workflow is merged, the test's
install/uninstall helpers should be re-pointed at that code rather than duplicating it.

## Related

- Feature spec: Repo Radius (#12078)
- Relates to #8096 (External data store for Radius), #8398 (Postgres DB initialization)
- Supersedes the unmerged prototype #11457

## Type of change

This pull request adds new features (state externalization commands) for Radius.

### DIFF
--- a/cmd/rad/cmd/root.go
+++ b/cmd/rad/cmd/root.go
@@ -80,6 +80,8 @@ import (
 	"github.com/radius-project/radius/pkg/cli/cmd/rollback"
 	rollback_kubernetes "github.com/radius-project/radius/pkg/cli/cmd/rollback/kubernetes"
 	"github.com/radius-project/radius/pkg/cli/cmd/run"
+	cmd_shutdown "github.com/radius-project/radius/pkg/cli/cmd/shutdown"
+	cmd_startup "github.com/radius-project/radius/pkg/cli/cmd/startup"
 	"github.com/radius-project/radius/pkg/cli/cmd/uninstall"
 	uninstall_kubernetes "github.com/radius-project/radius/pkg/cli/cmd/uninstall/kubernetes"
 	"github.com/radius-project/radius/pkg/cli/cmd/upgrade"
@@ -344,6 +346,12 @@ func initSubCommands() {
 
 	initCmd, _ := radinit.NewCommand(framework)
 	RootCmd.AddCommand(initCmd)
+
+	startupCmd, _ := cmd_startup.NewCommand(framework)
+	RootCmd.AddCommand(startupCmd)
+
+	shutdownCmd, _ := cmd_shutdown.NewCommand(framework)
+	RootCmd.AddCommand(shutdownCmd)
 
 	envCreateCmd, _ := env_create.NewCommand(framework)
 	previewCreateCmd, _ := env_create_preview.NewCommand(framework)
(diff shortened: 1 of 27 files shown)


sylvainsf:
Closing to re-open from an upstream branch so workflow changes run on the PR.

Issue #12244, opened 2026-06-24
Direct-module recipes: securely flow developer secret values into module parameters


willdavsmith:
## Overview of feature request

The direct-module recipe feature ([radius-project/radius#11876](https://github.com/radius-project/radius/issues/11876)) lets a platform engineer point a Recipe at an unmodified IaC module (Azure Verified Module, Terraform module) and map a developer resource onto its parameters with `{{context.*}}` expressions. This works for non-secret properties, but **there is no way to flow developer-authored secret material into a module parameter.**

Cloud provisioning modules require the **literal secret value** at deploy time:

- AVM `avm/res/db-for-postgre-sql/flexible-server` → `administratorLoginPassword` (`@secure() string`)
- `terraform-aws-modules/rds/aws` → `password` (sensitive variable)

But `context.*` exposes only **non-secret** data (`context.resource.properties.*`, `context.resource.connections.<name>.properties.*`). A recipe author has no expression that resolves to the value of a `Radius.Security/secrets` resource the developer referenced. So today the recipe cannot satisfy the module's secret input.

### Concrete gap

`Radius.Data/postgreSqlDatabases` lets a developer wire credentials via a secret:

```bicep
// app.bicep (developer)
resource dbsecret 'Radius.Security/secrets@...' = {
  name: 'db-credentials'
  properties: { data: { USERNAME: {...}, PASSWORD: {...} } }
}
resource pg 'Radius.Data/postgreSqlDatabases@...' = {
  properties: {
    environment: env
    size: 'S'
    database: 'appdb'
    secretName: dbsecret.name   // <-- developer's intended credential
  }
}
```

The recipe author wants to write, in `platform.bicep`:

```bicep
parameters: {
  administratorLogin:         '{{context.resource.connections.credentials.secrets.username}}'
  administratorLoginPassword: '{{context.resource.connections.credentials.secrets.password}}'
}
```

There is no such path. In the `resource-types-verification` test repo we work around this by **fabricating** the credential: Azure generates a password in the workflow and passes the same literal to both the module and the decorative secret; AWS sets `manage_master_user_password = true` so nothing flows. The developer's `secretName` is effectively ignored — the secret↔module contract is faked, which means the direct-module feature's secret story is unverified.

### Why this is non-trivial (and a security concern)

[...]

PR #12249, opened 2026-06-25
Securely flow developer secrets into direct-module recipe parameters

willdavsmith:
# Description

Implements **Phase 1** of #12244: securely flow developer-authored secrets into the parameters of direct-module recipes. Secret material is loaded for secret-typed connections, resolved into recipe parameters only as whole values, tagged as secure, and routed through each driver's native secret-handling mechanism.

## Changes by layer

- **Types** (`pkg/recipes/types.go`): adds a tainted `Secrets map[string]string` (`json:"-"`) lane to `recipes.ConnectedResource`, so secret material is never serialized alongside regular connection data.
- **Engine** (`pkg/recipes/engine`): `enrichConnectionSecrets` loads secret material for secret-typed connections (`Applications.Core/secretStores`, `Radius.Security/secrets`) using the engine's existing `SecretsLoader`. Loading is best-effort and non-fatal.
- **Param resolver** (`pkg/recipes/paramresolver`): `ResolveParameterExpressions` now returns `ResolvedParameters{Values, SecureKeys}, error`. It resolves `context.resource.connections.<name>.secrets.<key>` from a separate secret lookup, enforces whole-value-only usage (a secret interpolated into a surrounding string is rejected with an error), and records secret-sourced parameters in `SecureKeys`.
- **Bicep driver** (`pkg/recipes/driver/bicep`): passes the resolved values to ARM, relying on `@secure()` scrubbing; Radius never logs the parameter map.
- **Terraform driver** (`pkg/recipes/terraform`): routes `SecureKeys` through a `sensitive = true` Terraform variable referenced as `${var.<name>}`.

This change is rebased onto main's recursive nested-ternary implementation (#12239) and preserves all of its ternary helpers.

## Testing

- **Resolver secret cases** (`Test_SecretExpressions`): whole-value resolution + secure tagging, surrounding-whitespace handling, nested-object tagging, non-secret params untagged, interpolation-into-a-string rejection, and missing-key passthrough.
- **Engine enrichment cases** (`Test_enrichConnectionSecrets`): secrets populated for secret-typed connections only, non-fatal load error, and nil-loader no-op.
- **Terraform `Test_SetModuleParams`**: secure params routed through sensitive variables while others are set inline, plus an unknown-module no-op.
- Existing recursive nested-ternary resolver tests (`Test_TernaryExpressions`) continue to pass.

Verified locally: `gofmt -l` (clean), `go build ./...`, `go vet ./pkg/recipes/...`, and `go test ./pkg/recipes/...` all pass.

## Out of scope / hard to test

- End-to-end secret flow into a live AVM `@secure()` / Terraform `sensitive` parameter needs a real secret store plus a cloud deploy; existing direct-module cloud tests are already manual/deferred.
- ARM `@secure()` and Terraform `sensitive` redaction are provider behaviors and aren't assertable in unit tests.
- Cleartext in the temporary `main.tf.json` working directory is inherent to Phase 1; Phase 2 (reference passthrough) avoids it.

Addresses #12244 (Phase 2 reference-passthrough remains open, so this does not close the issue).

---

> **Note:** A second commit explored the `secretName` / `Radius.Security/secrets` reference path (`context.resource.secrets.<key>`) but has been reverted in this PR — the #12244 owner parked that path pending a scope decision (Phase 1 vs deferred Phase 2 reference passthrough). The reverted work remains recoverable in this branch's history. This PR's effective change is the connection-based Phase 1 deliverable described above.


willdavsmith:
Superseded by #12286, which consolidates this connection-based secrets work, the #12254 secret-binding references, and the new retain-encrypted-at-rest changes into a single draft PR against `main`. The branch history is preserved. Closing in favor of #12286.

PR #12254, opened 2026-06-25
Resolve secretName / Radius.Security/secrets references into direct-module recipe parameters

willdavsmith:
> **Part 2 of 2** · stacked on **#12249** (Part 1: connection-based secrets) · this is the **`secretName` → `Radius.Security/secrets`** half ("Option A": read the deployed Kubernetes Secret via `status.outputResources`). **Review/merge #12249 first** — this PR's diff is shown against its branch (Part 2 only, 16 files, +942/−6) and will be retargeted to `main` once Part 1 lands.

# Description

This PR adds the second developer-facing path for handing a secret to a direct-module recipe: a resource **property that references a `Radius.Security/secrets` resource by name** (marked `x-radius-secret-reference`, e.g. `secretName`), resolved into recipe parameters as `context.resource.secrets.<key>`. It builds on the connection-based path from #12249.

## Why this is Phase 1 (not Phase 2)

Phase 2 = **external** secret stores (Azure Key Vault / AWS Secrets Manager / Vault) where Radius holds only a reference. This PR reads Radius's **own Kubernetes materialization** of a developer-supplied secret — the same consumption model `Applications.Core/secretStores` already uses in production — so it is in-scope Phase 1. Scoped to Kubernetes-backed `Radius.Security/secrets`; externally-backed secrets remain Phase 2.

## How it works

Because a secret's sensitive data is redacted from the database before recipe execution, the cleartext is read on demand from the secret's **deployed Kubernetes Secret** via `status.outputResources` (never the redacted `Properties.Data`).

- **Schema** (`pkg/schema`): adds the `x-radius-secret-reference` annotation and `ExtractSecretReferenceFieldPaths`, which returns the property paths that name a secret.
- **Controller** (`pkg/portableresources/backend/controller`): resolves each referenced property value (a secret name) into a sibling `Radius.Security/secrets` resource ID scoped to the consuming resource's resource group, threaded onto `ResourceMetadata.SecretReferences`. A malformed reference is an error (fail-closed).
- **Engine** (`pkg/recipes/engine`): `enrichSecretReferences` loads the referenced secret's material into the recipe's tainted `Secrets`. **Fail-closed**: a referenced secret that cannot be loaded fails the deployment rather than passing a literal `{{…}}` placeholder to the module.
- **Loader** (`pkg/dynamicrp/secretsloader`, new): a type-aware dispatcher. `Radius.Security/secrets` is read from its backing Kubernetes Secret (located via the resource's `core/Secret` output resource); all other secret store types delegate to the existing `Applications.Core/secretStores` loader. Cleartext is reachable only by the in-process engine — never re-persisted to the database and never exposed via a GET.
- **Param resolver / drivers**: exposes `context.resource.secrets.<key>` with the same whole-value-only + secure-tagging rules as the connection path (ARM `@secure()` / Terraform `sensitive = true`).

## Testing

- **Loader** (`Test_DispatchingLoader_*`): reads cleartext from the backing K8s Secret, honors a requested-key filter, routes UDT vs store types and merges results, and fails closed when there is no output resource / the Secret is absent / a requested key is missing / the resource can't be read.
- **Engine** (`Test_enrichSecretReferences`): populates `recipe.Secrets`; fails closed on loader error / missing data / unconfigured loader; empty references are a no-op.
- **Schema** (`TestExtractSecretReferenceFieldPaths`), **controller** (`Test_buildSecretReferences`, `Test_stringValueAtPath`), and **resolver** (`Test_SecretExpressions`, extended for `resource.secrets.<key>` resolution, secure-tagging, and whole-value-only enforcement).
- `gofmt -l` (clean), `go build ./...`, `go vet`, and `go test` are green across all affected packages; existing #12249 tests are unaffected.

## Out of scope

- External (Key Vault / Secrets Manager / Vault) secret references — Phase 2.
- Docs — planned follow-up.

Addresses #12244 (Phase 2 reference-passthrough remains open, so this does not close the issue).

### DIFF
--- a/pkg/dynamicrp/secretsloader/secretsloader_test.go
+++ b/pkg/dynamicrp/secretsloader/secretsloader_test.go
@@ -1,0 +1,257 @@
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package secretsloader
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	v1 "github.com/radius-project/radius/pkg/armrpc/api/v1"
+	"github.com/radius-project/radius/pkg/armrpc/rpctest"
+	"github.com/radius-project/radius/pkg/components/database"
+	"github.com/radius-project/radius/pkg/components/kubernetesclient/kubernetesclientprovider"
+	"github.com/radius-project/radius/pkg/dynamicrp/datamodel"
+	"github.com/radius-project/radius/pkg/recipes"
+	"github.com/radius-project/radius/pkg/recipes/configloader"
+	resources_kubernetes "github.com/radius-project/radius/pkg/ucp/resources/kubernetes"
+)
+
+const (
+	testSecretID    = "/planes/radius/local/resourceGroups/test-rg/providers/Radius.Security/secrets/db-secret"
+	testStoreID     = "/planes/radius/local/resourceGroups/test-rg/providers/Applications.Core/secretStores/store"
+	testNamespace   = "test-namespace"
+	testK8sSecret   = "db-secret-k8s"
+	testInvalidID   = "not-a-valid-id"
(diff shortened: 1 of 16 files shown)


pkg/dynamicrp/secretsloader/secretsloader_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/radius-project/radius/pkg/components/kubernetesclient/kubernetesclientprovider"
+	"github.com/radius-project/radius/pkg/dynamicrp/datamodel"
+	"github.com/radius-project/radius/pkg/recipes"
+	"github.com/radius-project/radius/pkg/recipes/configloader"
+	resources_kubernetes "github.com/radius-project/radius/pkg/ucp/resources/kubernetes"
+)
+
+const (
+	testSecretID    = "/planes/radius/local/resourceGroups/test-rg/providers/Radius.Security/secrets/db-secret"
+	testStoreID     = "/planes/radius/local/resourceGroups/test-rg/providers/Applications.Core/secretStores/store"
```
lakshmimsft, line 42: checking: why are we referring to Applications.Core/secretsStores in this pr. This type is planned to be deprecated in the future.

willdavsmith:
Superseded by #12286. The consolidated PR also replaces this PR's "read the deployed Kubernetes Secret" approach with retaining the encrypted value at rest and decrypting from Radius's own store (multi-cluster-safe; the Kubernetes fallback was removed). Closing in favor of #12286.

PR #12344, opened 2026-07-08
feat(dynamic-rp): secure handling of recipe secret outputs

### LINKED ISSUE #12347: Secure handling of recipe secret outputs

Radius should handle recipe-generated **secret outputs** securely. Today, when a recipe (Bicep or Terraform, including AVM modules) returns a secret value (connection string, access/account key, API key, etc.), that value is persisted as plaintext on the owning resource's `properties`/`status` and surfaced in the application graph and logs.

This proposes materializing **declared** secret outputs into a managed `Radius.Security/secrets` resource and exposing only a read-only `secret { name, id }` reference on the owner — never plaintext. Consumers bind the secret by name via a container `valueFrom.secretKeyRef`.

Scope is **Scenario 3** (module-generated secret output) of the recipe-secrets design. Secret **inputs** (Scenario 1) are already handled via `x-radius-sensitive` and are out of scope. Scenario 2 (re-fetchable output) is a follow-up.

### Acceptance criteria

- Resource types can declare secret outputs via a top-level read-only `secrets:` block; `secret`/`secrets` are treated as framework-owned properties.
- Declared recipe secret outputs are materialized into a managed `Radius.Security/secrets` named `<owner>-secrets`, created/updated/deleted with the owner.
- The owner exposes only `properties.secret = { name, id }` — no secret plaintext in resource properties, status, the application graph, or logs.
- A `secretOutputs` recipe-pack mapping forces named module outputs (e.g. AVM `primaryConnectionString`) into the secret set regardless of module classification.
- Consumers can bind the managed secret via a container `valueFrom.secretKeyRef` — both whole-value and embedded `$(VAR)` composition.
- No-leak is enforced by tests (`rad resource show`, `rad app graph`).
- Proven end-to-end against a real cloud provider.

### Additional context

- Implementation PR: radius-project/radius#12344.
- Enabling recipe fix: radius-project/resource-types-contrib#224 (container env ordering) — merged.
- End-to-end validation across 7 resource types (kafka, redis, mongodb, rabbitmq, storage, search, llm): radius-project/resource-types-verification#32–#38.
- Follow-ups: registration-time `secrets`-block schema validation + `readOnly` input enforcement; Scenario 2 (re-fetchable output); converting the shipped built-in types in `resource-types-contrib` to declare `secrets:` blocks.

willdavsmith:
# Description

Implements secure handling of **recipe secret _outputs_** (Scenario 3 of the recipe-secrets design). A recipe-generated secret (a connection string, storage/access key, API key, etc.) is materialized into a managed `Radius.Security/secrets` resource and is **never persisted as plaintext** on the owning resource, in the application graph, or in logs. The owner exposes only a read-only `secrets.name` reference; consumers bind by name via a container env `valueFrom.secretKeyRef`.

**What's included**

- **Schema `secrets:` block** (`pkg/schema/secrets.go`): resource types declare a top-level read-only `secrets` object with a reserved `name` reference; `GetSecretsBlock` / `ValidateSecretsBlock` helpers. `secrets` is treated as a framework-owned basic property.
- **Managed-secret materializer** (`pkg/dynamicrp/backend/secret`): creates / updates / deletes a `Radius.Security/secrets` per owner via the generic resources client. The managed secret name is derived deterministically as `<owner-name>-<short-hash>-secrets`, where a short hash of the owner's fully-qualified resource ID disambiguates owners that share a name but differ in resource type or scope.
- **Processor leak-fix** (`pkg/dynamicrp/backend/processor/dynamicresource.go`): once a type declares a `secrets` block, materializes **all** of the recipe's secret outputs into the managed secret (the declared keys are advisory and do not filter), sets the read-only `properties.secrets.name` reference, and **stops writing secret plaintext** into resource properties/status; cascade-deletes the managed secret on delete.
- **Nested `outputs.secrets` recipe-pack mapping** for direct (AVM) modules — an entry under the reserved `secrets` key inside a recipe's `outputs` (e.g. `outputs: { host: 'name', secrets: { connectionString: 'primaryConnectionString' } }`) forces the named module output into the secret set regardless of how the module classified it. The API models `outputs` as an untyped map; the versioned↔datamodel converter splits the nested `secrets` object back out (typespec + swagger + generated models + datamodel + converter + `recipes` configloader).
- **No-leak verification**: integration test asserts a recipe secret output never appears in the `rad resource show` GET/LIST body; functional test (`sensitive_fields_test.go`) asserts secret values never appear in the `rad resource show` response.
- **Docs**: secret-output section in `docs/architecture/extensibility.md`.

Secret _inputs_ (Scenario 1) are unchanged — already handled by `x-radius-sensitive`.

## End-to-end validation

Proven end-to-end against **real Azure** across 7 resource types in `resource-types-verification`. Each run confirms the recipe secret output is materialized into a managed `Radius.Security/secrets` (`state=Succeeded`), the owner exposes only the read-only `properties.secrets.name` reference, **no plaintext appears on the owner**, and a container consumes the secret via `secretKeyRef`:

| Type | PR | Base provisioning | Consumer round-trip |
|---|---|---|---|
| kafka | #32 | ✅ | kcat whole-value + quarkus `$(VAR)` JAAS (explicit in-pod substitution assert) → Event Hubs |
| redis | #37 | ✅ | live POST/GET → Azure Managed Redis |
| mongodb | #36 | ✅ | mongo-express → Cosmos DB |
| rabbitmq | #38 | ✅ | bento AMQP send/receive → Service Bus |
| storage | #35 | ✅ | sftpgo → Blob (multi-key secret; selects `accountKey`) |
| search | #34 | ✅ | hurl → Azure AI Search |
| llm/models | #33 | ✅ | LiteLLM → Azure OpenAI chat |

## Type of change

Fixes: #12347

## Related PRs

- `radius-project/resource-types-contrib` #224 (merged) — container recipe env ordering fix (emit secret/`valueFrom` env vars before `value` vars), which enables the developer-side `$(VAR)` secret-composition pattern this feature relies on.
- `radius-project/resource-types-verification` #32 (kafka reference) and #33–#38 (llm, search, storage, mongodb, redis, rabbitmq) — the 7-type end-to-end validation above; each pins the `radius` submodule to this branch.

## Not yet included (follow-ups)

- Registration-time `secrets`-block schema validation and `readOnly` input enforcement.
- Durable managed-secret provisioning status (requeue/verify the managed secret exists before marking the owner Succeeded) — see review discussion.
- Scenario 2 (re-fetchable secret output: store an identifier and re-resolve the live value from the cloud provider at bind time).
- Converting the shipped built-in types in `resource-types-contrib` to declare `secrets:` blocks (the verification repo converted its test manifests to prove the flow).

## Contributor checklist

    - [x] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [ ] Yes
    - [ ] Not applicable
    - [x] Yes 
    - [ ] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable


.github/workflows/secrets-handling-tests.yaml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          go-version-file: go.mod
+          cache-dependency-path: go.sum
+          cache: true
+
+      - name: Run secret-handling tests
+        # `make test-secrets` runs go vet and the unit tests for the secret-handling packages. It
+        # deliberately avoids cluster tooling (KUBEBUILDER_ASSETS / helm) so it stays fast and is
+        # runnable locally with the same command.
+        run: make test-secrets
```
willdavsmith: let's remove this and the make command

build/test.mk
```diff
@@ -50,6 +50,24 @@ GOTEST_TOOL ?= go tool gotestsum $(GOTESTSUM_OPTS) --
...
+# secretOutputs mappings, and the dynamic-rp processor/controllers that wire them together. Keep this
+# list in sync when the secret-handling code moves. These packages do not need cluster tooling
+# (KUBEBUILDER_ASSETS / helm), so `test-secrets` stays fast for focused local and CI feedback.
+SECRETS_PACKAGES ?= \
+	./pkg/schema/... \
+	./pkg/resourceutil/... \
+	./pkg/recipes/... \
+	./pkg/corerp/api/v20250801preview/... \
+	./pkg/corerp/datamodel/... \
+	./pkg/dynamicrp/...
```
willdavsmith: remove these

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +88,116 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
-	// Add the secret values to the resource properties if they are part of the schema.
-	for key, value := range secretValues {
-		if slices.Contains(resourceProps, key) {
-			resource.Properties[key] = value.Value
+// materializeDeclaredSecrets routes recipe secret outputs whose keys are declared in the schema `secrets`
+// block into a managed Radius.Security/secrets resource, and exposes a read-only reference to that resource
+// on the owner. Secret values are never written onto the owner resource. When the resource type declares no
+// `secrets` block, recipe secret outputs are dropped rather than persisted.
+func (d *DynamicProcessor) materializeDeclaredSecrets(ctx context.Context, resource *datamodel.DynamicResource, schema map[string]any, secretValues map[string]rpv1.SecretValueReference) error {
+	declaredKeys, ok := schemautil.GetSecretsBlock(schema)
```
lakshmimsft: Secrets should be created off the recipe result's  secrets  block, not based on the type schema, which can contain only a subset of the recipe outputs including secrets. Current implementation silently deletes these sensitive values if they're not in the type schema. This is not expected.

pkg/dynamicrp/backend/secret/materializer.go
```diff
@@ -0,0 +1,177 @@
...
+
+// NewMaterializer creates a Materializer that talks to UCP using the provided ARM client options.
+func NewMaterializer(armClientOptions *arm.ClientOptions) Materializer {
+	return &clientMaterializer{armClientOptions: armClientOptions}
+}
+
+// ManagedSecretName derives the deterministic name of the managed Radius.Security/secrets resource for an
+// owner resource of the given name.
+func ManagedSecretName(ownerName string) string {
+	return ownerName + managedSecretNameSuffix
```
lakshmimsft: ownername + suffix may not be unique enough since different resourcetypes in same resource group can have same name. consider hash of id + name or something similar for more uniqueness

pkg/dynamicrp/backend/secret/materializer.go
```diff
@@ -0,0 +1,177 @@
...
+
+	client, err := generated.NewGenericResourcesClient(SecuritySecretsResourceType, ownerID.RootScope(), &aztoken.AnonymousCredential{}, m.clientOptions())
+	if err != nil {
+		return Result{}, err
+	}
+
+	// BeginCreateOrUpdate performs the accepted PUT synchronously (dynamic-rp persists the resource and
+	// queues its own async provisioning). We intentionally do not poll to completion to avoid holding a
+	// worker while the managed secret's own async operation waits for one.
+	if _, err := client.BeginCreateOrUpdate(ctx, secretName, generated.GenericResource{
```
lakshmimsft, line 178: since we're not polling here, there is a potential gap where this is marked Succeeded but the secret provisioning fails behind the scenes. Later when a consumer tries to bind to the secret it will fail/not behave as expected. There may be an approach to set the result status to requeue/provisioningState to provisioning so the call comes back and there is code to check for existence of the secrets before we mark it succeeded..but that may need a discussion

.github/workflows/secrets-handling-tests.yaml
willdavsmith: Done — removed `.github/workflows/secrets-handling-tests.yaml` in 651e1edcc. These packages are covered by the standard unit-test suite (`make test` / unit-tests.yaml).

build/test.mk
willdavsmith: Done — removed the `test-secrets` target and `SECRETS_PACKAGES` in 651e1edcc.

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +88,116 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
 		return err
 	}
 
-	err = resource.ApplyDeploymentOutput(rpv1.DeploymentOutput{DeployedOutputResources: outputResources, ComputedValues: computedValues, SecretValues: secretValues})
-	if err != nil {
+	// Persist output resources and computed values. Secret values are never persisted on the owner
+	// resource: they are materialized into a managed Radius.Security/secrets resource below.
+	if err := resource.ApplyDeploymentOutput(rpv1.DeploymentOutput{DeployedOutputResources: outputResources, ComputedValues: computedValues}); err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
willdavsmith: This is safe here — `go.mod` targets Go 1.26, and since Go 1.22 the `for … range` loop variable is scoped per-iteration, so `&value` is a distinct address each iteration. The processor tests record multiple computed values (host/port/database/username) and assert each is distinct, which passes. No change needed.

test/functional-portable/dynamicrp/noncloud/resources/sensitive_fields_test.go
```diff
@@ -249,4 +250,53 @@ func verifySensitiveFieldRedaction(
 		"K8s Secret should contain the decrypted connectionConfig url")
 	require.Equal(t, expectedConnectionConfigToken, string(k8sSecret.Data["connectionConfigToken"]),
 		"K8s Secret should contain the decrypted connectionConfig token")
+
+	// --- No-leak verification: secret values must not surface through `rad resource show` or the app graph ---
+	verifySecretsNotExposed(ctx, t, ct, appName, resourceTypeName, resourceName,
+		[]string{expectedPassword, expectedAPIKey, expectedCredentialSecret, expectedConnectionConfigToken})
```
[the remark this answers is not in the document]
willdavsmith: Fixed in 651e1edcc — `expectedConnectionConfigURL` is now included in the `verifySecretsNotExposed` no-leak check.

.github/workflows/secrets-handling-tests.yaml
```diff
@@ -0,0 +1,87 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------
+
+# yaml-language-server: $schema=https://www.schemastore.org/github-workflow.json
+---
+name: Secrets Handling Tests
```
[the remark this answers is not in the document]
willdavsmith: Resolved as moot — the `secrets-handling-tests.yaml` workflow is removed in 651e1edcc.

.github/workflows/secrets-handling-tests.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+# Focused, fast-feedback CI for the recipe secret input/output handling feature. The full unit-test
+# suite (unit-tests.yaml -> `make test`) still covers these packages; this workflow runs only when
+# the secret-handling code paths change so contributors get quick, targeted signal. The test logic
+# lives in the `test-secrets` Make target so it runs identically on developer machines and in CI.
+on:
+  workflow_dispatch:
+  push:
```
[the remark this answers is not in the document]
willdavsmith: Resolved as moot — the workflow is removed in 651e1edcc.

build/test.mk
```diff
@@ -50,6 +50,24 @@ GOTEST_TOOL ?= go tool gotestsum $(GOTESTSUM_OPTS) --
...
+
+.PHONY: test-secrets
+test-secrets: ## Runs go vet and unit tests for the recipe secret-handling packages (fast; no cluster tooling required)
+	go vet $(SECRETS_PACKAGES)
+	CGO_ENABLED=1 $(GOTEST_TOOL) $(SECRETS_PACKAGES) $(GOTEST_OPTS)
```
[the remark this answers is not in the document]
willdavsmith: Resolved as moot — the `test-secrets` target is removed in 651e1edcc; these packages run in the standard unit-test suite.

pkg/dynamicrp/backend/secret/materializer.go
willdavsmith: fixed. `ManagedSecretName` now derives the name from the owner's name plus a short hash of its fully-qualified resource ID (<name>-<sha256(id)[:8]>-secrets)

pkg/dynamicrp/backend/processor/dynamicresource.go
willdavsmith: materializeRecipeSecrets now materializes the recipe's full secrets block rather than only the schema-declared keys

pkg/dynamicrp/backend/secret/materializer.go
willdavsmith: that's true. I think for now we can handle this syncronously and if we see issues we can re-visit. what do you think?

docs/architecture/extensibility.md
```diff
@@ -415,15 +415,101 @@ The engine
...
+1. Reads the declared secret output keys from the schema's `secrets` block via
+   [`schema.GetSecretsBlock`](../../pkg/schema/secrets.go), which returns the
+   read-only sub-properties excluding the reserved `name`.
+2. Collects the recipe secret outputs whose keys match those declared keys.
+   Secret outputs with no matching declared key are dropped, never stored.
+3. Materializes a single Radius-managed `Radius.Security/secrets` resource for
+   the owner via
+   [pkg/dynamicrp/backend/secret](../../pkg/dynamicrp/backend/secret). The
+   managed secret is named `<owner>-secrets`, carries the owner's environment
+   and application, and holds the secret key/values as its sensitive `data`.
```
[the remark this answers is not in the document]
willdavsmith: Fixed in 5667fa69. Updated the numbered steps in `docs/architecture/extensibility.md` to match the implementation: the processor materializes all recipe secret outputs once a `secrets` block is declared (declared keys are advisory, not a filter), and the managed secret name is derived via `secret.ManagedSecretName` as `<owner-name>-<short-hash>-secrets`.

docs/architecture/extensibility.md
```diff
@@ -486,6 +572,10 @@ up before the Radius resource record is removed.
 - Dynamic recipe output only becomes user-visible resource properties when
   those properties exist in the registered schema and are not basic Radius
   properties such as `application` or `environment`.
+- Recipe **secret** outputs are never stored on the owning resource. They are
+  materialized into a managed `Radius.Security/secrets` resource, and the owner
+  exposes only the reserved read-only `secrets.name` reference. Secret outputs
+  not declared in the schema's `secrets` block are dropped.
```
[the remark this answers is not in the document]
willdavsmith: Fixed in 5667fa69. Updated this bullet: once a type declares a `secrets` block, all of the recipe's secret outputs are materialized (declared keys don't filter); a type with no `secrets` block has its secret outputs dropped.

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +92,125 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
+	// object so we don't clobber a declared block.
+	if resource.Properties == nil {
+		resource.Properties = map[string]any{}
+	}
+	secrets, ok := resource.Properties[schemautil.SecretsBlockPropertyName].(map[string]any)
+	if !ok {
+		secrets = map[string]any{}
+		resource.Properties[schemautil.SecretsBlockPropertyName] = secrets
+	}
+	secrets[schemautil.SecretNameReferenceKey] = result.Name
```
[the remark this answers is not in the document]
willdavsmith: Good catch — the PR description was stale. The reference is `properties.secrets.name` only; `id` was intentionally dropped since consumers bind by name via `secretKeyRef`. I've updated the PR description (and the architecture doc) to say `secrets.name` throughout rather than `{ name, id }` / `properties.secret`. The code already matches (exposes only the name), so no code change.

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +92,125 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
+// A resource type opts into secret materialization by declaring a `secrets` block in its schema; that block
+// is also what makes `secrets.name` a referenceable read-only property. Once a type opts in, ALL of the
+// recipe's secret outputs are materialized — not just those whose keys are declared in the block. The
+// declared keys document the type's expected secret surface but do not filter what is materialized, so a
+// recipe (for example a direct AVM module) that emits an additional sensitive output does not silently lose
+// it. When the resource type declares no `secrets` block, recipe secret outputs are dropped rather than
+// persisted.
+func (d *DynamicProcessor) materializeRecipeSecrets(ctx context.Context, resource *datamodel.DynamicResource, schema map[string]any, secretValues map[string]rpv1.SecretValueReference) error {
+	if _, ok := schemautil.GetSecretsBlock(schema); !ok {
+		return nil
```
lakshmimsft: If the type schema doesn't expose secrets but the recipe outputs include secrets, they can be dropped silently here. checking..is that expected? ..even if the secrets were deployed they would not be referenced, hence the question.

willdavsmith: that's expected - the type schema has to expose the secret key for it to be available

pkg/dynamicrp/backend/secret/materializer.go
```diff
@@ -0,0 +1,189 @@
...
+	return &clientMaterializer{armClientOptions: armClientOptions}
+}
+
+// ManagedSecretName derives the deterministic name of the managed Radius.Security/secrets resource for an
+// owner resource. It combines the owner's name with a short hash of the owner's fully-qualified resource ID
+// so that owners that share a name but differ in resource type or scope — which is allowed within a single
+// resource group — never collide on the same managed secret. The name is stable for a given owner, so
+// Materialize and Delete derive the same value.
+func ManagedSecretName(ownerID resources.ID) string {
+	sum := sha256.Sum256([]byte(strings.ToLower(ownerID.String())))
```
lakshmimsft, line 113: consider lowercasing (ownerID.Name() ) and length-bound (63) so it's not invalid k8s secret name

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +92,125 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
 	}
 
+	data := make(map[string]string, len(secretValues))
+	for key, ref := range secretValues {
+		data[key] = ref.Value
+	}
+	if len(data) == 0 {
```
lakshmimsft, line 172: I'm wondering if this is a possible but uncommon scenario worth addressing where this return retains the previously-materialized  Radius.Security/secrets  resource. If a resource materialized a managed secret on a prior deploy and a later update returns no secret outputs ( len(data) == 0 ), we return without deleting the old managed secret or clearing the stale  properties.secrets.name  reference — so the secret resource is orphaned and the owner keeps a dangling reference
Suggested fix: before returning, if the owner currently carries a  secrets.name  reference, cascade-delete the managed secret and clear the reference (mirroring the  Delete  path):

pkg/dynamicrp/backend/secret/materializer.go
willdavsmith: fixed, will be under 63 chars now

pkg/dynamicrp/backend/processor/dynamicresource.go
willdavsmith: fixed this path, shouldn't be leaving orphaned secrets now

pkg/dynamicrp/backend/secret/materializer.go
```diff
@@ -0,0 +1,222 @@
...
+// clientOptions returns a shallow copy of the configured ARM client options with the API version overridden
+// to the version Radius.Security/secrets supports, so the shared options are not mutated.
+func (m *clientMaterializer) clientOptions() *arm.ClientOptions {
+	options := &arm.ClientOptions{}
+	if m.armClientOptions != nil {
+		options = &arm.ClientOptions{ClientOptions: m.armClientOptions.ClientOptions}
+	}
+	options.APIVersion = securitySecretsAPIVersion
+	return options
+}
```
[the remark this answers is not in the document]
willdavsmith: Good catch — real bug. Fixed in bcd672ac: `clientOptions()` now copies the full `arm.ClientOptions` value and overrides only `APIVersion`, instead of rebuilding a fresh struct from just the embedded `ClientOptions`. That preserves `DisableRPRegistration` (and other sibling fields like `AuxiliaryTenants`). Added a test asserting `DisableRPRegistration`/`AuxiliaryTenants` survive the per-request copy and that the shared options aren't mutated.

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -36,12 +38,30 @@ var ErrNoSchemaFound = errors.New("no schema found for resource type")
...
+// backs this resource's declared secret outputs, if one was materialized.
 func (d *DynamicProcessor) Delete(ctx context.Context, resource *datamodel.DynamicResource, options processors.Options) error {
-	return nil
+	if d.SecretMaterializer == nil {
+		return nil
+	}
+	// Only resources that materialized a managed secret carry the secret reference; skip the delete call
+	// otherwise to avoid spurious requests.
+	secrets, ok := resource.Properties[schemautil.SecretsBlockPropertyName].(map[string]any)
+	if !ok {
```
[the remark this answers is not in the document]
willdavsmith: Reading from a nil map is safe in Go — it returns the zero value, so `resource.Properties[key]` on a nil map yields `nil`, and the comma-ok assertion `nil.(map[string]any)` returns `(nil, false)`, taking the early `return nil`. Only writes to a nil map panic, and there are none on this path. Verified with a quick repro. So there's no panic risk here; leaving the code as-is (the comma-ok already handles a nil/absent `Properties`).

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +92,154 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
+// presence of `properties.secrets.name` is the signal that a managed secret was previously materialized
+// (mirroring the Delete path); if it is absent there is nothing to clean up.
+func (d *DynamicProcessor) clearStaleManagedSecret(ctx context.Context, resource *datamodel.DynamicResource) error {
+	secrets, ok := resource.Properties[schemautil.SecretsBlockPropertyName].(map[string]any)
+	if !ok {
+		return nil
+	}
+	if _, ok := secrets[schemautil.SecretNameReferenceKey]; !ok {
+		return nil
+	}
```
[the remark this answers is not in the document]
willdavsmith: Same as the Delete path: this helper only reads `Properties` via a comma-ok type assertion, which is safe on a nil map (returns `(nil, false)` → early `return nil`) — it never reaches the `delete(...)` calls when `Properties` is nil. Reads and `delete` on nil maps don't panic in Go (only writes do), so there's no panic risk on a partially-populated resource. Verified with a repro; leaving as-is.

pkg/recipes/terraform/execute.go
```diff
@@ -406,12 +406,15 @@ func (e *executor) generateConfig(ctx context.Context, tf *tfexec.Terraform, opt
...
-		// module output so the values are available in the Terraform state for output mapping.
+	} else if len(options.EnvRecipe.Outputs) > 0 || len(options.EnvRecipe.SecretOutputs) > 0 {
+		// Direct module with an outputs and/or secretOutputs mapping: generate an output block for each
+		// referenced module output so the values are available in the Terraform state for output mapping.
 		if err = tfConfig.AddMappedOutputs(options.EnvRecipe.Name, options.EnvRecipe.Outputs, loadedModule.OutputSensitivity); err != nil {
 			return "", err
 		}
+		if err = tfConfig.AddMappedOutputs(options.EnvRecipe.Name, options.EnvRecipe.SecretOutputs, loadedModule.OutputSensitivity); err != nil {
+			return "", err
+		}
```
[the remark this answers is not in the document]
willdavsmith: Good catch — real leak risk. Fixed in 2dd2b4a0. `AddMappedOutputs` now takes a `forceSensitive` flag, and the `SecretOutputs` call passes `true`, so every generated secret-output block is `sensitive = true` regardless of the module's own declaration. That means a module output the module didn't mark sensitive (e.g. AVM `primaryConnectionString`) is still redacted in Terraform's stdout/stderr (which we stream into logs). The plain `Outputs` call keeps per-output sensitivity (`false`). Added a test asserting a non-sensitive module output is forced sensitive under the secret mapping.

pkg/dynamicrp/backend/processor/dynamicresource.go
```diff
@@ -72,73 +92,164 @@ func (d *DynamicProcessor) Process(ctx context.Context, resource *datamodel.Dyna
...
+	if err := resource.ApplyDeploymentOutput(rpv1.DeploymentOutput{DeployedOutputResources: outputResources, ComputedValues: computedValues}); err != nil {
 		return err
 	}
 
-	err = addOutputValuestoResourceProperties(ctx, options.UcpClient, resource, computedValues, secretValues)
+	// Fetch the resource type schema once and use it for both computed-value filtering and secret routing.
+	schema, err := getResourceSchema(ctx, options.UcpClient, resource)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
willdavsmith: Fixed in 2dd2b4a0. `Process` now calls `schema.ValidateSecretsBlock` right after fetching the schema and fails fast on a malformed block — non-object `secrets`, non-string sub-properties, or a reserved `name` that isn't readOnly — instead of opting a mis-shaped schema into materialization and writing an object into `properties.secrets`. `ValidateSecretsBlock` already existed but wasn't wired into this path; it is now. Added a test covering a `secrets`-as-string schema returning an error with no materialization.

brooke-hamilton (approved):

PR #12359, opened 2026-07-09
feat(tools): add automated tool updater

### LINKED ISSUE #8901: Put tool and framework versions in one place

DariuszPorowski:
# Description

This pull request adds an automated updater for Radius's external CLI tool versions and SHA-256 checksums.

The change:

- Adds `build/tools.yaml` as the canonical tool manifest and commits the generated `build/tools.generated.mk` Make include.
- Adds the `tool-updater` Go command with manifest validation, release-source checks, checksum refreshes, and Terraform version synchronization.
- Adds the scheduled `.github/workflows/update-tools.yaml` workflow, which opens or updates a PR using the Dependabot-manager GitHub App credentials.
- Preserves existing `make install-<tool>` targets and uses a stable `bin/tool-updater` path to avoid Windows temporary-executable elevation blocks.
- Documents the manifest schema and supported values in `internal/tooling/README.md`.

## Type of change

Fixes: #issue_number

## Contributor checklist

    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable
    - [ ] Yes
    - [x] Not applicable

## Validation

- `go test ./internal/tooling ./cmd/tool-updater`
- `go vet ./internal/tooling ./cmd/tool-updater`
- GitHub Actions `actionlint`
- Markdown lint and table formatting


internal/tooling/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Commands
+
+Run the updater through Make:
+
+```sh
+make update-tools
+```
+
+Make builds the updater into `bin/` before executing it. This stable path is important on Windows, where security software may block the temporary executable created by `go run`.
```
brooke-hamilton, line 32: The Commands section is ambiguous about which target does what, and it doesn't mention the `make install-<tool>` targets at all — so a reader can't tell "update the pinned versions" from "install a tool locally." Suggest spelling out the two distinct workflows:

- `make update-tools` — checks each tool's release source and **advances its pinned version in `build/tools.yaml` to the latest available release** (tools marked `update: false` stay pinned), recomputes the SHA-256 checksums, regenerates the committed `build/tools.generated.mk`, and syncs the Terraform `versionFiles`. It reaches the network and edits files; it does **not** install anything onto your machine.
- `make install-<tool>` (e.g. `make install-yq`, `make install-kubectl`) — downloads and installs that tool's pinned version into a user-owned `bin` dir, using the metadata from the generated include. This is the one to run when you actually want the binary.

A one-line note that `make update-tools` also regenerates `build/tools.generated.mk` as a side effect (there's no separate "regenerate" make target — the `generate-make` subcommand shown below is the binary-level equivalent) would remove the last bit of ambiguity.

.github/scripts/update-tools-pr.sh
```diff
@@ -0,0 +1,67 @@
...
+require_environment
+cd "${REPO_ROOT}"
+
+git config user.name "${GIT_USER_NAME}"
+git config user.email "${GIT_USER_EMAIL}"
+git checkout -B "${PR_BRANCH}"
+git add -A
+git commit --signoff -m "${PR_TITLE}"
+gh auth setup-git
+git push --force-with-lease origin "HEAD:${PR_BRANCH}"
```
lakshmimsft: `git push --force-with-lease` will reject updates after the automation branch has been created and therefore fail on every run after the first.
`actions/checkout` only fetches the checked-out base branch here, so the runner has no `origin/automation/update-tools` remote-tracking ref. Without an explicit lease value, Git treats the destination as expected not to exist; once the remote branch does exist, the push is rejected as stale. That means the first automated PR can be created, but subsequent scheduled runs cannot update it.

Please look up the destination SHA first and pass it explicitly.

.github/scripts/update-tools-pr_test.sh
```diff
@@ -0,0 +1,97 @@
...
+git -C "${REPOSITORY}" config user.name "Test User"
+git -C "${REPOSITORY}" config user.email "test@example.com"
+echo "original" >"${REPOSITORY}/tools.yaml"
+git -C "${REPOSITORY}" add tools.yaml
+git -C "${REPOSITORY}" commit -q -m "initial"
+git -C "${REPOSITORY}" remote add origin "${REMOTE}"
+git -C "${REPOSITORY}" push -q -u origin main
+
+mkdir -p "${FAKE_BIN}"
+cat >"${FAKE_BIN}/gh" <<'EOF'
```
lakshmimsft, line 66: The test fixture is a full local clone, so `origin/automation/update-tools` exists and the lease resolves — which is why it misses the bug above. Reproduce the CI topology (`git clone --single-branch --branch main`) so the second `run_script "42"` case exercises the real ref set.

internal/tooling/manifest.go
```diff
@@ -0,0 +1,617 @@
...
+		return false, err
+	}
+	contents, err := updateManifestYAML(path, manifest)
+	if err != nil {
+		return false, fmt.Errorf("update tool manifest: %w", err)
+	}
+	return writeIfChanged(path, contents)
+}
+
+func updateManifestYAML(path string, manifest Manifest) ([]byte, error) {
```
lakshmimsft, line 450: updateManifestYAML correctly mutates only the version and checksum nodes, but it then serializes the entire YAML document with yaml.Marshal, which rewrites unrelated formatting across build/tools.yaml so the first real automated version/checksum update will also produce a noisy whole-file formatting diff.

Could we preserve the existing formatting where practical—at minimum use an encoder with the repository’s two-space indent and retain the leading `---` marker when it was present in the original file?

Please also add a regression test that changes one version/checksum in the committed manifest and verifies the resulting diff is limited to those values rather than reformatting unrelated lines.

.github/scripts/update-tools-pr.sh
DariuszPorowski: good call, done :)

.github/scripts/update-tools-pr_test.sh
DariuszPorowski: done

internal/tooling/manifest.go
DariuszPorowski: done

build/tools.yaml
```diff
@@ -0,0 +1,366 @@
...
+      urlTemplate: https://releases.hashicorp.com/terraform/{version}/terraform_{version}_SHA256SUMS
+      format: standard
+    versionFiles:
+      - path: .terraform-version
+        format: plain
+      - path: pkg/recipes/terraform/version.go
+        format: replace
+        prefix: 'var terraformVersion = "'
+        suffix: '"'
+      - path: deploy/Chart/values.yaml
```
lakshmimsft, line 249: The Terraform  versionFiles  list covers  .terraform-version ,  pkg/recipes/terraform/version.go , and  deploy/Chart/values.yaml , but misses a fourth consumer:  deploy/Chart/tests/terraform_init_test.yaml  asserts  pattern: 'TERRAFORM_VERSION="1\.14\.9"'  at lines 88 and 95.
Those assertions are checked against the rendered chart, so once the updater bumps Terraform and syncs  values.yaml , the rendered value changes and both assertions fail — breaking  test-helm , which  make test  depends on. Terraform has updates enabled and resolves through the HashiCorp checkpoint endpoint on a weekly cron, so this will fire on the first release bump.
The effect is that the automation's first real Terraform update opens a PR that fails its own CI and needs a manual fix, which is the outcome the manifest sync was meant to avoid. Could this file either be added as a synchronized consumer, or the assertions rewritten so they don't hardcode the literal version?

.github/scripts/update-tools-pr_test.sh
```diff
@@ -0,0 +1,116 @@
...
+
+readonly REPOSITORY="${TEST_ROOT}/repository"
+readonly REMOTE="${TEST_ROOT}/remote.git"
+readonly SEED="${TEST_ROOT}/seed"
+readonly FAKE_BIN="${TEST_ROOT}/bin"
+readonly GH_LOG_PATH="${TEST_ROOT}/gh.log"
+
+git init -q --bare "${REMOTE}"
+git init -q -b main "${SEED}"
+git -C "${SEED}" config user.name "Test User"
```
lakshmimsft, line 47: The fixture configures  user.name  and  user.email  but doesn't disable commit signing. For anyone with  commit.gpgsign = true  set globally,  git commit  at line 51 blocks on a gpg passphrase prompt and the test hangs indefinitely rather than failing. I hit this locally; it passes with a clean global config.
 build/test.mk  now makes  test  depend on  test-update-tools-pr , so this silently wedges  make test  with no output — a failure mode that's considerably harder to diagnose than an error. CI is unaffected, since runners have no signing config, so it'd only ever bite contributors. Worth noting the repo requires DCO  --signoff , which is unrelated to gpg signing, so signing here is purely a local preference the fixture should be robust against.
Could both test repositories force signing off? Note it needs to cover the clone at line 58 as well, not just the seed — the script under test runs its own  git commit --signoff  inside  ${REPOSITORY} , so disabling it only on  ${SEED}  would just move the hang.

lakshmimsft (approved):

PR #12339, opened 2026-07-07
add icons support

nithyatsu:
# Description

This PR implements the core infrastructure for Resource Type Icons (Spec #003), enabling resource type authors to associate SVG icons with their types during registration. The icons are stored directly in the control plane and content-addressed via SHA-256 hashes.

### Setup fixtures

```bash
cat > /tmp/rt.yaml <<'YAML'
namespace: MyCompany.Resources
types:
  testResources:
    description: This is a test resource type.
    apiVersions:
      "2025-01-01-preview":
        schema: {}
YAML

cat > /tmp/rt-multi.yaml <<'YAML'
namespace: MyCompany.Multi
types:
  a: { apiVersions: { "2025-01-01-preview": { schema: {} } } }
  b: { apiVersions: { "2025-01-01-preview": { schema: {} } } }
YAML

cat > /tmp/icon.svg <<'SVG'
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"><circle cx="50" cy="50" r="40" fill="red"/></svg>
SVG

cat > /tmp/bad-root.svg <<'HTML'
<html><body>not an svg</body></html>
HTML

cat > /tmp/scripty.svg <<'SVG'
<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>
SVG

cat > /tmp/onload.svg <<'SVG'
<svg xmlns="http://www.w3.org/2000/svg" onload="alert(1)"/>
SVG

cat > /tmp/foreignobj.svg <<'SVG'
<svg xmlns="http://www.w3.org/2000/svg"><foreignObject><div/></foreignObject></svg>
SVG

cat > /tmp/external-href.svg <<'SVG'
<svg xmlns="http://www.w3.org/2000/svg"><image href="https://example.com/x.png"/></svg>
SVG

python3 -c "print('<svg xmlns=\"http://www.w3.org/2000/svg\">' + 'x' * 33000 + '</svg>')" > /tmp/big.svg
```

### Happy path

```bash
# 1. Register with icon (FR-004, FR-010, FR-012)
./drad resource-type create testResources -f /tmp/rt.yaml --icon /tmp/icon.svg

# 2. Verify server-computed hash matches SHA-256 of the file (FR-010)
EXPECTED=$(shasum -a 256 /tmp/icon.svg | awk '{print $1}')
[...]

PR #12396, opened 2026-07-14
Static graph icons + default icon

nithyatsu:
## Summary

- Static (modeled) graph now emits iconHash and a matching icons map from the CLI-embedded manifest — rad app graph app.bicep renders icons with no control plane.

- Default-icon substitution at registration time: types registered without an SVG get the product default hash stamped on the record 

- Graceful degradation — "icon absence is not an error": broken embedded asset logs and continues, callers set icon Hash to nil.

Note: static graphs can have only default icons for non default types, since we need control plane to register the tyep with an icon. All default.yaml types should render with icons in both  static and run time graphs. 

Fixes #12339 

## How to test

Contsruct an app.bicep with RRTs and and built in types and some types of your own. 
rad app graph /Users/nithya/radius/my-radius-recipes/deploy/app.bicep should produce app-graph.json that includes icons.


brooke-hamilton (commented):
A couple of minor doc/comment nits — nothing blocking. Nice, well-tested change.

pkg/cli/graph/modeled.go
```diff
@@ -25,20 +25,29 @@ import (
...
-	defaultResourceGroup     = "default"
-	applicationsResourceType = "Applications.Core/applications"
-	environmentsResourceType = "Applications.Core/environments"
-	recipePacksResourceType  = "Radius.Core/recipePacks"
+	defaultPlane         = "local"
+	defaultResourceGroup = "default"
 )
 
+// excludeResourceTypes are the resource types
+// are not graph members.
```
brooke-hamilton: Small wording tweak: this reads a little incompletely. Consider `excludeResourceTypes are the resource types that are not graph members.` so it's a complete sentence.

pkg/cli/graph/modeled.go
```diff
@@ -241,18 +299,18 @@ func stringAt(m map[string]any, key string) string {
...
 // buildModeledResource converts a single ARM JSON resource entry into an
 // ApplicationGraphResource. It returns (nil, nil) for entries that should
-// not appear as graph nodes (Radius applications and environments are
-// containers rather than graph members, and recipe packs are catalog
-// resources defined only in Radius.Core that hold reusable recipes).
+// not appear as graph nodes (applications and environments are containers
+// rather than graph members, and recipe packs are catalog resources that
+// hold reusable recipes). Both the Applications.Core and Radius.Core
+// spellings of applications/environments are excluded — see
+// containerResourceTypes.
```
brooke-hamilton: Doc reference points at `containerResourceTypes`, which isn't a symbol here. The actual names are the `excludeResourceTypes` var and the `isContainerResourceType` helper — worth updating so the reference resolves.

willtsai:
Recommend looking into these Copilot findings regarding validation of SVG data against malicious code:

**Only partially.** `ValidateIcon` rejects common active-content vectors:

- `<script>` and `<foreignObject>`
- `on*` event handlers
- External `href`/`xlink:href`
- Malformed XML, non-SVG roots, and oversized SVGs

It is **not a complete SVG sanitizer**. It permits `<style>`, style attributes, `data:` URLs, and other URL-bearing SVG/CSS attributes. Also, embedded static icons are hashed but never passed through `ValidateIcon`, and gathered graph SVGs are not revalidated. The endpoint’s CSP helps direct HTTP responses, but not necessarily consumers rendering SVG bytes from the graph’s inline `icons` map.

nithyatsu:
> Recommend looking into these Copilot findings regarding validation of SVG data against malicious code:
> 
> **Only partially.** `ValidateIcon` rejects common active-content vectors:
> 
> * `<script>` and `<foreignObject>`
> * `on*` event handlers
> * External `href`/`xlink:href`
> * Malformed XML, non-SVG roots, and oversized SVGs
> 
> It is **not a complete SVG sanitizer**. It permits `<style>`, style attributes, `data:` URLs, and other URL-bearing SVG/CSS attributes. Also, embedded static icons are hashed but never passed through `ValidateIcon`, and gathered graph SVGs are not revalidated. The endpoint’s CSP helps direct HTTP responses, but not necessarily consumers rendering SVG bytes from the graph’s inline `icons` map.

Addressed these issues in the validation.

willtsai:
Some more findings, please check to see if these can also be addressed:

The PR now rejects `<style>`, style attributes, external paint URLs, and validates all embedded icons in CI—good improvements.

`pkg/ucp/datamodel/icon_validation.go`
&nbsp;&nbsp;&nbsp;&nbsp;Line 159: URL-bearing attributes remain uncovered. For example, `marker-start="url(https://attacker/...)"` and `cursor="url(...)"` pass validation and may trigger network requests. The string scan is also bypassable with CSS escapes such as `fill="u\72l(https://attacker/...)"`. Use a complete URL-capable attribute allowlist and restricted CSS-value parsing before claiming validated SVGs never access the network.

Separately, the earlier integrity gap remains: gathered SVG bytes are still not re-hashed and compared with their advertised `iconHash`.

pkg/ucp/datamodel/icon_validation.go
```diff
@@ -88,11 +107,19 @@ func ValidateIcon(icon []byte) error {
...
 			}
+			if strings.EqualFold(name, "style") {
+				return fmt.Errorf("icon <%s> element has style attribute, which is not allowed", local)
+			}
 			if strings.EqualFold(name, "href") {
 				if err := validateHrefValue(local, attr.Value); err != nil {
 					return err
 				}
 			}
+			if isURLBearingAttr(name) {
```
willtsai, line 133: Please reject SVG animation elements as well. SMIL can mutate an attribute after this validation runs—for example, `<set attributeName="href" to="https://evil.example/x"/>` can turn an initially safe fragment reference into an external fetch at render time. Since icons are static assets, rejecting `animate`, `animateMotion`, `animateTransform`, `set`, and `discard` is safer than trying to validate every `from`/`to`/`by`/`values` mutation path. Please add rejection tests for at least `set` targeting `href` and `animate` targeting a URL-bearing presentation attribute, and update the documented validation list.

willtsai (commented):
Suggested tightening SVG validation to reject SMIL animation elements that can mutate checked attributes after validation.

pkg/ucp/datamodel/icon_validation.go
willtsai: Copilot suggested something like:
```
switch strings.ToLower(local) {
case "animate", "animatemotion", "animatetransform", "set", "discard":
    return fmt.Errorf("icon contains animation element <%s>, which is not allowed", local)
}
```

pkg/ucp/datamodel/icon_validation.go
```diff
@@ -79,20 +107,34 @@ func ValidateIcon(icon []byte) error {
...
 		}
+		if strings.EqualFold(local, "style") {
+			return fmt.Errorf("icon contains a <style> element, which is not allowed")
+		}
 		if strings.EqualFold(local, "foreignObject") {
 			return fmt.Errorf("icon contains a <foreignObject> element, which is not allowed")
 		}
+		if isSMILAnimationElement(local) {
+			return fmt.Errorf("icon contains a <%s> animation element, which is not allowed (SMIL can mutate attributes at render time and defeat this validator)", local)
+		}
```
willtsai, line 118: Please reject SVG animation elements as well. SMIL can mutate an attribute after this validation runs—for example, `<set attributeName="href" to="https://evil.example/x"/>` can turn an initially safe fragment reference into an external fetch at render time. Since icons are static assets, rejecting `animate`, `animateMotion`, `animateTransform`, `set`, and `discard` is safer than trying to validate every `from`/`to`/`by`/`values` mutation path. Please add rejection tests for at least `set` targeting `href` and `animate` targeting a URL-bearing presentation attribute, and update the documented validation list.

willtsai (dismissed):
thanks for addressing the icon validation feedback.

brooke-hamilton (approved):

PR #12519, opened 2026-07-23
fix(cli): stop bicep from inheriting rad's stderr handle

### LINKED ISSUE #12516: runBicepRaw leaks rad's stderr to the bicep child, hanging callers that pipe rad's stderr (Windows)
### Steps to reproduce

1. On Windows, install `rad` and bicep (`rad bicep download`).
2. From a parent process, spawn `rad` while **capturing its stderr through a pipe** and wait for the stdout/stderr streams to close. For example, in Node.js:

   ```js
   const child = spawn("rad", ["version", "--cli", "--output", "json"], {
     stdio: ["ignore", "pipe", "pipe"], // stderr piped
   });
   child.on("close", () => { /* never fires promptly */ });
   ```

3. `rad version --cli` internally shells out to the bicep binary:
   `writeCliVersionOnly` → `getCliVersionInfo()` → `bicep.Version()` → `runBicepRaw("--version")` in `pkg/cli/bicep/build.go`.
4. Observe that the parent's stderr stream never emits `close`/EOF when `rad` exits — it stays open until the **bicep grandchild** also exits. On Windows (where the child tree lives in the parent's Job Object) this deadlocks the parent, which hangs (or stalls until it hits its own timeout / kills the process tree).

### Observed behavior

`runBicepRaw` assigns the child bicep process the parent's real stderr handle:

```go
c := exec.Command(binPath, args...)
c.Stderr = os.Stderr          // bicep inherits rad's stderr (fd 2)
stdout, err := c.StdoutPipe() // stdout captured via rad's own pipe (fine)
```

Because bicep is given `os.Stderr` directly, it **inherits rad's real stderr handle**. When a caller launched `rad` with a *piped* stderr, the bicep grandchild now holds a write-end of that pipe. The caller's stderr stream therefore does not reach EOF until **bicep** exits — not just when `rad` exits. Any consumer that waits on rad's stderr stream to close will hang on Windows.

Note that stdout is **not** affected: rad already captures bicep's stdout through its own `StdoutPipe()`, so that handle is never inherited by anything external. Only stderr leaks.

`rad version --cli` is a common trigger because `--cli` skips the control-plane/Helm check but still calls `bicep.Version()`, so even a simple version probe spawns bicep.

### Desired behavior

bicep should never inherit the caller's stderr handle. rad should route bicep's stderr through a pipe **rad** owns and copy it to `os.Stderr` itself (exactly as it already does for stdout), so the only writer to the caller's stderr is rad. Then the caller observes stream close as soon as `rad` exits, regardless of bicep's lifecycle — while the live-streaming behavior of bicep's stderr is preserved.

Proposed change in `runBicepRaw` (replace `c.Stderr = os.Stderr`):

```go
stderr, err := c.StderrPipe()
if err != nil {
    return nil, fmt.Errorf("failed to create pipe: %w", err)
}
if err = c.Start(); err != nil { /* ... */ }

var wg sync.WaitGroup
wg.Add(1)
go func() {
    defer wg.Done()
    _, _ = io.Copy(os.Stderr, stderr)
}()

buf := bytes.Buffer{}
_, _ = io.Copy(&buf, stdout)

wg.Wait() // drain stderr before Wait() closes the pipe
if err = c.Wait(); err != nil { /* ... */ }
```

### Workaround

Callers can spawn `rad` with stderr set to `ignore` (not piped), and/or use detached/process-group spawning plus a hard timeout and process-tree kill. This is what the Radius Copilot Canvas extension does. But any consumer that legitimately wants to capture rad's stderr remains exposed to the hang.

### rad Version

Reproduced against `main` (built from source). Not version-specific — the `c.Stderr = os.Stderr` assignment in `runBicepRaw` is long-standing, so all recent releases are affected.

### Operating system

Windows 11, x64. The deadlock is Windows-specific (child processes share the parent's Job Object / console, and a pipe's read end only reaches EOF once every inherited write-end handle is closed). The underlying stderr-handle leak itself is cross-platform; Windows is just where it manifests as a hang.

### Additional context

- Call chain: `rad version --cli` → `writeCliVersionOnly` → `getCliVersionInfo()` → `bicep.Version()` → `runBicepRaw("--version")` (`pkg/cli/bicep/build.go`).
- Discovered while adding a "rad CLI must be ≥ latest release" version check to the Radius Copilot Canvas extension (radius-project/ai-extensions#115), which spawns `rad version --cli` and must not hang on Windows.
- A fix PR against `runBicepRaw` will be linked to this issue.

nicolejms:
## Summary

- route Bicep stderr through a rad-owned pipe instead of passing through the caller's stderr handle
- preserve live stderr streaming while preventing grandchild handle inheritance on Windows
- drain stdout and stderr before waiting for the Bicep process

## Validation

- `gofmt -l pkg/cli/bicep/build.go`
- `go build ./pkg/cli/...`
- `go vet ./pkg/cli/bicep/...`
- `go test ./pkg/cli/bicep/...`

Fixes #12516

### DIFF
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -74,9 +74,9 @@ func runBicepRaw(args ...string) ([]byte, error) {
 
 	var wg sync.WaitGroup
 	wg.Go(func() {
-    if _, err := io.Copy(os.Stderr, stderr); err != nil {
-        _, _ = io.Copy(io.Discard, stderr)
-    }
+		if _, err := io.Copy(os.Stderr, stderr); err != nil {
+			_, _ = io.Copy(io.Discard, stderr)
+		}
 	})
 
 	// copy to our buffer, we don't really need to observe
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -51,7 +51,7 @@ func runBicepRaw(args ...string) ([]byte, error) {
 
 	stdout, err := c.StdoutPipe()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create pipe: %w", err)
+		return nil, fmt.Errorf("failed to create stdout pipe: %w", err)
 	}
 
 	// Route bicep's stderr through a pipe we own instead of handing it os.Stderr
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -66,6 +66,8 @@ func runBicepRaw(args ...string) ([]byte, error) {
 	}
 
 	if err = c.Start(); err != nil {
+		_ = stdout.Close()
+		_ = stderr.Close()
 		return nil, fmt.Errorf("failed executing %q: %w", fullCmd, err)
 	}
 
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -62,7 +62,8 @@ func runBicepRaw(args ...string) ([]byte, error) {
 	// leaking the caller's handle to the grandchild.
 	stderr, err := c.StderrPipe()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create pipe: %w", err)
+		_ = stdout.Close()
+		return nil, fmt.Errorf("failed to create stderr pipe: %w", err)
 	}
 
 	if err = c.Start(); err != nil {
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -74,7 +74,9 @@ func runBicepRaw(args ...string) ([]byte, error) {
 
 	var wg sync.WaitGroup
 	wg.Go(func() {
-		_, _ = io.Copy(os.Stderr, stderr)
+    if _, err := io.Copy(os.Stderr, stderr); err != nil {
+        _, _ = io.Copy(io.Discard, stderr)
+    }
 	})
 
 	// copy to our buffer, we don't really need to observe
--- a/pkg/cli/bicep/build.go
+++ b/pkg/cli/bicep/build.go
@@ -33,8 +33,8 @@ import (
 // https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
 const SemanticVersionRegex = `(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
 
-// Run bicep with the given args and return the stdout. The stderr
-// is not capture but instead redirected to that of the current process.
+// Run bicep with the given args and return the stdout. Stderr is piped
+// through a private pipe and forwarded to the current process's stderr.
 func runBicepRaw(args ...string) ([]byte, error) {
 	if installed, _ := IsBicepInstalled(); !installed {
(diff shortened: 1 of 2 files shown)


pkg/cli/bicep/build.go
```diff
@@ -47,25 +48,43 @@ func runBicepRaw(args ...string) ([]byte, error) {
...
+
+	if err = c.Start(); err != nil {
+		_ = stdout.Close()
+		_ = stderr.Close()
 		return nil, fmt.Errorf("failed executing %q: %w", fullCmd, err)
 	}
 
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		_, _ = io.Copy(os.Stderr, stderr)
```
DariuszPorowski: Preserve draining after stderr forwarding fails

`io.Copy` stops on the first write error. If the caller closes rad's stderr pipe, the forwarding goroutine exits while Bicep's private stderr pipe remains open and undrained. A sufficiently verbose Bicep process can then fill that pipe and block forever, while rad waits for stdout EOF and never reaches `Cmd.Wait`.

Continue draining to `io.Discard` after a forwarding error, or explicitly close the pipe and handle the resulting Bicep write failure. Draining is less disruptive.

```suggestion
    if _, err := io.Copy(os.Stderr, stderr); err != nil {
        _, _ = io.Copy(io.Discard, stderr)
    }
```

pkg/cli/bicep/build_windows_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	stderrDone := make(chan []byte, 1)
+	go func() {
+		data, _ := io.ReadAll(stderr)
+		stderrDone <- data
+	}()
+
+	waitDone := make(chan error, 1)
+	go func() {
+		waitDone <- cmd.Wait()
```
DariuszPorowski: Make the regression test fail against the old implementation

The test does not distinguish this fix from the base behavior:

- `Version()` synchronously waits for the fake Bicep process, so Bicep has already exited before the helper exits.
- The test calls `cmd.Wait()` while `io.ReadAll(stderr)` is active and awaits `Wait` first. This violates the documented `StderrPipe` contract because `Wait` closes the pipe.
- The `ReadAll` error is discarded, so forced pipe closure can be mistaken for genuine EOF.

Keep the fake Bicep process alive using explicit handshakes, make the helper exit, and require a nil-error EOF before calling Wait or releasing Bicep. The base implementation should remain blocked because Bicep inherited the caller’s handle; this implementation should reach EOF.

DariuszPorowski (changes requested):
The central approach is correct: `StderrPipe` prevents Bicep from directly inheriting rad’s real stderr handle, and concurrent stdout/stderr draining preserves live output without the normal pipe-buffer deadlock. It therefore addresses the mechanism described in issue #12516, but the error-path regression and non-discriminating test should be fixed before merge.

pkg/cli/bicep/build_windows_test.go
nicolejms: @copilot  address and fix this omment

pkg/cli/bicep/build.go
```diff
@@ -47,25 +48,45 @@ func runBicepRaw(args ...string) ([]byte, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to create pipe: %w", err)
+		return nil, fmt.Errorf("failed to create stdout pipe: %w", err)
 	}
 
-	err = c.Start()
+	// Route bicep's stderr through a pipe we own instead of handing it os.Stderr
+	// directly. If bicep inherits the parent process's real stderr handle, a caller
+	// that launched rad with a piped stderr won't observe that pipe close until
+	// bicep also exits -- which hangs stderr stream-close waits on Windows. Copying
```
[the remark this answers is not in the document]
nicolejms: @copilot fix and add to batch

DariuszPorowski (approved):
🚀

PR #12563, opened 2026-07-28
Materialize schema property defaults before the resource is saved

### LINKED ISSUE #12532: [Bug]: Azure recipe size ternary leaks literally to ARM when properties.size is unset (postgres + redis)
### What type of issue is this?

Bug report

### Summary

Azure recipes use a Go-template ternary `{{ context.resource.properties.size == "S" ? ... : ... }}` to pick a tier/SKU. When `size` is set the expression resolves; when `size` is **unset**, the entire ternary is emitted **literally** into the ARM template, and ARM rejects it against the property's `allowedValues`. This affects `postgreSqlDatabases` (`tier`/`skuName`/`version`) and `redisCaches` (`skuName`).

### Steps to reproduce

1. Deploy the [eShop fork](https://github.com/Reshrahim/eShop) ([`.radius/app.bicep`](https://github.com/Reshrahim/eShop/blob/main/.radius/app.bicep)) with `Radius.Data/postgreSqlDatabases` and `Radius.Data/redisCaches` resources that **do not** set `properties.size`.
2. Run the deploy.
3. Observe the recipe deployment fail.

### Expected behavior

The recipe tolerates an unset `size` (e.g. defaults it before the ternary), or the resource types default `size`, so a valid tier/SKU is always emitted.

### Actual behavior

`RecipeDeploymentFailed` / `InvalidTemplate`: the literal string `{{context.resource.properties.size == "S" ? ... : ...}}` is rejected by ARM `allowedValues` for the postgres `tier` and redis `skuName`.

### Root cause

The recipe uses `{{ }}` ternaries referencing `context.resource.properties.size`. With `size` unset, the template engine leaves the ternary unrendered and passes it through literally. Affects postgres `tier`/`skuName`/`version` (`aks-recipepack.bicep` ~L189–190, L151) and redis `skuName` (~L36).

### Workaround (what unblocked it locally)

Set `size:'S'` on every `Radius.Data` resource.

### Screenshots or logs

- Run 29948244197 — postgres `tier` failed (size unset): https://github.com/Reshrahim/eShop/actions/runs/29948244197
- Run 29949004864 — after `size:'S'`, all 4 postgres + rabbitMQ succeeded; redis still failed until its `size` was also set: https://github.com/Reshrahim/eShop/actions/runs/29949004864

### Version or commit

`rad version: edge / 5b3b50e / bicep 0.42.1`

### Environment

macOS; Radius Copilot canvas extension; Azure `aks-recipepack`; deploy via GitHub Actions.

AzureMike:
## Summary

A resource type schema can declare a `default` for a property, and Radius never applied it. `schema.ApplyDefaults` fills absent properties from the default declared in the schema, and a new `UpdateFilter` on the dynamic resource PUT path runs it before the resource is saved.

Explicit values are never overwritten, and required properties are never defaulted, so leaving one out still fails validation instead of being quietly satisfied.

## Reason for change

Fixes #12532

Deploying a resource without setting an optional property puts a literal template expression into the ARM request. The Azure recipe pack writes parameters as expressions over the resource:

```
skuName: '{{context.resource.properties.size == "S" ? "Balanced_B0" : "Balanced_B1"}}'
```

Leave `size` unset and there's nothing to substitute, so the raw `{{...}}` string is handed to ARM as the SKU name and rejected against `allowedValues`.

The cause sits a level below the recipe. Validation checked a resource against its schema without filling anything in, so every `default:` key in every resource type has been inert — including the ones `objectStorage`, `mongoDatabases`, `models`, `kafka` and `rabbitMQ` already declare.

With the defaults from radius-project/resource-types-contrib#260 declared, the existing recipe expressions resolve on their own, and neither the resolver nor the recipe pack needs to change. That covers ten optional properties across the Azure pack, not just `size`:

| Type | Property | Unset today | With defaults |
| --- | --- | --- | --- |
| `redisCaches` | `size` | leaks | `Balanced_B0` |
| `postgreSqlDatabases` | `size` | leaks | `Standard_B1ms` / `Burstable` |
| `mySqlDatabases` | `version` | leaks | `8.0.21` |
| `models` | `model` | leaks | `gpt-5-mini` |
| `mongoDatabases` | `database` | leaks | `mongo_db` |
| `mySqlDatabases` | `database` | leaks | `mysql_db` |
| `postgreSqlDatabases` | `database` | leaks | `postgres_db` |
| `rabbitMQ` | `queue` | leaks | `jobs` |
| `kafka` | `topic` | leaks | `events` |
| `objectStorage` | `containerName` | leaks | `data` |

Six of those defaults already existed, so half the fix is Radius finally reading what contrib already wrote down.

Two things reviewers should know. The filter sits ahead of the encryption filter, so a defaulted sensitive field is still encrypted. And defaults are reapplied on every write rather than carried forward from the previous version of the resource, which is what full-replace PUT semantics ask for — the consequence is that changing a declared default later changes what an existing resource gets on its next deployment, so a default is part of a resource type's API contract.

## How to test

`go test ./pkg/schema/... ./pkg/dynamicrp/...` covers the new code directly. `pkg/schema/defaults_test.go` asserts that explicit values survive, required properties are never defaulted, read-only properties are skipped, absent nested objects aren't created, and map and slice defaults are deep copied so one resource can't mutate what the next one sees.

End to end, all ten optional properties above were run through `ApplyDefaults` and the parameter resolver against the real contrib schemas and the real `recipepack/azure/aks-recipepack.bicep` expressions. Every one emits a literal `{{...}}` without defaults and resolves to the value in the table with them, on the unmodified resolver and the unmodified recipe expressions.

`go vet` and `gofmt` are clean, and `pkg/schema/...`, `pkg/dynamicrp/...`, `pkg/recipes/...`, `pkg/cli/manifest/...` and `pkg/ucp/integrationtests/resourceproviders/...` all pass, including after rebasing onto the kin-openapi 0.144.0 bump in #12561.

Built-in resource types pick this up once radius-project/resource-types-contrib#260 merges and `make update-resource-types` bumps the pin in `go.mod`, since the manifests under `deploy/manifest/built-in-providers` are copies CI regenerates from it.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/schema/defaults.go` | New. `ApplyDefaults` walks a schema and fills absent properties from their declared `default`, returning the number applied. |
| `pkg/schema/defaults_test.go` | New. Covers explicit values, required properties, read-only properties, nested objects, and deep copying of map and slice defaults. |
| `pkg/dynamicrp/frontend/defaultsfilter.go` | New. `UpdateFilter` that fetches the resource type schema and applies defaults before the resource is saved, attaching the filled map only when something was applied. |
| `pkg/dynamicrp/frontend/routes.go` | Register the defaults filter ahead of the encryption filter. |


pkg/dynamicrp/frontend/defaultsfilter.go
```diff
@@ -0,0 +1,93 @@
...
+		newResource *datamodel.DynamicResource,
+		oldResource *datamodel.DynamicResource,
+		options *controller.Options,
+	) (rest.Response, error) {
+		return applySchemaDefaults(ctx, newResource, ucpClient)
+	}
+}
+
+// applySchemaDefaults fills unset properties from the "default" values declared in the resource type schema.
+func applySchemaDefaults(
```
kachawla, line 59: Should we add a few tests for this filter too? `ApplyDefaults` is well covered, but this path isn’t. It would be good to cover schema lookup errors, nil `Properties`, and make sure defaults are applied before encryption.

pkg/schema/defaults.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		if readOnly, ok := fieldSchema["readOnly"].(bool); ok && readOnly {
+			continue
+		}
+
+		// A supplied object is descended into even when required, since its own optional
+		// properties may declare defaults.
+		if existing, present := properties[name]; present {
+			if nested, ok := existing.(map[string]any); ok {
```
kachawla: Should we also apply defaults to objects inside arrays and values under `additionalProperties`? It looks like those cases would currently be skipped.

kachawla (commented):
Looks good overall. Couple of comments.

pkg/dynamicrp/frontend/defaultsfilter.go
AzureMike: Good point, added.

pkg/schema/defaults.go
AzureMike: Good point, fixed.

kachawla (approved):

PR #12566, opened 2026-07-28
feat: support oci hashes for tag

nithyatsu:
## Summary

Pins the OCI tag of every default recipe in the CLI-seeded default recipe pack to the per-namespace commit SHA recorded in `deploy/manifest/defaults.yaml`, replacing the previous "semver on release, `latest` on edge" scheme.

Getting there needed a home for the code that reads those pins. `deploy/manifest` could not be it: `//go:embed` cannot name a path outside its own package directory, so the embed directive has to sit beside the assets — but the assets are deployment artifacts owned by the build tooling (`build/scripts/sync-resource-types.sh` rewrites them, `build/docker.mk` copies them into the `ucpd` image, `build/generate.mk` reads them). This PR splits that package in two: `deploy/manifest` keeps the `//go:embed` and nothing else, and a new `pkg/defaults` owns all parsing of those bytes — both the icon lookups that already lived there and the new revision pins. Icon consumers move over with no behavior change.

### Tag resolution — before / after

| Build | Before | After |
| ----- | ------ | ----- |
| Edge / dev | `latest` (mutable) | `edge` (mutable, renamed) |
| Release | `version.Release()`, e.g. `0.58.0` — one tag shared by every recipe | the immutable commit SHA `defaults.yaml` pins that recipe's namespace to, resolved per recipe |

Namespaces are pinned independently in `defaults.yaml`, so recipes in a single build can now carry different tags. If a namespace has no `resourceTypes` entry — or the resource type is malformed — the tag falls back to `edge`, so a mis-configured build still emits a valid OCI reference rather than a broken one.

Nothing outside the CLI changes: this only affects the recipe-pack seed written by `rad`.

## Reason for change

A released `rad` installed whatever `latest` / `<semver>` happened to point at, not the recipes published from the `resource-types-contrib` revision that release was actually built against. `defaults.yaml` already records that revision per namespace (kept current by `make update-resource-types`), and the recipes publishing pipeline already stamps the same commit SHA on each namespace's OCI artifacts. This change consumes that pin so a released CLI installs exactly the recipes matching its own revision.

## How to test

The release path is covered by ordinary unit tests — no ldflags build required. `resolveRecipeTag` takes `isEdge` as a parameter precisely so both channel branches are reachable from a normally-built test binary:

```bash
go test ./pkg/cli/recipepack/... -run Test_resolveRecipeTag -v
```

Four subtests, none skipped: edge → `edge`, release → the pinned SHA for `Radius.Compute`, and the two fallbacks (malformed resource type, unpinned namespace) → `edge`.

Icon behavior should be unchanged — `pkg/defaults` is a move of existing logic, so the regression surface is the relocated icon tests plus the graph and resource-provider suites:

```bash
go test ./pkg/defaults/... ./pkg/cli/graph/... \
  ./pkg/corerp/frontend/controller/applications/... ./pkg/ucp/...
```

To eyeball the release tags without a release build, compare the SHA asserted by `Test_resolveRecipeTag` against the `resourceTypes` entries in `deploy/manifest/defaults.yaml` (currently `158be19…` for every namespace).

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `deploy/manifest/embed.go` | New. Assets-only package: exports the `//go:embed` `FS` plus the `DefaultIconPath`, `DefaultsYAMLPath`, and `BuiltInIconsDir` path constants. Holds no behavior. |
| `deploy/manifest/icons.go` | Deleted. Its parsing logic moves to `pkg/defaults`. |
| `pkg/defaults/defaults.go` | New. Package doc for the defaults contract; declares the `Pin` type and the `catalog` mirror of `defaults.yaml`, and parses the embedded bytes once in `init()`. A read or parse failure is logged to stderr and leaves the lookups empty rather than aborting — every accessor has a documented "not found" answer. |
| `pkg/defaults/icons.go` | New. Icon half of the old `deploy/manifest/icons.go`: `Icon`, `LookupIcon`, `DefaultIcon`, `DefaultIconHash`, `SplitResourceType`, and the `loadIcons` init helper. Preserves the existing per-type → product-default → `nil` fallback chain. |
| `pkg/defaults/pins.go` | New. `ResourceTypePin(namespace)` exposes the `resourceTypes` section of `defaults.yaml`; entries missing a name or ref are logged and skipped. |
| `pkg/defaults/icons_test.go` | Moved from `deploy/manifest/icons_test.go` and repointed at the new package. |
| `pkg/defaults/pins_test.go` | New. Covers pin lookup hits, misses, and the skip-malformed-entry path. |
| `pkg/cli/recipepack/recipepack.go` | `GetCoreTypesRecipeInfo` computes `isEdge` once and calls the new `resolveRecipeTag(resourceType, isEdge)` per recipe. The helper returns `edge` for edge builds and otherwise resolves the namespace to its pinned SHA via `defaults.ResourceTypePin`, falling back to `edge` on a malformed type or missing pin. Doc comment rewritten for the new contract. |
| `pkg/cli/recipepack/recipepack_test.go` | Edge test asserts the `:edge` suffix. `Test_resolveRecipeTag` is table-driven over both channels plus both fallback paths — no `t.Skip`, so the release branch is exercised by ordinary `go test`. |
| `pkg/cli/graph/modeled.go`, `modeled_test.go` | Icon lookups repointed from `deploy/manifest` to `pkg/defaults` (`Lookup` → `LookupIcon`, `DefaultHash` → `DefaultIconHash`). |
| `pkg/corerp/frontend/controller/applications/v20250801preview/getgraph.go`, `graphicons.go`, `graphicons_test.go` | Same repoint for the runtime graph pipeline. |
| `pkg/ucp/api/v20231001preview/resourcetype_conversion.go`, `resourcetype_conversion_test.go` | Same repoint for the resource-type conversion layer. |
| `pkg/ucp/frontend/controller/resourceproviders/geticon.go`, `geticon_test.go` | Same repoint for the icon endpoint. |
| `pkg/ucp/initializer/service.go`, `service_test.go` | Same repoint for default resource-type registration. |
| `pkg/ucp/integrationtests/resourceproviders/summary_test.go` | Same repoint. |
| `test/functional-portable/ucp/noncloud/resourceprovider_test.go` | Same repoint. |
| `docs/architecture/application-graph.md` | Icon-resolution sections updated to reference `pkg/defaults` and the new accessor names, and to explain why the assets stay embedded from `deploy/manifest`. |


deploy/manifest/sources.go
```diff
@@ -0,0 +1,100 @@
...
+// the immutable ref recorded in defaults.yaml under `sources`. It is
+// populated once at package init and treated as read-only afterwards,
+// mirroring the concurrency model of the `builtIns` map in icons.go.
+var sourceRefs map[string]string
+
+// sourcesFile matches the subset of defaults.yaml consumed here. Other
+// top-level fields (defaultRegistration, plus future additions) are
+// tolerated by yaml.Unmarshal and ignored.
+type sourcesFile struct {
+	Sources []sourcePin `json:"sources"`
```
DariuszPorowski: check: https://github.com/radius-project/radius/blob/main/deploy/manifest/defaults.yaml
```suggestion
	resourceTypes []sourcePin `json:"resourceTypes"`
```

kachawla:
@copilot update the PR description to capture current state of the changes

kachawla (dismissed):

DariuszPorowski (approved):

kachawla (approved):

PR #12577, opened 2026-07-29
Surface GitHub OIDC enterprise-claim failures in the Azure verify workflow

ryanwaite:
## Summary

Adds a diagnostic to the Azure credential-verify workflow (`.github/extension/verify-azure.yml`) so that when Azure Login (OIDC) is rejected over GitHub's `enterprise` claim (Entra `AADSTS7002381`), the run explains the cause and the fix in plain language instead of only surfacing the raw error code.

The diagnostic uses **progressive disclosure** and is **silent on success**, so the common case — a hobbyist deploying a personal repo to a non-enforcing Azure tenant, where an empty enterprise claim is normal and harmless — sees no extra output.

- **Inspect GitHub OIDC token claims** (new first step): decodes the runner's OIDC token locally (the token itself is never printed) and logs the non-secret claims — `repository`, `repository_owner`, `sub`, `enterprise`, `aud` — to the collapsed step log as a debugging breadcrumb. It writes nothing to the job summary and exports an `enterprise_empty` output. `set -e` is intentionally omitted so this best-effort step can never abort verification.
- **Explain OIDC enterprise-claim failure** (new, failure-only step): runs only when Azure Login failed **and** the enterprise claim was empty (the `AADSTS7002381` signature), writing a plain-language explanation plus fix to the job summary. It is skipped on success.

## Reason for change

The `AADSTS7002381` enterprise-claim rejection is a tenant-side federated-credential policy (common in Microsoft corporate tenants). When it fires, users see an opaque Azure error code with no indication that the cause is a missing GitHub `enterprise` claim or how to resolve it. This has been repeatedly hard to diagnose in practice.

An empty enterprise claim is only fatal when the target tenant enforces the policy — which cannot be known before login — so the diagnostic deliberately **warns/explains rather than hard-fails**, and only when the failure actually occurs. `azure/login` remains the authority on credential success.

Fixes #

## How to test

Run the **Radius - Verify Credentials** workflow for an Azure environment:

1. **Personal repo → non-enforcing tenant (success path):** the run concludes green and the job summary shows only the existing "✅ Credentials Verified" content — no enterprise-claim text. The decoded claims are visible only if you expand the collapsed "Inspect GitHub OIDC token claims" step log.
2. **Repo whose OIDC token has an empty `enterprise` claim → enforcing tenant (failure path):** Azure Login fails with `AADSTS7002381`, and the job summary now includes an "⚠️ Azure Login failed on the GitHub enterprise claim" section explaining the cause and the fix.

Validation performed locally: YAML parses; both embedded `run:` blocks pass `shellcheck`; the base64url JWT-payload decode was round-tripped against sample tokens with and without the `enterprise` claim.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/verify-azure.yml` | Add a silent OIDC token-claims decode step (step-log only, exports `enterprise_empty`) before Azure Login, and a failure-only step that explains an `AADSTS7002381` enterprise-claim rejection in the job summary. |
| `.github/extension/README.md` | Document the two new verify-azure steps and their silent-on-success behavior. |


.github/extension/verify-azure.yml
```diff
@@ -16,13 +16,86 @@ jobs:
...
+        # (AADSTS7002381). On a successful run this step is skipped and adds nothing, so
+        # the common hobbyist path (personal repo, non-enforcing tenant) stays silent.
+        if: failure() && steps.azure_login.outcome == 'failure' && steps.oidc_claims.outputs.enterprise_empty == 'true'
+        run: |
+          {
+            echo "## ⚠️ Azure Login failed on the GitHub enterprise claim"
+            echo ""
+            echo "Azure Login (OIDC) failed and this repository's GitHub OIDC token had an empty \"enterprise\" claim."
+            echo ""
+            echo "If the Azure error includes AADSTS7002381, the target tenant is enforcing an enterprise-claim requirement on its federated-credential policy (common in Microsoft corporate tenants). GitHub only includes the enterprise claim for repositories owned by an account or organization that belongs to a GitHub Enterprise — personal-account repositories cannot satisfy this policy."
```
kachawla: Could we change the step name and heading to say this is a **possible** enterprise-claim problem?

The workflow only knows that Azure Login failed and the `enterprise` claim was empty. The login could have failed for another reason, such as incorrect Azure settings. Calling this the cause may send users toward the wrong fix.

For example: “Possible GitHub enterprise-claim mismatch.” Users can then confirm the cause by checking whether the Azure error contains `AADSTS7002381`.

.github/extension/README.md
```diff
@@ -21,10 +21,12 @@ Each workflow runs on `ubuntu-latest`. No long-lived cloud secrets are stored 
...
 
-1. **Authenticate via OIDC.** Runs `azure/login` with the environment's Azure variables.
-2. **Verify access.** Runs `az account show` to confirm the login.
-3. **Verify AKS access.** Sets up `kubelogin`, fetches cluster credentials with `az aks get-credentials`, converts the kubeconfig, and runs `kubectl cluster-info`.
-4. **Summary.** Writes the environment and provider to the job summary.
+1. **Inspect OIDC token claims (diagnostic breadcrumb).** Decodes the runner's GitHub OIDC token locally (the token is never printed) and logs the claims — `repository`, `repository_owner`, `sub`, `enterprise`, `aud` — to the collapsed step log. It writes nothing to the job summary, so a successful run shows no extra output. This step never fails the job.
+2. **Authenticate via OIDC.** Runs `azure/login` with the environment's Azure variables.
+3. **Explain enterprise-claim failure (failure-only).** Runs only when Azure Login failed and the OIDC token's `enterprise` claim was empty — the signature of a tenant-side enterprise-claim policy (`AADSTS7002381`). It writes a plain-language explanation and fix to the job summary. On a successful run it is skipped and adds nothing.
```
kachawla: Same here

Could we avoid calling this “the signature” of `AADSTS7002381`?

A failed Azure Login and an empty `enterprise` claim do not prove that the claim caused the failure. Azure Login could have failed for another reason, so this wording may point users to the wrong fix.

Please describe it as a **possible** enterprise-claim problem and tell users to check for `AADSTS7002381` in the Azure error to confirm it.

.github/extension/verify-azure.yml
ryanwaite: Good call — updated in 39c0790.

Renamed the step to "Report possible GitHub enterprise-claim mismatch" and changed the summary heading to "## ⚠️ Possible GitHub enterprise-claim mismatch". The body is now conditional: it says the combination *may* mean the enterprise claim caused the failure, lists the other things that can fail Azure Login (wrong client/tenant/subscription id, missing federated credential, insufficient role assignments), and adds a "Confirm first" line telling users to check the Azure Login error for `AADSTS7002381` — with an explicit note that the fix below does not apply if it is absent.

.github/extension/README.md
ryanwaite: Fixed in 39c0790 — dropped the "signature of" framing here too.

Step 3 is now: "**Report possible enterprise-claim mismatch (failure-only).** Runs only when Azure Login failed and the OIDC token's `enterprise` claim was empty. That combination is a hint, not a diagnosis — Azure Login can fail for unrelated reasons — so the job summary asks the user to confirm by checking whether the Azure Login error contains `AADSTS7002381` before applying the suggested fix."

Same wording change was applied to the code comments in `verify-azure.yml`.

kachawla (approved):

PR #12590, opened 2026-07-30
refactor: icon and resource type handling to use new defaults package

DariuszPorowski:
- Introduced a new `defaults` package to centralize default icon and resource type pin management.
- Updated various components to replace direct references to `productmanifest` with the new `defaults` package.
- Ensured that all icon handling logic now utilizes `defaults.DefaultIcon()` and `defaults.DefaultIconHash()` for consistency.
- Adjusted tests to reflect changes in the way default icons are accessed and validated.
- Improved documentation within the `defaults` package to clarify the purpose and usage of default icons and resource type pins.


pkg/defaults/defaults.go
```diff
@@ -0,0 +1,130 @@
...
+// handle, so a broken asset degrades the product instead of killing it.
+func init() {
+	// The ucplog logger is not available here: init() runs at package-import
+	// time, before any context.Context exists and before the ucp logger has
+	// been configured.
+	logger := log.New(os.Stderr, "defaults: ", log.LstdFlags)
+
+	var parsed catalog
+	if raw, err := manifestassets.FS.ReadFile(manifestassets.DefaultsYAMLPath); err != nil {
+		logger.Printf("read defaults.yaml: %s; icon and pin lookups disabled", err)
```
kachawla, line 123: Should we return if parsing defaults.yaml fails? Right now we say the lookups are disabled, but still load them from parsed, which could contain partially parsed values.

kachawla (approved):

pkg/defaults/defaults.go
DariuszPorowski: fixed

kachawla (approved):

PR #12605, opened 2026-08-03
Fast-fail rad deploy when app.bicep is missing on the branch

sylvainsf:
## Description

The `run-rad-commands` composite action compiles `app.bicep` with `bicep build` before deploying. When the application file is missing on the deployed branch/commit, this surfaced as a confusing Bicep **compile** error rather than a clear "file not found" signal — which is exactly the failure mode reported for deploys dispatched before the generated `app.bicep` was committed and pushed.

This PR adds an explicit existence check in `load_declared_app_params` that **fast-fails before the Bicep compiler is invoked**, with an actionable message:

> `Application file .radius/app.bicep not found on this branch/commit. Generate it (ask Copilot to "create app.bicep") and commit it to the deployed branch before deploying.`

The gate runs on both the default deploy path and custom commands that target the app file.

This is defense-in-depth on the workflow side. The authoritative fix (never dispatch a deploy against a commit missing `app.bicep`, and disable the Deploy button when it is absent) is tracked in the extension: radius-project/ai-extensions#238. See also radius-project/ai-extensions#113.

## Type of change

- Bug fix (non-breaking change which fixes an issue / improves an error message)

## Testing

- Added a unit test in `deploy-parameters_test.sh` asserting the gate:
  - fast-fails on a missing app file,
  - never invokes the Bicep compiler,
  - emits the "not found on this branch/commit" message,
  - does not populate the parameter cache.
- `bash .github/extension/actions/run-rad-commands/deploy-parameters_test.sh` passes.
- `shellcheck` clean on the modified script.

## Related

- radius-project/ai-extensions#113
- radius-project/ai-extensions#238

### DIFF
--- a/.github/extension/actions/run-rad-commands/deploy-parameters.sh
+++ b/.github/extension/actions/run-rad-commands/deploy-parameters.sh
@@ -22,7 +22,16 @@ load_declared_app_params() {
     # error below. This commonly happens when a deploy is dispatched before the
     # generated app.bicep has been committed and pushed to the deployed branch.
     if [[ ! -f "${app_file}" ]]; then
-        echo "::error::Application file ${app_file} not found on this branch/commit. Generate it (ask Copilot to \"create app.bicep\") and commit it to the deployed branch before deploying." >&2
+        # Escape workflow-command metacharacters in the interpolated path before
+        # embedding it in the ::error:: annotation: '%' is the escape character
+        # (must be first), and CR/LF would otherwise break or inject the command.
+        # The assignments are intentionally unquoted so the single quotes around
+        # the patterns/replacements are treated as quoting, not literal text.
+        local escaped_app_file="${app_file}"
+        escaped_app_file=${escaped_app_file//'%'/'%25'}
+        escaped_app_file=${escaped_app_file//$'\r'/'%0D'}
+        escaped_app_file=${escaped_app_file//$'\n'/'%0A'}
+        echo "::error::Application file ${escaped_app_file} not found on this branch/commit. Generate it (ask Copilot to \"create app.bicep\") and commit it to the deployed branch before deploying." >&2
         return 1
     fi
 
--- a/.github/extension/actions/run-rad-commands/deploy-parameters.sh
+++ b/.github/extension/actions/run-rad-commands/deploy-parameters.sh
@@ -17,6 +17,15 @@ load_declared_app_params() {
         return 0
     fi
 
+    # Fast-fail with a clear message when the application file is missing on the
+    # deployed branch/commit, rather than surfacing a confusing Bicep compile
+    # error below. This commonly happens when a deploy is dispatched before the
+    # generated app.bicep has been committed and pushed to the deployed branch.
+    if [[ ! -f "${app_file}" ]]; then
+        echo "::error::Application file ${app_file} not found on this branch/commit. Generate it (ask Copilot to \"create app.bicep\") and commit it to the deployed branch before deploying." >&2
+        return 1
+    fi
+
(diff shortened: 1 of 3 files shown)


.github/extension/actions/run-rad-commands/deploy-parameters.sh
```diff
@@ -17,6 +17,15 @@ load_declared_app_params() {
         return 0
     fi
 
+    # Fast-fail with a clear message when the application file is missing on the
+    # deployed branch/commit, rather than surfacing a confusing Bicep compile
+    # error below. This commonly happens when a deploy is dispatched before the
+    # generated app.bicep has been committed and pushed to the deployed branch.
+    if [[ ! -f "${app_file}" ]]; then
```
brooke-hamilton, line 24: This gate is unreachable in the Azure/AWS deploy path, so the confusing failure it's meant to replace still happens.

`action.yml`'s first step ("Verify app namespace on target cluster", line 47) runs before `Run rad commands` and does:

```bash
BICEP_APP_NAME=$(grep -oP "name:\s*'\K[^']+" ".radius/app.bicep" 2>/dev/null | head -1)
```

Composite `shell: bash` defaults to `bash --noprofile --norc -e -o pipefail`. With a missing `.radius/app.bicep`, `grep` exits 2, `pipefail` propagates it, and `set -e` kills the step — with stderr swallowed by `2>/dev/null`, so the job dies with a bare exit 2 and no message at all. Verified locally:

```
case 1: app.bicep missing            -> exit=2 (BICEP_APP_NAME never assigned)
case 2: app.bicep exists, no name:   -> exit=1 (the BICEP_APP_NAME="app" fallback is dead code)
case 3: app.bicep exists with name   -> exit=0
```

`run-rad-commands-azure.yml`/`-aws.yml` set `RADIUS_TARGET_KUBECONFIG` to a real file in the same `deploy` job, so `[ -f "$TARGET_KUBECONFIG" ]` is true and this is the normal path, not an edge case.

Suggest making that step tolerant so it can't pre-empt this gate, e.g.:

```bash
BICEP_APP_NAME=$(grep -oP "name:\s*'\K[^']+" "$APP_FILE" 2>/dev/null | head -1 || true)
```

(also fixes the dead `"app"` fallback, and uses `$APP_FILE` instead of the hardcoded `.radius/app.bicep`). `publish-deploy-status/action.yml:33` already uses the `|| true` form.

sylvainsf:
Thanks both — addressed in eeb1c62.

**@brooke-hamilton (unreachable gate):** You're right, and the analysis was spot on. The "Verify app namespace on target cluster" step ran first and, with `set -e -o pipefail`, a missing `app.bicep` made `grep` exit 2 and killed the step with no message — pre-empting this gate on the normal Azure/AWS path. Fixed by making that grep tolerant (`… | head -1 || true`), which also makes the `BICEP_APP_NAME="app"` fallback reachable, and switched it to `$APP_FILE` instead of the hardcoded `.radius/app.bicep`. Added a test asserting the `|| true` form is present so it can't regress.

**@Copilot (unescaped `::error::`):** Fixed — the app-file path is now escaped (`%`→`%25`, CR→`%0D`, LF→`%0A`, `%` first) before being embedded in the annotation, matching the `::group::` title escaping in `action.yml`. Added a test that a `%` in the path is emitted as `%25`.

`deploy-parameters_test.sh` and `shellcheck` pass. The `upgrade-noncloud` failure was on the prior commit and is unrelated to this change (this PR only touches the `run-rad-commands` composite action, which that lane doesn't exercise); CI re-runs on the new commit.

brooke-hamilton (approved):

PR #12610, opened 2026-08-04
Fix deploy-status publisher never publishing, and failing successful deploys

### LINKED ISSUE #12613: publish-deploy-status passes absolute paths to ORAS, so Canvas receives no deployed graph
## Summary

The shared `publish-deploy-status` action generates `deploy-graph.json` after a successful Radius deployment, but its `oras push` command passes five absolute paths under a temporary directory. ORAS rejects those paths before uploading anything, which leaves the expected graph package absent.

The failure follows this sequence:

1. The Canvas-generated deployment workflow runs `rad deploy`.
2. The **Run rad commands** step succeeds.
3. `publish-deploy-status` generates the deployed graph in a directory returned by `mktemp -d`.
4. The action passes `${STATUS_DIR}/...` absolute paths to `oras push`.
5. ORAS rejects every file with `absolute file path detected`.
6. No graph artifact reaches GHCR, so the Canvas **Deployed** tab has no `deploy-graph.json` to render and shows **Nothing deployed yet**.

## Steps to reproduce

1. Create a Radius environment through Canvas with GHCR-backed state.
2. Deploy an application through the generated `run-rad-commands` workflow.
3. Let `rad deploy` complete successfully.
4. Observe the **Publish deployed graph and status** step.
5. Open the Canvas **Deployed** tab after the workflow completes.

[GitHub Actions run 30860602195](https://github.com/AzureMike/getting-started-todo-app/actions/runs/30860602195) reproduced the failure. Its **Run rad commands** step completed successfully, and the logs contain:

```text
Deployment complete.
```

The following **Publish deployed graph and status** step generated the graph for `todo-app`, then failed:

```text
Generating deployed graph for app 'todo-app' in environment 'mikeazure-radius-test'...
Error: absolute file path detected. If it's intentional, use
--disable-path-validation flag to skip this check:
/tmp/tmp.U3wleWv2Qe/deploy-graph.json,
/tmp/tmp.U3wleWv2Qe/deploy-progress.log,
/tmp/tmp.U3wleWv2Qe/deploy-activity.log,
/tmp/tmp.U3wleWv2Qe/deploy-controlplane.log,
/tmp/tmp.U3wleWv2Qe/deploy-state.txt
```

The deployment step succeeded and the publisher step failed in the same job.

## Observed behavior

The action creates its files with:

```bash
STATUS_DIR=$(mktemp -d)
```

It then passes absolute paths to ORAS:

```bash
oras push "$TARGET_REF" \
  --artifact-type application/vnd.radius.deploy-status.v1+json \
  "${STATUS_DIR}/deploy-graph.json:application/json" \
  "${STATUS_DIR}/deploy-progress.log:text/plain" \
  "${STATUS_DIR}/deploy-activity.log:text/plain" \
  "${STATUS_DIR}/deploy-controlplane.log:text/plain" \
  "${STATUS_DIR}/deploy-state.txt:text/plain"
```

The run had no explicit `RADIUS_GRAPH_REGISTRY`, so the action derived the target registry by replacing `radius-state` with `radius-graph` in:

```text
ghcr.io/azuremike/getting-started-todo-app-radius-state-mikeazure-radius-test-adbfbe974a7b
```

The expected graph package is therefore:

```text
ghcr.io/azuremike/getting-started-todo-app-radius-graph-mikeazure-radius-test-adbfbe974a7b
```

That package doesn't exist. The Canvas reader derives the same registry, tries to pull `deploy-graph.json` from its OCI artifact, and falls back to an empty graph when neither that artifact nor the legacy status branch contains one. The UI then renders **Nothing deployed yet**.

## Desired behavior

A successful graph generation should publish all five status files to the derived graph package, and Canvas should be able to read `deploy-graph.json` from that artifact.

The ORAS invocation must use a supported path form. Two direct options are:

- Run `oras push` from `STATUS_DIR` and pass relative file names.
- Pass `--disable-path-validation`, which ORAS documents as `skip path validation` and which the observed error explicitly recommends for intentional absolute paths.

Publishing should only report success after the artifact exists. If publishing fails, the warning should preserve the real ORAS error and state that the Canvas graph wasn't published.

## Workaround

Making the publish step non-blocking prevents this reporting failure from marking the deployment itself as failed, but it doesn't create the artifact and doesn't fix the empty **Deployed** tab.

A repository-level workaround must change the ORAS invocation to accept the generated files, then retain the existing registry, tag, media type, and layer titles expected by the Canvas reader.

## rad Version

Not captured. The `rad` deployment and graph-generation commands succeeded; the failure occurs in the subsequent ORAS invocation inside the shared GitHub composite action.

## Operating system

GitHub-hosted `ubuntu-latest` runner.

## Additional context

Open [radius#12610](https://github.com/radius-project/radius/pull/12610) already changes this action. Its current patch catches `oras push` errors, but it still passes the same absolute paths, which would make the workflow green without publishing the graph. That PR should also use relative paths or `--disable-path-validation` and close this issue.

brooke-hamilton:
## Description

Bug fixes and small cleanups to the deploy-status publisher, in preparation for getting deployed graph polling working. These are defects I hit while testing the GHCR publish path.

Fixes #12613 — the publisher never actually published anything, so the canvas **Deployed** tab showed "Nothing deployed yet" even after a fully successful deploy. Two other fixes here cause a **successful deploy to be reported as failed**.

## What changed

**1. Push the status files from their own directory so ORAS accepts them** (`publish-deploy-status/action.yml`)

`oras push` was given absolute `${STATUS_DIR}/...` paths and rejected all five files, so no artifact was ever created:

```
Error: absolute file path detected. If it's intentional, use --disable-path-validation
flag to skip this check: /tmp/tmp.U3wleWv2Qe/deploy-graph.json, ...
```

Now runs the push from inside `STATUS_DIR` in a subshell with bare file names. Note that `--disable-path-validation` is **not** an equivalent fix: ORAS derives each layer's `org.opencontainers.image.title` from the path it's given, so absolute paths yield titles like `/tmp/tmp.xoPITBhblC/deploy-graph.json` — containing a random `mktemp` component. The canvas looks these up by exact title, so that option makes the push succeed while leaving the artifact permanently unreadable.

**2. Publish deploy status even when the deploy fails** (`run-rad-commands-azure.yml`, `run-rad-commands-aws.yml`)

The publish step was gated on `if: success()`, so a failed deploy published nothing and the Deployed tab stayed empty — which is when it's most useful. Changed to `if: !cancelled()`, which still skips cancellation, where the control plane is already being torn down.

**3. Don't fail the deploy when publishing fails** (`publish-deploy-status/action.yml`)

`oras push` ran unguarded under `set -e`, so a publishing error failed the whole job even though the deployment succeeded. Publishing status is best-effort reporting, so it now warns and exits 0, matching every other failure path in that action. The warning includes the actual ORAS error rather than always blaming `packages: write`.

**4. Generate valid OCI tags for non-ASCII application names** (`publish-deploy-status/action.yml`)

Tag sanitization used `sed -E 's/[^a-z0-9._-]+/-/g'`, but in the runner's UTF-8 locale GNU sed leaves multi-byte characters intact. An app named `café` produced the tag `aks-dev-café-latest`, which is invalid per the OCI spec, so `oras push` failed. Sanitizing under `LC_ALL=C` strips them bytewise → `aks-dev-caf-latest`, which also matches what the canvas reader already computed for the same app.

**5. Give the three sibling status files distinct content** (`publish-deploy-status/action.yml`)

`deploy-progress.log`, `deploy-activity.log`, and `deploy-controlplane.log` were all copies of `rad-commands-result.json` — three identical blobs (same digest, deduplicated by the registry) that told the reader nothing. They now carry per-resource provisioning state, the rad command result, and control-plane health respectively.

**6. Warn when the app name comes from the `rad app list` fallback** (`publish-deploy-status/action.yml`)

The canvas derives the artifact tag from the first single-quoted `name:` literal in the app bicep and has no such fallback, so those runs publish to a tag it will never request. Previously this surfaced only as an empty tab with no explanation.

## How to test

`test-publisher-e2e.sh` extracts the **real** script body from the action YAML (so there's no copy to drift) and runs it against a **real** `oras` and a **real** local `registry:2`, then asserts the artifact exists and its layer titles match what the canvas reads:

```
=== was the artifact actually published? ===
   artifact EXISTS
   artifactType: application/vnd.radius.deploy-status.v1+json
   canvas contract check:
      OK      deploy-graph.json
      OK      deploy-progress.log
      OK      deploy-activity.log
      OK      deploy-controlplane.log
      OK      deploy-state.txt
```

Confirmed it fails against the pre-fix version (`*** ARTIFACT NOT PUBLISHED ***`) and passes after. Earlier harnesses stubbed `oras`, which is exactly why #12613 slipped through.

Also validated with `shellcheck -s bash` on the extracted body, YAML parsing of all three modified files, and a scenario harness covering:

| Scenario | Before | After |
|---|---|---|
| Deploy succeeded | **published nothing** (#12613) | publishes, exit 0 |
| Deploy failed | **skipped entirely** | publishes, exit 0 |
| App named `café` | **invalid tag, job fails** | `aks-dev-caf-latest`, exit 0 |
| Name from `rad app list` | silently wrong tag | warns, exit 0 |
| `oras push` returns 403 | **job fails** | warns with real error, exit 0 |

## Follow-ups (not in this PR)

- `verify-ghcr-push` only checks `RADIUS_STATE_REGISTRY`, but the publisher writes to the derived graph package. The check should cover the package actually being written to.
- `RADIUS_GRAPH_REGISTRY` is also read by the `rad` CLI (`cmd/rad/cmd/root.go:262`) to select the modeled graph archive backend, so setting it for the Deployed tab silently switches that archive from git to OCI. Two unrelated concerns sharing one variable.
- In `ai-extensions`, the reader still fetches `deploy-state.txt` and the three logs from the `radius-deploy-status` orphan branch, which this workflow no longer creates. The graph renders, but deploy state and logs won't.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/actions/publish-deploy-status/action.yml` | Push from `STATUS_DIR` with relative paths (#12613); surface the real ORAS error; don't fail the deploy on push errors; `LC_ALL=C` tag sanitization; distinct content for the three status files; warn on fallback-derived app name |
| `.github/extension/run-rad-commands-azure.yml` | Publish step `if: success()` → `if: !cancelled()` |
| `.github/extension/run-rad-commands-aws.yml` | Publish step `if: success()` → `if: !cancelled()` |
| `.github/extension/README.md` | Document publish-on-failure, best-effort publishing, and what each status file now contains |


AzureMike:
Hey @brooke-hamilton, I think this [issue](https://github.com/radius-project/radius/issues/12613) might be relevant to this PR?

.github/extension/actions/publish-deploy-status/action.yml
```diff
@@ -48,21 +50,45 @@ runs:
...
         EXIT_CODE="0"
         if [ -f "$RESULT_FILE" ]; then
           OUTCOME=$(jq -r '.outcome // "unknown"' "$RESULT_FILE" 2>/dev/null || echo "unknown")
           EXIT_CODE=$(jq -r '.exitCode // 0' "$RESULT_FILE" 2>/dev/null || echo "0")
+        fi
+
+        # deploy-progress.log - per-resource provisioning state from the live
+        # control plane, the closest thing to per-node status available here.
+        if ! rad resource list -a "$APP_NAME" -o json 2>/dev/null \
+          | jq -r '.[]? // .items[]? | "\(.name)\t\(.type)\t\(.properties.provisioningState // "Unknown")"' \
```
sylvainsf: Checked the output shape on this one: `rad resource list -o json` marshals a `[]generated.GenericResource` array directly (`pkg/cli/cmd/resource/list/list.go`, the final `WriteFormatted` on `resourceList`), so there is no `.items`/`.value` object wrapper. `.[]?` iterates the array correctly and `// .items[]?` is just dead code, not broken. So the Copilot suggestion to add `.value[]?` does not apply to the real output, safe to dismiss. Only thing worth doing here is dropping the dead `// .items[]?` for clarity, and the whole pipe is guarded with a placeholder fallback anyway, so a shape surprise degrades gracefully.

sylvainsf (commented):
Nice cleanup. Five real defects, each with a clear root cause, and the reasoning in the comments is easy to follow. A few thoughts, nothing blocking on correctness.

What I like:
- `if: !cancelled()` is the right operator. Publish runs after Run rad commands but before `rad shutdown`/teardown (both `if: always()`), so the control plane is still up when it queries. Ordering is safe.
- The `LC_ALL=C` fix is the best part. Root cause is dead on: GNU sed in the runner's UTF-8 locale leaves multibyte chars intact so `[^a-z0-9._-]` never strips the `é`, and forcing bytewise handling lines the producer up with the canvas JS reader. Applying it to `tr` too avoids locale casefolding surprises.
- Best-effort `oras push` with `exit 0` now matches every other degradation path in the action, and the success echo is correctly only on the fall-through.
- Distinct content for the three sibling files kills the identical-blob problem.

Two asks:

1. Please commit the test harness. The matrix in the description is exactly what I'd want guarding this (the `café` sanitization and the publish-on-failure/exit-0 behavior), but the diff is only YAML/README, so none of it is enforced. We already have precedent for committing this style of extracted-script test in `deploy-parameters_test.sh` (#12605). Validated-once-locally will drift.

2. On the Copilot suppressed comment about `.[]? // .items[]?`: I checked the output shape and it does not apply, see the inline note. Fine to dismiss it.

Nits, take or leave:
- The new `rad app graph` / `rad resource list` calls run against a possibly-wedged control plane on the failure path. Guarded for non-zero exit but not for a hang. Not a regression (existing code has the same property), but a `timeout` wrapper would harden it.
- Failure-path publish is still gated on `rad app graph` succeeding (line 46 exits 0 without publishing if it fails), which slightly undercuts the "most useful on failure" goal when the graph call is what broke. Acceptable, maybe a one-line README note.

CI: the red `corerp-cloud` failure looks unrelated (this PR only touches extension YAML/README, that lane does not exercise `publish-deploy-status`), but worth a glance against a clean `main` run before merge rather than assuming. Everything else is green and `pr:standard` is on.

.github/extension/actions/publish-deploy-status/action.yml
brooke-hamilton: Fixed in be9d64db2 — dropped the `// .items[]?` alternative.

Confirmed your reading of the output shape: `resourceList` is a `[]generated.GenericResource` passed straight to `WriteFormatted` (`pkg/cli/cmd/resource/list/list.go:189,231`), so it marshals as a bare array with no wrapper. Agreed the Copilot `.value[]?` suggestion doesn't apply.

One thing worth flagging — the alternative wasn't just dead, it was mildly harmful. Testing the two filters against real-shaped payloads:

| input | `.[]? // .items[]?` | `.[]?` |
|---|---|---|
| populated array | 2 rows | 2 rows (identical) |
| **empty array** | `jq: error: Cannot index array with string "items"` | clean, no output |
| `{"items":[...]}` | `jq: error: Cannot index array with string "properties"` | n/a |

On an empty array — a deploy that hasn't created resources yet, or failed early — jq evaluated the alternative, errored, and exited non-zero, sending the pipeline down the `if !` failure branch. Same end state thanks to the placeholder fallback, but via an error path. And the alternative didn't even work for the wrapped shape it was ostensibly there to handle.

So `.[]?` alone is both simpler and strictly better behaved. Added a comment recording why there's no wrapper, so it doesn't get "defensively" re-added.

brooke-hamilton:
@AzureMike yes that's a real bug in this PR's code path, and it's now fixed in 1d07a0163. Thanks for the writeup, the run link and the exact ORAS error made it straightforward to reproduce.

I confirmed the failure against a local registry using the action's exact invocation:

```
Error: absolute file path detected. If it's intentional, use --disable-path-validation
flag to skip this check: /tmp/tmp.cNlxHwnoVf/deploy-graph.json, ...
```

**One correction to the issue.** The two fixes you listed aren't equivalent — only the relative-path one works. ORAS derives each layer's `org.opencontainers.image.title` from the path it's given, and the canvas looks files up by exact title:

| fix | resulting layer title | canvas finds `deploy-graph.json`? |
|---|---|---|
| `cd $STATUS_DIR` + bare names | `deploy-graph.json` | **yes** |
| `--disable-path-validation` | `/tmp/tmp.xoPITBhblC/deploy-graph.json` | **no** |

With `--disable-path-validation` the push succeeds, the step goes green, and the artifact is permanently unreadable because the title embeds a random `mktemp` component that differs every run. That would convert today's loud failure into a silent one — strictly worse for debugging. So I went with `cd` + bare names (in a subshell, so it doesn't leak into the rest of the step).

**On your point about this PR making it green without publishing** — fair, and worth being precise about. The non-blocking change is still right on its own terms: publishing status is reporting, and it shouldn't fail a deployment that genuinely succeeded. But you're correct that it would have masked this. Two things address that: the warning now includes the actual ORAS error instead of always guessing `packages: write`, and the artifact is really published, so the green run means something.

**Why my testing missed it.** My round-trip test did `cd "$STATUS_DIR"` and passed relative names while claiming to be "the producer's exact invocation" — so it quietly tested the fixed form and validated an artifact shape the real code never produced. The scenario harness stubbed `oras` with a script that echoed its arguments and exited, so it never validated paths at all. Both gave me false confidence.

Replaced with `test-publisher-e2e.sh`, which extracts the real script body from the action YAML and runs it against a real `oras` and a real local registry, then asserts the artifact exists and its layer titles match what the canvas reads. Verified it fails on the pre-fix code and passes after:

```
=== was the artifact actually published? ===
   artifact EXISTS
   artifactType: application/vnd.radius.deploy-status.v1+json
   canvas contract check:
      OK      deploy-graph.json
      OK      deploy-progress.log
      OK      deploy-activity.log
      OK      deploy-controlplane.log
      OK      deploy-state.txt
```

This PR closes #12613.

sylvainsf (approved):
looks great

PR #12611, opened 2026-08-04
fix(ci): rebuild resource-types bot branch from main instead of rebasing

DariuszPorowski:
## Summary

Unwedges the `Update Resource Types` workflow, which currently fails on every `repository_dispatch` from `resource-types-contrib` (for example [run 30929498155](https://github.com/radius-project/radius/actions/runs/30929498155)).

## Reason for change

**Root cause.** The `Prepare cumulative update branch` step ran `git rebase origin/main` on the long-lived `bot/update-resource-types` branch. That branch's pending commit (`defcfb363`) was written when `deploy/manifest/defaults.yaml` still used `sources:` / `namespace:`. Main has since moved to `resourceTypes:` / `name:` / `tag:` / `recipePacks:` (#12567), so the old one-line pin bump can no longer be replayed textually and the rebase conflicts:

```text
CONFLICT (content): Merge conflict in deploy/manifest/defaults.yaml
error: could not apply defcfb363... chore: refresh resource type manifests ... (release: Radius.Compute)
```

Because the branch is shared and long-lived, this is not transient — every subsequent dispatch fails the same way.

**Proposed fix.** Rebasing is the wrong tool here: `make update-resource-types` regenerates everything it writes, so the branch's only durable state is its *pin set*, not its commit. The branch is now rebuilt from `origin/main` on every dispatch, and the still-pending pins are replayed as extra `RESOURCE_TYPES_PINS` / `RECIPE_PACKS_PINS` input. Cumulative behaviour is preserved and this conflict class disappears.

Supporting changes:

- Pins are only carried forward while the branch still has an **open PR**, so a leftover already-merged branch cannot drag pins backwards.
- Only pins that actually differ from `main` are carried, so the reported carried-forward list is accurate.
- `sources` / `namespace` are accepted as pre-rename aliases, which recovers the currently stuck `Radius.Compute` pin (`cfe0e011…`) instead of silently dropping it. This alias is transitional and can be removed once the branch has been rebuilt by a successful run.
- An entry with a `tag` is carried by tag so the tag survives re-resolution; a bare SHA re-pins without a network round trip.
- The push now uses an explicit `--force-with-lease=refs/heads/<branch>:<sha>` (SHA captured before the rebuild). The bare form leases against a remote-tracking ref this job never creates, which would have rejected every push after the first.
- `Install yq` moved ahead of branch preparation, since `install-yq` only appends its install dir to `GITHUB_PATH` and `yq` is therefore on `PATH` from the *next* step onwards.

## How to test

1. `actionlint .github/workflows/update-resource-types.yaml` — clean.
2. `yamllint` — no new violations (the three >120 char lines are pre-existing `github-script` body lines).
3. The `yq` + `jq` pin-merge logic was exercised locally against the real stuck branch plus five synthetic scenarios, all producing the expected merged pin sets:
   - real stuck branch (legacy `sources`/`namespace` schema) → carries `Radius.Compute` only, plus the payload pin
   - branch identical to `main` → nothing carried
   - pending tagged namespace + pending recipe pack, recipe-pack-only payload → both carried, tag preserved as the ref
   - payload also covers the pending name → payload wins
   - no pending branch, with and without a payload → payload only / empty
4. End-to-end verification happens on the next `resource-types-contrib` dispatch: the job should now complete and refresh the PR on `bot/update-resource-types` rather than failing in `Prepare cumulative update branch`.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/workflows/update-resource-types.yaml` | Replace the conflict-prone `git rebase` with a rebuild-from-`main` strategy that replays pending pins as make input; gate carry-forward on an open PR and on pins differing from `main`; accept pre-rename `sources`/`namespace` spellings; use an explicit `--force-with-lease` value; move `Install yq` before branch preparation; report carried-forward entries in the commit message and PR body. |

### DIFF
--- a/.github/workflows/update-resource-types.yaml
+++ b/.github/workflows/update-resource-types.yaml
@@ -233,7 +233,8 @@ jobs:
 
           pending='{}'
           if [ -n "${head_sha}" ] && [ -n "$(
-            gh pr list --repo "${GITHUB_REPOSITORY}" --head "${PR_BRANCH}" \
+            gh pr list --repo "${GITHUB_REPOSITORY}" \
+              --head "${GITHUB_REPOSITORY%%/*}:${PR_BRANCH}" \
               --state open --json number --jq '.[].number'
           )" ]; then
             git fetch origin "${PR_BRANCH}"
--- a/.github/workflows/update-resource-types.yaml
+++ b/.github/workflows/update-resource-types.yaml
@@ -250,25 +250,30 @@ jobs:
           base="$(yq -o=json -I=0 '.' "${defaults}")"
 
           # carried <section> <legacy_section> <payload_json> echoes the branch
-          # pins that main does not already have and the payload does not
-          # already cover, as a [{name, ref}] array. `sources`/`namespace` are
-          # the pre-rename spellings of `resourceTypes`/`name`; accepting them
-          # keeps a branch written against the older schema readable. An entry
-          # with a tag is carried by tag so the tag survives re-resolution; a
-          # bare SHA is re-pinned without a network round trip.
+          # pins the payload does not already cover and main does not already
+          # record with the same SHA *and* tag, as a [{name, ref}] array. The
+          # tag has to be part of that comparison: a release can tag a commit
+          # main is already pinned to, and comparing SHAs alone would discard
+          # that tag-only pin. `sources`/`namespace` are the pre-rename
+          # spellings of `resourceTypes`/`name`; accepting them keeps a branch
+          # written against the older schema readable. An entry with a tag is
+          # carried by tag so the tag survives re-resolution; a bare SHA is
+          # re-pinned without a network round trip.
           carried() {
             jq -c --arg section "$1" --arg legacy "$2" \
               --argjson payload "$3" --argjson base "${base}" '
               [ (.[$section] // .[$legacy] // [])[]
                 | {name: (.name // .namespace),
                    ref: (if (.tag // "") == "" then .ref else .tag end),
-                   sha: .ref}
+                   sha: .ref,
+                   tag: (.tag // "")}
                 | select(.name != null and .ref != null)
                 | . as $e
                 | select(($payload | any(.name == $e.name)) | not)
                 | select(
                     (($base[$section] // [])
-                      | any(.name == $e.name and .ref == $e.sha)) | not
+                      | any(.name == $e.name and .ref == $e.sha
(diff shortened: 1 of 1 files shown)


.github/workflows/update-resource-types.yaml
```diff
@@ -197,47 +201,133 @@ jobs:
...
-          else
-            git checkout -B "${PR_BRANCH}" origin/main
+          head_sha="$(
+            git ls-remote origin "refs/heads/${PR_BRANCH}" | awk '{print $1}'
+          )"
+
+          pending='{}'
+          if [ -n "${head_sha}" ] && [ -n "$(
+            gh pr list --repo "${GITHUB_REPOSITORY}" \
+              --head "${GITHUB_REPOSITORY%%/*}:${PR_BRANCH}" \
```
lakshmimsft: `gh pr list --head` does not accept the `owner:branch` form. `gh pr
    list` queries GraphQL `headRefName`, which matches a bare branch name only, so
    this filter always returns empty. Verified against this repo with gh 2.80.0:

        gh pr list --repo radius-project/radius --head "dp/inherent-gibbon-aqua"                 -> 12611
        gh pr list --repo radius-project/radius --head "radius-project:dp/inherent-gibbon-aqua"  -> (empty)

    Effect: the open-PR guard never passes, `pending` stays `{}`, and NO pins are
    ever carried forward. The workflow still succeeds, so this fails silently --
    and it drops the stuck `Radius.Compute` pin (`cfe0e011…`) this PR is meant to
    recover. That defeats the cumulative behavior the change is built around.

    Both of these work (verified):
        gh api "repos/${GITHUB_REPOSITORY}/pulls?state=open&head=${GITHUB_REPOSITORY%%/*}:${PR_BRANCH}" --jq '.[].number'
    or keep gh pr list and filter the owner explicitly:
        gh pr list --repo "${GITHUB_REPOSITORY}" --head "${PR_BRANCH}" --state open \
          --json number,headRepositoryOwner \
          --jq '.[] | select(.headRepositoryOwner.login=="'"${GITHUB_REPOSITORY%%/*}"'") | .number'

    Note the REST `head=owner:branch` form is what the earlier copilot review comment was
    describing; it is correct for the API but not for `gh pr list`.

DariuszPorowski: done

lakshmimsft (approved):

PR #12616, opened 2026-08-04
Add context.azure.resourceNameHash for globally unique Azure recipe names

AzureMike:
## Summary

Azure Recipes currently pass raw Radius names to services that require globally unique names, so common names such as `redis` and `rabbitmq` fail with `NameInUse`.

Direct Bicep and Terraform modules receive concrete parameters after Radius resolves Recipe context, which means they can't use Bicep's `uniqueString()` with a Radius resource ID. This change adds `context.azure.resourceNameHash`, which returns the first 16 lowercase hexadecimal characters of the SHA-256 hash of the lowercased Azure provider scope ID, a NUL separator, and the lowercased Radius resource ID. The provider scope is the resource-group ID when configured and otherwise the subscription ID.

The Azure Recipe Pack can combine this value with service-specific prefixes, giving each resource a stable Azure name without adding wrapper modules or limiting the fix to Bicep.

## Validation

- `go test ./pkg/recipes/...`

Supports radius-project/ai-extensions#128


pkg/recipes/paramresolver/resolver.go
```diff
@@ -243,6 +246,16 @@ func buildContextLookup(ctx *recipecontext.Context) map[string]string {
 		"context.environment.id":   ctx.Environment.ID,
 	}
 
+	// This stable hash avoids Azure name collisions while keeping redeployments stable.
+	// See https://github.com/radius-project/ai-extensions/issues/128.
```
kachawla: nit: I'd recommend not putting github issue links in the code unless it's a TODO. We should add any relevant context in the comment here if we think more information is needed to explain the code.

kachawla (commented):
Was there a reason for computing this directly in `buildContextLookup` instead of adding `resourceNameHash` to the Azure recipe context schema? I was wondering whether it would be cleaner to compute it when the context is created and have the lookup map only expose the existing value. That would keep the context object and the `context.*` expression namespace consistent.

kachawla (approved):

PR #12622, opened 2026-08-05
chore(ci): harden workflows against template injection and cache poisoning

DariuszPorowski:
# Summary

Hardens the GitHub Actions surface against two classes of findings reported by [`zizmor`](https://docs.zizmor.sh/):

- **Template injection** (`template-injection`) — four composite actions interpolated `${{ inputs.* }}` and `${{ steps.*.outputs.* }}` expressions directly into `run:` script bodies. Expression expansion happens *before* the shell sees the script, so an input containing shell metacharacters is pasted in verbatim and executed. Every such usage is now bound to a step-level `env:` variable and referenced as an ordinary shell variable, which is subject to normal quoting rules.
- **Cache poisoning** (`cache-poisoning`) — jobs that publish artifacts restored dependency caches that an earlier workflow run holding a valid `GITHUB_TOKEN` could have poisoned. Cache restoration is now disabled on `actions/setup-go` in the affected jobs, and `actions/setup-node`'s implicit package-manager cache is disabled in the docs-publishing workflow.

`zizmor .github/actions` now reports no findings.

## Reason for change

These are the two highest-signal classes of workflow security findings in the repository, and both have well-understood remediations that do not change workflow behavior.

The template-injection changes are the more important half: `install-git-http-backend` previously expanded `${{ inputs.git-password }}` straight onto a command line, and the KinD/registry actions expanded registry names and paths into heredocs and `docker` invocations.

Note the tradeoff on the cache-poisoning half: disabling Go module/build cache restoration in `build` and `unit-tests` trades CI wall-clock time for removing the poisoned-cache execution path. Reviewers should weigh whether that is the right call for `unit-tests` in particular, since it does not publish artifacts.

## How to test

1. `zizmor .github/actions` — expect "No findings to report."
2. `actionlint .github/workflows/build.yaml .github/workflows/unit-tests.yaml .github/workflows/publish-docs.yaml` — no new findings (two pre-existing shellcheck `info`/`style` notes in `build.yaml` are unrelated to the changed lines).
3. CI must stay green. The behavioral risk is concentrated in the composite actions, so the functional-test workflows that call `create-kind-cluster`, `create-local-registry`, `install-flux`, and `install-git-http-backend` are the ones to watch — specifically that the local registry is reachable from the KinD nodes and that the generated TLS cert paths still resolve.

## File change summary

| File                                                | Summary of change                                                                                                                                                                     |
| --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `.github/actions/create-kind-cluster/action.yaml`      | Bind `registry-port`, `registry-name`, `registry-server`, and `temp-cert-dir` to step-level `env:` and reference them as shell variables in both the insecure and secure cluster steps. |
| `.github/actions/create-local-registry/action.yaml`    | Same treatment for the cert generation, trust-store, `/etc/hosts`, and secure/insecure registry steps, including the `create-temp-cert-dir` step output.                                |
| `.github/actions/install-flux/action.yaml`             | Pass `version` via `env:` instead of interpolating it into the script invocation.                                                                                                       |
| `.github/actions/install-git-http-backend/action.yaml` | Pass git username, password, namespace, and image via `env:` — the password is no longer expanded onto the command line.                                                                |
| `.github/workflows/build.yaml`                         | `cache: false` on both `actions/setup-go` steps.                                                                                                                                       |
| `.github/workflows/unit-tests.yaml`                    | `cache: false` on `actions/setup-go`.                                                                                                                                                  |
| `.github/workflows/publish-docs.yaml`                  | `cache: false` on `actions/setup-go` and `package-manager-cache: false` on `actions/setup-node`.                                                                                        |


.github/workflows/build.yaml
```diff
@@ -227,7 +227,7 @@ jobs:
         with:
           go-version-file: go.mod
           cache-dependency-path: go.sum
-          cache: true
+          cache: false
```
lakshmimsft, line 230: copilot finding: The hardening is incomplete. `functional-test-noncloud.yaml`
    still has `cache: true` at lines 159 and 281, yet its `build` job uploads artifacts
    (lines 191–205) and its `tests` job runs `make docker-push` (line 352) — exactly the
    publish-with-restored-cache pattern this PR is removing from build.yaml.
    `functional-test-cloud.yaml` (389, 699) and `long-running-azure.yaml` (167) are in
    the same shape.

.github/workflows/unit-tests.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
         with:
           go-version-file: go.mod
           cache-dependency-path: go.sum
-          cache: true
+          cache: false
```
lakshmimsft, line 61: suggestion: This job only uploads test-result XML and a Codecov report — nothing downstream consumes them as a build input, so the poisoned-cache path has no meaningful sink. Disabling the Go
    module *and* build cache on the repo's most frequently run workflow is a real
    wall-clock cost on every PR. Suggestion: Prefer a scoped `# zizmor: ignore[cache-poisoning]`
    with a comment explaining why the risk is accepted.

.github/workflows/build.yaml
DariuszPorowski: done

.github/workflows/unit-tests.yaml
DariuszPorowski: done

lakshmimsft:
@copilot re-review pr

lakshmimsft (approved):

PR #12639, opened 2026-08-10
Fixed garbage text leaking into terminal for rad deploy, rad delete, and rad app delete

### LINKED ISSUE #12541: rad deploy/rad delete leak terminal capability-query responses to the shell
### Steps to reproduce

1. Use a terminal that supports mode 2026 (synchronized output), mode 2027, and/or the kitty keyboard protocol (for example a recent build of iTerm2, WezTerm, Ghostty, or kitty).
2. Run a deployment with the interactive progress display:

```bash
rad deploy app.bicep
```

3. Wait for the deployment to finish and return to the shell prompt.

### Observed behavior

After the progress display exits, stray characters are printed at the shell prompt and the terminal bell rings:

```text
❯ rad deploy new-todolist.bicep
Building new-todolist.bicep...
Deploying template 'new-todolist.bicep' into environment '.../environments/default' from workspace 'default'...

Deployment In Progress...

..                   frontend        Radius.Compute/containers
Completed            todolist        Radius.Core/applications

Deployment Complete

Resources:
    frontend        Radius.Compute/containers
    todolist        Radius.Core/applications

❯ 2026;2$y2027;4$y1u
```

The trailing `2026;2$y2027;4$y1u` are the terminal's replies to capability-detection queries that Bubble Tea v2 emits on startup:

| Query Bubble Tea sends | Terminal reply | Meaning |
| ---------------------- | -------------- | ------- |
| `CSI ? 2026 $p`        | `2026;2$y`     | mode 2026 = synchronized output |
| `CSI ? 2027 $p`        | `2027;4$y`     | mode 2027 = grapheme clustering |
| kitty keyboard `CSI ? u` | `1u`         | keyboard-protocol flags |

The bell (`BEL`) is part of the same handshake.

### Root cause

The interactive progress listener for `rad deploy` and `rad delete` creates its Bubble Tea program with an empty, non-TTY input reader in `pkg/cli/output/progress/progress.go`:

```go
program := tea.NewProgram(
    newModel(l.progressChan),
    tea.WithInput(strings.NewReader("")), // empty, non-TTY reader
    tea.WithoutSignalHandler(),
)
```

Bubble Tea writes the capability queries to **stdout** (the real terminal), and the terminal answers back on **stdin**. Because the program was given an empty `strings.Reader` instead of the real stdin, Bubble Tea never consumes the answers, so they remain in the terminal's input buffer and are echoed to the prompt after the program exits.

This was introduced by the Bubble Tea v1 → v2 upgrade in #12168. The empty-reader approach was intended to keep the terminal in cooked mode so that Ctrl+C still terminates the process during a deployment; it preserves Ctrl+C but leaks the query responses.

### Desired behavior

`rad deploy` and `rad delete` should not leak terminal capability-query responses to the shell and should not ring the bell. The progress display should consume its own query responses (for example by using the real stdin so Bubble Tea reads the answers it solicits), while still allowing Ctrl+C to interrupt the operation.

### Workaround

The stray characters are cosmetic; pressing Enter clears the prompt. There is no functional impact on the deployment or deletion itself.

### rad Version

❯ rad version
CLI Version Information:
edge      ca1568e   0.42.1    ca1568eee71011dabae50d96f2fe6b4964bbbdab

Control Plane Information:
STATUS     VERSION
Installed  edge

### Operating system

macOS, arm64 (Apple silicon). Reproduces in terminals that support modes 2026/2027 or the kitty keyboard protocol.

### Additional context

- Affected code: `pkg/cli/output/progress/progress.go` (`interactiveListener.Run`).
- Same failure mode as the `rad init` progress display, which is being fixed in #12540. The fix there was to let Bubble Tea use the real stdin (so it consumes its own query responses) and handle Ctrl+C explicitly in the model. The same approach applies here.

### LINKED ISSUE #12638: rad app delete leaks terminal capability-query responses to the shell
### Steps to reproduce

1. Use a terminal that supports mode 2026 (synchronized output), mode 2027, and/or the kitty keyboard protocol (for example a recent build of iTerm2, WezTerm, Ghostty, or kitty).
2. Run an application deletion with the interactive progress display:

```bash
rad app delete todolist
```

3. Wait for the deletion to finish and return to the shell prompt.

### Observed behavior

After the progress display exits, stray characters are printed at the shell prompt and the terminal bell rings:

```text
❯ rad app delete todolist
Deleting application 'todolist' from environment 'default'...

Completed2$y^[[?2027;4$y^[[?1u...                  todolist        Applications.Core/applications

Applications.Core/applications/todolist deleted

~/radius/test
❯ 2026;2$y2027;4$y1u
```

The trailing characters (`2026;2$y2027;4$y1u`, along with escaped sequences like `2$y^[[?2027;4$y^[[?1u` interleaved into the progress output) are the terminal's replies to capability-detection queries that Bubble Tea v2 emits on startup.

### Root cause

`rad app delete` uses `delete.DeleteApplicationWithProgress` (`pkg/cli/delete/delete.go`), which creates an interactive progress listener via `progress.NewListener` in `pkg/cli/output/progress/progress.go`. Same root cause as #12541.

### Desired behavior

`rad app delete` should not leak terminal capability-query responses to the shell and should not ring the bell, while still allowing Ctrl+C to interrupt. Same fix as #12541/#12540.

### Workaround

The stray characters are cosmetic; pressing Enter clears the prompt. No functional impact on the deletion itself.

### rad Version

❯ rad version
CLI Version Information:
edge      ca1568e   0.42.1    ca1568eee71011dabae50d96f2fe6b4964bbbdab

### Operating system

macOS, arm64 (Apple silicon).

### Additional context

Affected code: `pkg/cli/output/progress/progress.go` shared with #12541 and #12540.

zachcasper:
## Summary

Fix `rad deploy`, `rad delete`, and `rad app delete` leaking terminal
capability-query responses (for example `2026;2$y2027;4$y1u`) to the shell after
the interactive progress display exits, and restore Ctrl+C support during those
operations.

## Reason for change

The interactive progress listener built its Bubble Tea program with an empty,
non-TTY input reader (`tea.WithInput(strings.NewReader(""))`). Bubble Tea v2
writes capability-detection queries (synchronized output mode 2026, grapheme
clustering mode 2027, and the Kitty keyboard protocol) to stdout on startup, and
the terminal answers on stdin. Because the program never read the real stdin, it
never consumed those answers: the terminal stayed in cooked mode and echoed the
replies, so they were interleaved into the progress output and left at the next
shell prompt (with a bell).

Switching the program to read the real stdin lets Bubble Tea consume its own
query responses (raw mode also disables echo). Because raw mode means Ctrl+C is
delivered as a keystroke rather than SIGINT, the model now handles Ctrl+C
explicitly and aborts the operation with exit code 130, matching the approach
used for `rad init` in #12540.

Fixes #12541
Fixes #12638

## How to test

1. Use a terminal that supports mode 2026/2027 or the Kitty keyboard protocol
   (recent iTerm2, WezTerm, Ghostty, or kitty).
2. Build the CLI: `make build-rad`.
3. Run `rad deploy app.bicep`, `rad delete <resource>`, and `rad app delete <app>`.
4. Confirm no stray characters (`2026;2$y2027;4$y1u`) appear at the shell prompt
   after the display exits and the bell no longer rings.
5. Start a deployment and press Ctrl+C; confirm the command aborts (exit code 130).
6. `go test ./pkg/cli/output/progress/...`

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/output/progress/progress.go` | Read the real stdin so Bubble Tea consumes its own capability-query responses; add a `ctrl+c` case to the model that marks it interrupted and quits; abort with `os.Exit(130)` on interrupt; update the `Run` doc comment. |
| `pkg/cli/output/progress/progress_test.go` | Add `Test_model_ctrlCMarksInterruptedAndQuits` covering the Ctrl+C interrupt handling. |


pkg/cli/output/progress/progress.go
```diff
@@ -65,16 +65,22 @@ type interactiveListener struct {
...
 func (l *interactiveListener) Run() {
-	program := tea.NewProgram(
-		newModel(l.progressChan),
-		tea.WithInput(strings.NewReader("")),
-		tea.WithoutSignalHandler(),
-	)
-	if _, err := program.Run(); err != nil {
+	program := tea.NewProgram(newModel(l.progressChan))
+	finalModel, err := program.Run()
+	if m, ok := finalModel.(*model); ok && m.interrupted {
```
lakshmimsft: Ctrl+C only works here when stdin is a terminal. Since  tea.WithoutSignalHandler()  was removed, Bubble Tea now catches SIGINT itself and returns  ErrInterrupted  —  m.interrupted  stays  false . That happens when stdin is piped instead of a terminal, and also on any  kill -INT . In that case we just print a warning and the deploy keeps going instead of stopping. Can we check the error too?
```
	m, _ := finalModel.(*model)
	if (m != nil && m.interrupted) || errors.Is(err, tea.ErrInterrupted) {
```

pkg/cli/output/progress/progress.go
```diff
@@ -65,16 +65,22 @@ type interactiveListener struct {
...
+		// The user pressed Ctrl+C during the deployment or deletion. The work runs
+		// in another goroutine that may be blocked in a call that cannot be
+		// canceled cooperatively, so abort the process the way Ctrl+C did before
+		// the display ran in raw mode. Bubble Tea has already restored the terminal
+		// by the time Run returns.
+		os.Exit(130) //nolint:forbidigo // Intentional: abort on Ctrl+C, matching rad init.
+	}
+	if err != nil {
```
lakshmimsft, line 85: We only drain the channel when there's an error, but Bubble Tea also exits with no error on SIGTERM. When that happens the channel is still open, and deploy/delete block forever once the 1-slot buffer fills — so  kill <pid>  hangs  rad deploy  instead of stopping it. Can we always drain, and only print the warning when there's an error?
suggestion:
```
		if err != nil {
		// Bubble Tea failed to start or exited before the channel was closed.
		fmt.Fprintf(os.Stderr, "Warning: progress display stopped: %v\n", err)
	}

	// The model is no longer reading from progressChan, so drain it here to keep
	// producers (deploy/delete) from blocking until it is closed....
    ...
```

lakshmimsft (approved):

PR #12640, opened 2026-08-10
Architecture-aware container image builds in deploy workflows

sylvainsf:
## Description

`Radius.Compute/containerImages` builds run in an in-cluster BuildKit that is compiled for the runner's architecture (amd64 on standard GitHub-hosted runners). When an app leaves `build.platforms` unset, the recipe defaults to a multi-arch build (`linux/amd64,linux/arm64`), so the arm64 half is produced under QEMU emulation, which is roughly an order of magnitude slower and prone to emulation crashes (see the analysis on radius-project/radius#12595). This PR lets the Azure and AWS deploy workflows build only the platform(s) the target cluster actually runs, while preserving multi-arch when it is genuinely needed.

This is the upstream half of the contract in radius-project/ai-extensions#300.

## Contract

Two template placeholders on the Azure and AWS `run-rad-commands` workflows, rendered by the extension:

| Placeholder | Extension default |
| --- | --- |
| `{{TARGET_CLUSTER_ARCH_MODE}}` | `${{ vars.RADIUS_BUILD_ARCH_MODE \|\| 'detect' }}` |
| `{{TARGET_CLUSTER_ARCH_FALLBACK_PLATFORMS}}` | `${{ vars.RADIUS_BUILD_PLATFORMS \|\| 'linux/amd64,linux/arm64' }}` |

## Behavior

Resolved by `compute-build-platforms.sh`, invoked by the shared `run-rad-commands` action after the target kubeconfig is configured and before the app is built/deployed:

| Mode | Cluster | Result |
| --- | --- | --- |
| `detect` | single-arch | build that one platform (no emulation) |
| `detect` | mixed / undetermined | fallback platform list |
| explicit list (contains `/`, e.g. `linux/amd64`) | n/a | honored verbatim, no detection |
| empty / unsubstituted placeholder | n/a | feature off, recipe default applies (existing behavior) |

The computed list is exported as `RADIUS_EFFECTIVE_BUILD_PLATFORMS` and injected as `--parameters platforms=<list>` only when the app declares a `platforms` parameter and it was not already supplied via `RADIUS_DEPLOY_PARAMS`, flowing through the same conditional path as the existing `app-image` parameter. Apps that do not declare `platforms`, and templates that do not render the placeholders, are unaffected.

**Mixed-arch is a first-class outcome, not an error.** A cluster with both amd64 and arm64 nodes resolves to the fallback multi-arch list so images stay portable; only single-arch clusters drop to a single platform to skip emulation.

## Design notes

- Kept the contract small and explicit. Recognized modes are `detect`, an explicit platform list, or empty/placeholder (off). An unrecognized keyword fails safe to the fallback list with a warning, documented in the workflow and script comments.
- Detection uses `kubectl get nodes -o jsonpath=...nodeInfo.architecture`; a failed probe degrades to empty, which the resolver treats as "undetermined" and maps to the fallback.
- No Radius runtime/CLI changes; this is entirely in the deploy workflow templates, the shared composite action, and a helper script.

## Testing

- `make test-build-platforms` — new `compute-build-platforms_test.sh`: mode/detection/override/fallback matrix (single-arch, mixed-arch, unknown-arch, x86_64/aarch64 aliases, normalization, unrecognized mode), plus assertions that both workflows carry the placeholders and wire the inputs, and that the action declares the inputs and runs the helper.
- `make test-run-rad-commands-action` — extended `deploy-parameters_test.sh`: `platforms` is injected only when declared, skipped when no effective list was computed, and never overrides a `RADIUS_DEPLOY_PARAMS`-supplied value.
- `shellcheck` clean on all shell; all three modified YAML files parse.

## Files

| File | Change |
| --- | --- |
| `.github/extension/actions/run-rad-commands/compute-build-platforms.sh` | New: resolves effective build platforms from mode/fallback/detected arches |
| `.github/extension/actions/run-rad-commands/compute-build-platforms_test.sh` | New: unit + wiring tests |
| `.github/extension/actions/run-rad-commands/action.yml` | New `build-arch-mode` / `build-fallback-platforms` inputs; detection step exporting `RADIUS_EFFECTIVE_BUILD_PLATFORMS` |
| `.github/extension/actions/run-rad-commands/deploy-parameters.sh` | Inject `platforms` when the app declares it and a list was computed |
| `.github/extension/actions/run-rad-commands/deploy-parameters_test.sh` | Coverage for the injection |
| `.github/extension/run-rad-commands-azure.yml` / `-aws.yml` | Placeholders + pass the two inputs to `run-rad-commands` |
| `build/test.mk` | `test-build-platforms` target, added to `test` |

## Related

- radius-project/ai-extensions#300
- radius-project/radius#12595

### DIFF
--- a/.github/extension/actions/run-rad-commands/action.yml
+++ b/.github/extension/actions/run-rad-commands/action.yml
@@ -35,6 +35,24 @@ inputs:
   registry-password:
     description: Password/token passed when the app declares the registryPassword parameter (feeds its Radius.Security/secrets registry Secret).
     required: true
+  build-arch-mode:
+    description: >-
+      Target-cluster architecture mode for Radius.Compute/containerImages builds.
+      'detect' probes the target cluster node architectures; a single-arch cluster
+      builds only that platform, a mixed or undetermined cluster falls back to
+      build-fallback-platforms. An explicit platform list (any value containing
+      '/', e.g. 'linux/amd64') is honored verbatim. Empty or an unsubstituted
+      '{{TARGET_CLUSTER_ARCH_MODE}}' placeholder disables the feature and preserves
+      the recipe's default platforms.
+    required: false
+    default: ""
+  build-fallback-platforms:
+    description: >-
+      Comma-separated platform list used when build-arch-mode is 'detect' and the
+      target cluster is mixed-arch or its architecture cannot be determined. Empty
+      defaults to 'linux/amd64,linux/arm64'.
+    required: false
+    default: ""
 
 runs:
   using: composite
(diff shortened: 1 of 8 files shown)


sylvainsf:
Thanks for the review. Addressed in bd0005c1.

- **`set -u` on the kubeconfig assignment (action.yml L106/L110):** Fixed. `TARGET_KUBECONFIG` now default-expands `${RADIUS_TARGET_KUBECONFIG:-}`, so an unset value no longer trips `set -u` before the empty case is handled; `detect` mode degrades cleanly to the fallback when no kubeconfig is present.
- **Malformed MODE contract header comment:** Fixed. Rewrote the block so the bullets and parenthesis are well-formed.
- **stderr capture in the test (L71):** This one is a false positive, so I left it. The idiom `"$(cmd 2>&1 >/dev/null)"` is the correct order for "capture stderr, discard stdout": `2>&1` first points fd2 at the command-substitution capture (currently fd1), then `>/dev/null` redirects fd1 only. Reversing to `>/dev/null 2>&1` would discard both. The test asserts the warning is captured and it passes, which confirms stderr is captured.

The `corerp-cloud` functional failure is unrelated to this change (this PR only touches the extension workflow YAML, the shared composite action, and a helper script; that lane does not exercise `run-rad-commands`). Rebasing on main will re-trigger it.

DariuszPorowski (dismissed):

brooke-hamilton (approved):

DariuszPorowski (approved):

PR #12676, opened 2026-08-13
Adds support to rad deploy for deploying a remote Bicep or JSON file over https

### LINKED ISSUE #12673: Support remote Bicep template URLs in `rad deploy`

Add support for using a remote Bicep template URL with `rad deploy`, specifically allowing commands such as `rad deploy https://github.com/radius-project/resource-types-contrib/files/recipe-packs/azure/aks-recipepack.bicep` to resolve and deploy the referenced file. The command should retrieve or otherwise resolve the remote file before compiling and deploying it, while preserving the existing behavior for local `.bicep` and `.json` files. `kubectl` already supports deploying resources from remote URLs today, so `rad deploy` should provide similar behavior for remote Bicep templates.

### Acceptance criteria

- The example HTTPS URL resolves and proceeds through Bicep compilation and deployment.
- Remote template resolution reports clear errors for invalid URLs, unavailable files, and download or access failures.
- Existing local `.bicep` and `.json` deployment workflows continue to work unchanged.
- Relative imports or other file dependencies used by the retrieved Bicep template are handled correctly, or are explicitly rejected with a clear error message.
- URL handling follows behavior users already expect from tools such as `kubectl`.

### Additional context

The current `rad deploy` implementation treats its positional template argument as a local filesystem path. The requested behavior is to accept a HTTPS URL directly. Example: `rad deploy https://github.com/radius-project/resource-types-contrib/files/recipe-packs/azure/aks-recipepack.bicep`.

zachcasper:
## Summary

Adds support for deploying a remote Bicep or ARM template by URL with `rad deploy`. The template argument may now be an `http(s)` URL in addition to a local path; the remote file is downloaded, then read (ARM JSON) or compiled (Bicep) and deployed, mirroring the behavior users expect from tools like `kubectl`. Existing local `.bicep`/`.json` workflows are unchanged.

Because Bicep discovers `bicepconfig.json` by walking up from the template's directory, a downloaded template placed in an isolated temp dir would fail to resolve extension declarations such as `extension radius`. To handle this, a `bicepconfig.json` is written next to the downloaded template — reusing the nearest one found by searching upward from the working directory, and otherwise falling back to the default Radius extensions config.

Remote downloads are bounded by a maximum size to avoid unbounded memory use from a large or malicious response, and remote read/parse failures include the original URL for clearer diagnostics.

## Reason for change

`rad deploy` previously treated its template argument as a local filesystem path only. Users want to deploy templates directly from a URL (e.g. a recipe pack in `resource-types-contrib`) without downloading them first.

Fixes #12673

## How to test

Unit tests:

    go test ./pkg/cli/bicep/... ./pkg/cli/setup/... ./pkg/cli/cmd/deploy/...

Manual:

1. `rad deploy https://<host>/template.bicep` — resolves, compiles, and deploys.
2. Run from a directory with a valid `bicepconfig.json` to reuse it; run from a directory without one to confirm the default Radius config is used.
3. `rad deploy https://<host>/template.json` — remote ARM JSON works.
4. `rad deploy ./local.bicep` and `./local.json` — local behavior unchanged.
5. Error cases each report a clear error: 404 URL, non-`.bicep`/`.json` URL, unreachable host, a template exceeding the maximum download size, and invalid remote JSON (error names the URL).

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/bicep/types.go` | Extend `PrepareTemplate` to accept `http(s)` URLs: download the template to a temp file (size-bounded via `io.LimitReader`), write a `bicepconfig.json` alongside it so extensions resolve, then read/compile and clean up. Remote read/parse errors are wrapped with the source URL. Adds helpers `isRemoteURL`, `downloadTemplate`, `writeBicepConfig`, and `findBicepConfig`. |
| `pkg/cli/cmd/deploy/deploy.go` | Document remote-URL support in the command long help and add a remote-URL example. |
| `pkg/cli/setup/application.go` | Export `getVersionedBicepConfig` as `GetVersionedBicepConfig` so the bicep package can reuse the default config generator. |
| `pkg/cli/bicep/types_test.go` | New unit tests (using `httptest`) covering URL detection, download success/404/unsupported-extension/connection-failure, max-size enforcement, `bicepconfig.json` discovery/copy/fallback, filesystem-error branches (`MkdirTemp`/`WriteFile`/`ReadFile`), remote ARM JSON handling, and remote JSON errors including the URL. |


pkg/cli/bicep/types.go
```diff
@@ -88,6 +125,122 @@ func (i *Impl) PrepareTemplate(filePath string) (map[string]any, error) {
 	return template, nil
 }
 
+// isRemoteURL reports whether filePath is an http or https URL. Local paths, including Windows
+// paths such as C:\foo.bicep, are not treated as remote URLs.
+func isRemoteURL(filePath string) bool {
+	parsed, err := url.Parse(filePath)
+	if err != nil {
```
brooke-hamilton, line 157: Detect the `http://`/`https://` prefix before parsing and return an invalid-URL error when parsing fails. As written, a malformed input such as `https://[::1/app.bicep` makes this function return false, so `PrepareTemplate` treats it as a local `.bicep` path and reports a misleading file-not-found error. Kubectl avoids that fallback by classifying the prefix first and then surfacing `url.Parse` failures. Please also reject parsed HTTP(S) URLs with no host and add malformed/no-host cases.

brooke-hamilton (commented):
🚀

pkg/cli/bicep/types.go
```diff
@@ -88,6 +125,122 @@ func (i *Impl) PrepareTemplate(filePath string) (map[string]any, error) {
...
+		return "", nil, fmt.Errorf("invalid template URL %q: %w", templateURL, err)
+	}
+
+	ext := path.Ext(parsed.Path)
+	if !strings.EqualFold(ext, ".bicep") && !strings.EqualFold(ext, ".json") {
+		return "", nil, fmt.Errorf("the provided URL %q must reference a .json or .bicep file", templateURL)
+	}
+
+	i.Output.LogInfo("Downloading template from %s...", templateURL)
```
brooke-hamilton: Redact URL credentials before logging or formatting errors. Private template downloads commonly use signed query parameters, and Go also supports Basic auth through URL userinfo; printing `templateURL` verbatim exposes those credentials in terminal and CI logs. Keep the full URL for the request, but use a display copy with userinfo and sensitive query values removed or redacted everywhere the URL is surfaced.

pkg/cli/bicep/types.go
```diff
@@ -88,6 +125,122 @@ func (i *Impl) PrepareTemplate(filePath string) (map[string]any, error) {
...
+
+	client := &http.Client{Timeout: remoteTemplateTimeout}
```
brooke-hamilton: Make this HTTP client injectable and build the request with the command context rather than constructing a client inside the download. That allows cancellation before the 60-second timeout and makes timeout/redirect behavior deterministic to test. Please also define `CheckRedirect` explicitly: direct `http://` input can remain supported for kubectl compatibility, but an `https://` request should not silently downgrade to HTTP during a redirect.

pkg/cli/bicep/types.go
```diff
@@ -88,6 +125,122 @@ func (i *Impl) PrepareTemplate(filePath string) (map[string]any, error) {
...
+	resp, err := client.Get(templateURL)
```
brooke-hamilton: Add a bounded retry policy for transient downloads. Kubectl retries transport errors and 5xx responses up to three attempts; a single network interruption currently makes `rad deploy` fail immediately. Retry only safe transient cases within the existing overall timeout, avoid permanent 4xx responses, and consider honoring `Retry-After` for 429 responses.

pkg/cli/bicep/types.go
```diff
@@ -88,6 +125,122 @@ func (i *Impl) PrepareTemplate(filePath string) (map[string]any, error) {
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to download template from %q: %w", templateURL, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
brooke-hamilton: Reject an explicit `text/html` response before writing it as a template. GitHub page URLs can return HTML with `200 OK`, which currently reaches the Bicep/JSON parser and produces a misleading compiler error; users need the raw-content URL instead. Avoid a strict allowlist because valid Bicep downloads commonly use `text/plain` or `application/octet-stream`, but HTML is unambiguously not a deployable template.

pkg/cli/bicep/types.go
zachcasper: Done in c09d31b28. isRemoteURL now classifies by the http:///https:// prefix (case-insensitive) instead of url.Parse, so a malformed URL is still routed to remote handling. downloadTemplate then rejects url.Parse failures (invalid template URL, with the raw reason extracted so no credentials leak) and empty-host URLs (missing host). Added tests: Test_downloadTemplate_MalformedURL, Test_downloadTemplate_MissingHost, and Test_PrepareTemplate_MalformedURLNotTreatedAsLocal (asserts no "could not find file" fallback).

pkg/cli/bicep/types.go
zachcasper: Done in c09d31b28. Added redactURL, which strips userinfo and redacts all query-parameter values; it's applied to a display string used for the download log line and every error message. urlParseReason also strips the raw URL that url.Error embeds. The full URL is still used for the request. Tests: Test_redactURL and Test_downloadTemplate_RedactsCredentials (asserts basic-auth password and a sig query value never appear).

pkg/cli/bicep/types.go
zachcasper: Done in c09d31b28. Added an injectable Impl.HTTPClient (defaults to a client that sets an explicit CheckRedirect refusing an https→http downgrade while allowing http→http). Threaded context.Context through the Bicep.PrepareTemplate interface and all callers, and the request is built with http.NewRequestWithContext, so cancellation works before the overall timeout. Tests: Test_newHTTPClient_RedirectPolicy and Test_downloadTemplate_ContextCancelled.

pkg/cli/bicep/types.go
zachcasper: Done in c09d31b28. fetchRemoteTemplate retries transient failures (transport errors, 5xx, and 429) up to 3 attempts within the overall timeout, honoring Retry-After (parseRetryAfter handles delay-seconds and HTTP-date). Permanent 4xx responses are not retried. Tests: Test_downloadTemplate_RetriesTransientFailures, Test_downloadTemplate_NoRetryOn4xx, Test_parseRetryAfter.

pkg/cli/bicep/types.go
zachcasper: Done in c09d31b28. attemptDownload parses the Content-Type and rejects text/html with a message pointing at the raw-file URL, before the body reaches the Bicep/JSON parser. No allowlist, so text/plain and application/octet-stream still work. Test: Test_downloadTemplate_RejectsHTML.

brooke-hamilton (approved):

PR #12697, opened 2026-08-17
Fix samples tutorial functional test for modernized Radius.* demo sample

sk593:
## Root cause

The repo-wide `Functional Tests - samples-noncloud` job is failing on `main` since the samples repo was updated [radius-project/samples#2645](https://github.com/radius-project/samples/pull/2645) ("Modernize demo sample to Radius.* resource types").

radius CI checks out the samples repo at `refs/heads/edge` (`.github/workflows/functional-test-noncloud.yaml`). That samples PR merged into `edge`, so CI now pulls a modernized `samples/demo/app.bicep` that:

- uses `Radius.Core/applications` + `Radius.Compute/containers` at `2025-08-01-preview` (was `Applications.Core`),
- has **no** `application` parameter (only `environment`),
- names the app/container `demo-${environmentName}` → `demo-tutorial` for the `tutorial` environment,
- removes the inline `Applications.Datastores/redisCaches` resource (redis now lives in a separate `app-redis.bicep` this test does not deploy).

The consuming `Test_FirstApplicationSample` still used the legacy `Applications.Core` types and app name `demo`, and passed `--application`, producing a deploy 404 (container created under `Radius.Compute/containers/demo-tutorial`, app looked up under `Applications.Core/applications/demo-tutorial`).

## Changes

**`testdata/tutorial-environment.bicep`**
- Modernized to `Radius.Core/environments@2025-08-01-preview` named `tutorial` with `properties.providers.kubernetes.namespace: 'tutorial'`.
- Removed the redis recipe registration and the now-unused `registry`/`version` params. `rad deploy` injects the default recipe pack, which supplies the `Radius.Compute/containers` recipe, so no explicit pack is needed.

**`tutorial_test.go` (`Test_FirstApplicationSample`)**
- Env-deploy executor no longer passes `registry`/`version` bicep params.
- App-deploy step drops `.WithApplication(...)`; keeps `.WithEnvironment("tutorial")`.
- Expected `RPResources` updated: application `demo-tutorial` (`radius.core/applications`) and container `demo-tutorial` (`radius.compute/containers`); removed the redis `db` expectation.
- Pod port-forward + K8s validation now target the `demo-tutorial` resource selector in the environment namespace `tutorial` (recipe-driven `Radius.Compute/containers` pods land in the env's Kubernetes namespace, matching existing modern container tests).

This unblocks the repo-wide `samples-noncloud` functional test failure.

## Validation

- `go build ./test/...` passes.
- `go vet ./test/functional-portable/samples/...` passes.
- End-to-end run requires a kind cluster with Radius installed and `RADIUS_SAMPLES_REPO_ROOT` pointing at a samples `edge` checkout; the logic mirrors the modernized sample and existing modern `Radius.Compute/containers` functional tests.

### DIFF
--- a/test/functional-portable/samples/noncloud/tutorial_test.go
+++ b/test/functional-portable/samples/noncloud/tutorial_test.go
@@ -29,6 +29,7 @@ import (
 	"testing"
 	"time"
 
+	clikubernetes "github.com/radius-project/radius/pkg/cli/kubernetes"
 	"github.com/radius-project/radius/pkg/kubernetes"
 	"github.com/radius-project/radius/test/rp"
 	"github.com/radius-project/radius/test/step"
@@ -118,6 +119,16 @@ func Test_FirstApplicationSample(t *testing.T) {
 			},
 		},
 	})
+
+	// Radius.Core/environments rejects a Kubernetes namespace that does not already
+	// exist. RPTest.CreateInitialResources only ensures the namespace named after the
+	// test (demo-tutorial), so the environment's "tutorial" namespace must be created
+	// here before the steps run, mirroring NewPreviewEnvPreSetup.
+	test.PreSetup = func(ctx context.Context, t *testing.T, ct rp.RPTest) {
+		nsClient, err := rp.DeploymentTargetK8sClient(ct.Options)
+		require.NoError(t, err)
+		require.NoError(t, clikubernetes.EnsureNamespace(ctx, nsClient, appNamespace))
+	}
 
 	test.Test(t)
 }
--- a/test/functional-portable/samples/noncloud/tutorial_test.go
+++ b/test/functional-portable/samples/noncloud/tutorial_test.go
@@ -96,6 +96,13 @@ func Test_FirstApplicationSample(t *testing.T) {
 						Type: validation.ComputeContainersResource,
 						App:  appName,
 					},
+					{
+						// The environment-deploy step skips validation and has no RPResources,
+						// so include the fixed-name tutorial environment here to validate it and
+						// ensure the cleanup loop deletes it after the application.
+						Name: "tutorial",
+						Type: validation.CoreEnvironmentsResource,
+					},
 				},
 			},
 			PostStepVerify: func(ctx context.Context, t *testing.T, ct rp.RPTest) {
--- a/test/functional-portable/samples/noncloud/tutorial_test.go
+++ b/test/functional-portable/samples/noncloud/tutorial_test.go
@@ -61,31 +61,33 @@ func Test_FirstApplicationSample(t *testing.T) {
 	relPathSamplesRepo, err := filepath.Rel(cwd, samplesRepoAbsPath)
 	require.NoError(t, err)
 	template := filepath.Join(relPathSamplesRepo, "samples/demo/app.bicep")
-	appName := "demo"
-	appNamespace := "tutorial-demo"
+	// The modernized sample (radius-project/samples#2645) names both the application
+	// and the container demo-${environmentName}, which resolves to demo-tutorial for the
+	// "tutorial" environment. The recipe-driven Radius.Compute/containers pod lands in the
+	// environment's Kubernetes namespace ("tutorial").
+	appName := "demo-tutorial"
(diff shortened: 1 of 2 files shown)


zachcasper (dismissed):

test/functional-portable/samples/noncloud/tutorial_test.go
```diff
@@ -61,39 +61,41 @@ func Test_FirstApplicationSample(t *testing.T) {
...
 	require.NoError(t, err)
 	template := filepath.Join(relPathSamplesRepo, "samples/demo/app.bicep")
-	appName := "demo"
-	appNamespace := "tutorial-demo"
+	// The modernized sample (radius-project/samples#2645) names both the application
+	// and the container demo-${environmentName}, which resolves to demo-tutorial for the
+	// "tutorial" environment. The recipe-driven Radius.Compute/containers pod lands in the
+	// environment's Kubernetes namespace ("tutorial").
+	appName := "demo-tutorial"
+	appNamespace := "tutorial"
```
brooke-hamilton, line 76: The `tutorial` namespace is never created, so step 1 will fail on a clean cluster.

`Radius.Core/environments` validates that the Kubernetes namespace already exists and returns 400 otherwise (`pkg/corerp/frontend/controller/environments/v20250801preview/createorupdateenvironment.go:117-121`, `"Namespace '%s' does not exist in the Kubernetes cluster"`). The old `Applications.Core/environments@2023-10-01-preview` type had no such check, which is why `namespace: 'tutorial'` worked before.

`RPTest.CreateInitialResources` only ensures the namespace named `ct.Name` (`test/rp/rptest.go:339`), and `ct.Name` is now `demo-tutorial`, not `tutorial`. So `rad deploy testdata/tutorial-environment.bicep` will be rejected before anything else runs. Note the functional tests were skipped on this PR, so CI hasn't exercised this path.

Every other modernized test handles this explicitly — `NewPreviewEnvPreSetup` calls `kubernetes.EnsureNamespace` before creating the env (`test/rp/rptest.go:263-269`), and `cli_test.go:566` does the same. Add a `PreSetup` here that ensures `tutorial` exists before the steps run:

```go
test.PreSetup = func(ctx context.Context, t *testing.T, ct rp.RPTest) {
	nsClient, err := rp.DeploymentTargetK8sClient(ct.Options)
	require.NoError(t, err)
	require.NoError(t, kubernetes.EnsureNamespace(ctx, nsClient, appNamespace))
}
```

brooke-hamilton (changes requested):
🚀

test/functional-portable/samples/noncloud/tutorial_test.go
sk593: Fixed in `ddfbb9cfe` — added a `PreSetup` that ensures the `tutorial` namespace exists before the steps run.

brooke-hamilton (dismissed):

test/functional-portable/samples/noncloud/tutorial_test.go
```diff
@@ -61,39 +62,41 @@ func Test_FirstApplicationSample(t *testing.T) {
...
-						Type: validation.ContainersResource,
+						Type: validation.CoreApplicationsResource,
 					},
 					{
-						Name: "db",
-						Type: validation.RedisCachesResource,
+						Name: appName,
+						Type: validation.ComputeContainersResource,
+						App:  appName,
 					},
```
kachawla, line 98: Could we include the `tutorial` environment in this step’s `RPResources`, after the application and container?

The environment-deploy step skips resource validation and has no `RPResources`, so the cleanup loop never deletes the fixed-name `Radius.Core/environments/tutorial` resource. Adding it here as `{Name: "tutorial", Type: validation.CoreEnvironmentsResource}` would validate it and preserve the correct cleanup order: application first, then environment.

sk593: Fixed in `cb2fb8c86` — added `{Name: "tutorial", Type: validation.CoreEnvironmentsResource}` to the app-deploy step's `RPResources` after the application and container.

lakshmimsft (approved):

PR #12698, opened 2026-08-17
fix(cli): apply URL redaction to all remote template display paths  

lakshmimsft:
## Summary
Applies the existing URL redaction to every place `rad` displays a template argument, so credentials
embedded in a remote template URL are no longer written to the terminal, to CI logs, or into
generated files.

pr #12676 added remote-template support (`rad deploy https://host/app.bicep`) along with a `redactURL`
helper, because private template URLs commonly carry an Azure SAS token as a signed query parameter
or basic-auth userinfo. That redaction was applied inside the download path only. The raw URL was
still formatted into the `Building %s...` step, several error messages, the `rad deploy` progress
text and failure error, the `rad bicep publish` messages, and the `rad app graph` compile output.

This change introduces `bicep.RedactTemplatePath` (a no-op for local paths, redacting for `http(s)`
URLs) and routes every display site through it, plus three related leaks found while wiring it up:

- `rad bicep generate-kubernetes-manifest` derived its output file name with `filepath.Base`/`Ext`
  on the raw URL. Because `filepath.Ext` takes the last dot, a query value containing a dot survived
  into a real file name on disk and into the log line: `app.bicep?sig=abc.yaml`. A new
  `bicep.TemplateFileName` parses the URL and drops the query and fragment entirely, so a file name
  is never derived from credential text.
- A download transport failure returned a `*url.Error`, whose message embeds the raw request URL.
  The redacted URL and the credential were printed side by side:
  `failed to download template from "…?sig=redacted": Get "…?sig=TOPSECRET": dial tcp: …`.
  Both transport and body-read errors now unwrap through the existing `urlParseReason` helper.
- `redactURL` preserved the URL fragment. It is never needed to fetch a template, so it is dropped.

No behavior changes for local file paths: `isRemoteURL` matches only `http://` and `https://`
prefixes, so local paths (including Windows `C:\...` paths and names containing `?`) pass through
untouched.

## Reason for change
A SAS token or basic-auth credential supplied to `rad deploy` was echoed to stdout and into CI job
logs. In CI the URL is typically injected from a secret or environment variable, so the CLI printing
it is often the first time the value is written to a retained log, and GitHub Actions secret masking
does not reliably cover dynamically generated SAS URLs. The redaction control added in #12676 was
intended to prevent exactly this, but was only wired into part of the code path.

Fixes: follow up to pr #12676 

## How to test

1. `rad deploy 'https://<host>/app.bicep?sig=SECRET'` — the build step, progress text, and any
   deployment failure show `sig=redacted`, never the token.
2. `rad deploy 'https://<unreachable-host>/app.bicep?sig=SECRET'` — the connection error is reported
   without the token.
3. `rad bicep generate-kubernetes-manifest 'https://<host>/app.bicep?sig=a.b'` — the generated file
   is `app.yaml`, and neither the file name nor the manifest contents contain the token.
4. `rad bicep publish` and `rad app graph` with a credentialed URL — all messages are redacted.
5. `rad deploy ./app.bicep` and `./app.json` — local output is unchanged.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/bicep/types.go` | Add exported `RedactTemplatePath` and `TemplateFileName`; use a redacted `displayPath` at all four display sites in `PrepareTemplate`; drop the fragment in `redactURL`; unwrap `*url.Error` via `urlParseReason` on the transport and body-read error paths. |
| `pkg/cli/cmd/deploy/deploy.go` | Redact the template path in both progress-text branches and in the deployment-failure error. |
| `pkg/cli/cmd/bicep/publish/publish.go` | Redact the file argument in the prepare error, both publish errors, and the success log. |
| `pkg/cli/cmd/app/graph/graph.go`, `pkg/cli/cmd/app/graph/preview/graph.go` | Redact the path in the `Compiling %s` log and the compile-failure error. |
| `pkg/cli/cmd/bicep/generatekubernetesmanifest/generatekubernetesmanifest.go` | Derive the destination file name and the manifest template name from `bicep.TemplateFileName` so URL query text never reaches a file name or the generated YAML. |
| `pkg/cli/bicep/types_test.go` | Add `Test_RedactTemplatePath`, `Test_TemplateFileName`, `Test_PrepareTemplate_RemoteErrorRedactsCredentials`, `Test_downloadTemplate_TransportErrorRedactsCredentials`, and a fragment case. |
| `pkg/cli/cmd/deploy/deploy_test.go` | Add a `Test_Run` subtest asserting the captured `deploy.Options.ProgressText` contains `sig=redacted` and not the token. |

### DIFF
--- a/pkg/cli/bicep/types.go
+++ b/pkg/cli/bicep/types.go
@@ -78,7 +78,9 @@ type Impl struct {
 func (i *Impl) PrepareTemplate(ctx context.Context, filePath string) (map[string]any, error) {
 	// A remote URL is downloaded to a temporary local file so it can be read or compiled like a
 	// local template. This mirrors the behavior users expect from tools such as kubectl.
-	originalPath := filePath
+	// displayPath is the only form of the template argument that may be shown to the user, so a
+	// credential embedded in a remote URL is never written to the terminal or to CI logs.
+	displayPath := RedactTemplatePath(filePath)
 	remote := isRemoteURL(filePath)
 	if remote {
 		localPath, cleanup, err := i.downloadTemplate(ctx, filePath)
@@ -92,11 +94,11 @@ func (i *Impl) PrepareTemplate(ctx context.Context, filePath string) (map[string
 	if strings.EqualFold(path.Ext(filePath), ".json") {
 		template, err := ReadARMJSON(filePath)
 		if err != nil && remote {
-			return nil, fmt.Errorf("failed to read remote template %q: %w", originalPath, err)
+			return nil, fmt.Errorf("failed to read remote template %q: %w", displayPath, err)
 		}
 		return template, err
 	} else if !strings.EqualFold(path.Ext(filePath), ".bicep") {
-		return nil, fmt.Errorf("the provided file %q must be a .json or .bicep file", originalPath)
+		return nil, fmt.Errorf("the provided file %q must be a .json or .bicep file", displayPath)
 	}
 
 	ok, err := IsBicepInstalled()
@@ -118,14 +120,14 @@ func (i *Impl) PrepareTemplate(ctx context.Context, filePath string) (map[string
 		return nil, fmt.Errorf("could not find file: %w", err)
 	}
 
-	step := i.Output.BeginStep("Building %s...", originalPath)
+	step := i.Output.BeginStep("Building %s...", displayPath)
 	bytes, err := i.Call("build", "--stdout", filePath)
 	if err != nil {
 		i.Output.CompleteStep(step)
 		if remote {
 			// The bicep compiler prints detailed diagnostics to stderr, so keep the wrapper
 			// error focused on identifying the remote source rather than guessing the cause.
-			return nil, fmt.Errorf("failed to build remote template %q: %w", originalPath, err)
+			return nil, fmt.Errorf("failed to build remote template %q: %w", displayPath, err)
 		}
 		return nil, fmt.Errorf("failed to build template: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


pkg/cli/bicep/types_test.go
```diff
@@ -457,6 +457,114 @@ func Test_redactURL(t *testing.T) {
...
+
+// A transport failure returns a *url.Error whose message embeds the raw request URL, so the
+// wrapped error must not carry the credential that the redacted display string removed.
+func Test_downloadTemplate_TransportErrorRedactsCredentials(t *testing.T) {
+	i := newTestImpl()
+	_, _, err := i.downloadTemplate(t.Context(), "https://nonexistent.invalid/app.bicep?sig=TOPSECRET")
+	require.Error(t, err)
+	require.NotContains(t, err.Error(), "TOPSECRET")
+	require.Contains(t, err.Error(), "sig=redacted")
+}
```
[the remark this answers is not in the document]
lakshmimsft: updated.

pkg/cli/bicep/types.go
```diff
@@ -167,11 +181,31 @@ func redactURL(raw string) string {
...
+// credential in it).
+func TemplateFileName(templatePath string) string {
+	if !isRemoteURL(templatePath) {
+		return filepath.Base(templatePath)
+	}
+	parsed, err := url.Parse(templatePath)
+	if err != nil {
+		return "template"
+	}
+	return path.Base(parsed.Path)
```
brooke-hamilton: This can fail on Windows because `TemplateFileName` may return a path containing backslashes, which Windows treats as directory separators. URL parsing decodes `%5C` into `\`, while `path.Base` recognizes only `/`. For example, a URL path containing `folder%5Capp.bicep` produces `folder\app.bicep` rather than `app.bicep`; `generate-kubernetes-manifest` then uses that result as its default output path and may write to an unexpected directory or fail if the directory does not exist. Normalize or reject backslashes before extracting the filename, and add a test with an encoded backslash.

lakshmimsft: updated

brooke-hamilton (approved):

PR #12703, opened 2026-08-18
Add advisory PR commit signature verification

brooke-hamilton:
## Summary

Add an advisory pull request workflow that checks every commit for a GitHub **Verified** cryptographic signature. A pinned `actions/github-script` step loads the trusted `.github/scripts/verify-commit-signatures.mjs` module, which paginates the commits, reports the result, and creates or updates a polite guidance comment that remains after all commits are verified. Maintainers can also run the workflow manually by providing a pull request number.

## Reason for change

New contributors are asked to use cryptographically signed commits, so we want to give an early indication that it is required via a comment on the PR. Automating that reminder makes the cryptographic signing requirement visible after each push while keeping it separate from the existing DCO `Signed-off-by` requirement.

The workflow remains non-blocking for unsigned commits. API or workflow failures still fail the job so broken verification remains visible.

> NOTE: When a user updates their commits to add signing and force pushes to the PR branch, the advisory message will not be deleted from the PR. This is by design so that PR authors can quote-reply the message to give status.

## Example message

Thank you for contributing to Radius.

The following commits do not currently show a GitHub **Verified** signature:

- [`19c7a59`](https://github.com/brooke-hamilton/radius/commit/19c7a59523fb76c3ce15ad640b90f90f8d2767b9)
- [`2a80f28`](https://github.com/brooke-hamilton/radius/commit/2a80f28dcfddc43e63571c1aadda0fa70cf8de10)

Please follow our [commit-signing guide](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-first-commit/first-commit-06-creating-a-pr/index.md#signing-your-commits) to configure GPG, SSH, or S/MIME signing. Then re-sign the affected commits and force-push the rewritten branch with `--force-with-lease`.

Cryptographic commit signing is separate from the DCO `Signed-off-by` line; both are required.


.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,106 @@
...
+            echo
+            printf '%s%s\n' \
+              "Cryptographic commit signing is separate from the DCO " \
+              "\`Signed-off-by\` line; both are required."
+          } > commit-signature-guidance.md
+
+          printf '%s%s%s\n' \
+            "::warning title=Unverified commit signatures::" \
+            "${UNVERIFIED_COUNT}" \
+            " commit(s) do not show a Verified signature."
```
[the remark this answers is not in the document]
brooke-hamilton: Fixed — removed the redundant `::warning` workflow command that printed a value derived from PR data to stdout. The preceding verification step still reports the count and short SHAs as ordinary log output.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Prepare signature guidance
+        if: steps.verify.outputs.count != '0'
+        env:
+          UNVERIFIED_COMMITS: ${{ steps.verify.outputs.unverified_commits }}
+        run: |
+          pr_author="$(
+            gh api "/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}" \
+              --jq '.user.login'
+          )"
```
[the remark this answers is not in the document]
brooke-hamilton: Won't fix

.github/scripts/verify-commit-signatures.sh
```diff
@@ -0,0 +1,115 @@
...
+        <<< "${response}" > /dev/null; then
+        echo "Error: pull request ${PR_NUMBER} was not found" >&2
+        return 1
+    fi
+
+    jq -c '
+      [
+        .[] |
+        .data.repository.pullRequest.commits.nodes[] |
+        select(
```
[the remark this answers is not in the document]
brooke-hamilton: Won't fix - this is a manually verified script

.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,101 @@
...
+            jq -r \
+              --arg origin "${GITHUB_SERVER_URL}" \
+              --arg repo "${GITHUB_REPOSITORY}" \
+              '.[] | "- [`\(.[:7])`](\($origin)/\($repo)/commit/\(.))"' \
+              <<< "${UNVERIFIED_COMMITS}"
+            echo
+            printf '%s%s%s\n' \
+              "Please follow our [commit-signing guide](${guide_url}) to " \
+              "configure GPG, SSH, or S/MIME signing. Then re-sign the affected " \
+              "commits and force-push the rewritten branch with \`--force-with-lease\`."
```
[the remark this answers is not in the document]
brooke-hamilton: Won't fix - the guidance is in the PR comment.

.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,101 @@
...
+            echo
+            printf '%s%s\n' \
+              "Cryptographic commit signing is separate from the DCO " \
+              "\`Signed-off-by\` line; both are required."
+          } > commit-signature-guidance.md
```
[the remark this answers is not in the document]
brooke-hamilton: Fixed — unverified SHAs now stay in a normal JSON file and are validated as an array of 40-character lowercase hex values before use. Only the literal `has_unverified=true|false` crosses `GITHUB_OUTPUT`, and the guidance step reads the validated file directly.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

.github/workflows/verify-commit-signatures.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Checkout signature verifier
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.base.sha || github.sha }}
+          sparse-checkout: |
+            .github/scripts/verify-commit-signatures.sh
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+
+      - name: Verify commit signatures
```
DariuszPorowski, line 40: idea: just thinking if native actions/github-script wouldn't it be a lighter solution with less code to maintain?

brooke-hamilton: Implemented — replaced the Bash verifier, checkout, `jq` steps, and sticky-comment action with one pinned `actions/github-script` step. It paginates all commits through GraphQL, reports the result, and creates or updates the retained guidance comment.

.github/scripts/verify-commit-signatures.mjs
```diff
@@ -0,0 +1,189 @@
...
+    "Cryptographic commit signing is separate from the DCO " +
+      "`Signed-off-by` line; both are required.",
+    "",
+    COMMENT_MARKER,
+  ].join("\n");
+}
+
+/** @param {import('@actions/github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+export default async ({ github, context, core }) => {
+  const prNumber = Number(process.env.PR_NUMBER);
```
DariuszPorowski: suggestion: map the workflow value to `INPUT_PR_NUMBER` and read it with `core.getInput("PR_NUMBER", { required: true })`

.github/scripts/verify-commit-signatures.mjs
```diff
@@ -0,0 +1,189 @@
...
+    "",
+    "Cryptographic commit signing is separate from the DCO " +
+      "`Signed-off-by` line; both are required.",
+    "",
+    COMMENT_MARKER,
+  ].join("\n");
+}
+
+/** @param {import('@actions/github-script').AsyncFunctionArguments} AsyncFunctionArguments */
+export default async ({ github, context, core }) => {
```
DariuszPorowski, line 133: Yes. Add one top-level try/catch, call `core.setFailed` with the message, and let helper functions throw to stop unsafe execution:

```
export default async ({ github, context, core }) => {
  try {
    // Existing verifier logic
  } catch (error) {
    core.setFailed(
      error instanceof Error ? error.message : `Unexpected error: ${String(error)}`
    );
  }
};
```

This produces a clean PR annotation without a stack-heavy "Unhandled error." Do not replace helper throws directly because `core.setFailed()` marks failure but does not stop execution.

DariuszPorowski (changes requested):

.github/scripts/verify-commit-signatures.mjs
brooke-hamilton: Implemented — the workflow now maps the PR number to `INPUT_PR_NUMBER`, and the module reads it with `core.getInput("PR_NUMBER", { required: true })` before validating it as a positive integer.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

.github/scripts/verify-commit-signatures.mjs
brooke-hamilton: Implemented — wrapped the default action entry point in one top-level `try/catch`. Helpers still throw immediately, and the catch reports a clean message through `core.setFailed`, including a safe fallback for non-Error values.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

DariuszPorowski (approved):

PR #12718, opened 2026-08-19
Enforce K8s namespace uniqueness for Radius.Core/environments

### LINKED ISSUE #12420: `Radius.Core/environments` incorrectly restricts a Kubernetes namespace to a single environment
### Steps to reproduce

1. Confirm there is a single resource group with a `Radius.Core/environments` using the `default` Kubernetes namespace:
   ```
   ❯ rad group list
   GROUP     ID
   default   /planes/radius/local/resourcegroups/default

   ❯ rad env list --preview
   RESOURCE  TYPE                      GROUP     STATE
   default   Radius.Core/environments  default   Succeeded

   ❯ rad env show default --preview
   RESOURCE  TYPE                      GROUP     STATE
   default   Radius.Core/environments  default   Succeeded

   PROVIDER    PROPERTIES
   kubernetes  namespace: 'default'
   ```
2. Attempt to create a second environment in the same resource group without specifying a namespace:
   ```
   ❯ rad env create test --preview
   Error: {
     "code": "Conflict",
     "message": "Environment /planes/radius/local/resourcegroups/default/providers/Radius.Core/environments/default with the same namespace (default) already exists"
   }
   ```
3. Create a new resource group and create an environment in it without specifying a namespace:
   ```
   ❯ rad group create test
   System.Resources/resourceGroups/test created

   ❯ rad env create test --preview -g test
   Radius.Core/environments/test created
   ```
4. Observe that the new environment was created successfully, despite sharing the `default` Kubernetes namespace with the first environment.

### Observed behavior

The one-environment-per-Kubernetes-namespace constraint is enforced within a single resource group, but **not** across resource groups. Creating a second resource group and creating an environment in it without specifying `--kubernetes-namespace` silently succeeds and defaults to the `default` namespace — the same namespace already occupied by the first environment.

This makes the constraint trivially bypassable: the same operation that is blocked in one resource group is silently permitted in another, with no warning that the namespace is already in use.

### Desired behavior

A deliberate, documented policy decision should be made and consistently enforced. The recommended approach is:

**Option A (recommended) — Remove the restriction:** Allow multiple environments per namespace, but require that recipes generate unique resource names (e.g., scoped by application or environment name) to avoid collisions. This accommodates teams that run multiple applications in a single namespace and aligns with common Kubernetes practices. Apply this policy uniformly — the current namespace uniqueness check should be removed entirely.

**Option B (alternative) — Enforce the restriction globally:** No two environments may share a Kubernetes namespace, regardless of which resource group they belong to. The namespace uniqueness check must be cluster-wide, not resource-group-scoped.

Either way, the current behavior — where the constraint is enforced within a resource group but silently bypassed across resource groups — is incorrect and must be fixed.

### Workaround

Explicitly specify a unique Kubernetes namespace when creating each environment: `rad env create <name> --preview -g <group> --kubernetes-namespace <unique-namespace>`

### rad Version

CLI Version Information:
0.59.0    v0.59.0   0.42.1    2bf2c25fcdde20d4cba1371618829bbbe1f9a997

### Operating system

_No response_

### Additional context

_No response_

lakshmimsft:
## Summary
Enforces the "one environment per Kubernetes namespace" constraint consistently for `Radius.Core/environments`, and makes the namespace immutable once set.

Previously the check was scoped to a single Radius resource group, so creating an environment in a *different* resource group silently bypassed it. Two environments then deployed into the same namespace and their Kubernetes resources collided — deploying the same application to both left only one Deployment and Service on the cluster.

The new rules for `Radius.Core/environments`:

| Operation | Behavior |
| --- | --- |
| Create | Namespace must not be claimed by any other environment |
| Update, namespace unchanged | No uniqueness check runs at all |
| Update, namespace changed | Rejected — `400 NamespaceImmutable` |
| Update, Kubernetes provider removed | Rejected — `400 NamespaceImmutable` |
| Update, claiming a namespace for the first time | Uniqueness check runs |

Immutability is what makes this safe to ship: because the namespace cannot change after create, an update that leaves it alone cannot introduce a conflict, so the check is skipped entirely. Writes that only touch other properties — `rad recipe register`, `rad env update`, and the environment updates cascaded by `rad recipepack delete` — never fail because of a namespace the caller did not intend to modify.

**`Applications.Core/environments` enforcement is deliberately unchanged.** It remains scoped to the current resource group and resource type. That type is on a deprecation path, and widening it would start failing writes on existing installs that already have duplicate namespaces. The only change there is the error surface: a distinct `NamespaceAlreadyInUse` code and a message naming the conflicting environment, replacing the raw conflict blob.

Note: With a plane-wide constraint, defaulting every environment to `default` would make the second `rad env create` fail unless users always passed `--kubernetes-namespace`.

## Reason for change
Fixes #12420
Per the discussion on the issue, the chosen direction was to enforce it consistently rather than remove it, plus a clearer error message.

## How to test

**Cross-resource-group conflict is now caught:**

```bash
rad group create g1 && rad group create g2
rad env create env1 -g g1 --kubernetes-namespace shared --preview
rad env create env2 -g g2 --kubernetes-namespace shared --preview   # now fails
```

Expect a `NamespaceAlreadyInUse` error naming `.../g1/providers/Radius.Core/environments/env1`. Before this change the second command succeeded.

**Namespace is immutable:**

```bash
rad env update env1 -g g1 --kubernetes-namespace other --preview   # rejected
rad env update env1 -g g1 --clear-kubernetes --preview             # rejected
```

**Unrelated writes are unaffected** (the important regression check):

```bash
rad recipe register myrecipe -e env1 -g g1 --template-kind bicep --template-path <path>
rad env update env1 -g g1 --azure-subscription-id <id> --azure-resource-group <rg> --preview
```

Both must succeed, including on installs that already have duplicate namespaces across groups.

**Legacy type is untouched:** repeat the first scenario without `--preview`; the second create must still succeed (group-scoped enforcement), only with a clearer error when it does conflict within one group.

Automated:

```bash
go test ./pkg/corerp/frontend/... ./pkg/cli/cmd/env/... ./pkg/cli/cmd/radinit/... ./pkg/cli/clients/... ./pkg/armrpc/...
```

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/corerp/frontend/controller/util/namespace.go` | **New.** `FindEnvironmentNamespaceConflict` searches the whole plane across both environment resource types, iterating all matches while skipping the resource under change and ACI environments. Plus the conflict and immutability message builders. |
| `pkg/corerp/frontend/controller/util/namespace_test.go` | **New.** Covers cross-group and cross-type conflicts, self-exclusion, iterating past self to a later conflict, ACI exemption, and the recursive plane-wide query shape. |
| `pkg/corerp/frontend/controller/util/query.go` | Adds `findResourcesRecursive` (sets `ScopeRecursive`). Deliberately passes no `QueryOptions`: the Postgres client applies filters in Go *after* the SQL `LIMIT`, so a limit would silently reduce this to a first-page-only check. |
| `pkg/corerp/frontend/controller/environments/v20250801preview/createorupdateenvironment.go` | Adds `validateNamespace` and `existingNamespace`, implementing the create-vs-update rules above. |
| `pkg/corerp/frontend/controller/environments/v20250801preview/createorupdateenvironment_test.go` | Adds uniqueness and immutability test suites; existing update/patch cases now assert **no** uniqueness query is issued. |
| `pkg/corerp/frontend/controller/environments/createorupdateenvironment.go` | Legacy controller: enforcement unchanged; returns the new error code and message. |
| `pkg/corerp/frontend/controller/environments/createorupdateenvironment_test.go` | Adds a test asserting the improved error **and** that the query stays non-recursive and single-type. |
| `pkg/armrpc/api/v1/errorcodes.go` | Adds `NamespaceAlreadyInUse` and `NamespaceImmutable`. |
| `pkg/armrpc/rest/results.go` | Adds `NewConflictResponseWithCode` and `NewBadRequestResponseWithCode`. |
| `pkg/cli/clients/errors.go`, `pkg/cli/clients/errors_test.go` | Adds `IsNamespaceAlreadyInUseError`, keyed on the error **code** rather than HTTP 409 — `PrepareResource` also returns 409 for in-flight provisioning. |
| `pkg/cli/cmd/env/create/create.go` | Friendly message on namespace conflict. |
| `pkg/cli/cmd/env/create/preview/create.go` | Friendly message on namespace conflict. The `default` namespace default is unchanged. |
| `pkg/cli/cmd/env/update/preview/update.go`, `pkg/cli/cmd/env/update/preview/update_test.go` | Rejects namespace changes and `--clear-kubernetes` client-side for fast feedback; the server enforces independently. |
| `pkg/cli/cmd/radinit/preview/environment.go` | Validates the namespace prompt with Kubernetes namespace rules instead of resource-name rules — the prompt asks for a namespace, but a value like `MyNamespace` previously passed the CLI and failed server-side. |

## Breaking changes scoped to new `Radius.Core/environments` (preview). No impact on `Applications.Core` users.

- The Kubernetes namespace is now immutable after an environment is created. Changing it requires creating a new environment.
- `--clear-kubernetes` is rejected on an environment that has a namespace.

### DIFF
--- a/pkg/armrpc/rest/results.go
+++ b/pkg/armrpc/rest/results.go
@@ -431,13 +431,14 @@ func NewBadRequestResponse(message string) Response {
 	}
 }
 
-// NewBadRequestResponseWithCode creates a BadRequestResponse with a specific error code, so that
-// clients can distinguish this failure from a generic validation error.
-func NewBadRequestResponseWithCode(code string, message string) Response {
+// NewBadRequestResponseWithCode creates a BadRequestResponse with a specific ARM error code in the
+// response body. The HTTP status is always 400; errorCode only sets the machine-readable code that
+// lets clients distinguish this failure from a generic validation error (v1.CodeInvalid).
+func NewBadRequestResponseWithCode(errorCode string, message string) Response {
 	return &BadRequestResponse{
 		Body: v1.ErrorResponse{
 			Error: &v1.ErrorDetails{
-				Code:    code,
+				Code:    errorCode,
 				Message: message,
 			},
 		},
@@ -614,14 +615,15 @@ func NewConflictResponse(message string) Response {
 	}
 }
 
-// NewConflictResponseWithCode creates a ConflictResponse with a specific ARM error code. Use this
-// when clients need to distinguish one kind of conflict from another, since callers otherwise
-// cannot tell a namespace collision apart from, for example, an in-flight provisioning conflict.
-func NewConflictResponseWithCode(code string, message string) Response {
+// NewConflictResponseWithCode creates a ConflictResponse with a specific ARM error code in the
+// response body. The HTTP status is always 409; errorCode only sets the machine-readable code that
+// lets clients tell one kind of conflict from another — for example a namespace collision
+// (v1.CodeNamespaceAlreadyInUse) versus an in-flight provisioning conflict, both of which are 409.
+func NewConflictResponseWithCode(errorCode string, message string) Response {
 	return &ConflictResponse{
 		Body: v1.ErrorResponse{
 			Error: &v1.ErrorDetails{
-				Code:    code,
+				Code:    errorCode,
 				Message: message,
 			},
 		},
--- a/pkg/armrpc/rest/results.go
+++ b/pkg/armrpc/rest/results.go
@@ -431,6 +431,19 @@ func NewBadRequestResponse(message string) Response {
 	}
 }
 
+// NewBadRequestResponseWithCode creates a BadRequestResponse with a specific error code, so that
+// clients can distinguish this failure from a generic validation error.
+func NewBadRequestResponseWithCode(code string, message string) Response {
+	return &BadRequestResponse{
+		Body: v1.ErrorResponse{
+			Error: &v1.ErrorDetails{
+				Code:    code,
+				Message: message,
+			},
+		},
+	}
+}
+
 // NewBadRequestARMResponse creates a BadRequestResponse with error message.
 func NewBadRequestARMResponse(body v1.ErrorResponse) Response {
 	return &BadRequestResponse{
@@ -595,6 +608,20 @@ func NewConflictResponse(message string) Response {
 		Body: v1.ErrorResponse{
 			Error: &v1.ErrorDetails{
 				Code:    v1.CodeConflict,
+				Message: message,
+			},
+		},
+	}
+}
+
+// NewConflictResponseWithCode creates a ConflictResponse with a specific ARM error code. Use this
+// when clients need to distinguish one kind of conflict from another, since callers otherwise
+// cannot tell a namespace collision apart from, for example, an in-flight provisioning conflict.
+func NewConflictResponseWithCode(code string, message string) Response {
+	return &ConflictResponse{
+		Body: v1.ErrorResponse{
+			Error: &v1.ErrorDetails{
+				Code:    code,
 				Message: message,
 			},
 		},
--- a/pkg/corerp/frontend/controller/util/query.go
+++ b/pkg/corerp/frontend/controller/util/query.go
@@ -40,9 +40,8 @@ func FindResources(ctx context.Context, rootScope, resourceType, filterKey, filt
 // findResourcesRecursive behaves like FindResources but applies rootScope recursively, so a plane
 // scope such as "/planes/radius/local" matches resources in every resource group beneath it.
 //
-// No QueryOptions are passed, which leaves MaxQueryItemCount unset and the underlying SQL LIMIT
-// null. That matters: the Postgres client applies Filters in Go *after* the query returns, so a
-// limit would silently reduce this to "check only the first page of resources".
+// No max item count is set, because this uniqueness check has to see every match: a limited query
+// returns only the first batch of resources, so a conflict beyond it would go unnoticed.
 func findResourcesRecursive(ctx context.Context, rootScope, resourceType, filterKey, filterValue string, databaseClient database.Client) (*database.ObjectQueryResult, error) {
 	query := database.Query{
 		RootScope:      rootScope,
--- a/pkg/corerp/frontend/controller/util/query.go
+++ b/pkg/corerp/frontend/controller/util/query.go
@@ -36,3 +36,24 @@ func FindResources(ctx context.Context, rootScope, resourceType, filterKey, filt
 	}
 	return databaseClient.Query(ctx, query)
 }
+
+// findResourcesRecursive behaves like FindResources but applies rootScope recursively, so a plane
+// scope such as "/planes/radius/local" matches resources in every resource group beneath it.
+//
+// No QueryOptions are passed, which leaves MaxQueryItemCount unset and the underlying SQL LIMIT
+// null. That matters: the Postgres client applies Filters in Go *after* the query returns, so a
+// limit would silently reduce this to "check only the first page of resources".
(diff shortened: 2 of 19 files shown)


pkg/armrpc/rest/results.go
```diff
@@ -431,6 +431,19 @@ func NewBadRequestResponse(message string) Response {
 	}
 }
 
+// NewBadRequestResponseWithCode creates a BadRequestResponse with a specific error code, so that
+// clients can distinguish this failure from a generic validation error.
+func NewBadRequestResponseWithCode(code string, message string) Response {
```
willdavsmith: this is so that we can return 409 errors to the user?

pkg/armrpc/rest/results.go
```diff
@@ -601,6 +614,20 @@ func NewConflictResponse(message string) Response {
 	}
 }
 
+// NewConflictResponseWithCode creates a ConflictResponse with a specific ARM error code. Use this
+// when clients need to distinguish one kind of conflict from another, since callers otherwise
+// cannot tell a namespace collision apart from, for example, an in-flight provisioning conflict.
+func NewConflictResponseWithCode(code string, message string) Response {
```
willdavsmith: do we need the code to be a parameter, since conflict response will always be 409?

pkg/corerp/frontend/controller/util/query.go
```diff
@@ -36,3 +36,24 @@ func FindResources(ctx context.Context, rootScope, resourceType, filterKey, filt
 	}
 	return databaseClient.Query(ctx, query)
 }
+
+// findResourcesRecursive behaves like FindResources but applies rootScope recursively, so a plane
+// scope such as "/planes/radius/local" matches resources in every resource group beneath it.
+//
+// No QueryOptions are passed, which leaves MaxQueryItemCount unset and the underlying SQL LIMIT
```
willdavsmith: I think we should remove this comment, it's misleading and assuming we're using postgres for the DB

pkg/corerp/frontend/controller/util/query.go
```diff
@@ -36,3 +36,24 @@ func FindResources(ctx context.Context, rootScope, resourceType, filterKey, filt
...
+// null. That matters: the Postgres client applies Filters in Go *after* the query returns, so a
+// limit would silently reduce this to "check only the first page of resources".
```
willdavsmith: is this a bug though? does databaseClient.Query expect callers to pass a maxLimit or does it handle paginating/whatever internally and return the entire output?

pkg/armrpc/rest/results.go
lakshmimsft: this is 400. updated comment and param to errorCode

pkg/armrpc/rest/results.go
lakshmimsft: The 409 is fixed, but the body's error code isn't. We need  NamespaceAlreadyInUse  specifically because PrepareResource also returns 409 for in-flight provisioning —  IsNamespaceAlreadyInUseError  keys on the code.

pkg/corerp/frontend/controller/util/query.go
lakshmimsft: It doesn't paginate internally — it returns one batch plus a  PaginationToken  the caller has to follow. Not setting a limit skips all that and returns every match on every backend, which is what this check needs. There is existing behavior in the Postgres client around filters plus a limit, but that's unrelated to this PR.

willdavsmith (approved):

PR #12719, opened 2026-08-19
Remove duplicated Repo Radius extension workflows

sk593:
## Description

The Repo Radius extension assets under `.github/extension/` (workflow templates, composite actions, and helper scripts) now live solely in [`radius-project/ai-extensions`](https://github.com/radius-project/ai-extensions/tree/main/.github/extension), which is their single source of truth. This PR removes the duplicated copy from `radius-project/radius` and the consumer-side wiring that only existed to serve it.

Companion work in ai-extensions:
- Issue: radius-project/ai-extensions#420
- PR (port + self-contained rework): radius-project/ai-extensions#424

## What changed

- **Deleted** `.github/extension/` — all workflows, composite actions, scripts, and their tests (43 files, including the `delete-environment.yml` env-delete workflow that also landed on `main` and now lives in ai-extensions).
- **Deleted** `build/scripts/verify-contrib-consumers.sh` and its `make verify-contrib-consumers` target (`build/resource-types.mk`); removed the corresponding step and the `.github/extension/**` / `verify-contrib-consumers.sh` path triggers from `.github/workflows/verify-resource-types-manifest.yaml`.
- **Trimmed** `build/scripts/test-sync-resource-types.sh` to the stable-first pin-selection tests. The catalog-consumer tests moved with the workflows to ai-extensions; `sync-resource-types.sh` itself is unchanged and still fully covered.
- **Removed** every extension-action test target from `build/test.mk` (`test-apply-custom-recipe-packs`, `test-run-rad-commands-action`, `test-command-outcome`, `test-azure-oidc-refresh`, `test-build-platforms`, `test-publish-deploy-status`, `test-verify-azure`, `test-extension-action-shell-syntax`, `test-teardown`, `test-deploy-progress`, `test-deploy-progress-uploader`) and trimmed the `test` aggregate prerequisites accordingly; also dropped the `make test-deploy-progress-uploader` step from `.github/workflows/lint.yaml`.
- **Removed** the two `.github/extension/*` Dependabot directory entries.
- **Repointed** the contributing docs and the deploy-workflow design note at the ai-extensions repository.

## What radius keeps

`deploy/manifest/defaults.yaml` stays in this repository. It remains the Radius-owned catalog of contrib refs; the ai-extensions composite actions now fetch it **by ref at runtime** rather than requiring the whole radius repo to be checked out (see ai-extensions#424).

## Follow-up

The consumer-side verification that is being dropped here (`verify-contrib-consumers.sh` plus the catalog-consumer unit tests) should be **ported to ai-extensions** so the moved workflows keep that safety net. Tracked as a follow-up on #424.

## Validation

- `make test-sync-resource-types` — passes.
- `make -n test` — resolves cleanly after removing the extension test targets.
- No remaining `.github/extension` references in the repo except (a) the intentional links to the ai-extensions repo in the contributing docs and design note, and (b) `specs/005-live-deploy-graph/plan.md`, which is a completed, historical design/plan artifact for a past feature and is intentionally left as-is (it records the state at the time it was written and is out of scope for this cleanup).


build/test.mk
```diff
@@ -53,47 +53,13 @@ GOTEST_OPTS ?=
 GOTEST_TOOL ?= go tool gotestsum $(GOTESTSUM_OPTS) --
 
 .PHONY: test
-test: test-get-envtools test-helm test-manage-radius-installation test-run-rad-commands-action test-azure-oidc-refresh test-build-platforms test-publish-deploy-status test-extension-action-shell-syntax test-deploy-progress test-verify-azure ## Runs unit tests, excluding kubernetes controller tests
+test: test-get-envtools test-helm test-manage-radius-installation ## Runs unit tests, excluding kubernetes controller tests
```
DariuszPorowski, line 56: issue(test,blocking): move the workflow contract tests before removing these targets

PR #424 does not invoke the ported shell tests and does not yet contain the complete current suite. Removing these targets drops coverage for deploy parameter handling, Azure token refresh, platform selection, status publication, action shell validity, live progress, and Azure verification; current Radius main also adds the repeat custom-recipe-pack test. Port and wire the complete suite in `ai-extensions` before removing the Radius targets.

.github/workflows/verify-resource-types-manifest.yaml
```diff
@@ -35,7 +35,7 @@
 #   - New entries in defaults.yaml without re-syncing
 #
 # recipePacks[] have no vendored copies to compare, but their stable-first pin
-# policy, tag-to-SHA consistency, and runtime catalog consumers are tested.
+# policy and tag-to-SHA consistency are tested.
```
DariuszPorowski, line 38: issue(test,blocking): preserve validation of the catalog's real workflow consumers

This change intentionally stops testing runtime catalog consumers, but the removed `verify-contrib-consumers.sh` does more than validate pin policy: it fetches recipe-pack files, checks OCI recipe tags, verifies Git recipe directories, and enforces that workflows resolve contrib sources only through `defaults.yaml`. PR #424 has no replacement, so a broken contrib publish would first surface in a user's deploy. Move this verifier to `ai-extensions` and run it against the fetched Radius catalog before removing this gate.

DariuszPorowski (commented):
Core workflow-migration findings only.

DariuszPorowski (approved):

build/test.mk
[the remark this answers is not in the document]
sk593: Good catch. `specs/005-live-deploy-graph/plan.md` is a completed, historical design/plan artifact for a past feature — it records the state at the time it was written and is out of scope for this cleanup, so I've left it as-is rather than rewriting a shipped spec. I've updated the PR description to call this out explicitly instead of claiming there are zero remaining references.

sk593: Agreed — the suite needs to be complete in ai-extensions before these targets are dropped. 8/10 were already ported; the two newest that landed on radius main after the original port (`command-outcome_test.sh`, `teardown_test.sh`) are being ported now into `radius-project/ai-extensions` at the same paths, where the `extension-selftests.yml` job auto-discovers and runs every `.github/extension/**/*_test.sh`. I'll link the ai-extensions PR here and resolve this once it merges.

.github/workflows/verify-resource-types-manifest.yaml
sk593: Agreed — this gate shouldn't disappear. `verify-contrib-consumers.sh` is being ported to `radius-project/ai-extensions` and reworked to fetch the radius catalog (`deploy/manifest/defaults.yaml`) by ref (reusing the runtime catalog fetch) rather than a local radius checkout, then run against the `.github/extension/**` tree in CI — so a broken contrib publish fails an ai-extensions check instead of surfacing in a user deploy. I'll link that PR here and resolve this once it merges.

build/test.mk
sk593: Port is up: radius-project/ai-extensions#653 (draft, commit `e92d53db`). It ports `command-outcome_test.sh` and `teardown_test.sh` plus their complete upstream action/workflow behavior into ai-extensions, where `extension-selftests.yml` auto-runs every `.github/extension/**/*_test.sh`. Catalog/parity ref pinned to immutable radius commit `745ce9cc`. Keeping this thread open — I'll resolve once #653 merges, and the radius deletions here stay gated until then.

.github/workflows/verify-resource-types-manifest.yaml
sk593: Port is up: radius-project/ai-extensions#653 (draft, commit `e92d53db`). It ports `verify-contrib-consumers` with hermetic PR-gate policy/wrapper tests plus authenticated recipe-pack / OCI / Git validation wired into the existing live PR workflow, running against the fetched radius catalog (pinned to immutable radius commit `745ce9cc`). Keeping this thread open — I'll resolve once #653 merges, and the radius deletions here stay gated until then.

build/test.mk
sk593: Resolved: radius-project/ai-extensions#653 merged (`10e9fb8c`). The complete shell-test suite now lives in ai-extensions — `command-outcome_test.sh` and `teardown_test.sh` are present at their action paths and run automatically via `extension-selftests.yml`, so coverage is preserved before these targets are removed here.

.github/workflows/verify-resource-types-manifest.yaml
sk593: Resolved: radius-project/ai-extensions#653 merged (`10e9fb8c`). `verify-contrib-consumers.sh` (plus a hermetic `verify-contrib-consumers_test.sh`) now lives in ai-extensions and runs against the fetched radius catalog with authenticated recipe-pack / OCI / Git validation wired into CI — so a broken contrib publish fails an ai-extensions check instead of surfacing in a user deploy, before this gate is removed here.

PR #12741, opened 2026-08-20
fix(recipes): reject mappings to undeclared module outputs

### LINKED ISSUE #12647: Recipe pack outputs mapping silently drops missing outputs, deleting properties.secrets and failing on the consumer
## Summary

When a recipe pack's `outputs` mapping names a module output that doesn't exist, Radius drops it silently. For a **secret** output the consequence is severe: the owning resource reports `OK`, its `properties.secrets` is removed, and the deployment fails later on the *consumer* with an error that points at the wrong resource and looks like a schema bug.

I'd like a mapped-but-missing output to fail loudly, and ideally to be caught before deploy.

## What happens today

Recipe pack (custom type backed by an AVM module):

```bicep
source: 'mcr.microsoft.com/bicep/avm/res/service-bus/namespace:0.12.0'
outputs: {
  secrets: {
    connectionString: 'primaryConnectionString'
  }
}
```

`primaryConnectionString` doesn't exist in 0.12.0 (added in 0.15.0). The chain:

1. [`ApplyOutputsMapping`](https://github.com/radius-project/radius/blob/main/pkg/recipes/util/outputs.go#L30) maps only keys that are present. A missing key is skipped — no error, no warning:
   ```go
   for propertyName, outputName := range secretOutputsMap {
       if val, ok := secrets[outputName]; ok {
           mappedSecrets[propertyName] = val
       } else if val, ok := values[outputName]; ok {
           mappedSecrets[propertyName] = val
       }
       // no else: silently absent
   }
   ```
2. `RecipeOutput.Secrets` is empty.
3. [`materializeRecipeSecrets`](https://github.com/radius-project/radius/blob/main/pkg/dynamicrp/backend/processor/dynamicresource.go#L166) cannot distinguish "this update legitimately produced no secrets" from "the mapping is broken", so it takes the clear path and **deletes `properties.secrets`**.
4. The backing resource completes as `OK`.
5. The consumer fails:
   ```
   InvalidTemplate: Unable to process template language expressions for resource
   'Radius.Compute/containers/todo-list-app' ... 'The language expression property
   'secrets' doesn't exist, available properties are 'application, codeReference,
   environment, provisioningState, queueName, recipe, status'.'
   ```

## Why this is worth fixing

The error names the **container**, and the missing property is on the **Service Bus resource**, whose own status is `OK`. Nothing mentions the recipe or the output mapping. The obvious reading is "my resource type schema is wrong", so the natural response is to go edit the type schema and re-register it — which changes nothing, because the schema was correct.

It took several deploy cycles (~12 min each) to trace this back to a module version pin. The information needed to diagnose it — "the mapping referenced `primaryConnectionString`, which the recipe did not return" — is known inside the driver at step 1 and then discarded.

Silent-drop is also the wrong default for secrets specifically: the failure mode is a resource that looks healthy but has no credential.

## Proposed behavior

**1. Fail the recipe execution when a mapped output is missing.** In `ApplyOutputsMapping` (or at its call site, which has the recipe/type context), collect unmatched mapping entries and return an error naming them and what *was* returned:

```
recipe for 'Radius.Resources/azureServiceBusNamespaces' declares secret output
'connectionString' -> 'primaryConnectionString', but the recipe returned no such
output (available: location, name, privateEndpoints, resourceGroupName,
resourceId, serviceBusEndpoint, systemAssignedMIPrincipalId)
```

That message alone would have ended this in one cycle. Note the available-outputs list makes the version mismatch self-evident.

**2. Distinguish "no secrets produced" from "mapping produced nothing".** If a type declares a `secrets` block and the recipe declares secret output mappings, then zero materialized secrets should be an error rather than a trigger for `clearManagedSecret`. Today those two states are conflated.

**3. (Optional, best value) Validate at registration.** For `kind: bicep` with a pinned module `source`, the module's interface is knowable at `rad recipe register` / recipe-pack deploy time — so a mapping that can never resolve could be rejected before any deploy is attempted.

If the silent-drop behavior is deliberate for *non-secret* optional outputs, I'd suggest scoping the hard failure to `secrets` mappings, and emitting at least a warning for values.

## Repro

1. Pin a recipe pack to a module version lacking an output you map under `outputs.secrets`
2. Consume it via `secretKeyRef` with `<resource>.properties.secrets.name`
3. Deploy — the backing resource reports `OK` and the container fails with the `InvalidTemplate` above

Observed on: https://github.com/kachawla/todo-list-app/actions/runs/31552351543

AzureMike:
## Summary

Radius now rejects recipe output mappings that name outputs the referenced Bicep or Terraform module does not declare. The drivers return `InvalidRecipeOutputs` with `setupError` before Bicep calls `CreateOrUpdate` or Terraform starts `apply`.

## Reason for change

The Service Bus recipe pack maps `primaryConnectionString`, but version `0.12.0` of the Service Bus AVM module does not declare that output:

```bicep
source: 'mcr.microsoft.com/bicep/avm/res/service-bus/namespace:0.12.0'
outputs: {
  secrets: {
    connectionString: 'primaryConnectionString'
  }
}
```

Radius previously omitted this Bicep mapping. The backing resource could report a healthy state without the expected secret, then the consuming resource failed with `InvalidTemplate`. That error pointed at the consumer instead of the invalid recipe mapping.

Both drivers already load module declarations before deployment. The Bicep driver now checks mappings against the top-level ARM template `outputs`. The Terraform driver checks them against the outputs found during module inspection. An invalid mapping reports the recipe name, resource type, bad mappings, and available module outputs:

```text
recipe "servicebus" for resource type "Radius.Resources/azureServiceBusNamespaces":
invalid outputs mapping: no declared module output matches
"secrets.connectionString" -> "primaryConnectionString"; available module outputs:
"location", "name", "resourceId", "serviceBusEndpoint"
```

The error orders invalid mappings and available outputs consistently.

This change is limited to declaration validation. Runtime output mapping, logging, sensitivity routing, duplicate mappings, wrapped `result` handling, and Terraform output generation remain unchanged. Terraform deletion skips only the new declaration check, so existing deletion behavior, including failures caused by stale mapped output references, stays the same.

Fixes #12647.

## How to test

```text
go test -count=1 ./pkg/recipes/util ./pkg/recipes/terraform/... ./pkg/recipes/driver/bicep ./pkg/recipes/driver/terraform ./pkg/rp/util/registrytest
```

The focused tests cover deterministic validation errors, Bicep rejection before the deployment client is called, Terraform `setupError` classification, and the Terraform deletion bypass.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/recipes/util/outputs.go` | Adds shared validation and deterministic diagnostics for undeclared output mappings. |
| `pkg/recipes/util/outputs_test.go` | Covers valid mappings, multiple invalid mappings, and modules with no declared outputs. |
| `pkg/recipes/driver/bicep/bicep.go` | Validates mappings against ARM template output declarations before deployment. |
| `pkg/recipes/driver/bicep/bicep_test.go` | Verifies validation and confirms that an invalid mapping does not call the deployment client. |
| `pkg/recipes/terraform/execute.go` | Validates mappings against inspected Terraform outputs before apply and skips the new check during deletion. |
| `pkg/recipes/terraform/execute_test.go` | Covers Terraform declaration validation and the deletion bypass. |
| `pkg/recipes/driver/terraform/terraform.go` | Classifies declaration failures as `InvalidRecipeOutputs` with `setupError`. |
| `pkg/recipes/driver/terraform/terraform_test.go` | Covers Terraform driver error classification. |

### DIFF
--- a/pkg/recipes/driver/bicep/bicep.go
+++ b/pkg/recipes/driver/bicep/bicep.go
@@ -18,6 +18,7 @@ package bicep
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	reflect "reflect"
 	"slices"
@@ -54,6 +55,7 @@ const (
 	deploymentPrefix = "recipe"
 	pollFrequency    = time.Second * 5
 	recipeParameters = "parameters"
+	recipeOutputs    = "outputs"
 )
 
 var _ driver.Driver = (*bicepDriver)(nil)
@@ -118,6 +120,13 @@ func (d *bicepDriver) Execute(ctx context.Context, opts driver.ExecuteOptions) (
 	}
 	metrics.DefaultRecipeEngineMetrics.RecordRecipeDownloadDuration(ctx, downloadStartTime,
 		metrics.NewRecipeAttributes(metrics.RecipeEngineOperationDownloadRecipe, opts.Recipe.Name, &opts.Definition, metrics.SuccessfulOperationState))
+
+	if err := validateOutputMappings(opts.Definition, recipeData); err != nil {
+		if _, ok := errors.AsType[*recipes_util.OutputMappingError](err); ok {
+			return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
+		}
+		return nil, recipes.NewRecipeError(recipes.RecipeLanguageFailure, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
+	}
 
 	// create the context object to be passed to the recipe deployment
 	recipeContext, err := recipecontext.New(&opts.Recipe, &opts.Configuration)
@@ -326,6 +335,32 @@ func (d *bicepDriver) GetRecipeMetadata(ctx context.Context, opts driver.BaseOpt
 	}
 
 	return recipeData, nil
+}
+
+func validateOutputMappings(definition recipes.EnvironmentDefinition, recipeData map[string]any) error {
+	if len(definition.Outputs) == 0 && len(definition.SecretOutputs) == 0 {
+		return nil
+	}
+
+	declaredOutputs := []string{}
+	rawOutputs, ok := recipeData[recipeOutputs]
+	if ok && rawOutputs != nil {
+		outputs, ok := rawOutputs.(map[string]any)
+		if !ok {
+			return errors.New("recipe outputs must be an object")
+		}
+
+		declaredOutputs = make([]string, 0, len(outputs))
+		for outputName := range outputs {
+			declaredOutputs = append(declaredOutputs, outputName)
+		}
+	}
+
+	if err := recipes_util.ValidateOutputsMapping(declaredOutputs, definition.Outputs, definition.SecretOutputs); err != nil {
+		return fmt.Errorf("recipe %q for resource type %q: %w", definition.Name, definition.ResourceType, err)
+	}
+
+	return nil
 }
 
 func hasContextParameter(recipeData map[string]any) bool {
--- a/pkg/recipes/driver/bicep/bicep.go
+++ b/pkg/recipes/driver/bicep/bicep.go
@@ -122,10 +122,7 @@ func (d *bicepDriver) Execute(ctx context.Context, opts driver.ExecuteOptions) (
 		metrics.NewRecipeAttributes(metrics.RecipeEngineOperationDownloadRecipe, opts.Recipe.Name, &opts.Definition, metrics.SuccessfulOperationState))
 
 	if err := validateOutputMappings(opts.Definition, recipeData); err != nil {
-		if _, ok := errors.AsType[*recipes_util.OutputMappingError](err); ok {
-			return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
-		}
-		return nil, recipes.NewRecipeError(recipes.RecipeLanguageFailure, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
+		return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
 	}
 
 	// create the context object to be passed to the recipe deployment
@@ -195,6 +192,9 @@ func (d *bicepDriver) Execute(ctx context.Context, opts driver.ExecuteOptions) (
 
 	recipeResponse, err := d.prepareRecipeResponse(opts.BaseOptions.Definition, resp.Properties.Outputs, resp.Properties.OutputResources)
 	if err != nil {
+		if _, ok := errors.AsType[*recipes_util.MissingOutputValuesError](err); ok {
+			return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, err.Error(), recipes_util.ExecutionError, recipes.GetErrorDetails(err))
+		}
 		return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, fmt.Sprintf("failed to read the recipe output %q: %s", recipes.ResultPropertyName, err.Error()), recipes_util.ExecutionError, recipes.GetErrorDetails(err))
 	}
 
@@ -342,12 +342,15 @@ func validateOutputMappings(definition recipes.EnvironmentDefinition, recipeData
 		return nil
 	}
 
-	declaredOutputs := []string{}
+	var declaredOutputs []string
 	rawOutputs, ok := recipeData[recipeOutputs]
 	if ok && rawOutputs != nil {
 		outputs, ok := rawOutputs.(map[string]any)
 		if !ok {
-			return errors.New("recipe outputs must be an object")
+			return fmt.Errorf(
+				"recipe %q for resource type %q: recipe outputs must be an object",
+				definition.Name,
+				definition.ResourceType)
 		}
 
 		declaredOutputs = make([]string, 0, len(outputs))
@@ -356,11 +359,12 @@ func validateOutputMappings(definition recipes.EnvironmentDefinition, recipeData
 		}
 	}
 
-	if err := recipes_util.ValidateOutputsMapping(declaredOutputs, definition.Outputs, definition.SecretOutputs); err != nil {
-		return fmt.Errorf("recipe %q for resource type %q: %w", definition.Name, definition.ResourceType, err)
-	}
-
-	return nil
+	return recipes_util.ValidateOutputsMapping(
+		definition.Name,
+		definition.ResourceType,
+		declaredOutputs,
+		definition.Outputs,
+		definition.SecretOutputs)
 }
 
 func hasContextParameter(recipeData map[string]any) bool {
@@ -452,32 +456,35 @@ func newProviderConfig(resourceGroup string, envProviders coredm.Providers) clie
 // collection. For us this mostly means Kubernetes resources - the user has to be explicit.
 func (d *bicepDriver) prepareRecipeResponse(definition recipes.EnvironmentDefinition, outputs any, resources []*armdeployments.ResourceReference) (*recipes.RecipeOutput, error) {
 	recipeResponse := &recipes.RecipeOutput{}
-	out, ok := outputs.(map[string]any)
-	if ok && len(out) > 0 {
-		hasOutputsMapping := len(definition.Outputs) > 0 || len(definition.SecretOutputs) > 0
-		_, hasResultOutput := out[recipes.ResultPropertyName]
+	out, _ := outputs.(map[string]any)
+	hasOutputsMapping := len(definition.Outputs) > 0 || len(definition.SecretOutputs) > 0
+	_, hasResultOutput := out[recipes.ResultPropertyName]
 
-		switch {
-		case hasOutputsMapping:
-			// Direct module with an outputs mapping — collect all ARM outputs flat (splitting
-			// secure-typed outputs into secrets), then apply the mapping.
-			values, secrets := collectARMOutputs(out)
-			recipeResponse.Values, recipeResponse.Secrets = recipes_util.ApplyOutputsMapping(values, secrets, definition.Outputs, definition.SecretOutputs)
-		case hasResultOutput:
-			// Wrapped recipe — use the existing 'result' output parsing.
-			if result, ok := out[recipes.ResultPropertyName].(map[string]any); ok {
-				if resultValue, ok := result["value"].(map[string]any); ok {
-					err := recipeResponse.PrepareRecipeResponse(resultValue)
-					if err != nil {
-						return &recipes.RecipeOutput{}, err
-					}
+	switch {
+	case hasOutputsMapping:
+		// Direct module with an outputs mapping — collect all ARM outputs flat (splitting
+		// secure-typed outputs into secrets), then apply the mapping.
+		values, secrets := collectARMOutputs(out)
+		mappedValues, mappedSecrets, err := recipes_util.ApplyOutputsMapping(values, secrets, definition.Outputs, definition.SecretOutputs)
+		if err != nil {
+			return &recipes.RecipeOutput{}, fmt.Errorf("recipe %q for resource type %q: %w", definition.Name, definition.ResourceType, err)
+		}
+		recipeResponse.Values = mappedValues
+		recipeResponse.Secrets = mappedSecrets
+	case len(out) > 0 && hasResultOutput:
+		// Wrapped recipe — use the existing 'result' output parsing.
+		if result, ok := out[recipes.ResultPropertyName].(map[string]any); ok {
+			if resultValue, ok := result["value"].(map[string]any); ok {
+				err := recipeResponse.PrepareRecipeResponse(resultValue)
+				if err != nil {
+					return &recipes.RecipeOutput{}, err
 				}
 			}
-		default:
-			// Direct module without a mapping — pass through all ARM outputs unchanged, routing
-			// secure-typed outputs to Secrets so they are not exposed as plain values.
-			recipeResponse.Values, recipeResponse.Secrets = collectARMOutputs(out)
 		}
+	case len(out) > 0:
+		// Direct module without a mapping — pass through all ARM outputs unchanged, routing
+		// secure-typed outputs to Secrets so they are not exposed as plain values.
+		recipeResponse.Values, recipeResponse.Secrets = collectARMOutputs(out)
 	}
 
 	recipeResponse.Status = &rpv1.RecipeStatus{
--- a/pkg/recipes/driver/bicep/bicep_test.go
+++ b/pkg/recipes/driver/bicep/bicep_test.go
@@ -557,7 +557,7 @@ func Test_Bicep_Execute_InvalidOutputMappingDoesNotDeploy(t *testing.T) {
 	require.Equal(t, &recipes.RecipeError{
 		ErrorDetails: v1.ErrorDetails{
 			Code:    recipes.InvalidRecipeOutputs,
-			Message: `recipe "mongo-azure" for resource type "Applications.Datastores/mongoDatabases": invalid outputs mapping: no declared module output matches "secrets.connectionString" -> "primaryConnectionString"; available module outputs: none`,
+			Message: `recipe "mongo-azure" for resource type "Applications.Datastores/mongoDatabases": invalid outputs mapping: no declared module output matches secrets["connectionString"] -> "primaryConnectionString"; available module outputs: none`,
 		},
 		DeploymentStatus: recipes_util.RecipeSetupError,
 	}, err)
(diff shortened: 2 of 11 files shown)


pkg/recipes/driver/bicep/bicep.go
```diff
@@ -119,6 +121,13 @@ func (d *bicepDriver) Execute(ctx context.Context, opts driver.ExecuteOptions) (
 	metrics.DefaultRecipeEngineMetrics.RecordRecipeDownloadDuration(ctx, downloadStartTime,
 		metrics.NewRecipeAttributes(metrics.RecipeEngineOperationDownloadRecipe, opts.Recipe.Name, &opts.Definition, metrics.SuccessfulOperationState))
 
+	if err := validateOutputMappings(opts.Definition, recipeData); err != nil {
```
kachawla, line 124: Nice placement - after the download but before the deployment call, so a bad mapping costs a registry pull and nothing else. Marking it `RecipeSetupError` is right too, since nothing was created and there is nothing to clean up.

pkg/recipes/driver/bicep/bicep.go
```diff
@@ -119,6 +121,13 @@ func (d *bicepDriver) Execute(ctx context.Context, opts driver.ExecuteOptions) (
...
+		if _, ok := errors.AsType[*recipes_util.OutputMappingError](err); ok {
+			return nil, recipes.NewRecipeError(recipes.InvalidRecipeOutputs, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
+		}
+		return nil, recipes.NewRecipeError(recipes.RecipeLanguageFailure, err.Error(), recipes_util.RecipeSetupError, recipes.GetErrorDetails(err))
```
kachawla: Should this fallback be `RecipeLanguageFailure`? The only other error the helper can return is the "outputs must be an object" case, which is a malformed template rather than anything about the Bicep language. `InvalidRecipeOutputs` would read closer to what actually happened, with the type check only choosing the message.

pkg/recipes/driver/bicep/bicep.go
```diff
@@ -328,6 +337,32 @@ func (d *bicepDriver) GetRecipeMetadata(ctx context.Context, opts driver.BaseOpt
 	return recipeData, nil
 }
 
+func validateOutputMappings(definition recipes.EnvironmentDefinition, recipeData map[string]any) error {
+	if len(definition.Outputs) == 0 && len(definition.SecretOutputs) == 0 {
```
kachawla, line 341: Good early return here. Wrapped `result` recipes and mapping-free direct modules never reach the new validation, so the blast radius on upgrade is limited to packs that actually declare a mapping.

pkg/recipes/driver/bicep/bicep.go
```diff
@@ -328,6 +337,32 @@ func (d *bicepDriver) GetRecipeMetadata(ctx context.Context, opts driver.BaseOpt
...
+		return nil
+	}
+
+	declaredOutputs := []string{}
```
kachawla: nit: the slice is initialized as an empty literal and then reassigned with `make` inside the branch. `var declaredOutputs []string` would say the same thing and drop one allocation on the path where the template has no outputs.

pkg/recipes/driver/bicep/bicep.go
```diff
@@ -328,6 +337,32 @@ func (d *bicepDriver) GetRecipeMetadata(ctx context.Context, opts driver.BaseOpt
...
+	rawOutputs, ok := recipeData[recipeOutputs]
+	if ok && rawOutputs != nil {
+		outputs, ok := rawOutputs.(map[string]any)
+		if !ok {
+			return errors.New("recipe outputs must be an object")
```
kachawla: Could this one carry the recipe name too? Every other failure in the function gets wrapped with `recipe %q for resource type %q`, but this returns bare, so an operator sees "recipe outputs must be an object" with nothing telling them which recipe or which type produced it.

pkg/recipes/driver/bicep/bicep.go
```diff
@@ -328,6 +337,32 @@ func (d *bicepDriver) GetRecipeMetadata(ctx context.Context, opts driver.BaseOpt
...
+		}
+
+		declaredOutputs = make([]string, 0, len(outputs))
+		for outputName := range outputs {
+			declaredOutputs = append(declaredOutputs, outputName)
+		}
+	}
+
+	if err := recipes_util.ValidateOutputsMapping(declaredOutputs, definition.Outputs, definition.SecretOutputs); err != nil {
```
kachawla: Should the comparison be case-insensitive? ARM treats template output names case-insensitively, and at runtime the mapping is resolved against the deployment response keys, so a mapping differing only in case works today and would now be rejected before deploy. I am going off ARM's naming rules rather than a repro, so worth a quick check.

pkg/recipes/driver/bicep/bicep_test.go
```diff
@@ -513,6 +526,104 @@ func Test_Bicep_GetRecipeMetadata_Error(t *testing.T) {
...
+		{
+			name: "rejects non-object outputs collection",
+			definition: recipes.EnvironmentDefinition{
+				Outputs: map[string]string{"host": "endpoint"},
+			},
+			recipeData:    map[string]any{recipeOutputs: "invalid"},
+			expectedError: "recipe outputs must be an object",
+		},
+		{
+			name:       "skips template output parsing when no mappings are configured",
```
kachawla, line 621: Might be worth one more case where `recipeData` has no `outputs` key at all. That is the realistic shape for a template that declares none, and it exercises the `ok` branch rather than the nil-value branch, ending in the "available module outputs: none" message.

pkg/recipes/driver/terraform/terraform.go
```diff
@@ -118,6 +118,10 @@ func (d *terraformDriver) Execute(ctx context.Context, opts driver.ExecuteOption
 	}
 
 	if err != nil {
+		// OutputMappingError is raised before apply and must be reported as a setup failure.
+		if _, ok := errors.AsType[*recipes_util.OutputMappingError](err); ok {
```
kachawla, line 122: Good call reclassifying this as a setup error instead of letting it ride as `RecipeDeploymentFailed` with `ExecutionError`. Nothing was applied, so the retry and cleanup semantics the RP derives from the status are now honest about there being no state to unwind.

pkg/recipes/driver/terraform/terraform_test.go
```diff
@@ -171,6 +173,47 @@ func Test_Terraform_Execute_DeploymentFailure(t *testing.T) {
 	verifyDirectoryCleanup(t, tfDriver.options.Path, armCtx.OperationID.String())
 }
 
+func Test_Terraform_Execute_InvalidOutputMapping(t *testing.T) {
+	// The executor returns OutputMappingError before apply. The driver should classify it as a
+	// setup error.
+	invalidMappingErr := recipes_util.ValidateOutputsMapping(
```
kachawla, line 176: Does this test the type check or the string? The error is built by hand and handed to the mock, so a case where the executor returns a plain error with similar text would show `errors.AsType` is what is doing the work.

pkg/recipes/terraform/execute.go
```diff
@@ -49,6 +50,14 @@ var (
 	ErrRecipeNameEmpty = errors.New("recipe name cannot be empty")
 )
 
+// outputMappingValidationMode controls declaration validation during configuration generation.
```
kachawla, line 53: nit: could this just be a `bool`? Only `Delete` opts out, so there are only ever two states. Though the named constants do read better at the call sites, so happy either way.

pkg/recipes/terraform/execute.go
```diff
@@ -330,6 +339,10 @@ func (e *executor) generateConfig(ctx context.Context, tf *tfexec.Terraform, opt
 		return "", err
 	}
 
+	if err := validateOutputMappings(options.EnvRecipe, loadedModule, validationMode); err != nil {
```
kachawla: Could this move into the mapping branch instead of running every time? The mapping is only used when the module has no `result` output, so a `result` module with a stale `outputs` field deploys fine today but would now fail before creating anything. Moving the check next to `AddMappedOutputs` would keep it on the path that actually uses the mapping.

pkg/recipes/terraform/execute.go
```diff
@@ -330,6 +339,10 @@ func (e *executor) generateConfig(ctx context.Context, tf *tfexec.Terraform, opt
...
+		return "", err
```
kachawla: The same module could also pass validation and still drop everything. If a `result` module's mapped names happen to match real outputs, `AddMappedOutputs` at `execute.go:422` never runs, so only `result` reaches state while `prepareRecipeResponse` still prefers the mapping. That is the silent drop from #12647, surviving the fix.

pkg/recipes/terraform/execute.go
```diff
@@ -437,6 +450,23 @@ func (e *executor) generateConfig(ctx context.Context, tf *tfexec.Terraform, opt
 	return secretSuffix, nil
 }
 
+func validateOutputMappings(definition *recipes.EnvironmentDefinition, module *moduleInspectResult, validationMode outputMappingValidationMode) error {
```
kachawla, line 461: Could the `recipe %q for resource type %q` wrapper live inside `ValidateOutputsMapping`? It is byte-identical in both drivers today, so passing the name and type in would stop them drifting on message format.

pkg/recipes/terraform/execute.go
```diff
@@ -437,6 +450,23 @@ func (e *executor) generateConfig(ctx context.Context, tf *tfexec.Terraform, opt
...
+	if validationMode == skipOutputMappingValidation {
+		return nil
+	}
+
+	declaredOutputs := make([]string, 0, len(module.OutputSensitivity))
```
kachawla, line 466: Should the declared names come from somewhere that cannot silently drop entries? `inspectModule` skips any output whose `tfconfig` entry is nil (`module.go:178`), so a skipped output would look undeclared here and fail the deploy rather than being quietly ignored. Vanishingly unlikely, but the sensitivity map is doing two jobs now.

pkg/recipes/terraform/execute_test.go
```diff
@@ -69,13 +69,33 @@ func TestGenerateConfig(t *testing.T) {
...
 			e := executor{}
-			_, err = e.generateConfig(ctx, tf, tc.opts)
+			_, err = e.generateConfig(ctx, tf, tc.opts, requireValidOutputMappings)
 			require.Error(t, err)
 			require.ErrorContains(t, err, tc.err)
 		})
 	}
 }
 
+func TestValidateOutputMappings(t *testing.T) {
```
kachawla, line 215: Could this run through `generateConfig` rather than the helper? Testing the helper proves the message but not that the run stops before providers, `init` and `apply` - the Bicep side has exactly that assertion with its deployment spy. Codecov's uncovered lines in this file look like they are pointing at the same gap.

pkg/recipes/terraform/execute_test.go
```diff
@@ -69,13 +69,33 @@ func TestGenerateConfig(t *testing.T) {
...
+	module := &moduleInspectResult{
+		OutputSensitivity: map[string]bool{
+			"endpoint": false,
+		},
+	}
+
+	err := validateOutputMappings(definition, module, requireValidOutputMappings)
+	require.EqualError(t, err, `recipe "service-bus" for resource type "Applications.Messaging/rabbitMQQueues": invalid outputs mapping: no declared module output matches "secrets.connectionString" -> "primaryConnectionString"; available module outputs: "endpoint"`)
+
+	err = validateOutputMappings(definition, module, skipOutputMappingValidation)
```
kachawla, line 231: Could this also drive `executor.Delete` directly? The bypass assertion is good, but it is checked a level below where it matters, and `Delete` is the path that strands real infrastructure if the skip ever regresses.

pkg/recipes/util/outputs.go
```diff
@@ -16,6 +16,63 @@ limitations under the License.
...
 package util
 
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// OutputMappingError reports mappings that refer to undeclared module outputs.
+type OutputMappingError struct {
```
kachawla, line 26: Would you consider exporting these two fields? A caller that wants the missing mappings has to regex the message string today, so if the CLI ever renders them itself, exported fields would save it the parsing.

pkg/recipes/util/outputs.go
```diff
@@ -16,6 +16,63 @@ limitations under the License.
...
+// Error describes invalid mappings and lists the module's declared outputs.
+func (e *OutputMappingError) Error() string {
+	return fmt.Sprintf(
+		"invalid outputs mapping: no declared module output matches %s; available module outputs: %s",
+		strings.Join(e.missingMappings, ", "),
+		formatOutputNames(e.availableOutputs))
+}
+
+// ValidateOutputsMapping checks every mapping against the module's declared outputs.
+func ValidateOutputsMapping(declaredOutputs []string, outputsMap map[string]string, secretOutputsMap map[string]string) error {
```
kachawla: Should the hard failure be scoped to secret mappings? The issue suggested failing on secrets and only warning on values. A dropped secret leaves a resource looking healthy with no credential, while a dropped value is cosmetic - but as written, a stale non-secret mapping stops a pack deploying on upgrade alone.

pkg/recipes/util/outputs.go
```diff
@@ -16,6 +16,63 @@ limitations under the License.
...
+	missingMappings := []string{}
+	for propertyName, outputName := range outputsMap {
+		if _, ok := declared[outputName]; !ok {
+			missingMappings = append(missingMappings, fmt.Sprintf("%q -> %q", propertyName, outputName))
+		}
+	}
+
+	for propertyName, outputName := range secretOutputsMap {
+		if _, ok := declared[outputName]; !ok {
+			missingMappings = append(missingMappings, fmt.Sprintf("%q -> %q", "secrets."+propertyName, outputName))
```
kachawla: nit: could the prefix come from a label argument rather than being built inline? A values mapping with a property named `secrets.connectionString` would render identically to a real secret mapping.

pkg/recipes/util/outputs.go
```diff
@@ -16,6 +16,63 @@ limitations under the License.
...
+
+	if len(missingMappings) == 0 {
+		return nil
+	}
+
+	availableOutputs := make([]string, 0, len(declared))
+	for outputName := range declared {
+		availableOutputs = append(availableOutputs, outputName)
+	}
+	sort.Strings(missingMappings)
```
kachawla: nit: `slices.Sort` would match the style already used elsewhere in the recipes packages and drop the `sort` import.

pkg/recipes/util/outputs.go
```diff
@@ -16,6 +16,63 @@ limitations under the License.
...
+	sort.Strings(availableOutputs)
+
+	return &OutputMappingError{
+		missingMappings:  missingMappings,
+		availableOutputs: availableOutputs,
+	}
+}
```
kachawla, line 92: This does not cover an output that is declared but comes back null at apply - a `count = 0` module, or a conditional Bicep output. `ApplyOutputsMapping` below still skips those silently, and `materializeRecipeSecrets` still reads it as "no secrets" and clears `properties.secrets`. Same bad outcome as the issue, just a narrower path.

pkg/recipes/util/outputs_test.go
```diff
@@ -20,8 +20,52 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+func Test_ValidateOutputsMapping(t *testing.T) {
```
kachawla, line 26: Nice that this locks in the sort order - easy thing to break by accident. Could you add a case with an empty output name? `AddMappedOutputs` already rejects those with a different message (`config/config.go:325`), so it would be worth settling which of the two should catch it.

pkg/recipes/driver/bicep/bicep.go
AzureMike: Agreed.

pkg/recipes/driver/bicep/bicep.go
AzureMike: Great catch

pkg/recipes/driver/terraform/terraform_test.go
AzureMike: Good point, I've fixed the tests.

pkg/recipes/terraform/execute.go
AzureMike: I agree a bool could work, but I used the named constants for readability/maintainability.

pkg/recipes/terraform/execute.go
AzureMike: Agreed

pkg/recipes/terraform/execute.go
AzureMike: I updated configuration generation so explicit mappings take precedence over `result`, matching `prepareRecipeResponse`. Mapped outputs now reach Terraform state instead of being silently dropped, while deletion behavior remains unchanged.

pkg/recipes/util/outputs.go
AzureMike: @kachawla So which behavior would you prefer here: fail all undeclared mappings, or fail secret mappings and warn for non-secret mappings?

pkg/recipes/util/outputs.go
AzureMike: I initially thought this was out of scope, but I have implemented it.

pkg/recipes/util/outputs.go
kachawla: Fail all of them - what you have now is good. Nothing has deployed at that point, so being strict just costs a one-line pack edit.

The runtime split you added is exactly right though - skip a null value, fail a null secret. Thanks for adding that.

kachawla (approved):

PR #12742, opened 2026-08-20
Fix custom recipe pack reconciliation on repeat deploys

### LINKED ISSUE #323: Custom recipe pack is dropped from the environment on every deploy after the first
## Summary

The `apply-custom-recipe-packs` composite action attaches a repo's custom recipe pack to the environment **only on the first deploy**. On every subsequent deploy the pack is silently dropped from the environment's recipe pack list, so any `Radius.Resources/*` custom type fails with:

```
"code": "Internal",
"message": "no recipe pack found with recipe for resource type \"Radius.Resources/azureServiceBusNamespaces\""
```

## Root cause

The action derives the packs to attach by diffing the pack list before and after deploying the custom pack:

```bash
NEW_PACKS=$(jq -nc --argjson before "$PACKS_BEFORE" --argjson after "$PACKS_AFTER" '$after - $before')
...
PACK_IDS=$(jq -nr --argjson existing "$EXISTING_PACKS" --argjson new "$NEW_PACKS" '($existing + $new) | unique | join(",")')
rad env update "$ENVIRONMENT" --recipe-packs "$PACK_IDS" --preview
```

This assumes deploying the custom pack always *creates* it. It doesn't:

- **First run** — the pack doesn't exist, so `$after - $before` yields it, and it gets attached.
- **Later runs** — `rad startup` restores the control-plane state, so the pack already exists. Deploying it is an idempotent update, `$after - $before` is `[]`, and `NEW_PACKS` is empty.

`EXISTING_PACKS` doesn't compensate, because the earlier "create environment and recipe pack" step recreates the environment with only `azure-avm` attached. So `existing + new` collapses to just `azure-avm`, and since `rad env update --recipe-packs` **replaces** the list (the action's own output warns `The existing recipe pack list will be replaced with the specified packs`), the custom pack is removed.

## Evidence

Same repo, same `custom-recipe-pack.bicep`, two consecutive runs:

Run 1 — https://github.com/kachawla/todo-list-app/actions/runs/31548618151
```
Attaching recipe packs to environment 'test': .../recipePacks/azure-avm,.../recipePacks/custom-recipe-pack
```
The Service Bus recipe resolved and ran (it failed later, for unrelated reasons).

Run 2 — https://github.com/kachawla/todo-list-app/actions/runs/31550360808
```
Resources:
    custom-recipe-pack Radius.Core/recipePacks
Attaching recipe packs to environment 'test': .../recipePacks/azure-avm
```
The pack was deployed successfully and then not attached, and the deploy failed with `no recipe pack found with recipe for resource type`.

Note run 2 deployed the pack fine — this is purely the attach diff being wrong.

## Impact

Any application using a custom resource type fails on its **second and every later deploy**, with an error that points at the resource type rather than at the attach step. It looks like a modeling error, so the natural reaction is to go edit `app.bicep`, which cannot fix it.

It also defeats the deploy repair loop: the loop is handed a failure it cannot repair from the application model.

## Suggested fix

Don't diff. The action already knows which packs the template defines, so attach them by identity instead of by "what appeared":

- Parse the deployed pack names out of the `rad deploy` result (or the compiled template) and use those as `NEW_PACKS`, so the set is the same whether the deploy created or updated them.
- Then `PACK_IDS=($EXISTING_PACKS + $DECLARED_PACKS) | unique` as today.

A regression test that runs the same deploy twice against restored state would have caught this.

## Workaround

Rename the pack resource in `custom-recipe-pack.bicep` before each deploy so it reads as new. This is obviously not something users should have to do.

sk593:
## Summary

- derive custom recipe-pack identities from the compiled authored Bicep instead of a global before/after list diff
- preserve existing environment recipe packs while attaching only the exact authored packs on first and restored-state redeploys
- add composite-action regression coverage and update the Repo Radius workflow documentation

## Testing

- `make test-apply-custom-recipe-packs`
- `shellcheck .github/extension/actions/apply-custom-recipe-packs/apply-custom-recipe-packs_test.sh`

Resolves radius-project/ai-extensions#323


.github/extension/actions/apply-custom-recipe-packs/action.yml
```diff
@@ -37,76 +37,130 @@ runs:
...
+              (.name | type) != "string" or
+              (.name | length) == 0 or
+              (.name | startswith("["))
+            ) then
+              error("recipe pack names must be non-empty string literals")
+            else
+              map(.name) | unique
+            end
+          end
+        ' "${COMPILED_TEMPLATE}")
```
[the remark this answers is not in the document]
sk593: Fixed in `e5b3623` and `c4b02d4`. The jq now extracts the name from `.properties.name` with a `.name` fallback, and accepts both object-keyed (languageVersion 2.0) and array-shaped (languageVersion 1.x) `.resources`.

.github/extension/actions/apply-custom-recipe-packs/apply-custom-recipe-packs_test.sh
```diff
@@ -0,0 +1,378 @@
...
+}
+
+single_pack_template() {
+    write_compiled_template '{
+      "custom": {
+        "type": "Radius.Core/recipePacks@2025-08-01-preview",
+        "name": "custom"
+      }
+    }'
+}
```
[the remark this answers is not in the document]
sk593: Fixed in `e5b3623`. All fixtures now use the symbolic-name layout (`properties.name`), and `c4b02d4` adds an array-shaped fixture too.

.github/extension/actions/apply-custom-recipe-packs/action.yml
```diff
@@ -37,76 +37,144 @@ runs:
...
+          if (.resources | type) as $shape |
+             ($shape != "object" and $shape != "array") then
+            error("compiled template resources must be an object or array")
+          else
+            [.resources[] |
+              select(
+                ((.type // "") | ascii_downcase) as $type |
+                $type == "radius.core/recipepacks" or
+                ($type | startswith("radius.core/recipepacks@"))
+              )
```
nicolejms: Non-blocking: this selector matches on type alone, so an `existing` recipe-pack reference in `custom-recipe-pack.bicep` would also be discovered and attached to the environment. In compiled ARM, `existing` resources are still emitted into `.resources` (languageVersion 2.0 marks them with `"existing": true`), so a template that merely *references* a pack it doesn't author would silently pull that pack into the environment's `recipePacks` list.

I suspect that may actually be intended — referencing a pre-existing shared pack and attaching it is a reasonable thing to want. But it's currently unspecified in the action's header comment and in the design note, and untested, so it reads as accidental rather than deliberate.

Could you add a regression case to `apply-custom-recipe-packs_test.sh` that pins whichever behavior you want? Something like:

```bash
reset_case
write_compiled_template '{
  "shared": {
    "import": "radius",
    "existing": true,
    "type": "Radius.Core/recipePacks@2025-08-01-preview",
    "properties": { "name": "shared" }
  },
  "custom": {
    "import": "radius",
    "type": "Radius.Core/recipePacks@2025-08-01-preview",
    "properties": { "name": "custom", "properties": { "recipes": {} } }
  }
}'
run_action
```

...then assert either that `shared` is attached (intended), or that it's filtered out — e.g. adding `select(has("existing") | not)` to the discovery jq. Either way the contract gets documented and locked down against future jq edits.

.github/extension/actions/apply-custom-recipe-packs/action.yml
```diff
@@ -37,76 +37,144 @@ runs:
...
+          jq -nr \
+            --argjson existing "${EXISTING_PACKS}" \
+            --argjson custom "${CUSTOM_PACKS}" \
+            '($existing + $custom) | unique | join(",")'
+        )
 
-        if [ -z "$PACK_IDS" ]; then
-          echo "No recipe packs found to attach to environment '$ENVIRONMENT'." >&2
+        if [[ -z "${PACK_IDS}" ]]; then
+          echo "No recipe packs found to attach to environment '${ENVIRONMENT}'." >&2
```
nicolejms: Non-blocking: this guard looks unreachable now. The discovery jq already `error`s when the authored pack set is empty, and `jq -er` validates each resolved ID is a non-empty string, so `CUSTOM_PACKS` is guaranteed non-empty by the time we reach here — which makes `PACK_IDS` non-empty too, regardless of what `EXISTING_PACKS` contains.

It was meaningful under the old before/after diff approach, where `NEW_PACKS` could legitimately come back as `[]`. Harmless to keep as a belt-and-braces check, but if you do, a one-line comment noting it's defensive would save the next reader from hunting for the input that triggers it.

nicolejms (commented):
Two non-blocking observations on the discovery/attach path. Neither should hold up the merge — the fix itself is correct and the redeploy path is well covered. The first one would benefit from a regression test to pin the intended contract.

.github/extension/actions/apply-custom-recipe-packs/action.yml
sk593: Removed the unreachable empty pack-list guard in `cbe2129ea`.

.github/extension/actions/apply-custom-recipe-packs/action.yml
sk593: Filtered `existing` recipe-pack references from discovery and added regression coverage in `cbe2129ea`.

nicolejms (dismissed):

brooke-hamilton (approved):

PR #12764, opened 2026-08-21
Fix Azure verification during RBAC propagation

ryanwaite:
## Summary

Updates the canonical Repo Radius Azure credential-verification workflow so newly-created role assignments can propagate without skipping end-to-end verification.

`azure/login` now authenticates at tenant scope with the action's exact `allow-no-subscriptions: true` input. The next step checks subscription visibility immediately, refreshes Azure CLI account state on every attempt, compares the configured subscription ID with `jq`, and selects it before the existing Azure credential, AKS, and GHCR checks continue unchanged. A step-level 12-minute timeout bounds the wall clock if an Azure CLI refresh hangs.

## Reason for change

This supersedes [radius-project/ai-extensions#444](https://github.com/radius-project/ai-extensions/pull/444). That PR avoids the propagation race by suppressing the immediate verification dispatch in the frontend after it creates a Contributor role assignment. Skipping the dispatch removes the end-to-end signal and puts Azure-specific workflow timing policy in the wrong repository. Handling propagation in `.github/extension/verify-azure.yml` keeps verification enabled and makes the canonical upstream workflow responsible for its own Azure subscription-visibility semantics.

The retry is bounded and is not a fixed startup delay:

- Check immediately after OIDC login.
- Retry up to 23 attempts with 5, 10, 20, then 30-second capped delays (10 minutes 5 seconds of total waiting before the final check).
- Run `az account list --refresh` on every attempt without printing its JSON response.
- Select only the configured subscription after an exact, case-insensitive ID match.
- Apply a 12-minute step timeout so a hung CLI call cannot fall through to GitHub's multi-hour job default.

Genuine OIDC, client, tenant, or federated-credential errors still fail immediately in `azure/login`. A missing subscription variable fails before discovery starts, and a direct nonzero exit from `az account list --refresh` fails immediately under `set -euo pipefail`. Azure CLI can internally absorb some subscription-discovery exceptions and return successfully with cached tenant-level state; persistent discovery failure therefore reaches the same bounded no-visible-subscription timeout as RBAC propagation or a missing role assignment. The timeout diagnostic names all three possibilities.

## How to test

Exact focused validation run after the review fixes:

- `make test-verify-azure` — passed (`Azure verification workflow tests passed`).
- `shellcheck --rcfile .github/linters/.shellcheckrc .github/extension/verify-azure_test.sh` — passed with no findings.
- `ruby -e "require 'yaml'; data=YAML.load_file('.github/extension/verify-azure.yml'); abort 'missing jobs' unless data['jobs']; puts 'workflow structure parsed'"` — passed (`workflow structure parsed`).
- `npx --yes markdown-table-formatter@1.7.0 '.github/extension/README.md' 'docs/contributing/contributing-deploy-environments.md' --check` — passed; no formatting changes required.
- `npx --yes markdownlint-cli2@0.23.2 '.github/extension/README.md' 'docs/contributing/contributing-deploy-environments.md' --config './.github/linters/.markdownlint-cli2.yaml'` — passed with 0 issues.
- `npx --yes cspell@9.6.1 lint --config ./.github/linters/.cspell.yml --no-progress '.github/extension/README.md' 'docs/contributing/contributing-deploy-environments.md'` — passed with 0 issues.

The deterministic workflow test extracts and executes the real retry `run:` block with stubbed `az` and `sleep` commands. It covers immediate visibility, case-insensitive safe matching, missing configuration, delayed visibility and exponential backoff, a retry budget of at least 10 minutes, timeout without selecting a different subscription, direct Azure CLI failure, and the `azure/login` input contract. Its named wait-step assertion also pins `timeout-minutes: 12` and the exact `AZURE_SUBSCRIPTION_ID: ${{ vars.AZURE_SUBSCRIPTION_ID }}` environment mapping.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/verify-azure.yml` | Authenticates without requiring immediate subscription visibility, then refreshes, retries for at least 10 minutes, matches, and selects the configured subscription under a 12-minute hard timeout. |
| `.github/extension/verify-azure_test.sh` | Adds deterministic behavior tests and named-step contract assertions for login inputs, retry duration, timeout, and subscription-variable wiring. |
| `build/test.mk` | Adds the focused test target to the standard unit-test prerequisites. |
| `.github/extension/README.md` | Documents tenant-scope login, bounded propagation/discovery retry, hard timeout, selection, and failure behavior. |
| `docs/contributing/contributing-deploy-environments.md` | Updates the contributor workflow and troubleshooting guidance for Azure RBAC propagation and subscription-discovery failure. |


.github/extension/verify-azure.yml
```diff
@@ -80,7 +80,47 @@ jobs:
...
+            exit 1
+          fi
+
+          delay_seconds="${INITIAL_DELAY_SECONDS}"
+          for ((attempt = 1; attempt <= MAX_ATTEMPTS; attempt++)); do
+            subscriptions="$(az account list --refresh --only-show-errors --output json)"
+            if jq -e --arg subscription_id "${AZURE_SUBSCRIPTION_ID}" \
+              'any(.[]; ((.id? // "") | ascii_downcase) == ($subscription_id | ascii_downcase))' \
+              <<<"${subscriptions}" >/dev/null; then
+              az account set --subscription "${AZURE_SUBSCRIPTION_ID}"
```
DariuszPorowski, line 116: praise: Match and select only the configured Azure subscription

The exact case-insensitive comparison prevents verification from succeeding against an unrelated visible subscription.

.github/extension/verify-azure_test.sh
```diff
@@ -0,0 +1,283 @@
...
+}
+
+if ! command -v jq >/dev/null 2>&1; then
+    fail "jq is required; run 'make install-jq'"
+fi
+if ! command -v python3 >/dev/null 2>&1; then
+    fail "python3 is required to parse verify-azure.yml"
+fi
+
+extract_wait_step() {
```
DariuszPorowski, line 34: praise(test): Execute the workflow's real retry block in tests

Extracting the shipped run block prevents the harness from drifting into a separate implementation.

DariuszPorowski (dismissed):
praise: Preserve end-to-end verification during Azure RBAC propagation

The tenant-scope login and bounded subscription discovery retry fix the propagation race without suppressing the Azure, AKS, or GHCR checks.

.github/extension/verify-azure.yml
```diff
@@ -80,7 +80,47 @@ jobs:
...
+        env:
+          AZURE_SUBSCRIPTION_ID: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+        run: |
+          set -euo pipefail
+
+          readonly MAX_ATTEMPTS=23
+          readonly INITIAL_DELAY_SECONDS=5
+          readonly MAX_DELAY_SECONDS=30
+
+          if [[ -z "${AZURE_SUBSCRIPTION_ID}" ]]; then
```
kachawla, line 101: Could we validate that `AZURE_SUBSCRIPTION_ID` is a GUID before entering the retry loop? This only rejects the empty value, so a typo such as a truncated ID is retried 23 times even though propagation can never make it match; previously `azure/login` attempted to select the configured subscription and failed immediately. A malformed environment value now consumes roughly 10 minutes before reporting a generic visibility/RBAC error.

ryanwaite: Good catch. Added canonical GUID validation before the retry loop in `24f44be1f`, so malformed nonempty values now fail immediately with a configuration-specific error. The focused test also verifies that invalid input calls neither `az` nor `sleep`. `make test-verify-azure`, ShellCheck, Bash syntax validation, and the workflow YAML parse all pass.

kachawla (approved):

PR #12765, opened 2026-08-21
fix(test): stop AWS functional test cleanup from failing without waiting

lakshmimsft:
## Summary

Attempts to fix the intermittent `Test_AWS_LogsLogGroup` failures in the `corerp-cloud`
functional test job by making AWS resource cleanup actually wait for the
CloudControl delete to finish, and by making the post-delete existence check
retry on transient errors instead of giving up after the first one.

## Reason for change
`Test_AWS_LogsLogGroup` has been failing intermittently in `corerp-cloud` since
#12706 (`chore(deps): bump the go-deps group`, merged 2026-08-18), which moved
`github.com/aws/aws-sdk-go-v2/service/cloudcontrol` from `v1.32.4` to `v1.32.5`.

Example failure:
[run 32398057952](https://github.com/radius-project/radius/actions/runs/32398057952/job/96521782534)

```
rptest.go:572: failed to delete radiusfunctionaltest-b143eea2-...: failed to
  delete resource radiusfunctionaltest-b143eea2-... after 5m0s: not found: WaiterLogger
rptest.go:589: checking existence of resource radiusfunctionaltest-b143eea2-... failed
  with err: operation error CloudControl: GetResource, https response error
  StatusCode: 400, InvalidRequestException: AWS::Logs::LogGroup Handler returned
  status FAILED: Log group cannot be found.
rptest.go:599: Error: Should be true
  Messages: AWS resource radiusfunctionaltest-b143eea2-... was present, should be not found
```

## How to test

The affected test only runs against real AWS in the `corerp-cloud` functional
test job, so the primary verification is CI:

1. Confirm `Run corerp-cloud functional tests` passes, and that
   `Test_AWS_LogsLogGroup` no longer logs `not found: WaiterLogger`.
2. Confirm the cleanup log now shows the waiter actually polling before
   `validating deletion of AWS resource ... (attempt 1/5)` is reached.

The waiter regression itself can be reproduced locally without AWS credentials,
against the versions pinned in `go.mod` (`cloudcontrol v1.32.5`,
`smithy-go v1.27.7`) — the failure happens while building the middleware stack,
before any credentials are used:

```go
c := cloudcontrol.NewFromConfig(aws.Config{
    Region:      "us-west-2",
    Credentials: credentials.NewStaticCredentialsProvider("AKIAFAKE", "fake", ""),
})
tok := "faketoken"

// Before this change: returns "not found: WaiterLogger" immediately.
before := cloudcontrol.NewResourceRequestSuccessWaiter(c, func(o *cloudcontrol.ResourceRequestSuccessWaiterOptions) {
    o.LogWaitAttempts = true
})
fmt.Println(before.Wait(ctx, &cloudcontrol.GetResourceRequestStatusInput{RequestToken: &tok}, 300*time.Second))

// After this change: actually issues the request (fails only on the fake credentials).
after := cloudcontrol.NewResourceRequestSuccessWaiter(c)
fmt.Println(after.Wait(ctx, &cloudcontrol.GetResourceRequestStatusInput{RequestToken: &tok}, 300*time.Second))
```

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `test/validation/aws.go` | Removed `LogWaitAttempts = true` from the `ResourceRequestSuccess` waiter in `DeleteAWSResource` so the waiter builds its middleware stack successfully and genuinely waits up to 5 minutes for the delete. Added a comment recording why the option must stay off. |
| `test/rp/rptest.go` | AWS deletion validation now retries on error across all `AWSDeletionRetryLimit` attempts instead of breaking on the first one, skips the trailing sleep on the final attempt, and includes the last observed error in the `require.Truef` failure message via a new `lastErr` variable. |

### DIFF
--- a/test/rp/rptest.go
+++ b/test/rp/rptest.go
@@ -575,28 +575,36 @@ func (ct RPTest) Test(t *testing.T) {
 					// Ensure that the resource is deleted with retries
 					notFound := false
 					baseWaitTime := 15 * time.Second
+					var lastErr error
 
 					for attempt := 1; attempt <= AWSDeletionRetryLimit; attempt++ {
 						t.Logf("validating deletion of AWS resource for %s (attempt %d/%d)", ct.Description, attempt, AWSDeletionRetryLimit)
 
 						// Use AWS CloudControl.Get method to validate that the resource is deleted
-						notFound, err = validation.IsAWSResourceNotFound(ctx, &resource, ct.Options.AWSClient)
+						notFound, lastErr = validation.IsAWSResourceNotFound(ctx, &resource, ct.Options.AWSClient)
 
 						if notFound {
 							t.Logf("AWS resource %s to be deleted was not found", resource.Identifier)
 							break
-						} else if err != nil {
-							t.Logf("checking existence of resource %s failed with err: %s", resource.Name, err)
-							break
-						} else {
-							// Wait with exponential backoff
-							waitTime := baseWaitTime * time.Duration(attempt)
-							t.Logf("waiting for %s before next attempt", waitTime)
-							time.Sleep(waitTime)
 						}
+
+						if lastErr != nil {
+							// A delete that is still in flight can surface as a transient error rather than
+							// ResourceNotFoundException, so keep retrying instead of failing immediately.
+							t.Logf("checking existence of resource %s failed with err: %s", resource.Name, lastErr)
+						}
+
+						if attempt == AWSDeletionRetryLimit {
+							break
+						}
+
+						// Wait with exponential backoff
+						waitTime := baseWaitTime * time.Duration(attempt)
+						t.Logf("waiting for %s before next attempt", waitTime)
+						time.Sleep(waitTime)
 					}
 
-					require.Truef(t, notFound, "AWS resource %s was present, should be not found", resource.Identifier)
+					require.Truef(t, notFound, "AWS resource %s was present, should be not found (last error: %v)", resource.Identifier, lastErr)
 					t.Logf("finished validation of deletion of AWS resource %s for %s", resource.Name, ct.Description)
 				} else {
 					t.Logf("skipping deletion of %s", resource.Name)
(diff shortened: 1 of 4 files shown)


test/rp/rptest.go
```diff
@@ -575,28 +575,36 @@ func (ct RPTest) Test(t *testing.T) {
...
+						notFound, lastErr = validation.IsAWSResourceNotFound(ctx, &resource, ct.Options.AWSClient)
 
 						if notFound {
 							t.Logf("AWS resource %s to be deleted was not found", resource.Identifier)
 							break
-						} else if err != nil {
-							t.Logf("checking existence of resource %s failed with err: %s", resource.Name, err)
+						}
+
+						if lastErr != nil {
```
brooke-hamilton, line 591: This PR addresses two separate cleanup failures, but only the first is fixed here: removing `LogWaitAttempts` fixes the SDK's immediate `not found: WaiterLogger` failure, while the post-delete `GetResource` path still misclassifies an already-absent `AWS::Logs::LogGroup`. In the failing runs, CloudControl returns `InvalidRequestException` with `HandlerErrorCode: InvalidRequest` and either "Log group cannot be found" or "The specified log group does not exist"; `IsAWSResourceNotFound` still returns false for that response. Retrying it five times therefore only delays the same assertion by 150 seconds. Please include the narrow missing-LogGroup classifier tracked by #12767 in this PR, with tests that leave unrelated `InvalidRequestException` errors as failures, so the PR fully resolves the flake described by its title.

brooke-hamilton (commented):
The waiter fix is valid, but the related missing-LogGroup classification path should be completed in this PR.

test/rp/rptest.go
lakshmimsft: added the classifier.

brooke-hamilton (dismissed):

brooke-hamilton (approved):

PR #12771, opened 2026-08-21
Add CI coverage for database.enabled=true install

### LINKED ISSUE #12227: Add CI coverage for database.enabled=true install
## Overview

PR #12214 added `database.enabled=true` support (PostgreSQL-backed control plane for Repo Radius / `rad startup` / `rad shutdown`). During review, two blocking bugs in the database-enabled path (a non-existent mirror image tag → `ImagePullBackOff`, and missing per-RP GRANTs → `permission denied for table resources`) reached the PR because **no CI job exercises `database.enabled=true`** — the default functional suite installs with `database.enabled=false`.

PR #12214 added cheap regression guards via `helm template` assertions (image reference + GRANT statements), but those only validate rendering, not an actual install-and-run.

## Acceptance criteria

- A CI job installs Radius with `database.enabled=true` on a kind cluster and confirms the control plane reaches a healthy state (pods Ready, no `ImagePullBackOff`, no `permission denied` crash).
- Ideally exercises a minimal deploy so the PostgreSQL store is actually written and read.
- Runs on PRs that touch the Helm chart database templates and/or the database provider code.

## Additional context

- Related: #8096 (External data store for Radius), #8398 (Postgres DB initialization)
- Source PR: #12214
- The env-gated end-to-end lifecycle test at `test/functional-portable/statestore` (`RADIUS_STATE_E2E`) already exercises install→deploy→shutdown→startup with `database.enabled=true`; this issue may be satisfiable by wiring that test into CI once the cluster-create workflow lands.

nellshamrell:
## Summary

Adds a dedicated `database-noncloud` leg to the non-cloud functional test workflow that installs Radius with the PostgreSQL-backed control plane (`--set database.enabled=true`) and runs a small test package against it.

The leg installs via the existing shared "Install Radius" step with one extra Helm flag, then runs two tests:

- **`Test_DatabaseEnabled_MinimalDeploy`** — the primary coverage. Deploys an application and container and validates both the Radius resources and the rendered Kubernetes objects. The write-then-read round-trips through the resource providers' own PostgreSQL identities, which is exactly the path that broke when the per-RP users lacked privileges on the `resources` table.
- **`Test_DatabaseEnabled_ControlPlaneHealthy`** — asserts the database StatefulSet is serving, every resource provider Deployment reaches `Available`, and no provider logged `permission denied for table`. It also guards that the suite is running against a database-enabled install rather than quietly passing against the default API server store.

## Reason for change

Two bugs reached review of #12214 because no CI job installed Radius with `database.enabled=true`:

1. The database StatefulSet referenced a registry-mirror postgres tag that did not exist → `ImagePullBackOff`.
2. The per-RP database users were never `GRANT`ed access to the shared `resources` table → `permission denied for table resources`, crash-looping the provider.

This is the third of three complementary layers, and deliberately the cheapest and most direct signal:

| Layer | What it guards |
| ----- | -------------- |
| `deploy/Chart/tests/database_test.yaml` | *Rendering* of both fixes, via helm-unittest. |
| `statestore-noncloud` leg | Also installs with `database.enabled=true`, but as part of a 40-minute destructive shutdown/startup lifecycle whose failures are ambiguous. |
| **`database-noncloud` leg (this PR)** | A plain install-and-deploy check, so a failure means the PostgreSQL-backed control plane itself is broken. |

Note that `rad install` waits on the Helm release, so both original bugs would fail the install step before these tests run. The lasting value is the deploy round-trip through the PostgreSQL store.

Fixes #12227

## How to test

In CI, the new `database-noncloud` matrix leg runs automatically on this PR.

Locally:

```bash
rad install kubernetes --set database.enabled=true
make test-functional-database-noncloud
```

The tests fail against a default (API server-backed) install, so the group is excluded from `test-functional-all-noncloud`.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/workflows/functional-test-noncloud.yaml` | Add `database-noncloud` to the matrix; add a leg-gated step writing `DATABASE_INSTALL_ARGS=--set database.enabled=true` to `$GITHUB_ENV`; append it to `RAD_COMMAND`. |
| `build/test.mk` | Add the `test-functional-database-noncloud` target, kept out of `test-functional-all-noncloud`. |
| `test/functional-portable/database/noncloud/database_deploy_test.go` | New `Test_DatabaseEnabled_MinimalDeploy`, modelled on the existing corerp `Test_Container`. |
| `test/functional-portable/database/noncloud/database_install_test.go` | New `Test_DatabaseEnabled_ControlPlaneHealthy` plus the package doc. |
| `test/functional-portable/database/noncloud/testdata/database-enabled-container.bicep` | Minimal application + container fixture for the deploy test. |
| `docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md` | Document the new group and its `database.enabled=true` install prerequisite. |

### DIFF
--- a/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
+++ b/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
@@ -71,6 +71,8 @@ rad install kubernetes --set database.enabled=true
 
 Running `make test-functional-database-noncloud` against a default (API server-backed) install fails, because the tests check that the PostgreSQL StatefulSet is present before asserting anything else.
 
+Use a cluster that has never had a default install. The tests also assert that the API server store holds no objects, which is how they prove the resource providers actually switched to PostgreSQL rather than only running it alongside. Switching providers does not migrate or delete rows already written to the API server store, so a cluster that was previously installed without `database.enabled=true` keeps those objects and fails that assertion even when the PostgreSQL-backed control plane is healthy. CI creates a fresh cluster for every run.
+
 For multicluster tests, create the namespace and Secret before installing Radius. The kubeconfig stored in the Secret must use an API-server address reachable from the Radius pods:
 
 ```bash
--- a/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
+++ b/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
@@ -46,7 +46,7 @@ make test-functional-daprrp
 make test-functional-datastoresrp
 ```
 
-To run a single group directly, call its `make` target — for example `make test-functional-corerp-noncloud` for the non-cloud Core RP tests, or `make test-functional-all-noncloud` for the standard non-cloud groups. The groups (`ucp`, `kubernetes`, `corerp`, `cli`, `msgrp`, `daprrp`, `datastoresrp`, `dynamicrp`, `samples`, `upgrade`, `multicluster`, and `statestore`) and the variants each group supports are defined in [`build/test.mk`](../../../../build/test.mk).
+To run a single group directly, call its `make` target — for example `make test-functional-corerp-noncloud` for the non-cloud Core RP tests, or `make test-functional-all-noncloud` for the standard non-cloud groups. The groups (`ucp`, `kubernetes`, `corerp`, `cli`, `msgrp`, `daprrp`, `datastoresrp`, `dynamicrp`, `samples`, `upgrade`, `multicluster`, `database`, and `statestore`) and the variants each group supports are defined in [`build/test.mk`](../../../../build/test.mk).
 
 You can also run or debug individual tests from VS Code.
 
@@ -54,13 +54,22 @@ You can also run or debug individual tests from VS Code.
 
 The aggregate `make test-functional-all-noncloud` target intentionally excludes these isolated groups:
 
-| Target                                       | Requirements and behavior                                                                                                                      |
-|----------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|
-| `make test-functional-multicluster-noncloud` | Requires a second Kubernetes cluster, a target-cluster Secret mounted into Radius, and `RADIUS_TEST_EXTERNAL_KUBECONFIG` for the test process. |
-| `make test-functional-statestore-noncloud`   | Destructive lifecycle test that installs, purges, and reinstalls Radius. Run it only on a dedicated cluster.                                   |
-| `make test-functional-upgrade-noncloud`      | Exercises the Radius upgrade path and performs its own install/upgrade lifecycle.                                                              |
+| Target                                       | Requirements and behavior                                                                                                                          |
+|----------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| `make test-functional-multicluster-noncloud` | Requires a second Kubernetes cluster, a target-cluster Secret mounted into Radius, and `RADIUS_TEST_EXTERNAL_KUBECONFIG` for the test process.     |
+| `make test-functional-database-noncloud`     | Requires Radius installed with `--set database.enabled=true` (PostgreSQL-backed control plane) instead of the default Kubernetes API server store. |
+| `make test-functional-statestore-noncloud`   | Destructive lifecycle test that installs, purges, and reinstalls Radius. Run it only on a dedicated cluster.                                       |
+| `make test-functional-upgrade-noncloud`      | Exercises the Radius upgrade path and performs its own install/upgrade lifecycle.                                                                  |
 
-The multicluster and statestore groups run as isolated CI legs in `functional-test-noncloud.yaml`; do not run them against a shared development cluster.
+The multicluster, database, and statestore groups run as isolated CI legs in `functional-test-noncloud.yaml`; do not run them against a shared development cluster.
+
+For database tests, install Radius with the PostgreSQL-backed control plane first:
+
+```bash
+rad install kubernetes --set database.enabled=true
+```
+
+Running `make test-functional-database-noncloud` against a default (API server-backed) install fails immediately, because the tests check that the PostgreSQL StatefulSet is present before asserting anything else.
 
 For multicluster tests, create the namespace and Secret before installing Radius. The kubeconfig stored in the Secret must use an API-server address reachable from the Radius pods:
 
--- a/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
+++ b/docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
@@ -69,7 +69,7 @@ For database tests, install Radius with the PostgreSQL-backed control plane firs
 rad install kubernetes --set database.enabled=true
 ```
 
-Running `make test-functional-database-noncloud` against a default (API server-backed) install fails immediately, because the tests check that the PostgreSQL StatefulSet is present before asserting anything else.
+Running `make test-functional-database-noncloud` against a default (API server-backed) install fails, because the tests check that the PostgreSQL StatefulSet is present before asserting anything else.
 
 For multicluster tests, create the namespace and Secret before installing Radius. The kubeconfig stored in the Secret must use an API-server address reachable from the Radius pods:
 
--- a/test/functional-portable/database/noncloud/database_install_test.go
+++ b/test/functional-portable/database/noncloud/database_install_test.go
@@ -18,22 +18,31 @@ limitations under the License.
 // PostgreSQL-backed control plane (`--set database.enabled=true`).
 //
 // The default functional suite installs with `database.enabled=false` (the Kubernetes API server
-// store), so nothing exercised the PostgreSQL path end to end. Two blocking bugs reached review of
-// PR #12214 as a result: the database StatefulSet pointed at a registry-mirror tag that did not
-// exist (ImagePullBackOff), and the per-RP database users were never granted access to the
-// `resources` table (`permission denied for table resources`). See issue #12227.
+// store). Two blocking bugs reached review of PR #12214 as a result: the database StatefulSet
+// pointed at a registry-mirror tag that did not exist (ImagePullBackOff), and the per-RP database
+// users were never granted access to the `resources` table (`permission denied for table
+// resources`). See issue #12227.
 //
-// The chart's helm-unittest suite (deploy/Chart/tests/database_test.yaml) guards the *rendering* of
-// both fixes. This package guards the *runtime*: the control plane must actually come up against
-// PostgreSQL, and a real deployment must round-trip through the PostgreSQL store.
+// This is one of three complementary layers, and deliberately the cheapest and most direct:
+//
+//   - deploy/Chart/tests/database_test.yaml guards the *rendering* of both fixes with helm-unittest.
+//   - The statestore-noncloud leg also installs with database.enabled=true, but as part of a
+//     40-minute destructive shutdown/startup lifecycle whose failures are ambiguous.
+//   - This package is a plain install-and-deploy check, so a failure here means the
+//     PostgreSQL-backed control plane itself is broken.
+//
+// Note that `rad install` waits on the Helm release, so both of the original bugs would already
+// fail the install rather than reaching these tests. The lasting value here is
+// Test_DatabaseEnabled_MinimalDeploy: a real deployment that round-trips through the PostgreSQL
+// store using the resource providers' own database identities.
 //
 // These tests assume Radius is ALREADY installed with the database enabled:
 //
 //	rad install kubernetes --set database.enabled=true
 //
-// Test_DatabaseEnabled_ControlPlaneHealthy fails immediately if it is not, so the suite cannot
-// quietly pass against an apiserver-backed control plane. In CI the `database-noncloud` matrix leg
-// of .github/workflows/functional-test-noncloud.yaml performs that install.
+// Test_DatabaseEnabled_ControlPlaneHealthy fails if it is not, so the suite cannot quietly pass
+// against an apiserver-backed control plane. In CI the `database-noncloud` matrix leg of
+// .github/workflows/functional-test-noncloud.yaml performs that install.
 package database
 
 import (
@@ -45,6 +54,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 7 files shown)


test/functional-portable/database/noncloud/database_install_test.go
```diff
@@ -0,0 +1,356 @@
...
+func Test_DatabaseEnabled_ControlPlaneHealthy(t *testing.T) {
+	ctx, cancel := context.WithTimeout(t.Context(), testTimeout)
+	defer cancel()
+
+	k8s := test.NewTestOptions(t).K8sClient
+
+	// The database StatefulSet must exist at all. If it does not, Radius was installed without
+	// database.enabled=true and the rest of this package proves nothing — fail loudly rather than
+	// pass a meaningless run.
+	statefulSet, err := k8s.AppsV1().StatefulSets(radiusNamespace).Get(ctx, databaseStatefulSetName, metav1.GetOptions{})
```
DariuszPorowski: suggestion(test,non-blocking): Assert the store, not only the StatefulSet

This infers "we're running against PostgreSQL" from the presence of the `database` StatefulSet. A cluster could have that StatefulSet running while the providers still use the apiserver store, and the suite would pass.

Reading the deployed resource back out of the `resources` table, or asserting the apiserver-store CRDs hold no rows for it, would make `Test_DatabaseEnabled_MinimalDeploy` prove the round-trip its own doc comment claims.

test/functional-portable/database/noncloud/database_install_test.go
```diff
@@ -0,0 +1,356 @@
...
+var controlPlaneComponents = []string{"ucp", "applications-rp", "dynamic-rp"}
+
+// Test_DatabaseEnabled_ControlPlaneHealthy asserts that the PostgreSQL-backed control plane came up
+// cleanly: the database is serving, every resource provider reached Available, and no provider was
+// denied access to its tables.
+//
+// It also acts as the guard that this package is running against the right install. Because
+// Test_DatabaseEnabled_MinimalDeploy calls t.Parallel() by way of the RPTest harness and pauses
+// before deploying, this test runs to completion first, so a run against a default
+// apiserver-backed control plane fails here rather than after a misleading successful deployment.
```
DariuszPorowski: suggestion(non-blocking,if-minor): Make the ordering dependency explicit rather than implicit

This comment is accurate: `database_deploy_test.go` sorts first, hits `t.Parallel()` at `test/rp/rptest.go:446`, and parks, so this test finishes first. But the chain rests on file names and on `RunSerial` staying false.

Calling the StatefulSet precondition from `Test_DatabaseEnabled_MinimalDeploy` too, via a package-level `sync.Once`, would make the guard hold regardless of ordering.

test/functional-portable/database/noncloud/database_install_test.go
```diff
@@ -0,0 +1,356 @@
...
+			require.Failf(t, "the resource providers never became available",
+				"still not Available after %s: %s. A resource provider that cannot reach its PostgreSQL tables crash-loops on startup.",
+				controlPlaneReadyTimeout, strings.Join(pending, ", "))
+		}
+
+		t.Logf("waiting for control-plane deployments: %s", strings.Join(pending, ", "))
+		select {
+		case <-ctx.Done():
+			require.Failf(t, "the resource providers never became available", "context ended while waiting: %v", ctx.Err())
+		case <-time.After(databaseReadyPoll):
```
DariuszPorowski: nitpick(non-blocking): Poll interval named for the wrong wait

`requireResourceProvidersAvailable` budgets itself with `controlPlaneReadyTimeout` but polls on `databaseReadyPoll`. A shared `pollInterval` constant would read better.

test/functional-portable/database/noncloud/database_install_test.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+
+	missing := []string{}
+	for _, container := range controlPlaneComponents {
+		if !scanned[container] {
+			missing = append(missing, container)
+		}
+	}
+	require.Emptyf(t, missing, "could not read current logs for %s, so the permission scan proved nothing", strings.Join(missing, ", "))
```
DariuszPorowski, line 394: praise: The scan cannot fail open

Tolerating individual log-read failures is right, and requiring that every expected provider's current log was actually read is the part that usually gets left out. A log scan that silently reads nothing is the normal way this kind of test rots into a no-op, and this one is built so it can't.

test/functional-portable/database/noncloud/database_install_test.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// requireDatabaseReady waits for the PostgreSQL StatefulSet to report a ready replica, giving up
+// immediately (rather than after the full timeout) if its container cannot pull its image. A bad
+// image reference is terminal, so there is nothing to wait for. This is the regression guard for
+// the ImagePullBackOff bug.
+//
+// The poll runs on the test goroutine rather than inside require.Eventually: Eventually evaluates
+// its condition in a worker goroutine, where a failed assertion only stops that goroutine and the
+// test would still burn the whole timeout and then report a misleading "never became ready".
```
DariuszPorowski, line 276: praise: Correct reasoning, written down where the next reader needs it

Avoiding `require.Eventually` for exactly this reason is right, and it saves the next person an investigation. Same for the package doc's admission that `rad install` already catches both original bugs. The install path does use Helm's `StatusWatcherStrategy` (`pkg/cli/helm/helmclient.go:102`, `225`), so that's accurate.

DariuszPorowski (commented):
Reviewed the leg end to end. The plumbing follows the existing `TARGET_CLUSTER_INSTALL_ARGS` pattern exactly, every Kubernetes identifier the tests depend on matches the chart (StatefulSet name and `app.kubernetes.io/name` label, the `part-of` pod labels, and container names on `ucp` / `applications-rp` / `dynamic-rp`), and the guard against running on an apiserver-backed install cannot fail open. `go vet` on the new package passes and the leg is green in 7m33s, in the same band as the cheapest existing legs.

Worth stating plainly: because `rad install` waits on the release with Helm's `StatusWatcherStrategy` and a 5m timeout, both original bugs fail the install step before any test runs. Most of the value here comes from the leg existing at all, plus the deploy round-trip in `Test_DatabaseEnabled_MinimalDeploy`. The package doc says as much, which I appreciated.

Everything below is non-blocking. Two operational follow-ups worth tracking: the new `Functional Tests - database-noncloud` check needs adding to branch protection, and the `statestore-noncloud` failure on this commit looks unrelated (that leg is untouched, skips the shared install, and sets its own flag in test code), though the run was still in progress so I could not read its log.

test/functional-portable/database/noncloud/database_install_test.go
nellshamrell: Good catch — this was the real gap. Added `requireAPIServerStoreUnused`, which lists the `resources.ucp.dev` CRD in `radius-system` and requires it to be empty. That's the negative form you suggested, and it's the assertion that actually distinguishes a real `database.enabled=true` install from a hybrid one, since ucp, applications-rp and dynamic-rp each pick their database provider independently.

Two things I wanted to get right:

It fails closed. Any list error, including a missing CRD or unresolvable kind, fails the test rather than passing. The CRD ships unconditionally in `deploy/Chart/crds/ucpd/`, so its absence means a broken install or a mistyped GVR — treating that as success would have turned this into an assertion about nothing, which is the same trap you flagged approvingly on the log scan.

It's also not vacuous. My first instinct was to order it after the Deployments report Available, but that reasoning was wrong: the UCP Deployment has no readiness probe and the API and manifest initializer start concurrently, so Available says nothing about whether the bootstrap writes landed. The real guarantee comes from the workflow, which waits for "Successfully registered manifests" and then runs `rad group create` and `rad env create` before any Go test — real writes through UCP and applications-rp that must land in whichever store is configured. That's written down in the helper's comment.

It's called in two places: at the end of the health test, and from the deploy test's `PostStepVerify`, which fires after validation and before deletion. The second is the decisive one, since it observes resources the test just created itself.

One caveat documented in the failure message: switching providers doesn't migrate pre-existing apiserver rows, so this expects a cluster that wasn't previously installed without the flag. CI creates a fresh KinD cluster, so it holds there. Verified green in CI.

test/functional-portable/database/noncloud/database_install_test.go
nellshamrell: Agreed — resting the guarantee on file names and on `RunSerial` staying false was too fragile. Extracted the check into `requireDatabaseInstalled`, backed by a package-level `sync.Once`, and both tests now call it.

One subtlety worth noting: it caches the *error*, not the assertion. Asserting inside `Once.Do` would fail only whichever test won the race and let the other pass silently — which is the same class of ordering bug this was meant to remove. Instead the lookup result is cached and each caller re-asserts on its own `t`.

In the deploy test it's called after `rp.NewRPTest` but before `test.Test(t)`, since `t.Parallel()` is only reached inside `Test` — so it still runs sequentially and fails fast. Also dropped the now-stale ordering explanation from the package doc rather than leaving a comment that implies correctness still depends on it.

Accepted trade-off: a transient API error on the first caller is cached and fails both tests. That's no worse than the single un-retried `Get` this replaces, and this precondition should fail fast rather than paper over a cluster it can't read.

test/functional-portable/database/noncloud/database_install_test.go
nellshamrell: Fixed. Renamed to `pollInterval` and documented it as the shared cadence for both waits, which poll at the same rate and differ only in what they're waiting for. No behavior change — same 5s value in both loops.

nellshamrell:
@DariuszPorowski thanks — all three addressed in a2edeadf1, and the full non-cloud functional run is green on that commit, including the new `database-noncloud` leg.

Two follow-ups from your summary review:

**Branch protection — correcting what I wrote here earlier.** I originally said the new `Functional Tests - database-noncloud` check still needed adding to the required set. That was wrong on both counts, so to save the next person the detour:

- The `main` ruleset requires only `Functional Test Run`, `Lint`, `Build Check` and `DCO`. `Functional Test Run` is published solely by `functional-test-cloud.yaml`, the only workflow using `LouisBrunner/checks-action`. **No non-cloud leg gates merges today**, so `database-noncloud` is in exactly the same position as `corerp-noncloud`, `statestore-noncloud` and the rest — not a gap introduced by this PR.
- The check name I cited is also the wrong one to require: `Functional Tests - database-noncloud` comes from the test-reporter step, which is gated on `github.repository == vars.RADIUS_REPOSITORY`, so it never appears on fork PRs.

Requiring the underlying job check `Run database-noncloud functional tests` instead would block docs-only PRs, because `tests` is guarded by `if: needs.changes.outputs.only_changed != 'true'` and the entire matrix is skipped in that case. The cloud workflow already solved this by publishing a success `Functional Test Run` when tests are skipped (`functional-test-cloud.yaml:608-620`). Making the non-cloud legs gating properly would need that same rollup-check treatment plus a decision about their known flakiness — a repo-wide change that belongs in its own issue, not here.

**The `statestore-noncloud` failure.** You were right that it was unrelated. It has now passed on this branch, and it was green on 11 of the last 12 runs on `main`. The leg is also structurally unreachable from this change: it skips the shared "Install Radius" step and sets its own flag in test code, and `DATABASE_INSTALL_ARGS` is written to `GITHUB_ENV`, which is job-scoped. Treating it as a flake.

DariuszPorowski (dismissed):
LGTM

docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
```diff
@@ -46,21 +46,30 @@ make test-functional-daprrp
...
+| `make test-functional-multicluster-noncloud` | Requires a second Kubernetes cluster, a target-cluster Secret mounted into Radius, and `RADIUS_TEST_EXTERNAL_KUBECONFIG` for the test process.     |
+| `make test-functional-database-noncloud`     | Requires Radius installed with `--set database.enabled=true` (PostgreSQL-backed control plane) instead of the default Kubernetes API server store. |
+| `make test-functional-statestore-noncloud`   | Destructive lifecycle test that installs, purges, and reinstalls Radius. Run it only on a dedicated cluster.                                       |
+| `make test-functional-upgrade-noncloud`      | Exercises the Radius upgrade path and performs its own install/upgrade lifecycle.                                                                  |
 
-The multicluster and statestore groups run as isolated CI legs in `functional-test-noncloud.yaml`; do not run them against a shared development cluster.
+The multicluster, database, and statestore groups run as isolated CI legs in `functional-test-noncloud.yaml`; do not run them against a shared development cluster.
```
kachawla, line 64: The sibling `docs/contributing/contributing-code/contributing-code-tests/README.md` still says only the multicluster, state-store lifecycle, and upgrade groups have additional setup or isolation requirements. Since `database` now has the same kind of hard prerequisite and is excluded from the aggregate target, could we add it to that summary too? Otherwise contributors entering through the tests README may miss the special install requirement.

kachawla (dismissed):

docs/contributing/contributing-code/contributing-code-tests/running-functional-tests.md
nellshamrell: Good catch — I updated `running-functional-tests.md` but missed its sibling. Fixed in 0a5da8666: the README summary now reads "The multicluster, database, state-store lifecycle, and upgrade groups have additional setup or isolation requirements."

I also grepped the rest of `docs/` for the same group list to make sure this was the only place that had drifted — it was.

kachawla (dismissed):

DariuszPorowski (dismissed):

nellshamrell:
@copilot+gpt-5.6-sol please review

nellshamrell:
## Approver review

**Head `1f38135f`** · merge base `96398ff8` · +604/−8 across 7 files · 69 checks green, 0 failures

I wrote this PR, so treat this as a structured self-review rather than independent sign-off — it can surface defects but it doesn't replace the two approvals already on here. Posting it because the central assertion in this PR is non-obvious and I'd rather the reasoning (and its preconditions) be written down where the next reader finds it.

### Orientation

Radius stores control-plane state in one of two backends. By default it uses the Kubernetes API server as a database — rows are custom resources of kind `Resource` in the `resources.ucp.dev` CRD. Setting `database.enabled=true` swaps that for a PostgreSQL StatefulSet, with each resource provider connecting as its own database user. Every functional leg ran the default, so the PostgreSQL path had no end-to-end coverage, and two blocking bugs reached review of #12214 as a result: a database image tag that didn't exist, and per-RP users never granted privileges on the shared `resources` table (#12227).

This is test and CI infrastructure only. No production code, no dependencies, no API surface. A defect here means a false-green or a flaky leg, never a production regression.

**The one decision that carries the PR:** how does the suite prove the control plane is genuinely PostgreSQL-backed, rather than merely running PostgreSQL next to an API-server-backed control plane? Everything else is scaffolding around that.

### Data and control flow

```mermaid
sequenceDiagram
    participant W as CI leg
    participant H as Helm/rad install
    participant CP as ucp / applications-rp / dynamic-rp
    participant PG as PostgreSQL
    participant AS as resources.ucp.dev (apiserver store)
    participant T as Go tests

    W->>H: rad install --set database.enabled=true
    H->>CP: configmaps select provider=postgresql
    W->>CP: wait for "Successfully registered manifests"
    W->>CP: rad group create / rad env create
    CP->>PG: bootstrap writes land here
    Note over AS: must stay empty
    W->>T: make test-functional-database-noncloud
    T->>CP: StatefulSet present? Deployments Available? logs clean?
    T->>CP: deploy app + container, validate, read back
    T->>AS: List → require empty
```

The workflow runs `rad group create` and `rad env create` (`functional-test-noncloud.yaml:562-569`) **before any Go test**. Those are real writes through UCP and applications-rp, which is what makes the emptiness assertion meaningful rather than vacuous.

### Key decisions

**Proving the store is actually PostgreSQL.** `requireAPIServerStoreUnused` (`database_install_test.go:181-197`) lists `resources.ucp.dev` in `radius-system` and requires it empty. It runs last in the health test and, decisively, from the deploy test's `PostStepVerify` hook (`database_deploy_test.go:69-71`) — where it observes a write the test itself just made, which is causal evidence rather than circumstantial.

A ready StatefulSet only proves Postgres is running; `ucp`, `applications-rp` and `dynamic-rp` each select their provider independently, so any one could still be writing to the API server. Three preconditions make emptiness a valid discriminator, and each is a potential silent fail-open, so I verified them rather than assuming:

- **Namespace matches.** All three configmaps hardcode `namespace: "radius-system"` in their apiserver branch — the same namespace the test lists. A mismatch would return empty and pass vacuously.
- **The CRD ships unconditionally.** `deploy/Chart/crds/ucpd/ucp.dev_resources.yaml` has no conditional guard, so absence means a broken install, not an unused store. That's why the list must succeed rather than tolerating `NotFound`.
- **Nothing else writes there.** `controller` keeps `provider: "apiserver"` unconditionally, which would break this if it wrote rows — but it never constructs a database client (no `databaseprovider`/`DatabaseClient` reference anywhere under `cmd/controller` or `pkg/controller`). The apiserver *queue* provider stays enabled in both modes but writes the separate `queuemessages.ucp.dev` CRD.

**Removing the ordering dependency.** `requireDatabaseInstalled` (`:146-156`) caches the lookup in a `sync.Once` but stores *the error, not the assertion*. Asserting inside `Once.Do` would fail only whichever test won the race and let the other pass silently. `t.Parallel()` is only reached at `test/rp/rptest.go:446`, inside `Test`, so the deploy test's call at `database_deploy_test.go:78` genuinely runs sequentially and fails fast.

**Manual polling instead of `require.Eventually`.** `Eventually` evaluates its condition in a worker goroutine, where `FailNow` stops only that goroutine — a terminal ImagePullBackOff would burn the full timeout and then report a misleading "never became ready".

**Reusing the shared install step.** `DATABASE_INSTALL_ARGS` is exported by a gated step and appended in the shared install, mirroring the existing `TARGET_CLUSTER_INSTALL_ARGS` pattern. `GITHUB_ENV` is job-scoped and each matrix leg is its own job, so there's no cross-leg bleed. `test-functional-all-noncloud` is an explicit prerequisite list, not a wildcard, so the new package is genuinely excluded rather than accidentally picked up.

### What the tests actually prove

| Behavior | Evidence | Strength |
|---|---|---|
| Database image is pullable | terminal fail-fast on ImagePullBackOff | Strong |
| Per-RP GRANTs exist | Deployments reach `Available` + log scan | Strong (`Available` is the detector, the scan adds diagnosis) |
| Control plane uses PostgreSQL | apiserver store empty, asserted after a test-made write | Strong |
| Write/read round-trip through PostgreSQL | deploy + resource/K8s validation | Strong — the durable value |
| Broader deployment behavior | not covered here | Intentional; other legs own it |

Worth restating the honest framing from the package doc: `rad install` waits on the Helm release, so both original bugs would already fail the install before reaching these tests. The lasting value is the deploy round-trip, not the two regression guards.

### Non-blocking findings

**N1 — `database_install_test.go:194-196`.** The assertion is *global* emptiness, so it couples this test to any component that could write a `Resource` to `radius-system`. `controller` is inert today, but its config still says `apiserver` unconditionally; if that ever becomes live, this leg goes red while PostgreSQL works fine. The failure message names the configmaps, so the cost is confusion rather than a wrong conclusion. The narrower fix, if it ever bites: assert that no `ResourceEntry.ID` matches the deployed application and container IDs instead of requiring the store globally empty.

**N2 — `running-functional-tests.md:64-72`.** The docs tell contributors to install with `--set database.enabled=true` but don't warn that a cluster previously installed *without* it retains stale apiserver rows, which fails the store assertion. The in-code failure message covers this, so it's discoverable at the point of failure; a sentence in the doc would save the detour.

**N3 — `database_install_test.go:100`.** `testTimeout` (7m) plus a slow deploy can approach the leg's 15m `FUNCTIONALTEST_TIMEOUT`, in which case Go panics with a goroutine dump instead of reporting clean failures. Only degrades diagnostics in an already-failing run.

None of these affect whether the change is correct today. Happy to fold N2 in here if a reviewer wants it before merge; N1 and N3 are better as follow-ups.

### Verification

- Local: `gofmt -l`, `go vet`, `go test -c`, and `go test -race -c` all clean on `1f38135f`.
- CI: 69 SUCCESS / 8 SKIPPED / 2 NEUTRAL / 0 FAILURE. `database-noncloud` has now passed twice, including against a freshly merged `main`.
- The two NEUTRAL results are GitHub's default-setup CodeQL and gosec, which fail repo-wide on `main` too (`go.mod` requires Go 1.27, the default-setup runner has 1.26.6). The repo's own `codeql.yml` passes.

nellshamrell:
@kachawla @DariuszPorowski I added some documentation and unfortunately, it dismissed your reviews. Do you mind re-approving when you are able?

DariuszPorowski (approved):

PR #12772, opened 2026-08-21
fix(test): recognize CloudControl InvalidRequestException for absent AWS::Logs::LogGroup

### LINKED ISSUE #12767: Flaky test: AWS CloudWatch LogGroup cleanup misclassifies missing resources
### Steps to reproduce

1. Run the scheduled cloud functional tests containing `Test_AWS_LogsLogGroup` or `Test_AWS_MultiIdentifier_Resource`.
2. Allow cleanup to delete an `AWS::Logs::LogGroup`.
3. Observe the subsequent CloudControl `GetResource` validation after the log group is absent.

Reproduced by scheduled runs [32436557418](https://github.com/radius-project/radius/actions/runs/32436557418), [32373614953](https://github.com/radius-project/radius/actions/runs/32373614953), and [32257359533](https://github.com/radius-project/radius/actions/runs/32257359533).

### Observed behavior

AWS CloudControl returns `InvalidRequestException` with a message such as `The specified log group does not exist` or `Log group cannot be found` when an `AWS::Logs::LogGroup` is already absent. `validation.IsAWSResourceNotFound` delegates to `awsclient.IsAWSResourceNotFoundError`, which only recognizes the typed CloudControl `ResourceNotFoundException`. The cleanup loop therefore receives `notFound == false` with an error, stops retrying, and fails the test with `AWS resource ... was present, should be not found` even though the log group is gone.

### Desired behavior

AWS cleanup validation should recognize CloudControl's missing-resource response for `AWS::Logs::LogGroup` as not found and complete successfully. The implementation should distinguish this known absence response from other `InvalidRequestException` errors rather than treating every invalid request as deletion success.

### Workaround

Rerun the scheduled functional test. This does not prevent recurrence.

### rad Version

N/A; this affects repository functional-test cleanup logic.

### Operating system

GitHub-hosted scheduled functional-test runner.

### Additional context

Affected generated failure issues:

- #12762
- #12740
- #12715

Relevant code:

- `test/validation/aws.go`: `IsAWSResourceNotFound`
- `pkg/ucp/aws/errorhandler.go`: `IsAWSResourceNotFoundError`
- `test/rp/rptest.go`: AWS deletion validation and assertion

pujitha24:
## Summary

Fixes flaky AWS functional-test cleanup for `AWS::Logs::LogGroup` resources: CloudControl's `GetResource` returns an untyped `InvalidRequestException` (not the typed `ResourceNotFoundException`) when the underlying log group is already absent, so the cleanup/deletion-validation helpers in `test/validation/aws.go` treated the resource as still present and failed with `AWS resource ... was present, should be not found`.

## Reason for change

`test/validation.IsAWSResourceNotFound` and `DeleteAWSResource` both delegated solely to `pkg/ucp/aws.IsAWSResourceNotFoundError`, which only recognizes the typed CloudControl `ResourceNotFoundException`. The `AWS::Logs::LogGroup` resource handler is an exception: on a missing log group it returns `InvalidRequestException` with a message such as "The specified log group does not exist" or "Log group cannot be found". The retry loop in `test/rp/rptest.go` then received `notFound == false` with a non-nil error, stopped retrying, and the test failed even though the log group was actually gone (see the linked scheduled-run failures in the issue).

Fixes #

## How to test

A resource-type-scoped helper `isAWSResourceNotFoundError` was added in `test/validation/aws.go`: it still delegates to the shared `awsclient.IsAWSResourceNotFoundError` for the typed exception (all resource types), and additionally recognizes the two known `InvalidRequestException` messages **only** when `resourceType == "AWS::Logs::LogGroup"`. This keeps the change scoped to the test-cleanup code path — the shared, production-facing `pkg/ucp/aws.IsAWSResourceNotFoundError` (used by 8 AWS proxy controller call sites for every resource type) is untouched, so there is no risk of misclassifying an unrelated resource's real `InvalidRequestException` as "not found" in production request handling.

Validation performed:
- `go build ./test/validation/... ./pkg/ucp/aws/...` — passes.
- `go test ./test/validation/... ./pkg/ucp/aws/... ./pkg/ucp/frontend/controller/awsproxy/...` — all pass.
- `go vet ./test/validation/...` and `gofmt -l` on changed files — clean.
- `golangci-lint run ./test/validation/...` — `0 issues`.
- Added `test/validation/aws_test.go` with a table-driven `Test_IsAWSResourceNotFoundError` covering: the typed exception for any resource type, both known log-group `InvalidRequestException` messages, an unrelated `InvalidRequestException` message under the log-group type (must stay `false`), the same matching message text under a *different* resource type (must stay `false`, proving the scoping gate works), a generic error, and a nil error. Verified with `git stash` that the log-group cases fail against the pre-fix code path and pass after the change.
- Not run: the live functional test `Test_AWS_LogsLogGroup` against a real AWS account (no AWS credentials/cluster available in this environment). The unit test above reproduces the exact defect (message-based absence not recognized) and proves the fix; the functional test would exercise the same code path end-to-end but requires live AWS infrastructure.
- Base branch (`main`) CI was green at the time of this change (`gh run list --branch main` showed the latest Unit Tests / CodeQL runs as `success`).

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `test/validation/aws.go` | Added `isAWSResourceNotFoundError`, gated on `resourceType == "AWS::Logs::LogGroup"`, to also recognize CloudControl's `InvalidRequestException` "log group does not exist" / "cannot be found" messages as not-found; used it from `DeleteAWSResource` and `IsAWSResourceNotFound` in place of the direct `awsclient.IsAWSResourceNotFoundError` call. |
| `test/validation/aws_test.go` | New table-driven unit test for `isAWSResourceNotFoundError` covering the typed exception, both log-group messages, the resource-type scoping gate, and negative cases. |

Fixes #12767

### DIFF
--- a/test/validation/aws.go
+++ b/test/validation/aws.go
@@ -18,12 +18,15 @@ package validation
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/cloudcontrol"
+	"github.com/aws/aws-sdk-go-v2/service/cloudcontrol/types"
 	awsclient "github.com/radius-project/radius/pkg/ucp/aws"
 	"github.com/radius-project/radius/pkg/ucp/resources"
 	resources_aws "github.com/radius-project/radius/pkg/ucp/resources/aws"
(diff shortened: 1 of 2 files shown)


lakshmimsft:
Thank you for contributing. i think there was a similar recent fix  in PR #12765 which is why there is a conflict in files. I will close this PR at this time. Thank you again.

Issue #11841, opened 2026-05-08
Flaky test: Run upgrade-noncloud functional tests: An unknown error was returned while testing Radius API status 503


sk593:
### Area for Improvement

The upgrade noncloud functional tests are seeing intermittent failures during execution

### Observed behavior

```
    rptest.go:177: Loaded workspace: kind-kind (Kubernetes (context=kind-kind))
    rptest.go:180: 
        	Error Trace:	/home/runner/work/radius/radius/test/rp/rptest.go:180
        	            				/home/runner/work/radius/radius/test/functional-portable/upgrade/upgrade_test.go:233
        	            				/home/runner/work/radius/radius/test/functional-portable/upgrade/upgrade_test.go:234
        	            				/opt/hostedtoolcache/go/1.26.1/x64/src/runtime/asm_amd64.s:1771
        	Error:      	Received unexpected error:
        	            	An unknown error was returned while testing Radius API status:
        	            	Status Code: 503
        	            	Response Headers:
        	            	  Audit-Id: 73f9b0ea-defe-42e3-850a-8093e4711e80
        	            	  Cache-Control: no-cache, private
        	            	  Content-Length: 20
        	            	  Content-Type: text/plain; charset=utf-8
        	            	  Date: Fri, 08 May 2026 13:45:02 GMT
        	            	  X-Content-Type-Options: nosniff
        	            	  X-Kubernetes-Pf-Flowschema-Uid: 211e83f3-102c-4be3-b085-b3168d0ac12a
        	            	  X-Kubernetes-Pf-Prioritylevel-Uid: e153f0a4-3684-4848-b180-47599071d202
        	            	Response Body:
        	            	service unavailable
        	            	
        	Test:       	Test_PreflightContainer/Disabled
        	Messages:   	failed to create ApplicationsManagementClient
    upgrade_test.go:138: 
        	Error Trace:	/home/runner/work/radius/radius/test/functional-portable/upgrade/upgrade_test.go:225
        	            				/home/runner/work/radius/radius/test/functional-portable/upgrade/upgrade_test.go:138
        	Error:      	Condition never satisfied
        	Test:       	Test_PreflightContainer/Disabled
        	Messages:   	Control plane did not become available within timeout
```

### Desired behavior

We should investigate why the scheduled tests are failing for this job 

### rad Version

latest

willdavsmith:
A 503 error is usually a platform issue or transient issue, but in this case, it happens too frequently to be so. We should take a closer look.

[...]

PR #12785, opened 2026-08-21
fix(chart): wait for Radius aggregated API readiness

brooke-hamilton:
## Summary

Make successful Helm install and upgrade completion the authoritative Radius readiness contract.

- Add an HTTPS UCP pod probe so UCP does not join Service endpoints before it is listening.
- Add a mandatory `post-install,post-upgrade` hook that waits for `/apis/api.ucp.dev/v1alpha3` through kube-apiserver.
- Reuse the Radius `pre-upgrade` binary via a new `wait-for-control-plane` subcommand with least-privilege RBAC, bounded retries, and retained failure diagnostics.

## Reason for change

Helm does not check Kubernetes `APIService` readiness. It can return while UCP is serving but kube-apiserver still reports the recreated Radius APIService as unavailable:

```text
t=+6ms     podReady=1/1  apiserviceAvailable=False  api=503
t=+3182ms  podReady=1/1  apiserviceAvailable=False  api=503
t=+4540ms  podReady=1/1  apiserviceAvailable=True   api=SERVING
```

The pod probe and Helm hook cover different readiness boundaries. The pod probe prevents routing to a process that is not listening; the hook verifies the full client path through kube-apiserver, the aggregation controller, APIService, Service, TLS, and UCP. A pod probe cannot wait on APIService availability because the APIService itself requires a Ready endpoint, which would create a circular dependency.

The hook image follows the current build through this precedence: explicit readiness override, explicit non-default pre-upgrade image, then a `pre-upgrade` image derived from UCP's registry and tag. Template-local defaults preserve direct Helm `--reuse-values` upgrades from older charts.

Related to #11841.

## How to test

Deterministic checks:

```text
go test ./cmd/pre-upgrade/... ./pkg/cli/cmd/install/kubernetes/... -count=1
go vet ./cmd/pre-upgrade/... ./pkg/cli/cmd/install/kubernetes/...
go build ./cmd/pre-upgrade/...
make test-helm
```

Live kind validation with images built from this branch confirmed:

- Fresh install, reinstall, and upgrade all returned with the aggregated endpoint immediately serving.
- Deleting the APIService caused two retryable 404s; the hook succeeded after restoration.
- A forced timeout reported the last error and retained the failed Job.
- Main's existing upgrade test passed 3/3 unmodified with this hook.
- Removing only this hook reproduced the original one-503/idle-timeout failure exactly.

All current CI checks pass, including cloud install legs and `upgrade-noncloud`.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `cmd/pre-upgrade/cmd/root.go` | Preserve preflight behavior while adding subcommand dispatch. |
| `cmd/pre-upgrade/cmd/wait_for_control_plane.go` | Poll the aggregated Radius API with bounded retries and error classification. |
| `cmd/pre-upgrade/cmd/wait_for_control_plane_test.go` | Cover retry, timeout, authorization, request-path, and command behavior. |
| `deploy/Chart/templates/control-plane-readiness/*` | Add the mandatory Helm hook Job, ServiceAccount, and least-privilege RBAC. |
| `deploy/Chart/templates/ucp/deployment.yaml` | Add the direct HTTPS UCP readiness probe. |
| `deploy/Chart/values.yaml` | Configure hook image and timing with backward-compatible defaults. |
| `deploy/Chart/tests/*readiness_test.yaml` | Verify UCP and Helm readiness rendering, RBAC, image precedence, and old-value compatibility. |
| `deploy/Chart/templates/networkpolicies.yaml` | Correct probe/network-policy guidance. |
| `pkg/cli/cmd/install/kubernetes/kubernetes.go` | Document that successful Helm completion guarantees aggregated API readiness. |

PR #12790 is stacked on this PR and contains the upgrade-test cleanup/optimization. After this PR merges, retarget #12790 to `main`.

### DIFF
--- a/deploy/Chart/templates/control-plane-readiness/job.yaml
+++ b/deploy/Chart/templates/control-plane-readiness/job.yaml
@@ -1,0 +1,48 @@
+{{- $appversion := include "radius.versiontag" . }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: radius-api-readiness
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    control-plane: radius-api-readiness
+    app.kubernetes.io/name: radius-api-readiness
+    app.kubernetes.io/part-of: radius
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-output-log-policy": hook-failed
+spec:
+  backoffLimit: 0
+  activeDeadlineSeconds: {{ add .Values.controlPlaneReadiness.timeoutSeconds 30 }}
+  ttlSecondsAfterFinished: {{ .Values.controlPlaneReadiness.ttlSecondsAfterFinished }}
+  template:
+    metadata:
+      labels:
+        control-plane: radius-api-readiness
+        app.kubernetes.io/name: radius-api-readiness
+        app.kubernetes.io/part-of: radius
+    spec:
+      restartPolicy: Never
+      serviceAccountName: radius-api-readiness
+      {{- if .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .Values.global.imagePullSecrets | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: radius-api-readiness
+          image: "{{ include "radius.image" (dict "image" .Values.preupgrade.image "tag" (.Values.preupgrade.tag | default .Values.global.imageTag | default $appversion) "global" .Values.global) }}"
+          imagePullPolicy: Always
+          args:
+            - wait-for-control-plane
+            - "--timeout={{ .Values.controlPlaneReadiness.timeoutSeconds }}s"
+            - "--poll-interval={{ .Values.controlPlaneReadiness.pollIntervalSeconds }}s"
+            - "--request-timeout={{ .Values.controlPlaneReadiness.requestTimeoutSeconds }}s"
+          terminationMessagePolicy: FallbackToLogsOnError
+          resources:
+            {{- toYaml .Values.preupgrade.resources | nindent 12 }}
+          securityContext:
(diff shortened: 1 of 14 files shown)


DariuszPorowski (dismissed):

deploy/Chart/templates/control-plane-readiness/job.yaml
```diff
@@ -0,0 +1,69 @@
...
+{{- $ttlSecondsAfterFinished := get $readiness "ttlSecondsAfterFinished" | default 300 }}
+{{- $readinessImage := get $readiness "image" | default "" }}
+{{- $readinessTag := get $readiness "tag" | default "" }}
+{{- if not $readinessImage }}
+  {{- if ne .Values.preupgrade.image "pre-upgrade" }}
+    {{- $readinessImage = .Values.preupgrade.image }}
+    {{- $readinessTag = $readinessTag | default .Values.preupgrade.tag }}
+  {{- else }}
+    {{- if contains "/" .Values.ucp.image }}
+      {{- $readinessImage = regexReplaceAll "/[^/]+$" .Values.ucp.image "/pre-upgrade" }}
```
kachawla, line 20: This drops an embedded tag from a supported full UCP image path. For example, `--set-string ucp.image=custom-registry.io/ucpd:custom-tag` renders UCP as `custom-registry.io/ucpd:custom-tag`, but this readiness hook becomes `custom-registry.io/pre-upgrade:latest`. The install can then pull the wrong image or fail because that tag does not exist. Please preserve the embedded UCP tag when deriving the readiness image and add a Helm unit test for this case.

brooke-hamilton: Fixed in `8f9a7781e`. The readiness image derivation now extracts a tag only from the final UCP path segment (so registry ports are not mistaken for tags), carries that tag to the sibling `pre-upgrade` image, and still lets an explicit readiness tag override it. Added the requested Helm regression test for `custom-registry.io/ucpd:custom-tag`; the full chart suite passes (131 tests).

willdavsmith (approved):

PR #9958, opened 2025-07-10
Adds support for mounting terraform binaries from a container.

sylvainsf:
# Description

This PR adds support for pre-mounting Terraform binaries from a container image to improve performance and reduce internet dependencies when executing Terraform recipes. Previously, Radius downloaded Terraform binaries at runtime for each recipe execution. This feature allows users to specify a container image containing Terraform binaries that will be copied to the Radius pods during initialization, eliminating the need for runtime downloads.

## Key Changes

- Added `--terraform-container` flag to `rad install kubernetes` command
- Extended Helm chart with `global.terraform` configuration section
- Modified applications-rp and dynamic-rp deployments to support init containers for Terraform binary mounting
- Updated Terraform install logic to check for pre-mounted binaries before downloading
- Added comprehensive documentation and unit tests

## Type of change

- This pull request adds or changes features of Radius and has an approved issue (issue link required).

Fixes: #9957

## Contributor checklist
Please verify that the PR meets the following requirements, where applicable:

- An overview of proposed schema changes is included in a linked GitHub issue.
    - [X] Yes 
- A design document PR is created in the [design-notes repository](https://github.com/radius-project/design-notes/), if new APIs are being introduced.
    - [X] Not applicable 
- The design document has been reviewed and approved by Radius maintainers/approvers.
    - [X] Yes 
- A PR for the [samples repository](https://github.com/radius-project/samples) is created, if existing samples are affected by the changes in this PR.
    - [X] Not applicable 
- A PR for the [documentation repository](https://github.com/radius-project/docs) is created, if the changes in this PR affect the documentation or any user facing updates are made.
    - [X] Yes 
- A PR for the [recipes repository](https://github.com/radius-project/recipes) is created, if existing recipes are affected by the changes in this PR.
    - [X] Not applicable

PR #12800, opened 2026-08-24
Test producer-managed secret connections end to end

willdavsmith:
## Summary

- adds focused functional coverage for a Recipe-backed Redis producer that returns ordinary outputs and `outputs.secrets`
- verifies Radius materializes the deterministic managed `Radius.Security/secrets` resource and backing Kubernetes Secret
- verifies one Container connection emits ordinary `CONNECTION_REDIS_HOST`/`PORT` values and secret-backed `CONNECTION_REDIS_URL` through `secretKeyRef` for regular and init containers
- proves the variables are usable at runtime without printing secret values and asserts the Pod spec contains no plaintext secret
- leaves the existing Container Secret fixture unchanged

## Compatibility

The required control-plane support from #12709 and Kubernetes Container Recipe support from radius-project/resource-types-contrib#300 are both merged. The dedicated Recipe Pack selects the published `ghcr.io/radius-project/kube-recipes/{containers,rediscaches,secrets}:edge` artifacts used by Radius edge/PR functional coverage, so the test exercises the coordinated Radius/contrib contract directly.

The backing Kubernetes Secret is validated by its deterministic resource name because managed Secrets do not carry selector labels matching the managed Radius Secret resource name. Post-step verification also reads the Secret directly and validates its `url` data key.

## Validation

- `gofmt -w test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go`
- `go test ./test/functional-portable/corerp/noncloud/resources -run '^$'`
- `go test ./test/validation`
- `~/.rad/bin/bicep format test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep`
- `~/.rad/bin/bicep build test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep --stdout`
- `git diff --check`
- fresh non-cloud functional workflow, including `corerp-noncloud`


test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
```diff
@@ -0,0 +1,116 @@
+extension radius
+
+@description('Specifies the image used by the connection-checking containers.')
+param magpieimage string
+
+@description('Specifies the published resource-types-contrib Recipe tag. Radius PR functional tests use edge, which must include resource-types-contrib#300.')
```
willdavsmith:
```suggestion
@description('Specifies the published resource-types-contrib Recipe tag.')
```

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
```diff
@@ -0,0 +1,116 @@
...
+param recipeTag string = 'edge'
+
+var applicationName = 'corerp-container-managed-secret'
+var environmentName = '${applicationName}-env'
+var namespace = applicationName
+
+// This dedicated pack makes the cross-repository dependency explicit. The test
```
willdavsmith:
```suggestion
```

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
```diff
@@ -0,0 +1,116 @@
...
+// can run after resource-types-contrib#300 publishes the Container Recipe to edge.
```
willdavsmith:
```suggestion
```

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
willdavsmith: Updated the parameter description to remove the stale #300 prerequisite.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
willdavsmith: Removed the stale cross-repository dependency comment now that both prerequisite PRs are merged.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
willdavsmith: Removed the stale publication prerequisite while retaining the explicit published `:edge` Recipe sources.

nithyatsu (commented):
Nice, focused test — the plaintext-leak assertions (producer `secrets` map has no `url`, and the pod spec doesn't contain the decoded value) are exactly the right checks, and the namespace/workspace wiring all holds up.

One substantive gap plus a few nits inline.

### Missing the delete half of "end to end"

`Materializer.Delete` is part of the managed-secret contract, but nothing here asserts that the managed `Radius.Security/secrets` resource and its backing Kubernetes Secret are actually removed when the producer/application is deleted.

Note that listing the managed secret in `RPResources` does **not** cover this: `validation.DeleteRPResource` no-ops for `radius.security/secrets` (it falls through to the "cleaned up via cascade delete" branch), so the delete path is never directly exercised or verified.

A `PostDeleteVerify` that asserts both are gone would close the highest-value remaining gap:

```go
PostDeleteVerify: func(ctx context.Context, t *testing.T, test rp.RPTest) {
    _, err := test.Options.ManagementClient.GetResource(ctx, "Radius.Security/secrets", managedSecretName)
    require.Error(t, err, "managed Radius.Security/secrets resource should be deleted with its owner")

    _, err = test.Options.K8sClient.CoreV1().Secrets(name).Get(ctx, managedSecretName, metav1.GetOptions{})
    require.True(t, apierrors.IsNotFound(err), "backing Kubernetes Secret should be deleted with its owner")
},
```

(`PostDeleteVerify` is on `RPTest`, not `TestStep`, and it's skipped under `RADIUS_TEST_FAST_CLEANUP=true` — worth confirming the non-cloud workflow doesn't set that, otherwise the assertion silently never runs in CI.)

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,196 @@
...
+			),
+			RPResources: &validation.RPResourceSet{
+				Resources: []validation.RPResource{
+					{
+						Name: name,
+						Type: validation.CoreApplicationsResource,
+					},
+					{
+						Name: producerName,
+						Type: "radius.data/redisCaches",
```
nithyatsu: Nit: this is the only hardcoded resource-type string in the file — every other entry uses a `validation.*` constant. Worth adding `DataRedisCachesResource = "radius.data/redisCaches"` to `test/validation/shared.go` alongside the existing `DataMySQLDatabasesResource` / `DataPostgreSQLDatabasesResource`, and using it here.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,196 @@
...
+						validation.NewK8sSecretForResourceWithResourceName(managedSecretName),
+						validation.NewK8sPodForResource(name, containerName),
+					},
+				},
+			},
+			PostStepVerify: func(ctx context.Context, t *testing.T, test rp.RPTest) {
+				producer, err := test.Options.ManagementClient.GetResource(ctx, "Radius.Data/redisCaches", producerName)
+				require.NoError(t, err)
+				require.Equal(t, producerName+"."+name+".svc.cluster.local", producer.Properties["host"])
+				require.Equal(t, "6379", fmt.Sprint(producer.Properties["port"]))
```
nithyatsu: Nit: this relies on the untyped JSON value round-tripping as `float64` and `fmt.Sprint` rendering it as `6379` rather than `6379.0` / `6.379e+03`. That holds today, but it's brittle for no benefit.

```suggestion
				require.EqualValues(t, 6379, producer.Properties["port"])
```

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,196 @@
...
+package resource_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	backendsecret "github.com/radius-project/radius/pkg/dynamicrp/backend/secret"
```
nithyatsu, line 26: Nit: importing `pkg/dynamicrp/backend/secret` couples this functional test to an internal naming detail. It's justified here — you need the name up front to populate `RPResources` — and you do cross-check it against `producer.Properties["secrets"]["name"]` below, which is the real contract. Worth a short comment at the `managedSecretName` assignment saying so, otherwise a future reader may reasonably think the test is just re-deriving the implementation and asserting nothing.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
```diff
@@ -0,0 +1,114 @@
...
+        command: [
+          '/bin/sh'
+          '-c'
+        ]
+        args: [
+          '''
+            test -n "$CONNECTION_REDIS_HOST"
+            test "$CONNECTION_REDIS_PORT" = "6379"
+            test -n "$CONNECTION_REDIS_URL"
+            case "$CONNECTION_REDIS_URL" in redis://*) ;; *) exit 1 ;; esac
```
nithyatsu: Nit: a bare `exit 1` gives no diagnostic. If the contrib recipe's `url` output ever stops matching `redis://*`, this surfaces as an opaque init-container CrashLoop / pod-startup timeout with nothing in the logs, which is painful to debug from a CI run.

Echo a marker to stderr before failing — cheap, and it can't leak the value:

```sh
case "$CONNECTION_REDIS_URL" in redis://*) ;; *) echo "CONNECTION_REDIS_URL has unexpected scheme" >&2; exit 1 ;; esac
```

Same applies to the `test -n` / `test =` lines above, though the URL shape check is the one most likely to break on an upstream recipe change.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
```diff
@@ -0,0 +1,114 @@
...
```
nithyatsu: Nit: same as the init container — `exit 1` with no diagnostic. Add an `echo ... >&2` before it so a recipe-contract change is debuggable from the pod logs rather than only from a startup timeout.

willdavsmith: Added the same non-secret failure diagnostics to the regular container checks.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Added a comment explaining that the deterministic name is required before deployment for the validation set, while the later producer metadata assertion independently verifies the public contract.

test/functional-portable/corerp/noncloud/resources/testdata/corerp-resources-container-managed-secret.bicep
willdavsmith: Added non-secret stderr diagnostics for missing host, unexpected port, missing URL, and unexpected URL scheme in the init container.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Replaced the string conversion and comparison with `require.EqualValues(t, 6379, producer.Properties["port"])`.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Rebased onto current `main`, which already defines `validation.DataRedisCachesResource`, and updated both the resource validation entry and management client lookup to use it.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,198 @@
...
+				require.EqualValues(t, 6379, producer.Properties["port"])
+
+				producerSecrets, ok := producer.Properties["secrets"].(map[string]any)
+				require.True(t, ok, "producer should expose managed secret metadata")
+				require.Equal(t, managedSecretName, producerSecrets["name"])
+				if _, leaked := producerSecrets["url"]; leaked {
+					require.FailNow(t, "producer state contains the secret output")
+				}
+
+				_, err = test.Options.ManagementClient.GetResource(ctx, "Radius.Security/secrets", managedSecretName)
```
nithyatsu: The `Radius.Security/secrets` response is discarded via `_, err =`, so this only asserts the resource exists. Asserting the returned properties don't carry the plaintext `url` would cover the API read path the same way line 110 covers the producer — it's the one remaining place a caller could read the value back.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,198 @@
...
+				require.Equal(t, "connectioncheck", initContainer.Name)
+				require.Equal(t, producerName+"."+name+".svc.cluster.local", requireEnvValue(t, initContainer, "CONNECTION_REDIS_HOST"))
+				require.Equal(t, "6379", requireEnvValue(t, initContainer, "CONNECTION_REDIS_PORT"))
+				requireManagedSecretEnv(t, initContainer, "CONNECTION_REDIS_URL", managedSecretName, "url")
+
+				podSpec, err := json.Marshal(deployment.Spec.Template.Spec)
+				require.NoError(t, err)
+				if bytes.Contains(podSpec, secretValue) {
+					require.FailNow(t, "pod spec contains the plaintext secret")
+				}
```
nithyatsu: The scan covers `deployment.Spec.Template.Spec`, so pod-template annotations/labels and the Deployment's own metadata (e.g. `kubectl.kubernetes.io/last-applied-configuration`) aren't included. Marshalling the whole `deployment` would close those.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+				pods, err := test.Options.K8sClient.CoreV1().Pods(name).List(ctx, metav1.ListOptions{
+					LabelSelector: metav1.FormatLabelSelector(deployment.Spec.Selector),
+				})
+				require.NoError(t, err)
+				require.Len(t, pods.Items, 1)
```
nithyatsu: `require.Len(t, pods.Items, 1)` uses the Deployment selector, which also matches pods from older ReplicaSets during a rollout or restart. Selecting on the current ReplicaSet, or asserting `>= 1` and picking a Ready pod, would be steadier.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+				appLogs, err := testutil.GetPodLogs(ctx, test.Options.K8sClient, name, pods.Items[0].Name, appContainer.Name)
+				require.NoError(t, err)
+				require.Contains(t, appLogs, "managed-secret-connection-ready")
+
+				initLogs, err := testutil.GetPodLogs(ctx, test.Options.K8sClient, name, pods.Items[0].Name, initContainer.Name)
+				require.NoError(t, err)
+				require.Contains(t, initLogs, "managed-secret-init-ready")
```
nithyatsu: Two things about `testutil.GetPodLogs` here.

First, reliability: its implementation (`test/testutil/testutil.go:418`) does a single `Read` into a 1MB buffer rather than draining to EOF, so it can return a partial or empty result before the container's `echo` is flushed. `require.Eventually` around these assertions would make it deterministic.

Second, and worth deciding separately: `GetPodLogs` currently has no other callers in the repo. It was added in #9958 and its original caller is gone, so it's been dead code on `main` since — this PR is its first exercised use. The other log readers here drain fully, via `io.ReadAll` (`test/functional-portable/database/noncloud/database_install_test.go:410`) or `.DoRaw` (`test/functional-portable/upgrade/upgrade_test.go:473`). So either we fix the helper to `io.ReadAll` (which also makes it dependable for the next caller), or we drop it and use one of the existing patterns and delete the dead helper. We could log a ticket for the cleanup if you'd rather not widen this PR.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
```diff
@@ -0,0 +1,198 @@
...
+			},
```
nithyatsu, line 170: Nothing currently asserts the managed Kubernetes Secret and the `Radius.Security/secrets` resource are cleaned up on delete. `PostDeleteVerify` (`test/rp/rptest.go:142`) is available for it. An orphaned Secret holding a live credential seems in scope for this test's intent.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Updated the test to retain the managed `Radius.Security/secrets` response, assert that `properties.url` is absent, and verify the returned properties do not contain the actual Redis URL from the backing Secret.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Added `PostDeleteVerify` coverage that waits for both the managed Radius secret resource to return HTTP 404 and its backing Kubernetes Secret to return `IsNotFound`.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Expanded the plaintext scan to marshal the complete Deployment, so Deployment metadata and the full pod template, including labels and annotations, are covered.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Fixed `GetPodLogs` to drain the stream with `io.ReadAll`, added a focused regression test covering a response larger than 1 MiB, and added eventual retries around both log marker assertions for Kubernetes log availability.

test/functional-portable/corerp/noncloud/resources/container_managed_secret_test.go
willdavsmith: Replaced the singleton selector assumption with current-ReplicaSet resolution using controller ownership and the Deployment revision, then select a non-deleting Running/Ready pod owned by that ReplicaSet.

nithyatsu (approved):

PR #12822, opened 2026-08-25
Cascade delete applications and resources when deleting a Radius.Core environment

### LINKED ISSUE #12486: Radius.Core/environments differs in deletion behaviour from Applications.Core/environments type
Deletion call on Applications.Core/environments enumerates every application in the environment and deletes each application (which in turn deletes its resources) before removing the environment itself (pkg/cli/clients/management.go:675 ):
```
applications, _ := amc.ListApplicationsInEnvironment(ctx, name)
for _, application := range applications {
    _, err := amc.DeleteApplication(ctx, *application.ID, false)  // cascade: app -> its containers
}
```
the new type Radius.Core/environments deletes only the environment resource. 
```
client := r.RadiusCoreClientFactory.NewEnvironmentsClient()
_, err := client.Delete(ctx, r.Workspace.Scope, r.EnvironmentName, ...)  // environment only
```

lakshmimsft:
## Summary
`rad env delete --preview` now deletes everything that lives in the environment instead of
only the environment resource itself, matching the behavior of `Applications.Core/environments`.

When deleting a `Radius.Core` environment, the CLI now:

1. Looks up the environment, and treats a missing environment as a successful no-op so the
   command is idempotent.
2. Enumerates the applications whose `properties.environment` references the environment,
   plus the resources reported both by the environment and by each of those applications.
3. Shows a confirmation prompt that states how many applications and resources will be
   deleted, instead of the previous hardcoded "environment is empty" message.
4. Deletes the resources concurrently, then the applications, then the environment.

A `--force` flag is also added to the preview command so resources stuck in a non-terminal
state can be deleted, mirroring `rad app delete --force`.

The cascade logic is shared with `rad app delete --preview`. That command previously carried
its own private copy of the resource-enumeration logic; it now uses the same helpers, which
removes the duplication.

## Reason for change

The `Radius.Core/environments` env  when used in `rad env delete --preview` deleted only the environment resource and silently orphaned every application and resource deployed into it. The confirmation prompt made this worse by always
claiming the environment was empty, so users consented to a deletion that did not describe what actually happened.

The work is unblocked by #12728, which made `fullyQualifyID` pass fully qualified resource IDs
through unchanged. Fully qualified `Radius.Core` IDs can now be handed to
`ListResourcesInApplication` and `ListResourcesInEnvironment` and reach their filters intact,
so the preview commands no longer need their own enumeration code.

Fixes #12486

## How to test
Manual, against a cluster with Radius installed:

1. Create a preview environment and deploy an application into it.

2. Confirm the cascade prompt reports the real contents, rather than claiming the environment
   is empty:

   ```bash
   rad env delete my-env --preview
   ```

3. Confirm the applications and their resources are gone, not just the environment:

   ```bash
   rad app list --preview
   rad resource list --preview
   ```

4. Confirm the command is idempotent — a second delete reports that the environment was not
   found and exits successfully:

   ```bash
   rad env delete my-env --preview
   ```

5. Confirm the flag wiring:

   ```bash
   rad env delete --help                 # lists both --preview and --force
   rad env delete my-env --force         # rejected: "--force requires preview mode"
   rad env delete my-env --preview --force
   ```

6. Confirm the legacy path is unchanged:

   ```bash
   rad env delete my-env
   ```

## File change summary
| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/cmd/previewdelete.go` | New. Shared preview delete helpers: `PreviewResourceID`/`PreviewApplicationID`/`PreviewEnvironmentID` ID builders, `ListPreviewApplicationsInEnvironment`, `MergeResourcesByID`, and `DeleteResourcesInParallel`. |
| `pkg/cli/cmd/previewdelete_test.go` | New. Unit tests for the ID builders, merge/dedupe behavior, and parallel deletion including 404 tolerance, skipping of malformed entries, error surfacing, and `force` passthrough. |
| `pkg/cli/cmd/env/delete/preview/delete.go` | Cascade implementation: enumerate applications and resources, delete resources then applications then the environment, accurate three-way confirmation prompt, no-op on a missing environment, new `--force` flag, and updated help text. |
| `pkg/cli/cmd/env/delete/preview/cascade_test.go` | New. Behavioral coverage of the cascade: missing environment, empty environment, full cascade with deletion ordering, cross-environment isolation, deduplication, all three prompt messages, declining the prompt, `--force`, and resource, application, and enumeration failures. |
| `pkg/cli/cmd/env/delete/preview/delete_test.go` | Removed the `Test_Run` case superseded by `cascade_test.go`, along with its now-unused imports. |
| `pkg/cli/cmd/app/delete/preview/delete.go` | Replaced the private `listResourcesOwnedByApplication`/`isResourceOwnedByApplication` duplicates with `ListResourcesInApplication` plus the shared `DeleteResourcesInParallel`. |
| `pkg/cli/cmd/app/delete/preview/delete_test.go` | Retargeted mocks from `ListAllResourceTypesNames`/`ListResourcesOfType` onto `ListResourcesInApplication`, and asserted the fully qualified application ID contract. |
| `cmd/rad/cmd/root.go` | Wired env delete through `wirePreviewSubcommandPreviewBase` so the preview command is the executed base. This is what makes `--force` visible at parse time in preview mode and rejected outside of it. |
| `cmd/rad/cmd/root_test.go` | Regression guards on the assembled command tree covering `--preview`/`--force` exposure and the rejection of `--force` without preview. |
| `pkg/cli/clients/management_test.go` | Added table-driven tests for `isResourceInApplication` and `isResourceInEnvironment`, including case-insensitive matching, preserving the coverage previously provided by the removed helpers in `app/delete/preview`. |

### DIFF
--- a/cmd/rad/cmd/root.go
+++ b/cmd/rad/cmd/root.go
@@ -381,8 +381,8 @@ func initSubCommands() {
 
 	envDeleteCmd, _ := env_delete.NewCommand(framework)
 	previewDeleteCmd, _ := env_delete_preview.NewCommand(framework)
-	wirePreviewSubcommand(envDeleteCmd, previewDeleteCmd)
-	envCmd.AddCommand(envDeleteCmd)
+	wirePreviewSubcommandPreviewBase(previewDeleteCmd, envDeleteCmd.RunE, "Use the Radius.Core preview implementation for environment delete", "force")
+	envCmd.AddCommand(previewDeleteCmd)
 
 	envListCmd, _ := env_list.NewCommand(framework)
 	previewListCmd, _ := env_list_preview.NewCommand(framework)
--- a/cmd/rad/cmd/root_test.go
+++ b/cmd/rad/cmd/root_test.go
@@ -449,6 +449,24 @@ func Test_EnvDelete_ExposesPreviewAndForceFlags(t *testing.T) {
 	require.NotNil(t, deleteCmd.Flags().Lookup("force"), "rad env delete must expose --force")
 }
 
+// Test_EnvDelete_ExposesFlagsLegacyRunnerReads guards the coupling created by registering the
+// preview command as the base for `rad env delete`: the legacy runner now validates against the
+// preview command's flag set. If any of these flags is dropped from the preview command, plain
+// `rad env delete` (without --preview) breaks at runtime with "flag accessed but not defined".
+// Test_EnvPreviewOnlyFlagsRejectedWithoutPreview cannot catch this, because it returns in the
+// --force guard before it ever reaches the legacy runner.
+func Test_EnvDelete_ExposesFlagsLegacyRunnerReads(t *testing.T) {
+	deleteCmd, _, err := RootCmd.Find([]string{"env", "delete"})
+	require.NoError(t, err)
+
+	// Read by the legacy runner's Validate, directly or via cli.RequireWorkspace,
+	// cli.RequireEnvironmentNameArgs and cli.RequireOutput.
+	for _, flag := range []string{"workspace", "group", "environment", "yes", "output"} {
+		require.NotNilf(t, deleteCmd.Flags().Lookup(flag),
+			"rad env delete must expose --%s: the legacy runner reads it during Validate", flag)
+	}
+}
+
 // Test_EnvPreviewOnlyFlagsRejectedWithoutPreview drives the real, fully-assembled command tree
 // and asserts that each preview-only flag is rejected when preview mode is off. This guards
 // against a preview-only flag being added to a command without also being registered in the
--- a/cmd/rad/cmd/root_test.go
+++ b/cmd/rad/cmd/root_test.go
@@ -437,6 +437,18 @@ func Test_ResourceList_ExposesPreviewFlag(t *testing.T) {
 	require.NotNil(t, listCmd.Flags().Lookup("preview"), "rad resource list must expose --preview")
 }
 
+// Test_EnvDelete_ExposesPreviewAndForceFlags asserts against the fully-assembled command tree that
+// `rad env delete` exposes both flags. The preview runner reads --force during Validate, so if the
+// wired command does not declare it, every `rad env delete --preview` invocation fails with
+// "flag accessed but not defined: force".
+func Test_EnvDelete_ExposesPreviewAndForceFlags(t *testing.T) {
+	deleteCmd, _, err := RootCmd.Find([]string{"env", "delete"})
+	require.NoError(t, err)
+	require.Equal(t, "delete", deleteCmd.Name())
+	require.NotNil(t, deleteCmd.Flags().Lookup("preview"), "rad env delete must expose --preview")
+	require.NotNil(t, deleteCmd.Flags().Lookup("force"), "rad env delete must expose --force")
+}
+
 // Test_EnvPreviewOnlyFlagsRejectedWithoutPreview drives the real, fully-assembled command tree
 // and asserts that each preview-only flag is rejected when preview mode is off. This guards
 // against a preview-only flag being added to a command without also being registered in the
@@ -452,6 +464,7 @@ func Test_EnvPreviewOnlyFlagsRejectedWithoutPreview(t *testing.T) {
 		{command: "create", flag: "recipe-packs", value: "p1"},
 		{command: "update", flag: "recipe-packs", value: "p1"},
 		{command: "update", flag: "clear-kubernetes", value: "true"},
+		{command: "delete", flag: "force", value: "true"},
 	}
 
 	for _, tc := range testcases {
--- a/pkg/cli/cmd/env/delete/preview/delete.go
+++ b/pkg/cli/cmd/env/delete/preview/delete.go
@@ -23,8 +23,12 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/radius-project/radius/pkg/cli"
+	"github.com/radius-project/radius/pkg/cli/clients"
+	generated "github.com/radius-project/radius/pkg/cli/clients_new/generated"
+	"github.com/radius-project/radius/pkg/cli/clierrors"
 	"github.com/radius-project/radius/pkg/cli/cmd"
 	"github.com/radius-project/radius/pkg/cli/cmd/commonflags"
+	"github.com/radius-project/radius/pkg/cli/connections"
 	"github.com/radius-project/radius/pkg/cli/framework"
 	"github.com/radius-project/radius/pkg/cli/output"
 	"github.com/radius-project/radius/pkg/cli/prompt"
@@ -35,6 +39,10 @@ import (
 const (
 	msgEnvironmentDeletedPreview  = "Radius.Core/environments/%s deleted"
 	msgEnvironmentNotFoundPreview = "Radius.Core/environments/%s not found"
+	msgDeletingResources          = "Deleting %d resource(s) in environment %s...\n"
+	msgDeletingApplications       = "Deleting %d application(s) in environment %s...\n"
+	msgDeletingApplication        = "  Deleting application %s..."
+	msgForceWarning               = "WARNING: Force deleting an environment. Resources in non-terminal states may leave orphaned external resources that require manual cleanup."
 )
 
 // NewCommand creates an instance of the command and runner for the `rad env delete --preview` command.
@@ -42,18 +50,40 @@ func NewCommand(factory framework.Factory) (*cobra.Command, framework.Runner) {
 	runner := NewRunner(factory)
 
 	cmd := &cobra.Command{
-		Use:     "delete",
-		Short:   "Delete environment (preview)",
-		Long:    `Delete environment using the Radius.Core preview API surface.`,
-		Args:    cobra.MaximumNArgs(1),
-		RunE:    framework.RunCommand(runner),
-		Example: `rad env delete myenv`,
+		Use:   "delete",
+		Short: "Delete environment",
+		Long: `Delete environment. Deletes the user's default environment by default.
+
+In preview mode, deleting an environment also deletes the applications in that environment
+and the resources deployed into it.`,
+		Args: cobra.MaximumNArgs(1),
+		RunE: framework.RunCommand(runner),
+		Example: `
+# Delete current environment
+rad env delete
+
+# Delete current environment and bypass confirmation prompt
+rad env delete --yes
+
+# Delete specified environment
+rad env delete my-env
+
+# Delete specified environment in a specified resource group
+rad env delete my-env --group my-env
+
+# Delete a Radius.Core environment and everything deployed into it
+rad env delete my-env --preview
+
+# Delete a Radius.Core environment, forcing deletion of resources in a non-terminal state
+rad env delete my-env --force --preview
+`,
 	}
 
 	commonflags.AddWorkspaceFlag(cmd)
 	commonflags.AddResourceGroupFlag(cmd)
 	commonflags.AddEnvironmentNameFlag(cmd)
 	commonflags.AddConfirmationFlag(cmd)
+	commonflags.AddForceFlag(cmd)
 	commonflags.AddOutputFlag(cmd)
 
 	return cmd, runner
@@ -64,19 +94,22 @@ type Runner struct {
 	ConfigHolder            *framework.ConfigHolder
 	Output                  output.Interface
 	InputPrompter           prompt.Interface
+	ConnectionFactory       connections.Factory
 	Workspace               *workspaces.Workspace
 	RadiusCoreClientFactory *corerpv20250801.ClientFactory
 
 	Confirm         bool
+	Force           bool
 	EnvironmentName string
 }
 
 // NewRunner creates a new instance of the preview delete runner.
 func NewRunner(factory framework.Factory) *Runner {
 	return &Runner{
-		ConfigHolder:  factory.GetConfigHolder(),
-		Output:        factory.GetOutput(),
-		InputPrompter: factory.GetPrompter(),
+		ConfigHolder:      factory.GetConfigHolder(),
+		Output:            factory.GetOutput(),
+		InputPrompter:     factory.GetPrompter(),
+		ConnectionFactory: factory.GetConnectionFactory(),
 	}
 }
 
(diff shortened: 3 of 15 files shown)


pkg/cli/cmd/env/delete/preview/delete.go
```diff
@@ -123,13 +166,69 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		return err
+	}
+
+	resourceLists := [][]generated.GenericResource{resourcesInEnvironment}
+	for _, application := range applications {
+		if application.ID == nil {
+			continue
+		}
+
+		resourcesInApplication, err := managementClient.ListResourcesInApplication(ctx, *application.ID)
```
kachawla: This enumerates resources once per application on top of the environment-wide pass on line 199, and each `ListResourcesInApplication` call re-runs `ListAllResourceTypesNames` and then, for every registered type, `getApiVersionsForResourceType` plus a full `ListByRootScope` page walk.

With A applications and T resource types that is (A+1)×T list calls plus (A+1)×T provider-summary calls, all serial, over data the environment pass already fetched.

Both filters (`isResourceInEnvironment` / `isResourceInApplication`) run client-side against the same `ListResourcesOfType` result, so a single pass per type — filtered against the environment ID and the set of application IDs — keeps the defensive two-direction semantics described on 195-198 at 1/(A+1) the cost.

pkg/cli/cmd/env/delete/preview/delete.go
```diff
@@ -140,11 +239,43 @@ func (r *Runner) Run(ctx context.Context) error {
 		}
 	}
 
-	client := r.RadiusCoreClientFactory.NewEnvironmentsClient()
-	_, err := client.Delete(ctx, r.Workspace.Scope, r.EnvironmentName, &corerpv20250801.EnvironmentsClientDeleteOptions{})
-	if err != nil {
-		// If this is a 404, treat as successful but with a different message
-		// We don't have the Is404Error helper wired to Radius.Core yet, so always surface the error.
+	if r.Force {
```
kachawla: The `--force` warning prints *after* the confirmation prompt on 218-240, so the user consents to the cascade before seeing that force may leave orphaned external resources behind. Move it above the prompt so the warning is in view when they answer.

`rad app delete --preview` has the same ordering, so this is a shared fix rather than something introduced here — but it matters more at environment scope, where the blast radius is every application.

pkg/cli/cmd/env/delete/preview/delete.go
```diff
@@ -140,11 +239,43 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	}
+
+	// Applications are deleted after their resources, so that an interrupted delete leaves the
+	// application behind as a recovery point rather than orphaning its resources.
+	if len(applications) > 0 {
+		r.Output.LogInfo(msgDeletingApplications, len(applications), r.EnvironmentName)
+
+		appClient := r.RadiusCoreClientFactory.NewApplicationsClient()
+		for _, application := range applications {
+			if application.Name == nil {
```
kachawla: An application with a nil `Name` is skipped silently and the environment is still deleted on line 271, orphaning it. Same class of issue as the ID/Type skip in `previewdelete.go`.

At minimum log a warning here. The cascade should not report success for something it did not delete.

pkg/cli/cmd/previewdelete.go
```diff
@@ -0,0 +1,133 @@
...
+func PreviewEnvironmentID(scope string, environmentName string) string {
+	return PreviewResourceID(scope, datamodel.EnvironmentResourceType_v20250801preview, environmentName)
+}
+
+// DeleteResourcesInParallel deletes the given resources concurrently, tolerating resources that
+// have already been deleted. Resources missing an ID or type are skipped. The ID of each
+// resource is logged before its deletion is started, because output.Interface implementations are
+// not guaranteed to be thread-safe and logging up front keeps the output deterministic.
+func DeleteResourcesInParallel(ctx context.Context, client clients.ApplicationsManagementClient, out output.Interface, resources []generated.GenericResource, force bool) error {
+	g, groupCtx := errgroup.WithContext(ctx)
```
kachawla: No concurrency bound. In the app-delete path this was naturally capped by one application's resource count; at environment scope it is every resource across every application, so this can spawn hundreds of simultaneous long-running delete pollers against the RP. Add `g.SetLimit(n)`.

Related, and worth a comment in the doc block: on the first failure `errgroup` cancels `groupCtx`, which abandons the in-flight pollers for every other delete. Their server-side outcome is unknown while the command surfaces a single error, and re-running is the only recovery path.

pkg/cli/cmd/previewdelete.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, resource := range resources {
+		if resource.ID == nil || resource.Type == nil {
```
kachawla: Resources missing an ID or Type are skipped silently, but `MergeResourcesByID` only drops nil-ID entries. A nil-Type resource is therefore counted in `len(resourcesToDelete)` and reported to the user in the confirmation prompt (`env/delete/preview/delete.go:222-227`), never deleted, and then orphaned when the environment is deleted.

The `skips resources without an ID or type` case in `previewdelete_test.go` locks the silent behavior in. Log a warning per skipped resource so the prompt count and the actual outcome cannot diverge unnoticed.

cmd/rad/cmd/root.go
```diff
@@ -381,8 +381,8 @@ func initSubCommands() {
 
 	envDeleteCmd, _ := env_delete.NewCommand(framework)
 	previewDeleteCmd, _ := env_delete_preview.NewCommand(framework)
-	wirePreviewSubcommand(envDeleteCmd, previewDeleteCmd)
-	envCmd.AddCommand(envDeleteCmd)
+	wirePreviewSubcommandPreviewBase(previewDeleteCmd, envDeleteCmd.RunE, "Use the Radius.Core preview implementation for environment delete", "force")
```
kachawla: The preview command is now the registered command, so the legacy `env_delete` runner validates against the *preview* command's flag set. Its `Validate` reads `workspace`, `group`, `environment`, `yes`, and `output` — if any of those is ever dropped from the preview command, plain `rad env delete` fails at runtime with `flag accessed but not defined`.

Nothing exercises that path today: `Test_EnvPreviewOnlyFlagsRejectedWithoutPreview` short-circuits in the `--force` guard before it ever reaches `legacyRunE`.

cmd/rad/cmd/root_test.go
```diff
@@ -437,6 +437,18 @@ func Test_ResourceList_ExposesPreviewFlag(t *testing.T) {
...
+// Test_EnvDelete_ExposesPreviewAndForceFlags asserts against the fully-assembled command tree that
+// `rad env delete` exposes both flags. The preview runner reads --force during Validate, so if the
+// wired command does not declare it, every `rad env delete --preview` invocation fails with
+// "flag accessed but not defined: force".
+func Test_EnvDelete_ExposesPreviewAndForceFlags(t *testing.T) {
+	deleteCmd, _, err := RootCmd.Find([]string{"env", "delete"})
+	require.NoError(t, err)
+	require.Equal(t, "delete", deleteCmd.Name())
+	require.NotNil(t, deleteCmd.Flags().Lookup("preview"), "rad env delete must expose --preview")
+	require.NotNil(t, deleteCmd.Flags().Lookup("force"), "rad env delete must expose --force")
```
kachawla: Extend this to assert the flags the legacy runner actually reads — `yes`, `output`, `group`, `environment`, `workspace` — not just `preview` and `force`. That is the cheap guard for the coupling `root.go:384` introduces between the legacy runner and the preview command's flag set.

pkg/cli/cmd/env/delete/preview/delete.go
kachawla: This lists resources once per application, on top of the environment-wide pass on line 199. Each `ListResourcesInApplication` call re-runs `ListAllResourceTypesNames`, then for every registered type calls `getApiVersionsForResourceType` and walks all pages of `ListByRootScope`.

With A applications and T resource types that is (A+1)×T list calls and (A+1)×T provider summary calls, all serial, over data the environment pass already fetched.

Both filters (`isResourceInEnvironment` and `isResourceInApplication`) run client side on the same `ListResourcesOfType` result. One pass per type, filtered against the environment ID and the set of application IDs, gives you the same two-direction coverage you describe on 195-198 at 1/(A+1) the cost.

pkg/cli/cmd/env/delete/preview/delete.go
kachawla: The `--force` warning prints after the confirmation prompt on 218-240. The user agrees to the cascade before seeing that force can leave orphaned external resources. Move it above the prompt.

`rad app delete --preview` has the same ordering, so this is a shared fix and not a regression here. It matters more at environment scope, where the blast radius is every application.

pkg/cli/cmd/env/delete/preview/delete.go
kachawla, line 255: An application with a nil `Name` is skipped with no output, and the environment is still deleted on line 271, so the application is orphaned. Same problem as the ID/Type skip in `previewdelete.go`.

Log a warning here at minimum. The cascade should not report success for something it did not delete.

pkg/cli/cmd/previewdelete.go
kachawla, line 77: No concurrency limit. In the app delete path this was capped by one application's resource count. At environment scope it is every resource across every application, so this can start hundreds of long-running delete pollers against the RP at once. Add `g.SetLimit(n)`.

Also worth a note in the doc comment: on the first failure `errgroup` cancels `groupCtx`, which drops the in-flight pollers for every other delete. Their server-side outcome is unknown while the command returns a single error, and re-running is the only way to recover.

pkg/cli/cmd/previewdelete.go
kachawla, line 81: Resources missing an ID or Type are skipped with no output, but `MergeResourcesByID` only drops nil-ID entries. So a nil-Type resource is counted in `len(resourcesToDelete)`, shown to the user in the confirmation prompt (`env/delete/preview/delete.go:222-227`), never deleted, then orphaned when the environment goes away.

The `skips resources without an ID or type` case in `previewdelete_test.go` locks that behavior in. Log a warning per skipped resource so the count in the prompt cannot quietly disagree with what actually happened.

cmd/rad/cmd/root.go
kachawla, line 384: The preview command is now the registered command, so the legacy `env_delete` runner validates against the preview command's flag set. Its `Validate` reads `workspace`, `group`, `environment`, `yes`, and `output`. If any of those is dropped from the preview command later, plain `rad env delete` fails at runtime with `flag accessed but not defined`.

Nothing covers that path today. `Test_EnvPreviewOnlyFlagsRejectedWithoutPreview` returns in the `--force` guard before it reaches `legacyRunE`.

cmd/rad/cmd/root_test.go
kachawla, line 449: Assert the flags the legacy runner actually reads (`yes`, `output`, `group`, `environment`, `workspace`), not just `preview` and `force`. That is the cheap guard for the coupling `root.go:384` creates between the legacy runner and the preview command's flag set.

pkg/cli/cmd/env/delete/preview/delete.go
lakshmimsft: updated both env and app

cmd/rad/cmd/root.go
lakshmimsft: Added  Test_EnvDelete_ExposesFlagsLegacyRunnerReads

cmd/rad/cmd/root_test.go
lakshmimsft: Done —  Test_EnvDelete_ExposesFlagsLegacyRunnerReads  asserts  workspace ,  group ,  environment ,  yes ,  output  on the assembled  RootCmd

pkg/cli/cmd/env/delete/preview/delete.go
lakshmimsft: added  msgSkippingApplication  and a warning in the loop, with a  cascade_test.go  case asserting i

pkg/cli/cmd/previewdelete.go
lakshmimsft: Added  maxParallelDeletes = 10  and  g.SetLimit(...) , can revisit if 10 is to be updated.

pkg/cli/cmd/env/delete/preview/delete.go
```diff
@@ -35,25 +38,52 @@ import (
...
-		Short:   "Delete environment (preview)",
-		Long:    `Delete environment using the Radius.Core preview API surface.`,
-		Args:    cobra.MaximumNArgs(1),
-		RunE:    framework.RunCommand(runner),
-		Example: `rad env delete myenv`,
+		Use:   "delete",
+		Short: "Delete environment",
+		Long: `Delete environment. Deletes the user's default environment by default.
+
+In preview mode, deleting an environment also deletes the applications in that environment
```
kachawla, line 57: This help text is the only place the cascade is written down. The behavior change is user facing (deleting an environment now deletes applications and resources), and `--force` is a new flag, so both need a docs repo update before release. Nothing in `docs/contributing/` or `docs/architecture/` covers `rad env delete`, so there is nothing to change in this repo.

Not blocking this PR. Filing a docs issue is enough.

pkg/cli/cmd/env/delete/preview/delete.go
```diff
@@ -35,25 +38,52 @@ import (
...
+# Delete a Radius.Core environment, forcing deletion of resources in a non-terminal state
+rad env delete my-env --force --preview
+`,
 	}
 
 	commonflags.AddWorkspaceFlag(cmd)
 	commonflags.AddResourceGroupFlag(cmd)
 	commonflags.AddEnvironmentNameFlag(cmd)
 	commonflags.AddConfirmationFlag(cmd)
+	commonflags.AddForceFlag(cmd)
```
kachawla, line 86: Unrelated to this line, leaving it here because there is no better anchor.

The PR title says "Cascade delete for resources", but the change does three things: it cascades to applications as well as resources, it adds `--force`, and it rewires `rad env delete` so the preview command is the registered base. The last one changes how the legacy path resolves flags, which is the least obvious part of the PR and is invisible from the title.

Something like "Cascade delete applications and resources when deleting a Radius.Core environment" would read better in the release notes and in git log.

kachawla (approved):
Two non blocking comments that I missed in my earlier review

PR #12824, opened 2026-08-25
Add depreciation warning to `rad deploy` for Application.* types

### LINKED ISSUE #12643: Add deprecation warning to legacy APIs

lakshmimsft:
## Summary
Adds a deprecation warning to `rad deploy` when a Bicep template uses a legacy `Applications.*` resource type with the `2023-10-01-preview` API version. Each deprecated type is reported alongside the extensible `Radius.*` type that replaces it.

Example output:

```text
WARNING: The following resource types use the deprecated Applications.* namespace with API version 2023-10-01-preview and will be removed in a future release:
  - Applications.Core/applications@2023-10-01-preview: use Radius.Core/applications@2025-08-01-preview instead.
  - Applications.Core/containers@2023-10-01-preview: use Radius.Compute/containers@2025-08-01-preview instead.
```

Types without a one-to-one replacement (`Applications.Core/extenders`, all `Applications.Dapr/*`) fall back to generic guidance directing users to define an equivalent resource type and register a recipe pack that provides it.

## Reason for change
The replacement types now exist as opposed to earlier and are registered by default in `deploy/manifest/defaults.yaml`. 
Notes:
- Emitted from `rad deploy` only. This is the authoring path, where the user can act on the warning immediately. Read paths such as `rad resource list` were deliberately excluded: they report resources already deployed, would re-fire on every read, and cannot carry an API version in their arguments — the repeat-noise-without-recourse pattern that motivated the original revert.

Fixes #12643

## How to test

1. Deploy any template that uses a legacy type, for example:

   ```bicep
   resource app 'Applications.Core/applications@2023-10-01-preview' = {
     name: 'my-app'
     properties: { environment: environment }
   }
   ```

   ```bash
   rad deploy app.bicep
   ```

   The warning is printed once, above the deployment progress output, and the deployment proceeds normally.

2. Confirm a modern template is silent — a template using only `Radius.*` types produces no warning.

3. Confirm module coverage — declare a legacy resource inside a Bicep `module`. It is still reported, because module contents compile into a nested `Microsoft.Resources/deployments` template.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/bicep/resources.go` | Adds `deprecatedTypeReplacements` (legacy → `Radius.*` mapping, kept in sync with `deploy/manifest/defaults.yaml`), the `DeprecatedResource` type, and `FormatDeprecationWarning`. Adds `collectDeprecatedResources`, which recurses into nested module templates and de-duplicates types case-insensitively, and sorts results for deterministic output. Splits inspection into `inspectEnvironmentResources` and `inspectDeprecatedResources` so each exported wrapper performs only the scan it needs. |
| `pkg/cli/bicep/resources_test.go` | Adds `Test_GetDeprecatedResources` and `Test_FormatDeprecationWarning`, covering the mapping, unmapped-type fallback, case-insensitive matching, non-matching API versions, non-Radius namespaces, missing API version, duplicate types, and resources nested inside single and doubly-nested modules. |
| `pkg/cli/cmd/deploy/deploy.go` | Emits the warning at the start of `Run()`, reusing the `TemplateInspectionResult` already computed once during `Validate()`. |
| `pkg/cli/cmd/deploy/deploy_test.go` | Adds a `Test_Run` case asserting the exact warning output for both a mapped and an unmapped deprecated type. |

### DIFF
--- a/pkg/cli/bicep/resources.go
+++ b/pkg/cli/bicep/resources.go
@@ -17,6 +17,8 @@ limitations under the License.
 package bicep
 
 import (
+	"fmt"
+	"sort"
 	"strings"
 )
 
@@ -26,7 +28,48 @@ const (
 
 	// legacyEnvironmentResourceType is the legacy resource type for Radius environments
 	legacyEnvironmentResourceType = "applications.core/environments"
+
+	// deprecatedNamespacePrefix is the namespace prefix of the legacy Radius resource types.
+	deprecatedNamespacePrefix = "applications."
+
+	// deprecatedAPIVersion is the API version of the legacy Radius resource types.
+	deprecatedAPIVersion = "2023-10-01-preview"
+
+	// replacementAPIVersion is the API version of the extensible Radius.* resource types that
+	// replace the legacy Applications.* resource types.
+	replacementAPIVersion = "2025-08-01-preview"
 )
+
+// deprecatedTypeReplacements maps a lowercased legacy Applications.* resource type to the
+// extensible Radius.* resource type that replaces it. Types that are absent from this map have no
+// direct one-to-one replacement and are reported with generic recipe pack guidance instead.
+//
+// Keep this in sync with the built-in resource types in deploy/manifest/defaults.yaml.
+var deprecatedTypeReplacements = map[string]string{
+	"applications.core/applications":         "Radius.Core/applications",
+	"applications.core/environments":         "Radius.Core/environments",
+	"applications.core/containers":           "Radius.Compute/containers",
+	"applications.core/gateways":             "Radius.Compute/routes",
+	"applications.core/volumes":              "Radius.Compute/persistentVolumes",
+	"applications.core/secretstores":         "Radius.Security/secrets",
+	"applications.datastores/mongodatabases": "Radius.Data/mongoDatabases",
+	"applications.datastores/rediscaches":    "Radius.Data/redisCaches",
+	"applications.datastores/sqldatabases":   "Radius.Data/sqlServerDatabases",
+	"applications.messaging/rabbitmqqueues":  "Radius.Messaging/rabbitMQ",
+}
+
+// DeprecatedResource describes a resource in a template that uses a legacy Applications.* resource
+// type with the deprecated API version.
+type DeprecatedResource struct {
+	// FullType is the resource type as authored, including the API version,
+	// e.g. "Applications.Core/containers@2023-10-01-preview".
+	FullType string
+
+	// Replacement is the extensible Radius.* resource type that replaces this type, including the
+	// API version, e.g. "Radius.Compute/containers@2025-08-01-preview". It is empty when no direct
+	// replacement exists.
+	Replacement string
+}
 
 // radiusNamespacePatterns lists namespace prefixes for resource type namespaces that belong to Radius.
 // Resource types matching these prefixes are routed through the Radius control plane, not Azure ARM.
@@ -42,6 +85,10 @@ type TemplateInspectionResult struct {
 
 	// EnvironmentResources contains the list of environment resources found in the template.
 	EnvironmentResources []map[string]any
+
+	// DeprecatedResources contains the resources that use a legacy Applications.* resource type
+	// with the deprecated API version, along with their replacements.
+	DeprecatedResources []DeprecatedResource
 }
 
 // ResourceTypeEntry represents a parsed resource type from a compiled Bicep/ARM template.
@@ -136,23 +183,38 @@ func HasOnlyRadiusResourceTypes(template map[string]any) bool {
 // The expected structure of resource in the template is:
 // {"resources": {"resourceName": {"type": "Applications.Core/containers@2023-10-01-preview", ...}}}
 func InspectTemplateResources(template map[string]any) TemplateInspectionResult {
-	result := TemplateInspectionResult{
-		ContainsEnvironmentResource: false,
-	}
+	containsEnvironment, environmentResources := inspectEnvironmentResources(template)
 
+	return TemplateInspectionResult{
+		ContainsEnvironmentResource: containsEnvironment,
+		EnvironmentResources:        environmentResources,
+		DeprecatedResources:         inspectDeprecatedResources(template),
+	}
+}
+
+// inspectEnvironmentResources walks the top-level resources of a compiled template and reports
+// whether an environment resource is present, along with the Radius.Core environment resources.
+//
+// This deliberately does not descend into nested module templates: the results drive deployment
+// behavior (whether the template creates its own environment, and default recipe pack injection),
+// so the scope must stay exactly as it has always been.
+func inspectEnvironmentResources(template map[string]any) (bool, []map[string]any) {
 	if template == nil {
-		return result
+		return false, nil
 	}
 
 	resourcesValue, ok := template["resources"]
 	if !ok {
-		return result
+		return false, nil
 	}
 
 	resources, ok := resourcesValue.(map[string]any)
 	if !ok {
-		return result
+		return false, nil
 	}
+
+	containsEnvironment := false
+	var environmentResources []map[string]any
 
 	for _, resourceValue := range resources {
 		resource, ok := resourceValue.(map[string]any)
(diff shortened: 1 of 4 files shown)


pkg/cli/bicep/resources.go
```diff
@@ -170,16 +232,112 @@ func InspectTemplateResources(template map[string]any) TemplateInspectionResult
...
+	}
+
+	// Symbolic-name templates use a map keyed by resource name; classic ARM templates and nested
+	// module templates may use an array instead, so handle both shapes.
+	var resourceValues []any
+	switch typed := resourcesValue.(type) {
+	case map[string]any:
+		for _, resourceValue := range typed {
+			resourceValues = append(resourceValues, resourceValue)
 		}
```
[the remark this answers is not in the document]
lakshmimsft: Fixed by iterating the map keys in sorted order in  collectDeprecatedResources. Added  Test_GetDeprecatedResources_IsDeterministic.

pkg/cli/bicep/resources.go
```diff
@@ -26,8 +28,49 @@ const (
...
+	// replacementAPIVersion is the API version of the extensible Radius.* resource types that
+	// replace the legacy Applications.* resource types.
+	replacementAPIVersion = "2025-08-01-preview"
 )
 
+// deprecatedTypeReplacements maps a lowercased legacy Applications.* resource type to the
+// extensible Radius.* resource type that replaces it. Types that are absent from this map have no
+// direct one-to-one replacement and are reported with generic recipe pack guidance instead.
+//
+// Keep this in sync with the built-in resource types in deploy/manifest/defaults.yaml.
```
kachawla: This sync pointer is wrong for two of the ten entries. `defaults.yaml`'s `defaultRegistration` only lists the contrib namespaces (`Radius.Compute`, `Radius.Data`, `Radius.Security`, `Radius.Messaging`, `Radius.AI`, `Radius.Storage`) — `Radius.Core/applications` and `Radius.Core/environments` come from `deploy/manifest/built-in-providers/{dev,self-hosted}/radius_core.yaml`. Someone following this comment would conclude those two entries are stale.

Related, on line 40: `replacementAPIVersion` is a single constant spanning six namespaces that `defaults.yaml` pins to independent upstream tags (`Radius.Compute/v0.1.0`, `Radius.Messaging/v0.1.1`). They all happen to be `2025-08-01-preview` today — I checked all ten and they resolve — but nothing fails if one moves, and the warning would then point users at a type that doesn't resolve.

A guard test that loads the embedded manifests and asserts every value in this map resolves to a registered type at `replacementAPIVersion` would make the mapping self-checking, and let this comment just point at that test.

pkg/cli/bicep/resources.go
```diff
@@ -170,16 +232,122 @@ func InspectTemplateResources(template map[string]any) TemplateInspectionResult
...
+			if deprecated, isDeprecated := newDeprecatedResource(resourceType); isDeprecated {
+				key := strings.ToLower(deprecated.FullType)
+				if _, alreadySeen := seen[key]; !alreadySeen {
+					seen[key] = struct{}{}
+					*out = append(*out, deprecated)
+				}
+			}
+		}
+
+		// Recurse into the inline template of a nested deployment (a Bicep module).
```
kachawla: The recursion isn't actually gated on `Microsoft.Resources/deployments` the way this comment says — it descends into any resource with a map-valued `properties.template`. `Applications.Core/extenders` accepts arbitrary user-defined properties, so an extender that happens to carry a `template` object would get walked.

Harmless in practice (worst case an extra warning line), but check `resource["type"]` before recursing so the code matches its comment.

pkg/cli/bicep/resources.go
```diff
@@ -188,11 +356,45 @@ func InspectTemplateResources(template map[string]any) TemplateInspectionResult
...
 // all environment resources found as maps.
 func GetEnvironmentResources(template map[string]any) []map[string]any {
-	return InspectTemplateResources(template).EnvironmentResources
+	_, environmentResources := inspectEnvironmentResources(template)
+	return environmentResources
+}
+
+// GetDeprecatedResources inspects the compiled Radius Bicep template's resources and returns the
+// resources that use a legacy Applications.* resource type with the deprecated API version.
+func GetDeprecatedResources(template map[string]any) []DeprecatedResource {
```
kachawla: `GetDeprecatedResources` has no production caller — `deploy.go:311` reads `r.TemplateInspectionResult.DeprecatedResources` directly. It's a one-line wrapper over `inspectDeprecatedResources`, which is in this same package, so the tests can call that instead.

Let's drop it rather than widen the `pkg/cli/bicep` API for test-only use. `FormatDeprecationWarning` does have a real caller, so that one should stay exported.

pkg/cli/bicep/resources.go
```diff
@@ -188,11 +356,45 @@ func InspectTemplateResources(template map[string]any) TemplateInspectionResult
...
+	fmt.Fprintf(&builder,
+		"WARNING: The following resource types use the deprecated Applications.* namespace with API version %s and will be removed in a future release:\n",
+		deprecatedAPIVersion)
+
+	for _, resource := range deprecated {
+		if resource.Replacement != "" {
+			fmt.Fprintf(&builder, "  - %s: use %s instead.\n", resource.FullType, resource.Replacement)
+		} else {
+			fmt.Fprintf(&builder,
+				"  - %s: no direct replacement. Define an equivalent resource type and register a recipe pack that provides it to your environment.\n",
```
kachawla: #11459 gave three reasons for the revert: the types weren't actually deprecated, there was no replacement, and there was no documentation. This PR fixes the first two well, but this fallback still tells users to "define an equivalent resource type and register a recipe pack" with nowhere to go.

Worth adding a docs URL here (and on the mapped-replacement line above) before this lands again — it's the part of the original objection that's still open, and it matters most for `Applications.Core/environments`, which was the loudest example in that issue.

pkg/cli/cmd/deploy/deploy.go
```diff
@@ -306,6 +306,13 @@ func (r *Runner) Run(ctx context.Context) error {
 	// Use the template that was prepared during validation
 	template := r.Template
 
+	// Warn about legacy Applications.* resource types before deploying, so the message is visible
+	// above the deployment progress output.
+	if warning := bicep.FormatDeprecationWarning(r.TemplateInspectionResult.DeprecatedResources); warning != "" {
```
kachawla: This also fires on `rad run`. `run.Runner` embeds `deploycmd.Runner` and `run.go:145` calls `r.Runner.Run(ctx)`, so the description's "Emitted from `rad deploy` only" isn't accurate.

The behavior itself is fine — `rad run` is equally an authoring path where the user can act on the warning — but given #11460 reverted the last attempt over exactly this blast-radius question, the description should say so, and a `rad run` test case would lock it in.

pkg/cli/bicep/resources.go
kachawla: This sync pointer is wrong for two of the ten entries. `defaults.yaml`'s `defaultRegistration` only lists the contrib namespaces (`Radius.Compute`, `Radius.Data`, `Radius.Security`, `Radius.Messaging`, `Radius.AI`, `Radius.Storage`). `Radius.Core/applications` and `Radius.Core/environments` come from `deploy/manifest/built-in-providers/{dev,self-hosted}/radius_core.yaml` instead, so someone following this comment would conclude those two entries are stale.

Related, on line 40: `replacementAPIVersion` is a single constant spanning six namespaces that `defaults.yaml` pins to independent upstream tags (`Radius.Compute/v0.1.0`, `Radius.Messaging/v0.1.1`). They all happen to be `2025-08-01-preview` today, and I checked that all ten resolve, but nothing fails if one moves. The warning would then point users at a type that doesn't resolve.

A guard test that loads the embedded manifests and asserts every value in this map resolves to a registered type at `replacementAPIVersion` would make the mapping self-checking, and let this comment just point at that test.

pkg/cli/bicep/resources.go
kachawla: The recursion isn't actually gated on `Microsoft.Resources/deployments` the way this comment says. It descends into any resource with a map-valued `properties.template`, and `Applications.Core/extenders` accepts arbitrary user-defined properties, so an extender that happens to carry a `template` object would get walked.

Harmless in practice (worst case an extra warning line), but check `resource["type"]` before recursing so the code matches its comment.

pkg/cli/bicep/resources.go
kachawla: `GetDeprecatedResources` has no production caller. `deploy.go:311` reads `r.TemplateInspectionResult.DeprecatedResources` directly, and this is a one-line wrapper over `inspectDeprecatedResources`, which is in this same package, so the tests can call that instead.

Let's drop it rather than widen the `pkg/cli/bicep` API for test-only use. `FormatDeprecationWarning` does have a real caller, so that one should stay exported.

pkg/cli/bicep/resources.go
kachawla: This tells users to "define an equivalent resource type and register a recipe pack" but gives them nowhere to go.

Worth adding a docs URL here, and on the mapped-replacement line above. A deprecation warning is only actionable if it points at the migration guidance.

pkg/cli/cmd/deploy/deploy.go
kachawla: This also fires on `rad run`. `run.Runner` embeds `deploycmd.Runner` and `run.go:145` calls `r.Runner.Run(ctx)`, so the description's "Emitted from `rad deploy` only" isn't accurate.

The behavior itself is fine, since `rad run` is equally an authoring path where the user can act on the warning, but the description should say so, and a `rad run` test case would lock it in.

pkg/cli/bicep/resources.go
kachawla: `Radius.Core/applications` and `Radius.Core/environments` aren't in `defaults.yaml`. Its `defaultRegistration` only covers the contrib namespaces; those two come from `built-in-providers/{dev,self-hosted}/radius_core.yaml`. Point the comment at both files.

All ten replacements resolve today, but nothing enforces that. A guard test over the embedded manifests would be worth adding.

pkg/cli/bicep/resources.go
kachawla: This recurses into any resource with a map-valued `properties.template`, not just `Microsoft.Resources/deployments`. `Applications.Core/extenders` takes arbitrary properties, so an extender carrying a `template` object would get walked.

Harmless today, but check `resource["type"]` so the code matches the comment.

pkg/cli/bicep/resources.go
kachawla: No production caller. `deploy.go:311` reads `r.TemplateInspectionResult.DeprecatedResources` directly, and this just wraps `inspectDeprecatedResources` from the same package, which the tests can call instead.

Drop it rather than widen the `pkg/cli/bicep` API for test-only use. `FormatDeprecationWarning` has a real caller, so that one stays.

pkg/cli/bicep/resources.go
kachawla, line 405: This tells users to define a type and register a recipe pack but gives them nowhere to go. Add a docs URL here and on the mapped-replacement line above.

pkg/cli/cmd/deploy/deploy.go
kachawla, line 311: This also fires on `rad run`, which embeds `deploycmd.Runner` and calls `r.Runner.Run(ctx)` at `run.go:145`. So "Emitted from `rad deploy` only" in the description isn't accurate.

The behavior is right, since `rad run` is also an authoring path. Update the description and add a `rad run` test case.

lakshmimsft: `rad run ` is to be deprecated and has a warning in there so i have not included it at this time.

pkg/cli/bicep/resources.go
lakshmimsft: Input for updated message from @willtsai / @Reshrahim /  @zachcasper  would be good here since the docs are currently on edge.

kachawla: Feel free to address this in a follow up PR once you have the docs link

kachawla (approved):

PR #12826, opened 2026-08-25
Document Recipe Pack preflight validation

AzureMike:
## Summary

- describe preflight validation for provider-specific Recipe restrictions
- compare Core validation with Recipe wrappers and authoring guidance
- define which changes belong in `radius` and `resource-types-contrib`

## Testing

Not run. Documentation-only change.

### DIFF
--- a/eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
+++ b/eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
@@ -1,0 +1,569 @@
+# Recipe Pack preflight validation
+
+- **Author**: Mike Azure (@AzureMike)
+
+## Problem
+
+Radius resource schemas work across providers. Because of this, a Radius schema can allow a value that a specific provider rejects.
+
+For example, the Azure Recipe Pack passes the PostgreSQL `database` property to an Azure child resource. Radius allows `database: 'postgres'`. Azure rejects it because every PostgreSQL Flexible Server already has a database named `postgres`.
+
+Azure can create the PostgreSQL server before it rejects the database. The Radius resource then fails, but the Azure server still exists and can incur charges.
+
+Radius needs to reject known provider-specific errors before it starts a Recipe.
+
+## Proposal
+
+Add optional validation rules to each Recipe definition. The Recipe Pack defines the rules because it chooses the provider and module.
+
+Radius runs the rules:
+
+1. After it selects the Recipe.
+2. Before it loads Recipe secrets.
+3. Before it starts the Bicep or Terraform driver.
+
+The rules cover fixed restrictions that Radius can check without calling the provider. Examples include string length, format, and reserved names.
+
+## Terms
+
+| Term | Meaning |
+| --- | --- |
+| Portable resource | A Radius resource that uses a Recipe to choose its provider implementation. |
+| Recipe Pack | A set of Recipe definitions, keyed by portable resource type. |
+| Preflight validation | Validation that runs after Recipe selection and before the Recipe driver starts. |
+| Provider-specific rule | A provider or module restriction that does not apply to every implementation of the portable resource. |
+| Recipe driver | The Radius component that runs a Bicep or Terraform Recipe. |
+| Authoring skill | Optional guidance for AI-assisted application authoring. It is not part of the Radius runtime. |
+
+## Issue
+
+[resource-types-contrib issue #299: Azure Recipe Pack accepts values that Azure rejects](https://github.com/radius-project/resource-types-contrib/issues/299)
+
+## Goals
+
+- Reject known provider-specific property errors before a Recipe can create provider resources.
+- Keep portable resource schemas independent of any provider.
+- Store provider rules in the Recipe Pack that owns them.
+- Version the rules with that Recipe Pack.
+- Use the same validation for Bicep and Terraform Recipes.
+- Return a `RecipeValidationFailed` error that names the property but does not include its value.
+- Keep current behavior for Recipe Packs without validation rules.
+
+## Non-goals
+
+- Predict region availability, quota, capacity, changing model availability, or temporary provider policy.
+- Add Azure rules to shared `Radius.Data`, `Radius.Storage`, or `Radius.Messaging` schemas.
+- Run custom validation code or make provider API calls during validation.
+- Validate Recipe parameters set by platform operators. The first version validates only the defaulted developer properties in `ResourceMetadata.Properties`.
+- Change delete behavior.
+- Change simulated Environment behavior.
+- Add validation to the authoring skill. The skill may include the same rules later, but runtime validation must work without it.
+- Change CI configuration.
+
+## User scenarios
+
+### Invalid PostgreSQL database name
+
+An application developer deploys a PostgreSQL resource with `database: 'postgres'` to an Environment that uses the Azure Recipe Pack.
+
+Radius rejects `properties.database` before Azure creates a PostgreSQL server.
+
+### Recipe Pack rules change with the provider
+
+An Azure Recipe Pack maintainer records documented rules for names, lengths, formats, and reserved values with the affected Recipes.
+
+If the pack changes its provider API or module version, the maintainer can update those rules without changing the portable resource type.
+
+## User experience
+
+The application resource stays provider-independent:
+
+```bicep
+resource database 'Radius.Data/postgreSqlDatabases@2025-08-01-preview' = {
+  name: 'orders'
+  properties: {
+    environment: environment
+    database: 'postgres'
+    username: 'appadmin'
+    password: password
+  }
+}
+```
+
+Radius fails the deployment before it starts the Recipe:
+
+```text
+RecipeValidationFailed: resource property "properties.database" is reserved by this Recipe's provider
+```
+
+The Recipe Pack contains the Azure rule:
+
+```bicep
+'Radius.Data/postgreSqlDatabases': {
+  kind: 'bicep'
+  source: 'mcr.microsoft.com/bicep/avm/res/db-for-postgre-sql/flexible-server:<version>'
+  validation: {
+    resourceProperties: {
+      database: {
+        minLength: 1
+        maxLength: 63
+        pattern: '^[A-Za-z_][A-Za-z0-9_-]{0,62}$'
+        disallowedValues: [
+          'postgres'
+          'azure_maintenance'
+          'azure_sys'
+          'template0'
+          'template1'
+        ]
+      }
+    }
+  }
+}
+```
+
+## Design
+
+### Runtime flow
+
+The Core RP checks the validation rules when it stores a Recipe Pack.
+
+For each deployment, the Recipe engine:
+
+1. Loads the selected Recipe.
+2. Checks the defaulted resource properties against the Recipe's rules.
+3. Stops with `RecipeValidationFailed` if a property is invalid.
+4. Loads Recipe secrets if the properties are valid.
+5. Starts the Bicep or Terraform driver.
+
+```mermaid
+flowchart LR
+    A[Defaulted resource properties] --> B[Load selected Recipe]
+    B --> C[Run validation rules]
+    C -->|valid| D[Load Recipe secrets]
+    D --> E[Bicep or Terraform driver]
+    E --> F[Cloud provider]
+    C -->|invalid| G[RecipeValidationFailed]
+```
+
+### Supported rules
+
+The first version supports these string rules:
+
+- Minimum length.
+- Maximum length.
+- A Go/RE2-compatible regular expression.
+- Allowed values.
+- Disallowed values.
+- Case-insensitive comparison for allowed and disallowed values.
+
+This design does not use JSON Schema. It does not allow executable validation code.
+
+## Options considered
+
+### Option 1: Radius checks rules from the Recipe Pack
+
+This option adds a validation mechanism to Radius and adds provider rules to each Recipe Pack.
+
+The Recipe Pack continues to point directly to the provider module:
+
+```text
+Recipe Pack -> Azure Verified Module -> Azure
+```
+
+The Recipe definition gains a `validation` field next to its existing `kind` and `source` fields. The field lists rules for the resource properties used by that Recipe.
+
+For example, the Azure PostgreSQL Recipe defines rules for `database` in `resource-types-contrib`. Those rules can set a minimum length, maximum length, format, and reserved names.
+
+When the Recipe Pack is installed, the Core RP stores the rules with the Recipe definition. When Radius later selects that Recipe for a deployment, the Recipe engine reads the rules and checks the defaulted resource properties.
+
+The deployment flow is:
+
+1. Radius selects the Recipe.
+2. Radius reads the `validation` rules from the Recipe definition.
+3. Radius checks the defaulted resource properties.
+4. If a property is invalid, Radius returns `RecipeValidationFailed` and stops.
+5. If the properties are valid, Radius loads Recipe secrets.
+6. Radius starts the Bicep or Terraform driver.
+7. The driver calls the provider module already named in `source`.
+
+The work is split between the repositories:
+
+- `radius-project/radius` defines the rule format, stores the rules in the Core RP, and runs one evaluator for Bicep and Terraform.
+- `radius-project/resource-types-contrib` defines the Azure-specific rules in the Azure Recipe Pack.
+
+No wrapper module is added. The Recipe Pack keeps its direct reference to the Azure Verified Module or Terraform module. Radius checks the values before either driver starts.
+
+Advantages:
+
+- Stops bad input before Bicep compilation.
+- Stops bad input before Terraform installation or initialization.
+- Stops bad input before Radius creates a provider deployment or cloud resource.
+- Gives Bicep and Terraform the same behavior.
+- Stores the rules with the pack and module version that need them.
+- Keeps direct references to Azure Verified Modules and community Terraform modules.
+- Covers deployments made through the API, CLI, GitOps, SDKs, and other clients.
+
+Disadvantages:
+
+- Requires a Core API change.
+- Requires generated client changes.
+- Requires config-loader and Recipe engine changes.
+- Requires Radius support to ship before Recipe Packs can use the rules.
+- Cannot check provider availability, quota, capacity, or other conditions that require a provider call.
+
+### Option 2: Validate inside Bicep and Terraform wrappers
+
+This option does not change Radius. The Recipe Pack points to a new wrapper module instead of pointing directly to the provider module.
+
+For example, the Azure PostgreSQL Recipe currently points directly to the Azure Verified Module:
+
+```text
+Recipe Pack -> Azure Verified Module -> Azure
+```
+
+With Option 2, it points to a module owned by the Recipe Pack:
+
+```text
+Recipe Pack -> Recipe Pack-owned wrapper -> Azure Verified Module -> Azure
+```
+
+The wrapper accepts the same values that the Recipe currently passes to the provider module. It checks those values first. If they are valid, it passes them to the provider module unchanged. If they are invalid, the wrapper fails the deployment.
+
+The Bicep and Terraform paths need separate wrappers:
+
+- A Bicep wrapper uses Bicep parameter rules, assertions, or ARM validation. It then calls the original Bicep module.
+- A Terraform wrapper uses variable validation or lifecycle preconditions. It then calls the original Terraform module.
+
+Radius treats the wrapper like any other Recipe source. It loads secrets, starts the driver, and gives the inputs to the wrapper. Radius does not know that the wrapper contains validation.
+
+This means validation runs inside the Bicep or Terraform execution:
+
+1. Radius selects the Recipe.
+2. Radius loads Recipe secrets.
+3. Radius starts the Bicep or Terraform driver.
+4. The driver loads and processes the wrapper.
+5. The wrapper checks the input.
+6. The wrapper either fails or calls the original provider module.
+
+The wrapper code and its tests belong with the provider Recipe Pack in `radius-project/resource-types-contrib`. Each wrapper must be published as an artifact, and the Recipe Pack must point to that artifact instead of the original module.
+
+Advantages:
+
+- Requires no Radius API changes.
+- Requires no Recipe engine changes.
+- Keeps provider checks in provider-owned modules.
+- Uses validation features that Bicep and Terraform module authors already know.
+
+Disadvantages:
+
+- Requires a wrapper for every direct Bicep and Terraform Recipe.
+- Adds work to publish, pin, test, and update each wrapper.
+- Implements the same rules twice.
+- Can produce different errors and timing in Bicep and Terraform.
+- Terraform must install and initialize before its validation runs.
+- Terraform may defer some preconditions when values are unknown.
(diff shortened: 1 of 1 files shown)


eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+## Problem
+
+Radius resource schemas work across providers. Because of this, a Radius schema can allow a value that a specific provider rejects.
+
+For example, the Azure Recipe Pack passes the PostgreSQL `database` property to an Azure child resource. Radius allows `database: 'postgres'`. Azure rejects it because every PostgreSQL Flexible Server already has a database named `postgres`.
+
+Azure can create the PostgreSQL server before it rejects the database. The Radius resource then fails, but the Azure server still exists and can incur charges.
+
+Radius needs to reject known provider-specific errors before it starts a Recipe.
```
willdavsmith, line 13: Maybe. it's a difficult problem because these provider-specific errors can change at any time from the provider. I think we should think about it as a best-effort thing, and also have clear ownership (e.g. if a terraform resource half-deploys like this, you can still delete it)

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+2. Before it loads Recipe secrets.
+3. Before it starts the Bicep or Terraform driver.
+
+The rules cover fixed restrictions that Radius can check without calling the provider. Examples include string length, format, and reserved names.
+
+## Terms
+
+| Term | Meaning |
+| --- | --- |
+| Portable resource | A Radius resource that uses a Recipe to choose its provider implementation. |
```
lakshmimsft, line 31: Maybe we call this Extensible/Radius resource. Portable was a term used in the past for differnt set of types, would not like to cause a confusion.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
AzureMike: Yes, it's worth considering if we even want to even do this validation in the first place, or assume that those who are authoring these recipes are aware of these provider restrictions.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+  minLength?: int32;
+  maxLength?: int32;
+  pattern?: string;
+  allowedValues?: string[];
+  disallowedValues?: string[];
+  ignoreCase?: boolean;
+}
+```
+
+Each `resourceProperties` key is the name of a top-level resource property.
```
kachawla: `resourceProperties` keys are top-level property names, so this contract cannot express a constraint on the resource name. Most Azure rejections in this class come from names the Recipe derives from `context.resource.name`: storage account (3-24, lowercase alphanumeric), Key Vault (3-24), Event Hub namespace (6-50). Same failure shape and same partial-create cost as the `postgres` case that motivates the design, and none of it is reachable from here.

Add a reserved key such as `resourceName`, or a sibling block next to `resourceProperties`.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+The first version supports these string rules:
+
+- Minimum length.
+- Maximum length.
+- A Go/RE2-compatible regular expression.
+- Allowed values.
+- Disallowed values.
+- Case-insensitive comparison for allowed and disallowed values.
+
+This design does not use JSON Schema. It does not allow executable validation code.
```
kachawla: This states the choice but not the reasoning, and the codebase already argues the other way. `pkg/schema/validator.go` runs OpenAPI/kin-openapi validation over UDT schemas with `minLength`, `maxLength`, `pattern`, and `enum`, and produces path-qualified errors. `dynamicresource.go:145` calls it on every PUT before the recipe path is selected.

A constrained JSON Schema subset reuses that evaluator and gets nested properties and numeric constraints for free. `disallowedValues` is `not: {enum: [...]}` and `ignoreCase` is `(?i)`, so nothing in the current rule set is lost.

That matters beyond elegance: `capacity`, `throughput`, and `sku` are the next rules anyone will want, and with this vocabulary adding them is a contract change on a shipped preview API. Either adopt the subset or record why the reuse was rejected.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+}
+```
+
+Each `resourceProperties` key is the name of a top-level resource property.
+
+The rules work as follows:
+
+- A property without a rule is allowed.
+- A missing or null optional property is ignored.
+- A property with a rule must be a string when present.
```
kachawla, line 376: what happens when it is not a string? Fail with RecipeValidationFailed, or skip the rule?

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+## Errors
+
+Invalid resource input returns `RecipeValidationFailed` with the current `RecipeSetupError` deployment status.
+
+The error includes a path such as `properties.database` and describes the failed rule. It does not include the submitted value. A future pack may validate a sensitive property, and that value must not appear in resource status, logs, or API errors.
+
+Invalid rule definitions fail the Recipe Pack create or update request with `BadRequest`. The error names the resource type and invalid rule.
+
+The runtime also checks for invalid rules that were stored before the Core RP added storage-time validation.
```
kachawla: This does not say what the runtime does when it finds one. Fail the deployment, or skip the rule and run the Recipe?

It is the difference between a bad pack blocking every deployment in an Environment and a bad pack silently disabling the validation it was installed for. Both are defensible; pick one and say so. Given the goal is preventing provider resource creation, fail closed with an error that names the pack rather than the developer's property seems more consistent, since the developer cannot fix a malformed rule.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+Cosmos DB Mongo rules need either complete published documentation or repeatable provider tests.
+
+Azure AI model, version, SKU, region, capacity, and quota combinations remain provider checks because they can change.
+
+## API design (`radius`)
+
+Add `validation` to `Radius.Core/recipePacks` in the next preview API version approved by maintainers.
+
+A Recipe Pack that uses validation must use the new API version. An older Core RP must reject the request instead of ignoring the new field.
```
kachawla: An already-shipped Core RP cannot be made to reject anything, so this reads as a requirement on the past. The mechanism is the api-version: an older RP does not serve the new preview version and rejects the request on that basis, which gets you the same guarantee.

Reword to describe what the api-version does rather than what old RPs must do. Line 519 has the correct reasoning already.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+| Recipe driver | The Radius component that runs a Bicep or Terraform Recipe. |
+| Authoring skill | Optional guidance for AI-assisted application authoring. It is not part of the Radius runtime. |
+
+## Issue
+
+[resource-types-contrib issue #299: Azure Recipe Pack accepts values that Azure rejects](https://github.com/radius-project/resource-types-contrib/issues/299)
+
+## Goals
+
+- Reject known provider-specific property errors before a Recipe can create provider resources.
```
kachawla: "Preflight" and "before a Recipe can create provider resources" hold for the resource being validated, not for the deployment. This runs in the async backend controller, same as the existing schema validation, so by the time it fires the deployment is underway and other resources in the graph may already be provisioned.

The resource lands in `Failed` with `RecipeSetupError` and the user gets a partial deployment. That is still a large improvement over an orphaned PostgreSQL server, but the goal should say per-resource so nobody reads this as pre-deployment validation.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+```bicep
+'Radius.Data/postgreSqlDatabases': {
+  kind: 'bicep'
+  source: 'mcr.microsoft.com/bicep/avm/res/db-for-postgre-sql/flexible-server:<version>'
+  validation: {
+    resourceProperties: {
+      database: {
+        minLength: 1
+        maxLength: 63
+        pattern: '^[A-Za-z_][A-Za-z0-9_-]{0,62}$'
```
kachawla: The pattern already bounds length at 1-63, so `minLength` and `maxLength` restate it. Two encodings of the same constraint drift the moment Azure changes the limit, and the sample gets copied into every pack.

Use the pattern for shape and the length rules for bounds, not both. Same example repeats at 313-315.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+- Use the same validation for Bicep and Terraform Recipes.
+- Return a `RecipeValidationFailed` error that names the property but does not include its value.
+- Keep current behavior for Recipe Packs without validation rules.
+
+## Non-goals
+
+- Predict region availability, quota, capacity, changing model availability, or temporary provider policy.
+- Add Azure rules to shared `Radius.Data`, `Radius.Storage`, or `Radius.Messaging` schemas.
+- Run custom validation code or make provider API calls during validation.
+- Validate Recipe parameters set by platform operators. The first version validates only the defaulted developer properties in `ResourceMetadata.Properties`.
```
kachawla: Reasonable scope for v1, but pack-set `sku`, `region`, and `capacity` are the other half of provider rejections, and the pack author writing the rules is the same person setting those parameters. Note it as a follow-up rather than leaving it as a flat exclusion.

Cross-property rules are also unexpressible here (Azure requires the admin username to differ from the database name, for one). Worth adding to this list explicitly so the omission is a decision rather than a gap.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+## Security
+
+The rules are data. They cannot run code, load remote content, or call a provider.
+
+Go regular expressions use linear-time matching. They do not allow denial of service through regular-expression backtracking.
+
+Existing Recipe Pack write permissions control who can set the rules for an Environment.
+
+Errors include the property path and failed rule. They do not include the property value.
```
kachawla, line 509: I'd add the logging constraint as well here, not just the error constraint

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+## CLI design
+
+No CLI changes.
+
+## Implementation by repository
+
+### `radius-project/radius`
+
+Generated Radius Bicep types include the validation fields.
```
kachawla: The `Radius.Core/recipePacks` doc comment in `typespec/Radius.Core/recipePacks.tsp` is user-facing reference material, with worked Bicep examples for `kind`, `source`, and `parameters`. Adding `validation` to the model without a matching section leaves the published reference incomplete.

Call it out in the radius-repo work item at line 335 so it does not get treated as generated-code churn.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+Use the next preview version approved by maintainers. Do not assign a date until maintainers choose the version.
+
+### Should `postgres` use Azure's existing database instead of failing?
+
+No.
+
+`Radius.Data/postgreSqlDatabases` creates and owns a database. Using an existing database would change ownership, deletion, security, and outputs. That behavior needs a separate design.
+
+### How should the pack handle changing Azure AI restrictions?
```
willdavsmith, line 553: This is a good case study for changing requirements. Azure AI models deprecate models pretty often. We should assume that the validation rules that we set will not be set in stone, and versions of the validation rules that we release are only valid for some amount of time (in the case of AI models, a short amount of time).

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+}
+```
+
+Radius fails the deployment before it starts the Recipe:
+
+```text
+RecipeValidationFailed: resource property "properties.database" is reserved by this Recipe's provider
+```
+
+The Recipe Pack contains the Azure rule:
```
willtsai, line 99: can you show what the full recipe pack would look like, and perhaps include an additional recipe? I'm wondering if it'll end up looking clunky or if many of the validation rules will be duplicated accross resource types.

AzureMike: It could look like the version below, where we could store the validation stuff in a json file. (see the validation line, 3rd row)

```
'Radius.Data/postgreSqlDatabases': {
        kind: 'bicep'
        source: 'mcr.microsoft.com/bicep/avm/res/db-for-postgre-sql/flexible-server:0.15.2'
        validation: recipeValidations['Radius.Data/postgreSqlDatabases']
        parameters: {
          name: 'pgsql-{{context.azure.resourceNameHash}}'
          administratorLogin: '{{context.resource.properties.username}}'
          administratorLoginPassword: '{{context.resource.properties.password}}'
          authConfig: {
            activeDirectoryAuth: 'Enabled'
            passwordAuth: 'Enabled'
          }
```

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+          'azure_maintenance'
+          'azure_sys'
+          'template0'
+          'template1'
+        ]
+      }
+    }
+  }
+}
+```
```
nithyatsu, line 325: recipe pack and recipe are both operator provided. I am not sure I understand why instead of authoring the recipe with these validations, we would want to separate validation into recipe-packs and then author recipes which might be invalidated by radius. I see the point about radius not having to deploy and then fail though.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+- **Author**: Mike Azure (@AzureMike)
+
+## Problem
+
+Radius resource schemas work across providers. Because of this, a Radius schema can allow a value that a specific provider rejects.
+
+For example, the Azure Recipe Pack passes the PostgreSQL `database` property to an Azure child resource. Radius allows `database: 'postgres'`. Azure rejects it because every PostgreSQL Flexible Server already has a database named `postgres`.
+
+Azure can create the PostgreSQL server before it rejects the database. The Radius resource then fails, but the Azure server still exists and can incur charges.
```
willdavsmith, line 11: there's 2 problems here:
1. we didn't do validation up-front (which is what this doc proposes)
2. Radius doesn't provide a good story for resources in a half-deployed state

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+- Does not cover hand-written Bicep.
+- Does not cover existing applications.
+- Does not cover API clients, SDKs, GitOps, or other tools that do not use the skill.
+- Can become outdated when provider rules or module versions change.
+- Depends on prompt context, token limits, model behavior, and the author following the advice.
+- Cannot guarantee that Radius stops bad input before it creates provider resources.
+
+## Decision
+
+Use Option 1.
```
willtsai, line 289: I wonder if we can do a combo of Option 1 and Option 3. That way we can ensure the agents model things correctly from the beginning (hence avoiding the problem at the source) while also keeping a safety net of validation failures. This also better models how things would be done in absence of an Agent (i.e. the rules put into the skills live inside the head of the platform engineer 😄)

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
kachawla, line 434: We can update it in place for this one as we haven't guaranteed backwards compatibility yet.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
AzureMike: Agreed

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+## Goals
+
+- Reject known provider-specific property errors before a Recipe can create provider resources.
+- Keep portable resource schemas independent of any provider.
+- Store provider rules in the Recipe Pack that owns them.
+- Version the rules with that Recipe Pack.
+- Use the same validation for Bicep and Terraform Recipes.
+- Return a `RecipeValidationFailed` error that names the property but does not include its value.
+- Keep current behavior for Recipe Packs without validation rules.
```
sk593, line 50: How are we keeping properties up to date with changes in the resource provider?

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+### Invalid PostgreSQL database name
+
+An application developer deploys a PostgreSQL resource with `database: 'postgres'` to an Environment that uses the Azure Recipe Pack.
+
+Radius rejects `properties.database` before Azure creates a PostgreSQL server.
+
+### Recipe Pack rules change with the provider
+
+An Azure Recipe Pack maintainer records documented rules for names, lengths, formats, and reserved values with the affected Recipes.
```
sk593, line 73: Have we thought about automatic detection for when these values change? It seems like we could easily get into a scenario where we restrict based on outdated rules

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+Store the Azure rules in a checked-in JSON file.
+
+Load that file from the Azure Recipe Pack's Bicep definition and assign the needed rules to each Recipe.
+
+Add tests for the Azure limits and reserved values. Document the source of each rule and the minimum Radius version required by the pack.
+
+## Errors
+
+Invalid resource input returns `RecipeValidationFailed` with the current `RecipeSetupError` deployment status.
```
DariuszPorowski, line 470: issue(operations,blocking): Preserve cleanup eligibility after a rejected update

An invalid update returns `RecipeSetupError`. The controller persists that status (`pkg/portableresources/backend/controller/createorupdateresource.go:229-250`), while deletion skips the Recipe driver whenever the stored status is `setupError` (`pkg/portableresources/backend/controller/deleteresource.go:78-81`). If the resource had previously deployed successfully, deletion will leave its prior output resources behind. Base cleanup on persisted output resources or retain separate state for the last successful deployment, then add an update-failure-delete regression test.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+A Recipe Pack that uses validation must use the new API version. An older Core RP must reject the request instead of ignoring the new field.
+
+The current preview API remains available for older packs without validation.
+
+This change adds no CLI commands or arguments. Existing deployment commands report `RecipeValidationFailed` through the current resource status and error paths.
+
+## CLI design
+
+No CLI changes.
```
DariuszPorowski, line 442: issue(compatibility,blocking): Prevent old API writes from erasing validation rules

`rad env create` and `rad env update` fetch and replace Recipe Packs through `2025-08-01-preview` to synchronize `referencedBy` (`pkg/cli/cmd/env/create/preview/create.go:453-477`, `pkg/cli/cmd/env/update/preview/update.go:557-593`). That model cannot represent `validation`; its converter drops the field, and `PrepareResource` does not merge old properties before replacement (`pkg/corerp/api/v20250801preview/recipepack_conversion.go:89-132`, `pkg/armrpc/frontend/controller/operation.go:126-157`). Move reference maintenance server-side, preserve unknown newer fields, or reject down-level PUT/PATCH operations on validation-bearing packs.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+  pattern?: string;
+  allowedValues?: string[];
+  disallowedValues?: string[];
+  ignoreCase?: boolean;
+}
+```
+
+Each `resourceProperties` key is the name of a top-level resource property.
+
+The rules work as follows:
```
DariuszPorowski, line 372: suggestion(compatibility,non-blocking): Define how multiple rules compose

State whether all configured rules must pass, whether empty allowed or disallowed lists are no-ops, and which list wins when a value appears in both after case folding. These semantics become part of the public API and need deterministic tests.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+- Test valid values.
+- Test missing optional properties.
+- Test Unicode length.
+- Test patterns.
+- Test allowed values.
+- Test disallowed values.
+- Test case-insensitive comparisons.
+- Test properties with the wrong type.
+- Test invalid rule definitions.
+- Check that invalid input returns `RecipeValidationFailed` and `RecipeSetupError` before Radius calls the secrets loader or either mock driver.
```
DariuszPorowski, line 490: suggestion(test,non-blocking): Prove sensitive values stay out of every failure surface

Add a sentinel sensitive property and assert that its value is absent from returned `ErrorDetails`, persisted status, and captured logs. This verifies the no-value security invariant through the actual error paths.

DariuszPorowski (commented):
The validation boundary is sound, but two lifecycle and compatibility gaps need resolution. Details are inline.

eng/design-notes/recipes/2026-08-recipe-pack-preflight-validation.md
```diff
@@ -0,0 +1,569 @@
...
+
+1. Radius selects the Recipe.
+2. Radius reads the `validation` rules from the Recipe definition.
+3. Radius checks the defaulted resource properties.
+4. If a property is invalid, Radius returns `RecipeValidationFailed` and stops.
+5. If the properties are valid, Radius loads Recipe secrets.
+6. Radius starts the Bicep or Terraform driver.
+7. The driver calls the provider module already named in `source`.
+
+The work is split between the repositories:
```
DariuszPorowski, line 189: praise: Keep provider rules beside the Recipe that selects the provider

This ownership avoids contaminating shared resource schemas and gives Bicep and Terraform one enforcement path.

PR #12836, opened 2026-08-26
Fix install.sh escalating to sudo for user-writable $HOME/.local/bin

### LINKED ISSUE #12835: install.sh escalates to sudo for $HOME/.local/bin when .local doesn't exist, leaving it root-owned and breaking later user-owned tool installs (yq)
## Summary

`deploy/install.sh` can install the `rad` CLI into a **root-owned** `$HOME/.local/bin`. Any later step that installs a user-owned tool into the same directory (e.g. `make install-yq` via `build/scripts/install-yq.sh`) then fails with `Permission denied`.

This surfaces in the extension deploy/delete workflows at the **"Load Radius contrib catalog"** step:

```
mv: cannot move '/tmp/tmp.XXXX/yq' to '/home/runner/.local/bin/yq': Permission denied
make: *** [build/tools.mk:47: install-yq] Error 1
##[error]Process completed with exit code 2.
```

Example failing run: https://github.com/sk593/aks-store-demo/actions/runs/33014623444

## Root cause

In `deploy/install.sh`, `needsSudo()` only checks the install dir and its **immediate** parent:

```bash
needsSudo() {
    local install_dir="$1"
    # Root never needs sudo
    if [[ ${EUID:-$(id -u)} -eq 0 ]]; then
        return 1
    fi
    # If the directory exists and is writable, no sudo needed
    if [[ -d "${install_dir}" && -w "${install_dir}" ]]; then
        return 1
    fi
    # If the parent directory is writable (dir doesn't exist yet), no sudo
    local parent_dir
    parent_dir=$(dirname "${install_dir}")
    if [[ -d "${parent_dir}" && -w "${parent_dir}" ]]; then
        return 1
    fi
    # Need elevated privileges
    return 0
}
```

When the default install dir is `$HOME/.local/bin` and **neither `.local/bin` nor `.local` exists yet** (common on a fresh GitHub-hosted runner), both the dir check and the immediate-parent check fail, so `needsSudo` returns `0`. `runAsRoot mkdir -p "$HOME/.local/bin"` then runs `sudo mkdir -p`, creating the whole `~/.local` tree **owned by root** — even though `$HOME` itself is writable and no elevation was actually required.

Subsequently `build/scripts/install-yq.sh` targets the same `$HOME/.local/bin`, so its `mkdir -p` silently succeeds (dir already exists) and the `mv` into the root-owned dir fails with `Permission denied`.

### Why it's intermittent

`install-yq.sh` early-returns if a matching `yq` is already on `PATH`. Runners with a pre-baked `yq` skip the download/`mv` entirely and never hit the failure; runners without it fall into the failing path. So the same workflow passes or fails depending on runner image contents.

## Proposed fix

Make `needsSudo()` walk up to the nearest **existing** ancestor and only escalate to sudo when that ancestor is not writable. For `$HOME/.local/bin` under a writable `$HOME`, this creates the tree as the current user (no sudo), so it stays user-owned and later installs into it succeed.

## Impact

Any consumer whose runner lacks a pre-created `~/.local` and installs a second user-owned tool into `~/.local/bin` after `rad` — including the extension's deploy and delete workflows (`setup-control-plane` + `load-contrib-catalog`).

sk593:
## Description

`deploy/install.sh` could install the `rad` CLI into a **root-owned** `$HOME/.local/bin`, which then broke any later user-owned tool install into the same directory (e.g. `make install-yq`) with `Permission denied`.

`needsSudo()` only checked the install dir and its **immediate** parent. When the default install dir is `$HOME/.local/bin` and neither `.local/bin` nor `.local` exists yet (common on a fresh GitHub-hosted runner), both checks fail, so the installer escalates to `sudo` and `sudo mkdir -p` creates the whole `~/.local` tree **root-owned** — even though `$HOME` is writable and no elevation was actually needed.

The next user-owned install into that dir (`build/scripts/install-yq.sh`) then does `mkdir -p` (silently succeeds, dir exists) and `mv` into the root-owned dir, failing with `Permission denied` at the **Load Radius contrib catalog** step of the extension deploy/delete workflows.

### Fix

Make `needsSudo()` walk up to the nearest **existing** ancestor and only escalate to `sudo` when that ancestor is not writable. For `$HOME/.local/bin` under a writable `$HOME`, the tree is created as the current user (no sudo) and stays user-owned, so later installs into it succeed.

## Type of change

- Bugfix

## Fixes

Fixes #12835

## Testing

Extracted the patched `needsSudo()` and exercised it against a temp filesystem:

| Case | install dir | before | after |
|------|-------------|--------|-------|
| Fresh writable `$HOME`, nested `.local/bin` missing | `$HOME/.local/bin` | needs sudo ❌ | **no sudo** ✅ |
| Existing writable install dir | `$HOME/.local/bin` | no sudo | no sudo ✅ |
| `/usr/local/bin` (not writable) | `/usr/local/bin` | sudo | sudo ✅ |
| Nearest existing ancestor not writable | `<ro>/sub/bin` | sudo | sudo ✅ |

`bash -n` passes; behavior is unchanged for the root, already-writable, and genuinely-needs-root cases.

## Why it was intermittent

`install-yq.sh` early-returns when a matching `yq` is already on `PATH`. Runners with a pre-baked `yq` skipped the failing `mv` entirely; runners without it hit it. Same workflow, different runner image contents.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -141,7 +141,8 @@ needsSudo() {
     # sudo is needed. This matters for $HOME/.local/bin when neither .local/bin
     # nor .local exists yet: the whole tree is under a writable $HOME, so
     # escalating to sudo here would create it root-owned and break later
-    # user-owned installs into the same dir (e.g. `make install-yq`).
+    # user-owned installs into the same dir (the build/scripts/install-*.sh
+    # tools all default to $HOME/.local/bin and install without sudo).
     local ancestor
     ancestor="${install_dir}"
     while [[ ! -d "${ancestor}" ]]; do
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -270,22 +270,38 @@ warnExistingRadiusElsewhere() {
     echo "============================================================================"
 }
 
+# Fetch a GitHub API URL. When GITHUB_TOKEN is set the request is authenticated
+# to avoid the unauthenticated rate limit (60 requests/hour per IP), which is
+# easily exhausted on shared CI runners and behind corporate NAT.
+githubApiGet() {
+    local url="$1"
+
+    if [[ "${RADIUS_HTTP_REQUEST_CLI}" == "curl" ]]; then
+        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+            curl -s -H "Authorization: Bearer ${GITHUB_TOKEN}" "${url}"
+        else
+            curl -s "${url}"
+        fi
+    else
+        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
+            wget -q --header="Authorization: Bearer ${GITHUB_TOKEN}" --header="Accept: application/json" -O - "${url}"
+        else
+            wget -q --header="Accept: application/json" -O - "${url}"
+        fi
+    fi
+}
+
 getLatestRelease() {
     local radReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
     local latest_release=""
+    local releases_json=""
+
+    releases_json=$(githubApiGet "${radReleaseUrl}")
 
     if [[ "${INCLUDE_RC}" == "true" ]]; then
-        if [[ "${RADIUS_HTTP_REQUEST_CLI}" == "curl" ]]; then
-            latest_release=$(curl -s "${radReleaseUrl}" | grep \"tag_name\" | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p')
-        else
-            latest_release=$(wget -q --header="Accept: application/json" -O - "${radReleaseUrl}" | grep \"tag_name\" | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p')
-        fi
+        latest_release=$(echo "${releases_json}" | grep \"tag_name\" | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p') || true
     else
-        if [[ "${RADIUS_HTTP_REQUEST_CLI}" == "curl" ]]; then
-            latest_release=$(curl -s "${radReleaseUrl}" | grep \"tag_name\" | grep -v rc | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p')
-        else
-            latest_release=$(wget -q --header="Accept: application/json" -O - "${radReleaseUrl}" | grep \"tag_name\" | grep -v rc | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p')
-        fi
+        latest_release=$(echo "${releases_json}" | grep \"tag_name\" | grep -v rc | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p') || true
     fi
 
     if [[ -z "${latest_release}" ]]; then
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -270,38 +270,22 @@ warnExistingRadiusElsewhere() {
     echo "============================================================================"
 }
 
-# Fetch a GitHub API URL. When GITHUB_TOKEN is set the request is authenticated
-# to avoid the unauthenticated rate limit (60 requests/hour per IP), which is
-# easily exhausted on shared CI runners and behind corporate NAT.
-githubApiGet() {
-    local url="$1"
-
-    if [[ "${RADIUS_HTTP_REQUEST_CLI}" == "curl" ]]; then
-        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
-            curl -s -H "Authorization: Bearer ${GITHUB_TOKEN}" "${url}"
-        else
-            curl -s "${url}"
-        fi
-    else
-        if [[ -n "${GITHUB_TOKEN:-}" ]]; then
-            wget -q --header="Authorization: Bearer ${GITHUB_TOKEN}" --header="Accept: application/json" -O - "${url}"
-        else
-            wget -q --header="Accept: application/json" -O - "${url}"
-        fi
-    fi
-}
-
 getLatestRelease() {
     local radReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
     local latest_release=""
-    local releases_json=""
-
-    releases_json=$(githubApiGet "${radReleaseUrl}")
 
     if [[ "${INCLUDE_RC}" == "true" ]]; then
-        latest_release=$(echo "${releases_json}" | grep \"tag_name\" | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p') || true
+        if [[ "${RADIUS_HTTP_REQUEST_CLI}" == "curl" ]]; then
+            latest_release=$(curl -s "${radReleaseUrl}" | grep \"tag_name\" | awk 'NR==1{print $2}' | sed -n 's/\"\(.*\)\",/\1/p')
+        else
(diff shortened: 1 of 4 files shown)


DariuszPorowski (dismissed):

deploy/install.sh
```diff
@@ -136,10 +136,24 @@ needsSudo() {
...
+    while [[ ! -d "${ancestor}" ]]; do
+        local parent
+        parent=$(dirname "${ancestor}")
+        # Stop at the filesystem root to avoid an infinite loop.
+        if [[ "${parent}" == "${ancestor}" ]]; then
+            break
+        fi
+        ancestor="${parent}"
+    done
+    if [[ -d "${ancestor}" && -w "${ancestor}" ]]; then
```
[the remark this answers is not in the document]
sk593: Fixed in `3fd9e2a`.

deploy/install.sh
```diff
@@ -136,10 +136,24 @@ needsSudo() {
...
-    if [[ -d "${parent_dir}" && -w "${parent_dir}" ]]; then
+    # Walk up to the nearest existing ancestor. If we can write to it, we can
+    # create the (possibly nested) install dir ourselves with `mkdir -p`, so no
+    # sudo is needed. This matters for $HOME/.local/bin when neither .local/bin
+    # nor .local exists yet: the whole tree is under a writable $HOME, so
+    # escalating to sudo here would create it root-owned and break later
+    # user-owned installs into the same dir (e.g. `make install-yq`).
+    local ancestor
+    ancestor="${install_dir}"
+    while [[ ! -d "${ancestor}" ]]; do
```
kachawla, line 148: No regression test. `deploy/test-install.sh:419-421` pre-creates `$HOME/.local/bin` specifically to dodge this bug - drop that `mkdir -p` and `test_default_dir_nonroot` reproduces #12835 and fails on `main`. `validate-installers.yaml` already gates on this file, so it runs here and on every future edit to `needsSudo`.

The `3fd9e2a` case is also uncovered: `test_nonwritable_dir_no_sudo_fails` chmods the dir *and* its parent, so it passes either way. Chmod only the install dir to pin it. Its comment at :436-437 ("it checks both") is now stale.

deploy/install.sh
```diff
@@ -136,10 +136,24 @@ needsSudo() {
...
-    # If the parent directory is writable (dir doesn't exist yet), no sudo
-    local parent_dir
-    parent_dir=$(dirname "${install_dir}")
-    if [[ -d "${parent_dir}" && -w "${parent_dir}" ]]; then
+    # Walk up to the nearest existing ancestor. If we can write to it, we can
+    # create the (possibly nested) install dir ourselves with `mkdir -p`, so no
+    # sudo is needed. This matters for $HOME/.local/bin when neither .local/bin
+    # nor .local exists yet: the whole tree is under a writable $HOME, so
+    # escalating to sudo here would create it root-owned and break later
+    # user-owned installs into the same dir (e.g. `make install-yq`).
```
kachawla: All thirteen `build/scripts/install-*.sh` default to `$HOME/.local/bin` and do `mkdir -p` + `mv` without sudo. yq was just first. Worth generalizing this comment and the description.

This also only prevents new breakage. Where `~/.local` is already root-owned (cached runner images, dev machines), the next `install-*.sh` fails the same way. Consider warning when `needsSudo` is true under `$HOME`, with `sudo chown -R "$(id -u):$(id -g)" ~/.local`. Non-blocking.

deploy/install.sh
```diff
@@ -136,10 +136,24 @@ needsSudo() {
         return 1
     fi
 
-    # If the parent directory is writable (dir doesn't exist yet), no sudo
-    local parent_dir
-    parent_dir=$(dirname "${install_dir}")
-    if [[ -d "${parent_dir}" && -w "${parent_dir}" ]]; then
+    # Walk up to the nearest existing ancestor. If we can write to it, we can
```
kachawla, line 139: This hit users too, not just CI: the documented `curl ... | bash` on a machine without `~/.local` prompted for sudo and left the tree root-owned. Same class as the installer notes in `v0.56.0.md` and `v0.56.1.md`. Nothing to change here - flagging it for the release-prep PR.

deploy/install.sh
sk593: Fixed in `a96aab41f`. Dropped the `mkdir -p` in `test_default_dir_nonroot` so it reproduces #12835, and `test_nonwritable_dir_no_sudo_fails` now chmods only the install dir (parent stays writable) and the stale "it checks both" comment is gone. Confirmed the non-writable test fails when the walk starts at `dirname(install_dir)`.

deploy/install.sh
sk593: Acknowledged — nothing to change here; carrying this over to the release-prep PR.

deploy/install.sh
sk593: Generalized the comment in `a96aab41f` to reference all `build/scripts/install-*.sh` tools rather than yq. Leaving the `sudo chown` warning suggestion out for now since you flagged it non-blocking.

deploy/test-install.sh
```diff
@@ -416,9 +416,10 @@ test_default_dir_nonroot() {
     local fake_home
     fake_home=$(make_test_dir "fakehome")
 
-    # Pre-create the expected default directory so needsSudo() sees a
-    # writable dir and doesn't try to escalate.
-    mkdir -p "${fake_home}/.local/bin"
+    # Intentionally do NOT pre-create $HOME/.local/bin. The installer must
+    # detect that the whole tree is under a writable $HOME and create it
+    # without sudo. Pre-creating it would mask a regression in needsSudo()
+    # (this is the #12835 scenario).
```
kachawla, line 422: This still passes on `main`. The bug is about *ownership*, and nothing here asserts ownership.

`assert_rad_installed` (:107-127) checks three things: `-f`, `-x`, and that `rad version --cli` runs. On `ubuntu-24.04` the `runner` user has passwordless sudo, so pre-fix: `needsSudo` returns true, `sudo mkdir -p` + `sudo cp` create the tree root-owned, and the binary lands `0755 root:root` - all three checks pass. Quick proof with any root-owned system binary:

```
$ ls -l /usr/bin/id
-rwxr-xr-x  2 root  wheel  /usr/bin/id
$ [[ -f /usr/bin/id && -x /usr/bin/id ]] && /usr/bin/id -u   # all pass
$ [[ -O /usr/bin/id ]]                                        # fails - the missing check
```

Bash's `-O` (owned by effective UID) is the fix, and it avoids the GNU/BSD `stat` split, which matters since CI is Linux but contributors run this on macOS:

```bash
    assert_rad_installed "${fake_home}/.local/bin"

    # The tree must be created as the current user. Without this the test
    # also passes when the installer sudo-creates it root-owned (#12835).
    local d
    for d in "${fake_home}/.local" "${fake_home}/.local/bin"; do
        if [[ ! -O "${d}" ]]; then
            echo "  ASSERT FAILED: ${d} is not owned by the current user"
            return 1
        fi
    done
```

Check `.local` too, not just `.local/bin` - `sudo mkdir -p` roots the whole tree, and `.local` is the one the other installers trip over.

Knock-on: pre-fix this also leaves root-owned dirs in `TEST_ROOT`. `cleanup` does `chmod u+rwx ... || true` (fails as non-root) then an unguarded `rm -rf`, under `set -euo pipefail` in an EXIT trap - so the run gets a confusing exit status on top. Goes away once the assertion is added.

sk593: Fixed in `c207b0376`. Added a `-O` (owned by effective UID) assertion on both `.local` and `.local/bin`, so the test now fails when the installer sudo-creates the tree root-owned. Used `-O` rather than `stat` to avoid the GNU/BSD split for macOS contributors.

kachawla (approved):

Issue #12837, opened 2026-08-26
Paginated `nextLink` leaks the RP's cluster-internal address, breaking any out-of-cluster client


nithyatsu:
## Bug

`nextLink` on a paginated LIST response leaks the serving RP's cluster-internal address, so any client that is not inside the cluster cannot follow it.

Observed from a GitHub Actions runner talking to a k3d control plane through the Kubernetes API-server proxy:

```text
Error: Get "http://dynamic-rp.radius-system:8082/planes/radius/local/resourcegroups/default/providers/Radius.Compute/containers?api-version=2025-08-01-preview&skipToken=MjAyNi0wOC0xOVQyMjo1MjoyNi4zNjc5MDha&top=10": dial tcp: lookup dynamic-rp.radius-system on 127.0.0.53:53: no such host
```

The first page succeeds over the proxied connection. The second page fails, because `rad` follows the returned absolute URL verbatim and `dynamic-rp.radius-system` is not resolvable outside the cluster.

Practical impact: `rad app delete` fails during **resource enumeration**, before anything is deleted, for any application with more than one page of resources of a single type (`top=10` by default). The application is left undeletable. This is not delete-specific though — it affects any paginated LIST for any non-in-cluster caller.

## Root cause

`GetNextLinkURL` builds the link from the inbound request, and `GetURLFromReqWithQueryParameters` takes the host straight from `req.Host`:

https://github.com/radius-project/radius/blob/main/pkg/armrpc/frontend/controller/util.go

```go
func GetURLFromReqWithQueryParameters(req *http.Request, qps url.Values) *url.URL {
	url := url.URL{
		Host:     req.Host,
		Scheme:   req.URL.Scheme,
		Path:     req.URL.Path,
		RawQuery: qps.Encode(),
	}
	...
}
```

When UCP proxies a LIST downstream to an RP, `req.Host` as seen by that RP is the internal service address (`dynamic-rp.radius-system:8082`), not the address the caller used. So the emitted `nextLink` is only valid from inside the cluster.

## The caller-visible host is already available

UCP's proxy forwards the original host and scheme downstream — see `pkg/ucp/proxy/testdata/arm/basic-roundtrip/downstream-request.json`:

```json
{
  "url": "http://example.com/some-path?query=yes",
  "method": "GET",
  "headers": {
    "X-Forwarded-Host": ["ucp.example.com"],
    "X-Forwarded-Proto": ["http"],
    "traceparent": ["traceid"]
  },
  "body": ""
[...]

PR #12839, opened 2026-08-26
Make control-plane services reachable for paginated deletes

nithyatsu:
## Summary

Adds a step to the shared `delete-resource` composite action that makes the Radius control-plane services reachable from the runner under their in-cluster DNS names, so `rad app delete` can follow a paginated resource listing.

## Reason for change

`rad` reaches the ephemeral control plane through the Kubernetes API-server proxy, but a paginated LIST returns a `nextLink` built from the serving RP's own address:

```text
Error: Get "http://dynamic-rp.radius-system:8082/planes/radius/local/resourcegroups/default/providers/Radius.Compute/containers?api-version=2025-08-01-preview&skipToken=...&top=10": dial tcp: lookup dynamic-rp.radius-system on 127.0.0.53:53: no such host
```

The CLI follows that URL verbatim, bypassing the proxy, and the in-cluster name does not resolve on a GitHub runner. The first page succeeds; the second fails. So deleting an application with more than one page of resources of a single type (`top=10`) fails while **enumerating**, before anything is deleted, leaving the application undeletable from the workflow.

This is a workflow-level mitigation so the delete workflows work today. The underlying leak — `GetURLFromReqWithQueryParameters` using `req.Host` and ignoring the `X-Forwarded-Host` UCP already sends downstream — is tracked in #12837 and fixed separately. Once that lands this step becomes redundant and can be removed.

Reported downstream as radius-project/ai-extensions#441.

## Implementation notes

- Placed in `delete-resource` rather than `setup-control-plane`: the earlier "Refresh external deployment target credentials" step restarts `applications-rp`, `dynamic-rp` and `bicep-de`, which would drop any forwards established earlier.
- Placed in the shared composite action rather than in `delete-azure.yml` / `delete-aws.yml`, so both providers and both the application and environment delete paths get it without duplication.
- Each service is forwarded onto the *same* port it uses in-cluster, so the aliased host:port from the `nextLink` matches.
- A port that is already in use is reported and skipped rather than aliased to whatever else is listening.
- A service with no resolvable port, or a forward that never becomes ready, warns and continues instead of failing the delete.
- The forwards are deliberately left running for the delete step; the runner reclaims them at job end.

## How to test

2. Deploy it to an AKS-backed environment via the extension deploy workflow.
3. Run the `Delete with Radius` workflow for that application.
4. The new step logs `✅ dynamic-rp reachable on 127.0.0.1:8082` (and similarly for `applications-rp` / `ucp`), and `rad app delete` pages through the full listing and completes instead of failing DNS resolution.

Validated locally:

- `.github/extension/actions/action-shell-syntax_test.sh` passes (19 run blocks across 10 action.yml files).
- `shellcheck -s bash` is clean on the extracted run block.
- The run block was executed against a stubbed `kubectl`/`sudo`/`tee` harness covering all three branches: a healthy service (forwarded and aliased), a service with no port (skipped), and a port already in use (warned, not aliased). Step exits 0 in every case and writes exactly the expected `/etc/hosts` lines.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/actions/delete-resource/action.yml` | Add a "Make control-plane services reachable for paginated listings" step before the delete that port-forwards `dynamic-rp`, `applications-rp` and `ucp` onto their in-cluster ports on loopback and aliases their in-cluster DNS names in `/etc/hosts`. |

### DIFF
--- a/.github/extension/actions/delete-resource/action.yml
+++ b/.github/extension/actions/delete-resource/action.yml
@@ -25,6 +25,64 @@ inputs:
 runs:
   using: composite
   steps:
+    # `rad` reaches the control plane through the Kubernetes API-server proxy, but a
+    # paginated LIST returns a nextLink built from the serving RP's own address (for
+    # example http://dynamic-rp.radius-system:8082/...). The CLI follows that URL
+    # verbatim, bypassing the proxy, and the in-cluster name does not resolve on the
+    # runner -- so deleting an application with more than one page of resources of a
+    # single type fails while enumerating, before anything is deleted. Alias the
+    # control-plane services onto loopback so the directly dialled nextLink still
+    # reaches them. Runs here rather than in setup-control-plane because the earlier
+    # target-credential refresh restarts these deployments, which would drop the
+    # forwards. The forwards are left running for the delete below; the runner
+    # reclaims them when the job ends.
+    - name: Make control-plane services reachable for paginated listings
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        port_open() {
+          # The subshell closes the descriptor when it exits, so this only probes.
+          (exec 3<>"/dev/tcp/127.0.0.1/$1") 2>/dev/null
+        }
+
+        for service in dynamic-rp applications-rp ucp; do
+          port=$(kubectl get service "${service}" -n radius-system \
+            -o jsonpath='{.spec.ports[0].port}' 2>/dev/null || echo "")
+          if [[ -z "${port}" ]]; then
+            echo "Skipping ${service}: no service port found."
+            continue
+          fi
+
+          # A port already in use would satisfy the readiness probe below and alias
+          # the service to whatever else is listening.
+          if port_open "${port}"; then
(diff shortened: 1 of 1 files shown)


nithyatsu:
Closing this — the approach is architecturally wrong, not just rough around the edges.

The step aliases `dynamic-rp.radius-system` (and friends) to loopback in `/etc/hosts` so the CLI can dial the leaked `nextLink`. But that makes the client talk to the resource provider **directly, bypassing UCP**. Resource providers are not a client-facing surface — UCP is the only component exposed to a CLI, and it owns plane resolution and response URL fix-up. Routing around it to make a listing paginate is not something we should ship in a shared composite action, even for CI only.

It also had concrete defects: no cleanup of the `/etc/hosts` edits or the background `kubectl port-forward` processes (harmless on an ephemeral runner, persistent damage on a self-hosted one), warn-and-continue on both failure paths, and it would be blocked on clusters with `networkPolicies.enabled=true`, where ingress from the node is only permitted on port 9443.

The real fix belongs in UCP and is tracked in #12837.

PR #12829, opened 2026-08-26
Revert default resource group and environment creation from rad install kubernetes

lakshmimsft:
## Summary
Reverts the default resource group and environment creation that `rad install kubernetes` performed,
returning it to a barebones control-plane install. Also removes the `--preview` flag from this command,
since its only effect was selecting the resource type (and default recipe pack) for the default
environment that is no longer created.

After this change:

- `rad install kubernetes` installs the Radius control plane .
- No `default` resource group, no `default` environment, no default recipe pack, and no environment
  targeting the `default` Kubernetes namespace.

This reverts the behavior introduced by #11870 (shipped in v0.59.0) and the `--preview` flag added on
top of it by #12504 (shipped in v0.60.0). Unrelated changes that landed in the same files since then
are preserved: the `t.Context()` test modernization from #12523 and the current `docs.radapp.io`
architecture URL.

## Reason for change

`rad install kubernetes` silently provisioning a default environment and recipe packs is undesirable
for anyone who intends to customize environments, namespaces, or recipe packs — notably the GitHub
Copilot app integration, where a default environment targeting the `default` Kubernetes namespace
conflicts with the intended setup.

Opinionated initialization belongs in `rad init`, where users expect an onboarding flow with defaults
preset. This restores a clean separation of responsibilities:

- `rad install kubernetes` — install the control plane only.
- `rad init` — install and/or initialize Radius with default resources and recipe packs.

Fixes #12827

## How to test
Manual, against a clean cluster:

rad install kubernetes
rad group show default   # expect: not found
rad env show default     # expect: not found

rad init                 # still creates the default group, environment, and recipe pack
rad group show default   # expect: exists
rad env show default     # expect: exists

Also confirm  rad install kubernetes --preview  now reports  unknown flag: --preview , and that the
existing install flags ( --reinstall ,  --chart ,  --set ,  --set-file ,  --kubecontext ,
 --skip-contour-install , and the  --contour-*  flags) are unchanged.

| File | Summary of change |
| ---- | ----------------- |
[...]

PR #12840, opened 2026-08-26
Fix corrupt state archive: self-heal the default group and guard rad shutdown

### LINKED ISSUE #12838: restore-state creates the default resource group before `rad startup`, which restores over it → deploy fails with resourcegroup \"default\" not found (regression from #12829)
## Summary

Since #12829 (*"Revert default resource group and environment creation from rad install kubernetes"*, merged 2026-08-26T17:48Z), the repo-based deploy pipeline fails deterministically at the environment-creation step with:

```
"code": "NotFound",
"message": "... resource group \"/planes/radius/local/resourcegroups/default\" not found"
```

Example failing runs (Azure): https://github.com/sk593/aks-store-demo/actions/runs/33015265288, https://github.com/sk593/aks-store-demo/actions/runs/33016468544

## Root cause

The shared state archive was getting corrupted (a `default`-group-less snapshot) and then never repaired, via two independent problems. `rad startup` (restore-state) restores the previous run's snapshot with `pg_dump --clean --if-exists` — a **wholesale drop-and-replace** of the control-plane PostgreSQL databases (`ucp`, `applications_rp`, `dynamic_rp`), not a merge.

**1. The `default` group was created *before* `rad startup`, so it never survived a bad restore.** The `restore-state` action created the group and then ran `rad startup`, which drops and replaces the databases — dropping any group the restored snapshot didn't contain. Before #12829, `rad install kubernetes` also created the `default` group **server-side** on every install, silently repairing it regardless of the archive; #12829 removed that, so once a group-less archive existed nothing re-created the group and every deploy failed:

```
21:43:55  rad group create default                 # created...
21:43:57  Restoring control-plane databases...      # ...then dropped by the wholesale restore
21:44:21  State restored successfully.
21:44:22  rad deploy $ENV_BICEP  ->  NotFound: resourcegroup "default" not found
```

**2. `rad shutdown` persisted state unconditionally, so a pre-startup failure blanked the archive.** `rad shutdown` (teardown) runs under `if: always()` but never checked whether `rad startup` restored state first. Any run that **failed before `rad startup` completed** (control-plane setup, cluster connect, OIDC projection, credential registration, or startup itself) still reached teardown, backed up the uninitialized databases, and overwrote the good archive. The next run restored that blank state wholesale and re-persisted it — self-perpetuating corruption.

## Fix (#12840)

A two-part fix; neither half is sufficient alone.

1. **Create the `default` group *after* `rad startup`.** The group is now (re)created on top of whatever the restore produced, so it is present for the deploy and persisted by the next `rad shutdown`. A group-less archive self-heals on the next successful run. (`rad startup` has no resource-group dependency, so nothing needs the group earlier.)
2. **Guard `rad shutdown` on a successful restore.** `restore-state` exposes a `state-restored` output, set to `true` only after `rad startup` succeeds; the workflow passes it into `teardown`'s `state-restored` input, and teardown skips `rad shutdown` (with a `::warning::`) when it is not `true`. A run that never reached startup can no longer overwrite the archive.

First-run seeding and persistence after a *later* deploy failure both still work: `rad startup` is a legitimate no-op on the first run but still succeeds and sets the output. The guard is pure workflow orchestration — it does not couple `rad startup` and `rad shutdown` as commands.

## Impact

All repo-based deploys that pin the composite actions to `@main`. Before this fix, any run that failed before `rad startup` could overwrite the durable state archive with blank state, corrupting it for every future run.

## Follow-ups (not required for this fix)

- #12847 — `rad shutdown` should refuse to persist a degenerate/empty snapshot.
- #12848 — add a `concurrency:` group to serialize workflows that share the state archive.

sk593:
## Description

The repo-based deploy pipeline (`run-rad-commands-azure.yml` / `run-rad-commands-aws.yml`, and the `delete-*` workflows) restores control-plane state at setup with `rad startup` and persists it at teardown with `rad shutdown`. `rad startup` restores the previous run's snapshot with `pg_dump --clean --if-exists` — a wholesale drop-and-replace of the control-plane PostgreSQL databases, not a merge.

Two independent problems corrupted the shared state archive and made deploys fail deterministically with `resource group "default" not found`:

1. **A corrupt (group-less) archive never repaired itself.** After #12829 stopped `rad install kubernetes` from re-seeding the `default` group server-side, the pipeline created it *before* `rad startup`. Since startup restores the databases wholesale, a group created before it is dropped whenever the restored snapshot lacks the group — so once the archive was group-less it stayed that way, and every deploy failed:

   ```
   21:43:55  rad group create default                 # created...
   21:43:57  Restoring control-plane databases...      # ...then dropped by the wholesale restore
   21:44:21  State restored successfully.
   21:44:22  rad deploy $ENV_BICEP  ->  NotFound: resource group ".../resourcegroups/default" not found
   ```

2. **A pre-startup failure overwrote the good archive with blank state.** `rad shutdown` ran unconditionally under `if: always()`. A workflow that failed *before* `rad startup` completed (setup, cluster connect, OIDC, credential registration, or startup itself) still hit teardown, backed up the uninitialized databases, and overwrote the archive. The next run restored that blank state wholesale and re-persisted it — self-perpetuating corruption.

## Fix

A two-part fix; neither half is sufficient alone.

**1. Create the `default` group *after* `rad startup` (self-heal).** Moving `rad group create`/`switch` below the restore means the group is (re)created on top of whatever the restore produced, so it is present for the deploy and persisted by the next `rad shutdown`. A group-less archive now heals itself on the next successful run. `rad startup` has no resource-group dependency, so nothing needs the group earlier.

**2. Guard `rad shutdown` on a successful restore.** `restore-state` exposes a `state-restored` output, set to `true` only after `rad startup` succeeds. The workflow passes it into `teardown`'s `state-restored` input; when it is not `true`, teardown skips `rad shutdown` (emitting a `::warning::`) so a run that never reached startup cannot overwrite the archive. This replaces the earlier `${RUNNER_TEMP}` file marker with an explicit action output→input, so the dependency is visible in the workflow YAML and fails safe if the steps are ever split across jobs.

This preserves the behaviors we *do* want:

- **First-run seeding.** `rad startup` is a no-op on the first run (empty archive) but still succeeds and sets the output, so `rad shutdown` runs and seeds the archive.
- **Persist after a later failure.** If startup succeeded and a *later* step fails (e.g. a partially-applied Terraform deploy), the output is `true`, so `rad shutdown` still runs and that work is not lost.

The guard is pure workflow orchestration — it does **not** couple `rad startup` and `rad shutdown` as commands.

### Also in this PR

- **`teardown`: `rad app list` → `rad app list --preview`.** The pipeline deploys `Radius.Core` applications, but the teardown status step queried the legacy `Applications.Core` plane and listed nothing. `--preview` matches the surface the rest of the pipeline uses, and a failed listing now surfaces a `::warning::` instead of being swallowed by `|| true`.

## Type of change

- Bugfix

## Fixes

Fixes #12838

## Testing

- `.github/extension/actions/teardown/teardown_test.sh` (wired into `build/test.mk` as `test-teardown`) extracts the actual `run:` blocks from the `restore-state` and `teardown` composite actions and executes them with stubbed `rad`/`git` (no cluster/CLI), asserting real behavior:
  - `restore-state` sets `state-restored=true` on `$GITHUB_OUTPUT` after `rad startup`, and creates the `default` group **after** startup (order asserted from the invocation log).
  - Negative path: when `rad startup` fails, the block exits non-zero and never sets the output.
  - First run: a no-op `rad startup` still sets the output.
  - `teardown` runs `rad shutdown` only when `state-restored == "true"`, and otherwise skips with a `::warning::`.
  - `teardown` lists apps via `rad app list --preview` and warns (does not swallow) on failure.
  - All four workflows wire `restore-state` → `teardown` within the same job.
  Verified by mutation that each invariant, when broken, fails the test.
- `make test-teardown` and `make test-extension-action-shell-syntax` pass; the test is `shellcheck`-clean.

## Follow-ups (out of scope)

Raised in review and intentionally deferred — neither is required to fix #12838:

- #12847 — `rad shutdown` should refuse to persist a degenerate/empty snapshot (defense-in-depth for control-plane degradation *after* a successful startup).
- #12848 — add a `concurrency:` group to serialize workflows that share the state archive.

## Notes

Example failing runs (before this fix): https://github.com/sk593/aks-store-demo/actions/runs/33015265288, https://github.com/sk593/aks-store-demo/actions/runs/33016468544


.github/extension/actions/restore-state/action.yml
```diff
@@ -49,3 +49,12 @@ runs:
...
         rad startup
+
+        # Record that state was successfully restored in this run. Teardown reads
+        # this marker to decide whether persisting state is safe: if the workflow
+        # fails before `rad startup` completes, the control-plane databases were
+        # never restored, and `rad shutdown` must not overwrite the durable archive
+        # with that uninitialized state. On the first ever run `rad startup` is a
+        # no-op restore but still succeeds, so the marker is written and teardown
+        # seeds the archive as expected.
+        touch "${RUNNER_TEMP}/radius-state-restored"
```
kachawla: The marker path is a bare literal duplicated here and in `teardown/action.yml:25`, with nothing tying the two together. Edit either string, or reorder `restore-state` out of a workflow, and teardown silently stops persisting while every run still reports green.

Could `restore-state` set a step output that the workflow passes to `teardown` as an input instead? That puts the dependency in the four workflow YAMLs where a reviewer can see it.

.github/extension/actions/teardown/action.yml
```diff
@@ -14,11 +14,18 @@ runs:
...
-        # a partially-applied Terraform run is not lost. It does not manage the
+        # Only persist state when `rad startup` successfully restored it earlier in
+        # this run (marker written by the restore-state action). If the workflow
+        # failed before startup ran, the control-plane databases were never
+        # restored, so backing them up now would overwrite the durable archive with
+        # blank/uninitialized state and corrupt it for every future run. rad
+        # shutdown still runs after a *later* failure (e.g. a partially-applied
+        # Terraform deploy) so that work is not lost. It does not manage the
         # cluster; the cleanup step tears k3d down afterwards.
+        if [ ! -f "${RUNNER_TEMP}/radius-state-restored" ]; then
```
kachawla: This depends on `restore-state` and `teardown` running in the same job, since `RUNNER_TEMP` is per job. True in all four callers today, but teardown is an `if: always()` step, so moving it to its own job is a plausible refactor and would disable persistence with no failing test.

Worth stating the same-job requirement in the comment here and in `restore-state`.

.github/extension/actions/teardown/action.yml
```diff
@@ -14,11 +14,18 @@ runs:
...
+          echo "rad startup did not complete in this run; skipping rad shutdown to preserve the existing state archive."
```
kachawla: `echo` lands in the step log only. Skipping persistence is an abnormal outcome, so could this be a `::warning::` annotation? A repo that quietly stopped persisting weeks ago looks identical to a healthy one in the run list.

.github/extension/actions/teardown/action.yml
```diff
@@ -14,11 +14,18 @@ runs:
...
+          exit 0
+        fi
         git config --global user.email "radius-deploy@users.noreply.github.com"
         git config --global user.name "radius-deploy"
         rad shutdown
```
kachawla: The marker proves `rad startup` ran, not that the databases are worth persisting. If the control plane degrades after a successful startup - a `radius-cp` pod crash-looping, `rad install kubernetes` re-run - `rad shutdown` still snapshots what is there and overwrites the archive.

Having the command refuse a degenerate snapshot (empty dump, zero resource groups) would cover that, and would cover callers outside these actions. Not blocking, but worth tracking before #12838 is treated as closed.

.github/extension/actions/teardown/action.yml
```diff
@@ -27,7 +34,9 @@ runs:
       if: always()
       shell: bash
       run: |
-        rad app list || true
+        # The pipeline deploys Radius.Core applications, so list them via the
+        # preview API surface to match; the legacy plane would show nothing.
+        rad app list --preview || true
```
kachawla: Good catch, and the comment explains why. Noting for the release notes that this is a second, unrelated change riding along with the guard.

.github/extension/README.md
```diff
@@ -121,8 +121,8 @@ The dispatcher routes to the matching provider workflow, which runs on `ubuntu-2
...
-16. **Tear down.** Runs `rad app list`, and always deletes the ephemeral `radius-cp` cluster. On failure, Radius and application logs are collected and uploaded as the `radius-logs` artifact (three-day retention).
+15. **Persist state (`rad shutdown`).** Backs the control-plane databases and Terraform recipe-state Secrets up to the state archive — the OCI-backed archive by default (pushed to GHCR, selected by the `RADIUS_STATE_*` variables), or the `radius-state` git orphan branch when `RADIUS_STATE_BACKEND=git`. This runs even when the deploy fails (`if: always()`), so a partially-applied Terraform run is not lost — but only when `rad startup` restored state earlier in the run. If the run fails before `rad startup` completes, this step is skipped so the durable archive is not overwritten with uninitialized state.
+16. **Tear down.** Runs `rad app list --preview` (the pipeline deploys `Radius.Core` apps, so the preview API surface is used), and always deletes the ephemeral `radius-cp` cluster. On failure, Radius and application logs are collected and uploaded as the `radius-logs` artifact (three-day retention).
```
kachawla: How does an already corrupted archive get repaired now? A corrupt archive that makes `rad startup` fail means the run skips `rad shutdown`, so the bad archive survives and the pipeline stays wedged until someone deletes it by hand. Before this change it would eventually be overwritten.

The trade is the right one, but the recovery procedure should be written down here.

.github/extension/README.md
```diff
@@ -163,7 +163,7 @@ This workflow also reads GitHub Actions **secrets** for image push and applicati
...
+`rad startup` and `rad shutdown` are kind-agnostic CLI commands that restore and back up all durable Radius state (control-plane PostgreSQL + Terraform recipe-state Secrets). These workflows use the OCI-backed state archive by default — the `RADIUS_STATE_*` variables select an OCI repository and the workflow logs in to GHCR before `rad startup`/`rad shutdown` — and fall back to the `radius-state` git orphan branch only when `RADIUS_STATE_BACKEND=git`. They do not manage cluster lifecycle — the workflow owns creating and destroying the ephemeral control plane around them. `rad startup` runs after the install (so `rad deploy` plans against prior state) and `rad shutdown` runs after the commands with `if: always()` (so state survives a failed deploy), but only persists when `rad startup` restored state earlier in the run — a run that fails before `rad startup` skips `rad shutdown` so it cannot overwrite the archive with uninitialized state.
```
kachawla: Also on state persistence: none of the four workflows declares a `concurrency:` group, so two runs against the same repo both restore the same archive and both `rad shutdown`. Last writer wins and the other run's work is dropped.

Not introduced here, but the same class of bug, and worth tracking separately from #12838.

.github/extension/actions/teardown/teardown_test.sh
```diff
@@ -0,0 +1,159 @@
...
+
+def run_block(script, runner_temp):
+    """Execute a run block with stubbed rad/git; return (stdout, rad invocations)."""
+    workdir = pathlib.Path(tempfile.mkdtemp())
+    bin_dir = workdir / "bin"
+    bin_dir.mkdir()
+    rad_log = workdir / "rad.log"
+
+    (bin_dir / "rad").write_text(
+        "#!/bin/bash\nprintf '%s\\n' \"$*\" >> \"${RAD_LOG}\"\n", encoding="utf-8"
```
kachawla: The stub always exits 0, so the negative case for invariant 1 is never exercised: `rad startup` failing must leave the marker absent. That is the assertion protecting the archive, and the other three are downstream of it.

`run_block` treats a non-zero block as a test failure, so this needs an `expect_failure` path.

.github/extension/actions/teardown/teardown_test.sh
```diff
@@ -0,0 +1,159 @@
...
+status_block = extract_run_block(teardown_action, "Show application status")
+
+# 1. restore-state writes the marker after rad startup, in a shared RUNNER_TEMP.
+runner_temp = pathlib.Path(tempfile.mkdtemp())
+marker = runner_temp / "radius-state-restored"
+_, restore_rad = run_block(restore_block, runner_temp)
+if "startup" not in restore_rad:
+    fail("restore-state block must call `rad startup`")
+if not marker.exists():
+    fail("restore-state block must write the state-restored marker after `rad startup`")
```
kachawla: The description leans on `rad startup` being a no-op that still succeeds on the first run against an empty archive. Nothing in the diff establishes that, and an always-zero stub cannot distinguish it. If it is wrong, the archive is never seeded. Could that case be asserted, or linked to a run showing it?

.github/extension/actions/teardown/teardown_test.sh
```diff
@@ -0,0 +1,159 @@
...
+
+# 3. With the marker present (from the restore above), teardown runs rad shutdown.
+_, persist_rad_present = run_block(persist_block, runner_temp)
```
kachawla: This hands both blocks the same `runner_temp`, which is the production assumption itself, so the test manufactures what it should verify. Asserting that every workflow using `teardown` also uses `restore-state` in the same `jobs.<id>.steps` list would cover it. `verify-azure_test.sh` already does wiring assertions of that shape.

.github/extension/actions/teardown/teardown_test.sh
```diff
@@ -0,0 +1,159 @@
...
+
+    if not any(chunk.strip() for chunk in body):
+        sys.exit(f"{action_file}: step '{step_name}' run block is empty")
+
+    return "\n".join(body).rstrip("\n") + "\n"
+
+
+def run_block(script, runner_temp):
+    """Execute a run block with stubbed rad/git; return (stdout, rad invocations)."""
+    workdir = pathlib.Path(tempfile.mkdtemp())
```
kachawla: nit: three `tempfile.mkdtemp()` directories per run, none cleaned up. `progress_test.sh` and `publish-deploy-status_test.sh` use one `mktemp -d` with `trap 'rm -rf' EXIT`. Worth matching so repeated local `make test` runs do not litter `/tmp`.

.github/extension/actions/restore-state/action.yml
```diff
@@ -49,3 +49,12 @@ runs:
         git config --global user.email "radius-deploy@users.noreply.github.com"
         git config --global user.name "radius-deploy"
         rad startup
```
kachawla, line 65: `rad group create default` runs in the step above, and `rad startup` is a wholesale `pg_dump --clean --if-exists` replace, so a group-less archive drops the group that was just created. That is the trace in the description, and `rad startup` succeeds there, so the marker is written and teardown persists as usual.

The guard never engages in that case, so archives already in that state are not repaired and #12838 stays open for them. `rad group create` is idempotent (`CreateOrUpdateResourceGroup`), so moving the create and switch below `rad startup` would let each run self-heal. Was that the branch name's intent?

.github/extension/actions/restore-state/action.yml
kachawla: The marker path is a bare literal duplicated here and in `teardown/action.yml:25`, with nothing tying the two together. Edit either string, or reorder `restore-state` out of a workflow, and teardown silently stops persisting while every run still reports green.

Could `restore-state` set a step output that the workflow passes to `teardown` as an input instead? That puts the dependency in the four workflow YAMLs where a reviewer can see it.

.github/extension/actions/teardown/action.yml
kachawla: This depends on `restore-state` and `teardown` running in the same job, since `RUNNER_TEMP` is per job. True in all four callers today, but teardown is an `if: always()` step, so moving it to its own job is a plausible refactor and would disable persistence with no failing test.

Worth stating the same-job requirement in the comment here and in `restore-state`.

.github/extension/actions/teardown/action.yml
kachawla: `echo` lands in the step log only. Skipping persistence is an abnormal outcome, so could this be a `::warning::` annotation? A repo that quietly stopped persisting weeks ago looks identical to a healthy one in the run list.

.github/extension/actions/teardown/action.yml
kachawla, line 45: The marker proves `rad startup` ran, not that the databases are worth persisting. If the control plane degrades after a successful startup - a `radius-cp` pod crash-looping, `rad install kubernetes` re-run - `rad shutdown` still snapshots what is there and overwrites the archive.

Having the command refuse a degenerate snapshot (empty dump, zero resource groups) would cover that, and would cover callers outside these actions. Not blocking, but worth tracking before #12838 is treated as closed.

.github/extension/actions/teardown/action.yml
kachawla: `|| true` makes a failed command and an empty app list look identical, which is how the legacy-plane query went unnoticed in the first place. Teardown runs at `always()`, so swallowing the error is reasonable, but could the failure branch echo something instead of nothing?

.github/extension/README.md
kachawla, line 125: The trace in the description has `rad startup` succeeding against a bad archive, so the guard never engages in that scenario. Worth being explicit here about what it does and does not cover, and how to reset an archive that is already corrupt, since nothing in the pipeline repairs one now that #12829 removed the server side seeding.

.github/extension/README.md
kachawla, line 166: Also on state persistence: none of the four workflows declares a `concurrency:` group, so two runs against the same repo both restore the same archive and both `rad shutdown`. Last writer wins and the other run's work is dropped.

Not introduced here, but the same class of bug, and worth tracking separately from #12838.

.github/extension/actions/teardown/teardown_test.sh
kachawla: The stub always exits 0, so the negative case for invariant 1 is never exercised: `rad startup` failing must leave the marker absent. That is the assertion protecting the archive, and the other three are downstream of it.

`run_block` treats a non-zero block as a test failure, so this needs an `expect_failure` path.

.github/extension/actions/teardown/teardown_test.sh
kachawla: The description leans on `rad startup` being a no-op that still succeeds on the first run against an empty archive. Nothing in the diff establishes that, and an always-zero stub cannot distinguish it. If it is wrong, the archive is never seeded. Could that case be asserted, or linked to a run showing it?

.github/extension/actions/teardown/teardown_test.sh
kachawla: This hands both blocks the same `runner_temp`, which is the production assumption itself, so the test manufactures what it should verify. Asserting that every workflow using `teardown` also uses `restore-state` in the same `jobs.<id>.steps` list would cover it. `verify-azure_test.sh` already does wiring assertions of that shape.

.github/extension/actions/teardown/teardown_test.sh
kachawla: nit: three `tempfile.mkdtemp()` directories per run, none cleaned up. `progress_test.sh` and `publish-deploy-status_test.sh` use one `mktemp -d` with `trap 'rm -rf' EXIT`. Worth matching so repeated local `make test` runs do not litter `/tmp`.

.github/extension/actions/restore-state/action.yml
sk593: Moved `rad group create`/`switch` below `rad startup` in `729e0a56c` so a group-less archive self-heals.

.github/extension/actions/restore-state/action.yml
sk593: Replaced the file marker with an explicit `state-restored` action output→input in `729e0a56c`, so the coupling is visible in the workflow YAML.

.github/extension/actions/teardown/action.yml
sk593: The same-job requirement is now stated in the teardown input description and step comment in `729e0a56c`, and a test asserts all four workflows wire it within one job.

.github/extension/actions/teardown/action.yml
sk593: Now emits a `::warning::` annotation when skipping persistence in `729e0a56c`.

.github/extension/actions/teardown/action.yml
sk593: Replaced `|| true` with a `::warning::` on the failure branch in `729e0a56c` so a failed listing is distinguishable from an empty one.

.github/extension/README.md
sk593: Rewrote the README state-persistence section in `729e0a56c` to spell out what the guard covers/doesn't and how to reset an already-corrupt archive (the group case self-heals).

.github/extension/actions/teardown/teardown_test.sh
sk593: Added a negative case in `729e0a56c` where the `rad` stub fails on `startup`, asserting the block exits non-zero and never writes `state-restored`.

.github/extension/actions/teardown/teardown_test.sh
sk593: Added an explicit first-run assertion in `729e0a56c` that a no-op `rad startup` still sets `state-restored=true`.

.github/extension/actions/teardown/teardown_test.sh
sk593: Added a wiring assertion in `729e0a56c` that every workflow using teardown has an `id: restore-state` step and passes `state-restored` within the same job.

.github/extension/actions/teardown/teardown_test.sh
sk593: Consolidated to a single `mktemp` scratch root cleaned up on exit in `729e0a56c`.

.github/extension/actions/teardown/action.yml
sk593: Agreed this is a real gap, but it's a different failure mode than #12838 (control-plane degradation *after* a successful startup), so it's out of scope for this PR. Tracked in #12847.

.github/extension/README.md
sk593: Good point — but #12838's failure is deterministic within a single run, so serializing overlapping runs is an unrelated robustness improvement rather than part of this fix. Tracked in #12848.

kachawla (dismissed):

kachawla (approved):

kachawla (approved):

PR #12841, opened 2026-08-26
Build paginated nextLink from the caller-visible host

### LINKED ISSUE #12837: Paginated `nextLink` leaks the RP's cluster-internal address, breaking any out-of-cluster client

nithyatsu:
## Summary

Builds a paginated `nextLink` from the caller-visible host (`X-Forwarded-Host` / `X-Forwarded-Proto`) instead of the serving provider's own `req.Host`.

## Reason for change

`GetURLFromReqWithQueryParameters` took the host straight from `req.Host`. When UCP proxies a LIST downstream to a resource provider, `req.Host` as seen by that provider is its cluster-internal service address, so the `nextLink` it emits is only usable from inside the cluster:

```text
Error: Get "http://dynamic-rp.radius-system:8082/planes/radius/local/resourcegroups/default/providers/Radius.Compute/containers?api-version=2025-08-01-preview&skipToken=...&top=10": dial tcp: lookup dynamic-rp.radius-system on 127.0.0.53:53: no such host
```

Any client that is not in-cluster gets the first page fine and then fails on the second. The user-visible symptom is `rad app delete` failing during **resource enumeration**, before anything is deleted, for an application with more than one page of resources of a single type (`top=10` by default) — leaving the application undeletable. It is not delete-specific: it affects every paginated LIST for every out-of-cluster caller.

UCP's proxy already forwards the caller-visible host and scheme downstream as `X-Forwarded-Host` and `X-Forwarded-Proto` (see `pkg/ucp/proxy/testdata/arm/basic-roundtrip/downstream-request.json`); the helper just ignored them.

Fixes #12837. Reported downstream as radius-project/ai-extensions#441.

## Implementation notes

- Fixing the shared helper covers both `GetNextLinkURL` callers (`pkg/armrpc/frontend/defaultoperation/listresources.go` and `pkg/dynamicrp/frontend/listresources.go`) and every RP that goes through them, with no CLI change.
- Falls back to `req.Host` / `req.URL.Scheme` / `DefaultSheme` exactly as before when the headers are absent or empty, so a direct (unproxied) request is unchanged.
- Takes the first entry of a comma-separated value, as set by a proxy chain.
- `X-Forwarded-Proto` handling follows the existing precedent in `pkg/armrpc/rest/results.go`.

Follow-up worth considering separately: `results.go` builds `Location` / `Azure-AsyncOperation` headers from `req.Host` in the same way and applies `X-Forwarded-Proto` but not `X-Forwarded-Host`, so async operation URLs likely have the same leak. Left out of scope here.

## How to test

`go test ./pkg/armrpc/frontend/controller/ -run 'TestGetURLFromReqWithQueryParameters|TestGetNextLinkURL' -v`

Validated locally:

- 11 new subtests pass, covering: unproxied request, missing scheme defaulting, forwarded host only, forwarded scheme only, both together, comma-separated proxy chain, empty forwarded headers, empty pagination token, and a full proxied `nextLink` round-trip asserting scheme/host/path/`api-version`/`skipToken`/`top`.
- Confirmed the new tests **fail without the production change** (6 subtests fail, including an explicit assertion that the link does not contain `dynamic-rp.radius-system`), so they are real regression tests rather than vacuous ones.
- `go build`, `go vet`, and `gofmt -l` clean.
- Full `./pkg/armrpc/...` and `./pkg/dynamicrp/frontend/...` suites pass with no failures.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/armrpc/frontend/controller/util.go` | Prefer `X-Forwarded-Host` / `X-Forwarded-Proto` over `req.Host` / `req.URL.Scheme` when building the URL, via a new `forwardedHeaderValue` helper that takes the first entry of a comma-separated value. |
| `pkg/armrpc/frontend/controller/util_test.go` | Add `TestGetURLFromReqWithQueryParameters` and `TestGetNextLinkURL` covering proxied and unproxied requests, proxy chains, empty headers, scheme defaulting, and the no-further-pages case. |

### DIFF
--- a/pkg/armrpc/frontend/controller/util.go
+++ b/pkg/armrpc/frontend/controller/util.go
@@ -115,13 +115,38 @@ func checkIfNoneMatchHeader(ifNoneMatchETag string, etag string) error {
 	return nil
 }
 
-// GetURLFromReqWithQueryParameters function builds a URL from the request and query parameters
+// forwardedHeaderValue returns the first entry of a possibly comma-separated
+// X-Forwarded-* header, or an empty string when the header is absent.
+func forwardedHeaderValue(req *http.Request, header string) string {
+	value := req.Header.Get(header)
+	if value == "" {
+		return ""
+	}
+
+	first, _, _ := strings.Cut(value, ",")
+	return strings.TrimSpace(first)
+}
+
+// GetURLFromReqWithQueryParameters function builds a URL from the request and query parameters.
+//
+// When a request reaches a resource provider through UCP's proxy, req.Host is the
+// provider's own cluster-internal address (for example dynamic-rp.radius-system:8082),
+// which the original caller cannot resolve. The proxy forwards the caller-visible host
+// and scheme as X-Forwarded-Host and X-Forwarded-Proto, so prefer those when present.
 func GetURLFromReqWithQueryParameters(req *http.Request, qps url.Values) *url.URL {
 	url := url.URL{
 		Host:     req.Host,
 		Scheme:   req.URL.Scheme,
 		Path:     req.URL.Path,
 		RawQuery: qps.Encode(),
+	}
+
+	if host := forwardedHeaderValue(req, "X-Forwarded-Host"); host != "" {
+		url.Host = host
+	}
+
+	if scheme := forwardedHeaderValue(req, "X-Forwarded-Proto"); scheme != "" {
+		url.Scheme = scheme
 	}
 
 	if url.Scheme == "" {
--- a/pkg/armrpc/frontend/controller/util_test.go
+++ b/pkg/armrpc/frontend/controller/util_test.go
@@ -18,8 +18,10 @@ package controller
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"net/http"
+	"net/url"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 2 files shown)


nithyatsu:
Closing this — it fixes the symptom at the wrong layer, and it does not actually fix the reported failure.

Preferring `X-Forwarded-Host` has a resource provider infer its own client-facing identity. It can't: UCP sets that header from its *own* inbound host (`pkg/ucp/proxy/types.go` → `SetXForwarded()`), and when the CLI reaches UCP through the aggregated APIService, the Kubernetes aggregator has already discarded the original host — `newReq.Host = location.Host` in `k8s.io/apiserver/pkg/util/proxy/proxy.go`, which is kubernetes/kubernetes#107435, the same bug `pkg/sdk/connection_location.go` already warns about. So this change just swaps `dynamic-rp.radius-system:8082` for the UCP service address; both are in-cluster and neither is reachable from an external client.

The correct fix follows what UCP already does for `Location` / `Azure-AsyncOperation`: rewrite the URL on the way out of UCP using the `Referer` header (`ProcessAsyncOperationHeaders`), then rebase the host client-side (`locationRewriteRoundTripper`). `nextLink` needs the same two-stage treatment; it was only missed because it lives in the response body rather than a header.

Tracked in #12837, with the correct fix coming separately.

PR #12842, opened 2026-08-27
Return a relative paginated nextLink so clients can reach the next page

### LINKED ISSUE #12837: Paginated `nextLink` leaks the RP's cluster-internal address, breaking any out-of-cluster client

nithyatsu:
# Description

`GetNextLinkURL` built the paginated `nextLink` from the inbound request host:

```go
url := url.URL{
    Host:   req.Host,        // ← the host *this server* was reached on
    Scheme: req.URL.Scheme,
    Path:   req.URL.Path,
    ...
}
```

A server has no reliable way to know the address the client actually dialed. The request may have been proxied by UCP, and in Kubernetes the API server aggregator explicitly replaces the original host with the address of the service it forwards to:

```go
// k8s.io/apiserver/pkg/util/proxy/proxy.go
newReq.Host = location.Host
```

(see kubernetes/kubernetes#107435 — the same limitation already documented in `pkg/sdk/connection_location.go`).

So the link came back as `http://dynamic-rp.radius-system:8082/planes/...`, a cluster-internal address no out-of-cluster client can reach. Any client that follows it fails. That is why `rad app delete` aborts partway through enumerating resources and leaves the application undeletable, as reported in radius-project/ai-extensions#441.

## Reason for change

Fixes #12837.

The bug only appears once a listing paginates, so it stays invisible until an application has more resources than one page holds — and then the application cannot be deleted at all.

## Approach

The server cannot know the client's address, so it should stop trying to guess it. This change emits a **relative** reference (path and query only) and lets the client resolve it against the endpoint it already reached — which is routable for that client by definition.

This requires no client change. Every generated client already funnels the link through azcore, which resolves a relative `nextLink` against its configured endpoint:

```go
// azcore/internal/exported/request.go — NewRequestForNextLink
if !hasScheme(nextLink) {
    nextLink = JoinPaths(endpoint, nextLink)
}
```

The path base is trimmed so the client supplies its own, which keeps both producers of `nextLink` consistent:

| Producer | Path the server sees | Emitted link |
|---|---|---|
| Resource provider behind the UCP proxy | `/planes/...` (UCP already trimmed its path base in `proxy.go`) | `/planes/...` |
| UCP serving its own list | `/apis/api.ucp.dev/v1alpha3/planes/...` (routes are mounted *under* the path base) | `/planes/...` |

UCP's configured path base (`/apis/api.ucp.dev/v1alpha3`) is exactly the path portion of the endpoint a Kubernetes connection builds, so repeating it would double-prefix the URL. Trimming it means both producers resolve to the same reachable URL. `v1.ParsePathBase` already implements this, so no configuration has to be plumbed through.

## Alternatives considered

Rewriting the link on the way out — either in the UCP proxy or in the SDK round tripper, mirroring how `Location` and `Azure-AsyncOperation` are handled — was implemented first and discarded. Both stages depend on the `Referer` header, whose host is *also* the address UCP itself was reached on and therefore wrong behind the aggregator. Rewriting also means parsing and re-serializing every paginated response body in the proxy. Not embedding an unknowable host in the first place is smaller and removes the failure mode rather than compensating for it.

## Auto-generated summary

| File | Change |
|---|---|
| `pkg/armrpc/frontend/controller/util.go` | `GetNextLinkURL` returns a relative reference, dropping the scheme and authority the server cannot know, and trims the path base so the client supplies its own. |
| `pkg/armrpc/frontend/controller/util_test.go` | Adds value-level coverage for `GetNextLinkURL`. |

## Testing

`GetNextLinkURL` had **no value-level coverage** before this change. The existing tests only asserted that a link was non-empty (`require.NotNil(t, actualOutput.NextLink)`), which is precisely why the leaked host went unnoticed. The new tests assert the link itself:

- the link is relative and never contains the inbound host, for a proxied RP, a UCP-served list, and an Azure-plane list;
- resolving it against the endpoint a Kubernetes connection builds yields the same reachable URL whether it came from the RP or from UCP;
- `api-version`, `top` and an opaque `skipToken` containing reserved characters survive the round trip;
- the last page yields an empty link.

All six subtests fail against the previous implementation and pass with this change. `./pkg/armrpc/...`, `./pkg/ucp/...`, `./pkg/dynamicrp/...`, `./pkg/sdk/...` and `./pkg/corerp/frontend/...` pass, and `golangci-lint` reports no issues.

### Manual verification

With a resource group holding more than one page of resources:

```console
kubectl get --raw "/apis/api.ucp.dev/v1alpha3/planes/radius/local/resourcegroups/<rg>/providers/Radius.Compute/containers?api-version=2025-08-01-preview&top=5" | jq -r .nextLink
```

Before: `http://dynamic-rp.radius-system:8082/planes/...`
After: `/planes/...`

Then confirm `rad app delete` completes against that resource group.

## Contributor checklist

- [x] Considered malicious use cases and addressed them.
- [x] Considered UX impact.
- [x] Considered security implications.
- [x] Added relevant tests.
- [ ] Documentation updated (no user-facing documentation describes the shape of `nextLink`).

### DIFF
--- a/pkg/armrpc/frontend/controller/util.go
+++ b/pkg/armrpc/frontend/controller/util.go
@@ -131,7 +131,19 @@ func GetURLFromReqWithQueryParameters(req *http.Request, qps url.Values) *url.UR
 	return &url
 }
 
-// GetNextLinkUrl function returns the URL string by building a URL from the request and the pagination token.
+// GetNextLinkURL returns the link to the next page of a paginated list response, or an empty
+// string when the current page is the last one.
+//
+// The link is deliberately a relative reference (path and query only) rather than an absolute URL.
+// A server has no reliable way to know the address the client actually dialed: the request may have
+// been proxied by UCP and, in Kubernetes, the API server aggregator replaces the original host with
+// the address of the service it forwards to (kubernetes/kubernetes#107435). Building the link from
+// the inbound host therefore leaks a cluster-internal address such as
+// http://dynamic-rp.radius-system:8082/... that no out-of-cluster client can reach.
+//
+// Emitting a relative reference lets the client resolve the link against the endpoint it already
+// reached, which is by definition routable for it. Clients generated against azcore do this
+// automatically in runtime.NewRequestForNextLink.
 func GetNextLinkURL(ctx context.Context, req *http.Request, paginationToken string) string {
 	if paginationToken == "" {
 		return ""
@@ -144,5 +156,14 @@ func GetNextLinkURL(ctx context.Context, req *http.Request, paginationToken stri
 	qps.Add("skipToken", paginationToken)
 	qps.Add("top", strconv.Itoa(serviceCtx.Top))
 
-	return GetURLFromReqWithQueryParameters(req, qps).String()
+	// Strip whatever the caller's routable base path is so the client can supply its own. UCP serves
+	// its own lists under a path base (/apis/api.ucp.dev/v1alpha3) which is already part of every
+	// client's endpoint, while a proxied resource provider sees a path that begins at /planes/ and
+	// so has nothing to trim.
+	relative := url.URL{
+		Path:     strings.TrimPrefix(req.URL.Path, v1.ParsePathBase(req.URL.Path)),
+		RawQuery: qps.Encode(),
+	}
+
+	return relative.String()
 }
--- a/pkg/armrpc/frontend/controller/util_test.go
+++ b/pkg/armrpc/frontend/controller/util_test.go
@@ -18,8 +18,10 @@ package controller
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"net/http"
+	"net/url"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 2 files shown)


nithyatsu (commented):
Verified the fix end-to-end against the current tree. Summary of what I checked and one nit worth addressing.

**Correctness in production: ✅**

- All Radius-facing generated clients (corerp v20231001preview + v20250801preview, daprrp, datastoresrp, messagingrp, ucp, and `pkg/cli/clients_new/generated/zz_generated_genericresources_client.go`) use `runtime.NewRequestForNextLink(ctx, GET, client.internal.Endpoint(), nextLink)`. The raw `runtime.NewRequest` call sites (`pkg/sdk/v20241101preview/*`, `pkg/sdk/clients/resourcedeploymentoperationsclient.go`) talk to Azure ARM only and are unaffected.
- Dynamic-rp resources like `Radius.Compute/containers` (the exact repro in #12837) are enumerated by the CLI through `GenericResourcesClient`, so the fix covers them.
- azcore `runtime.NewRequestForNextLink` (v1.23.0) joins a scheme-less link via `JoinPaths(endpoint, nextLink)`. `pkg/sdk/connection_kubernetes.go` sets endpoint to `<host>/apis/api.ucp.dev/v1alpha3`, so the client dials `<host>/apis/api.ucp.dev/v1alpha3/planes/...` — the URL UCP actually serves through the aggregated APIService.

**Test run**

Checked out `util.go` + `util_test.go` from this PR on top of `upstream/main` and ran the affected tests:

```
=== RUN   TestGetNextLinkURL_IsRelativeSoTheClientSuppliesTheAddress
--- PASS (3 subtests)
=== RUN   TestGetNextLinkURL_ResolvesAgainstTheEndpointTheClientDialed
--- PASS (2 subtests)
=== RUN   TestGetNextLinkURL_CarriesThePaginationState
--- PASS
=== RUN   TestGetNextLinkURL_IsEmptyOnTheLastPage
--- PASS
PASS
ok      github.com/radius-project/radius/pkg/armrpc/frontend/controller 0.464s
```

**Nit — `TestGetNextLinkURL_ResolvesAgainstTheEndpointTheClientDialed` models the client with `url.URL.ResolveReference`, but the real client uses `runtime.JoinPaths`. They diverge:**

Given endpoint `https://kubernetes.example.com/apis/api.ucp.dev/v1alpha3` and the emitted link `/planes/radius/local/.../Radius.Compute/containers?…`:

| Method | Resolved URL |
| --- | --- |
| `url.URL.ResolveReference` (used in the test) | `https://kubernetes.example.com/planes/radius/local/.../containers?…` |
| `runtime.JoinPaths` (used by the CLI via `NewRequestForNextLink`) | `https://kubernetes.example.com/apis/api.ucp.dev/v1alpha3/planes/radius/local/.../containers?…` |

Per RFC 3986, `ResolveReference` treats an absolute-path reference as replacing the base path — so it drops `/apis/api.ucp.dev/v1alpha3`. `JoinPaths` concatenates. The URL that actually reaches UCP through the aggregated APIService is the `JoinPaths` one; the current test asserts the `ResolveReference` one. Both look plausible in isolation, but a future change that produced `planes/...` (no leading slash) would silently break production while the test kept passing.

Suggest replacing the `base.ResolveReference(...)` block with `runtime.JoinPaths(endpoint, link)` and updating the expected string to include `/apis/api.ucp.dev/v1alpha3`. That closes the loop between what the server emits and what the client dials.

nithyatsu:
Good catch, and it was more than a nit — thanks. Fixed in 7dbd142.

You're right that `ResolveReference` and `JoinPaths` diverge, and the test was asserting the wrong side of that divergence. I confirmed it by running both against azcore v1.23.0 rather than reasoning about it:

```
NewRequestForNextLink -> https://kubernetes.example.com/apis/api.ucp.dev/v1alpha3/planes/radius/local/.../containers?api-version=2025-08-01-preview&skipToken=page2&top=10
ResolveReference      -> https://kubernetes.example.com/planes/radius/local/.../containers?api-version=2025-08-01-preview&skipToken=page2&top=10
```

So the production behaviour was correct all along, but the test asserted a URL the API server does not route — it had quietly dropped the path base.

Rather than swap in `runtime.JoinPaths`, I went one step further and called `runtime.NewRequestForNextLink` itself, since that is the exact function the generated clients invoke and it also exercises `EncodeQueryParams` on the way through. That removes the modelling step entirely instead of replacing one model with a closer one:

```go
req, err := runtime.NewRequestForNextLink(t.Context(), http.MethodGet, endpoint, link)
require.NoError(t, err)

require.Equal(t,
    "https://kubernetes.example.com/apis/api.ucp.dev/v1alpha3/planes/radius/local/resourcegroups/default/providers/Radius.Compute/containers?api-version=2025-08-01-preview&skipToken=page2&top=10",
    req.Raw().URL.String())
```

I also added an explicit assertion for the failure mode you called out, since it is a property of the emitted link rather than of the resolution:

```go
require.True(t, strings.HasPrefix(link, "/"), "link must start at the root")
```

Verified both regressions are now caught by deliberately breaking the implementation:

| Injected regression | Failing subtests |
| --- | --- |
| Previous absolute-URL implementation | 5 |
| Link emitted without a leading slash (`planes/...`) | 3 |

The second one previously passed, which is precisely the gap you identified. `gofmt`, `golangci-lint` and `./pkg/armrpc/...` are clean.

Thanks also for independently checking the generated-client call sites and confirming the raw `runtime.NewRequest` ones are Azure-ARM only — that was the main residual risk I had flagged in the description, so it is good to have it closed.

DariuszPorowski (approved):
LGTM

PR #12849, opened 2026-08-27
Refuse to persist an empty control-plane snapshot in rad shutdown

### LINKED ISSUE #12847: rad shutdown should refuse to persist a degenerate/empty state snapshot
## Background

Follow-up from PR #12840 (review by @kachawla).

The repo-based deploy pipeline gates `rad shutdown` on a `state-restored` signal from the restore-state action: teardown only persists state when `rad startup` restored it earlier in the same run. This prevents a run that fails *before/at* `rad startup` from overwriting the durable state archive with uninitialized state.

## Gap

The `state-restored` signal proves `rad startup` ran, **not** that the resulting control-plane databases are healthy and worth persisting. If the control plane degrades *after* a successful startup (e.g. a pod crash-loop, or `rad install` is re-run mid-pipeline), `rad shutdown` will still snapshot that degraded state and push it to the archive — corrupting it for future runs. The workflow-level guard cannot detect this because startup already succeeded.

## Proposed fix

Make `rad shutdown` itself refuse to persist a degenerate snapshot — e.g. bail out (non-zero, or a no-op with a clear message) when the dump is empty or contains zero resource groups — so the durable archive is never overwritten with obviously-bad state regardless of how the command was invoked.

## Scope

Not required to fix #12838 (the two-part fix in #12840 resolves that failure). This is defense-in-depth for a different failure mode and is intentionally out of scope for #12840.

pujitha24:
## Summary

`rad shutdown` now refuses to commit a control-plane backup to the durable
state archive when the backed-up `ucp` database contains zero rows in the
`resources` table (the table that stores every UCP resource, including
resource groups). It bails out with a clear error and leaves the existing
archive untouched, instead of silently persisting the degenerate snapshot.

## Reason for change

The repo-based deploy pipeline restores durable state with `rad startup` and
persists it with `rad shutdown`. A prior fix (#12840) added a workflow-level
guard so `rad shutdown` only runs when `rad startup` actually restored state
earlier in the same run — but that guard only proves `rad startup` ran, not
that the control plane is still healthy by the time `rad shutdown` runs. If
the control plane degrades *after* a successful startup (e.g. the PostgreSQL
pod crash-loops, or `rad install` is re-run mid-pipeline), `rad shutdown`
would still dump and commit an empty database, corrupting the shared archive
for every future run. This is defense-in-depth inside the command itself, so
the archive is protected regardless of how `rad shutdown` is invoked.

Fixes #

## How to test

```
go test ./pkg/cli/pgbackup/... ./pkg/cli/cmd/shutdown/... -v
```

All tests pass, including:
- `pkg/cli/pgbackup/pgbackup_test.go`: `Test_IsControlPlaneEmpty_*` — unit
  tests for the new `IsControlPlaneEmpty` helper against hand-built
  `COPY ... FROM stdin;` fixtures (no data rows, no `resources` table at all,
  data rows present, missing file).
- `pkg/cli/cmd/shutdown/shutdown_test.go`: `Test_Run_EmptyControlPlaneStopsBeforeCommit`
  — asserts `Run` returns an error and never reaches `BackupTerraform` or
  `session.Commit` when the database backup is empty.

## Validation

- `go build ./pkg/cli/pgbackup/... ./pkg/cli/cmd/shutdown/...` passes.
- `go test ./pkg/cli/pgbackup/... ./pkg/cli/cmd/shutdown/... -v` passes (all
  cases, old and new).
- Fail-then-pass check: with only the two source files (`pgbackup.go`,
  `shutdown.go`) reverted and the test files left in place, the package fails
  to build (`undefined: IsControlPlaneEmpty`) and the new shutdown test fails
  because `Commit` is called unexpectedly — confirming the added check is
  what prevents the commit.
- `go vet ./pkg/cli/pgbackup/... ./pkg/cli/cmd/shutdown/...` is clean;
  `gofmt -l` reports no formatting issues on the changed files.
- `golangci-lint run` (pinned repo version v2.13.1, from `build/tools.yaml`)
  on both changed packages reports 0 issues.
- The `IsControlPlaneEmpty` parsing logic (matching the `pg_dump` plain-format
  `COPY <table> (<cols>) FROM stdin;` ... rows ... `\.` block) was verified
  against a real, locally-run PostgreSQL 16 instance: a `resources` table was
  created with the exact schema this codebase writes
  (`pkg/components/database/postgres/postgresclient.go`), and real
  `pg_dump --data-only` output was captured for both an empty table and a
  populated one (including a row with escaped backslashes/tabs), then fed
  through the function directly — both cases classified correctly. This
  environment has no live Radius cluster or `kubectl`, so the change was not
  exercised end-to-end through `rad shutdown` against a real control plane;
  the above is the closest available proxy for that.
- Branch HEAD is the current tip of `upstream/main` (commit `745ce9cc0`), and
  `upstream/main`'s own CI (Unit Tests, CodeQL) is green as of this writing.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `pkg/cli/pgbackup/pgbackup.go` | Add `IsControlPlaneEmpty(stateDir)`, which detects a `ucp.sql` dump with zero rows in the `resources` table (or no such table at all). |
| `pkg/cli/pgbackup/pgbackup_test.go` | Add unit tests for `IsControlPlaneEmpty` covering empty, missing-table, populated, and missing-file cases. |
| `pkg/cli/cmd/shutdown/shutdown.go` | Call `IsControlPlaneEmpty` after the database backup and before committing; refuse to persist and return a clear error when the backup is empty. |
| `pkg/cli/cmd/shutdown/shutdown_test.go` | Make `fakeStateBackupClient.BackupDatabases` write a realistic `ucp.sql` fixture; add `Test_Run_EmptyControlPlaneStopsBeforeCommit`. |

Fixes #12847


sk593:
The validation is thorough and the approach is sound. A couple of suggestions to strengthen the PR further:

1. **Test coverage for schema-qualified table names:** The regex already handles variants like `COPY myschema.resources (...)` and `COPY "public"."resources" (...)`, but these aren't explicitly tested. Adding a unit test for schema-qualified names would round out coverage and confirm the edge case works as intended.

This is an optional enhancement; the PR is solid as-is. The test fixtures are realistic and accurately represent real `pg_dump` output, and the integration test correctly verifies that empty backups block before commit.

sk593:
Hi @pujitha24, there's some open comments that need to be resolved before we can merge. If they have been addressed, please reply/resolve the comments. Thanks!

PR #12850, opened 2026-08-27
Add concurrency groups to repo-based deploy/delete workflows

### LINKED ISSUE #12848: Add concurrency: group to repo-based deploy/delete workflows to serialize shared state-archive access
## Background

Follow-up from PR #12840 (review by @kachawla).

The four repo-based deploy/delete workflows (`run-rad-commands-aws.yml`, `run-rad-commands-azure.yml`, `delete-aws.yml`, `delete-azure.yml`) share a single durable state archive (the OCI package selected by `RADIUS_STATE_*`, or the `radius-state` git orphan branch). Each run does `rad startup` (restore) → commands → `rad shutdown` (persist).

## Gap

None of the workflows declares a `concurrency:` group. If two runs against the same repo/environment overlap, they race on the shared archive: one run's `rad shutdown` can overwrite state the other run restored, or two runs can push conflicting snapshots, corrupting the archive.

## Proposed fix

Add a `concurrency:` group (keyed by workflow + environment/ref, with an appropriate `cancel-in-progress` policy) to serialize runs that share a state archive.

## Scope

Not required to fix #12838 — that failure is deterministic within a single run. This is an unrelated robustness improvement and is intentionally out of scope for #12840.

pujitha24:
## Summary

Add a `concurrency:` group to the four repo-based deploy/delete reusable
workflow templates (`run-rad-commands-aws.yml`, `run-rad-commands-azure.yml`,
`delete-aws.yml`, `delete-azure.yml`) so runs against the same GitHub
Environment are serialized instead of racing on the shared state archive.

## Reason for change

These four workflows each restore Radius control-plane state with
`rad startup` and persist it again with `rad shutdown`, against a single
durable state archive per environment (an OCI package or the `radius-state`
git orphan branch). None of them declared a `concurrency:` group, so two
overlapping runs targeting the same environment — e.g. two deploys, or a
deploy and a delete — could race on that archive: one run's `rad shutdown`
overwriting state another run just restored, or two runs pushing conflicting
snapshots and corrupting it.

The added group key intentionally omits `github.workflow` so that a deploy
and a delete against the same environment also serialize against each other,
since both touch the same archive; keying only by the workflow name would
still leave that pair racing. `cancel-in-progress` is left `false` so a
queued run waits its turn instead of a newer run cancelling one that is
already mid-flight, which would abandon it before `rad shutdown` can persist
its state.

Fixes #

## How to test

These are reusable-workflow (`workflow_call`) YAML templates that Radius
copies into a user's repository; they have no Go unit-test target. Validated
statically:

- `actionlint .github/extension/{run-rad-commands-aws,run-rad-commands-azure,delete-aws,delete-azure}.yml`
  — no new findings. The one pre-existing shellcheck `SC1090` warning in the
  two `run-rad-commands-*` files is unrelated to this change and reproduces
  identically on `main` (confirmed via `git stash`).
- `python3 -c "import yaml; yaml.safe_load(open(f).read())"` for all four
  files — parses cleanly, including the new `concurrency` block.
- Confirmed placement matches this repo's own convention for reusable and
  top-level workflows (`concurrency:` immediately after `permissions:` and
  before `env:`/`jobs:`, as in `build.yaml` / `unit-tests.yaml` /
  `functional-test-cloud.yaml`).

Not run: an actual overlapping-run reproduction against a live GitHub
Actions runner and real state archive, which isn't reproducible locally.
GitHub Actions' documented behavior is that a `concurrency:` group declared
inside a reusable `workflow_call` workflow is honored for that workflow's
own job(s), and in this repo's case these four files are copied verbatim
into the user's `.github/workflows/` and invoked via a local `uses: ./...`
path (per `.github/extension/README.md`), which is the simplest, well
supported form of this feature.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/run-rad-commands-aws.yml` | Added `concurrency:` group keyed by environment, `cancel-in-progress: false`. |
| `.github/extension/run-rad-commands-azure.yml` | Added `concurrency:` group keyed by environment, `cancel-in-progress: false`. |
| `.github/extension/delete-aws.yml` | Added `concurrency:` group keyed by environment, `cancel-in-progress: false`. |
| `.github/extension/delete-azure.yml` | Added `concurrency:` group keyed by environment, `cancel-in-progress: false`. |

Fixes #12848


pujitha24:
While checking on this I noticed #12719 merged into main and removed `.github/extension/` from this repo entirely — those four workflow templates now live solely in radius-project/ai-extensions (radius-project/ai-extensions#424). That's why this is showing as conflicting: the files this PR modifies don't exist on main anymore. The concurrency gap from #12848 is still real over there (I checked ai-extensions' copy of `run-rad-commands-aws.yml` and it has no `concurrency:` block either), but porting this fix belongs in that repo, not here. I haven't touched anything in this PR — didn't want to guess at closing it without checking with you first. Want me to close this in favor of opening the equivalent fix in ai-extensions, or is there a reason to keep it open here?

sk593:
Hi @pujitha24, thanks for the contribution. We have changed our workflow setup so this PR will require some updates. I will respond here again once we have more info on how to update the PR. Thank you for your patience!

sk593:
Hi @pujitha24, we moved all of our workflows to the [ai-extensions](https://github.com/radius-project/ai-extensions) repository. Please reopen this PR in that repository for review. Thanks!

pujitha24:
Makes sense — closing this here since `.github/extension/` moved wholesale to ai-extensions (radius-project/ai-extensions#424 hosts these workflows now). I'll open the equivalent concurrency-group fix against that repo's copy of these templates.

PR #12853, opened 2026-08-27
Fix empty environment list in deploy status control-plane diagnostic

lakshmimsft:
## Summary

`publish-deploy-status` runs `rad env list -o json` to capture control-plane health
into the `deploy-controlplane.log` status artifact. `rad env list` without `--preview`
dispatches to the legacy `Applications.Core/environments` surface, but this pipeline
creates its environment as a `Radius.Core/environments` resource. The diagnostic
therefore always logged an empty list.

This adds `--preview` to that call so the log reports the environments the pipeline
actually created, and adds a test guard so the flag cannot be dropped again.

## Reason for change

The extension consistently uses the Radius.Core preview surface: the environment is
created as `Radius.Core/environments` via Bicep (`run-rad-commands-azure.yml`,
`run-rad-commands-aws.yml`), and every other read-side command already passes
`--preview`.

 The result could be a diagnostic artifact that silently reports no
environments on every run — most misleading exactly when someone is using it to debug a
failed deploy, since an empty list reads as "the control plane has no environments"
rather than "this command looked in the wrong place".

The existing test asserted only that the `# rad env list` *header* was present, so it
guarded the `echo` rather than the command, and stayed green with the bug in place.

## How to test

Unit tests:

```bash
bash .github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh
bash .github/extension/actions/action-shell-syntax_test.sh
```

To confirm the new assertions actually reach the changed code, revert `action.yml` and
re-run — the test must fail with `rad: env list must use --preview`:

```bash
git stash push -- .github/extension/actions/publish-deploy-status/action.yml
bash .github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh   # expected: FAIL
git stash pop
```

End to end: run a deploy workflow and download the
`radius-deploy-status-<environment>-<app>` artifact. `deploy-controlplane.log` should
now list the environment the run created under `# rad env list --preview`, where it
previously showed an empty array.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/extension/actions/publish-deploy-status/action.yml` | Pass `--preview` to `rad env list` so the control-plane diagnostic reads `Radius.Core/environments`; update the log header to match and add a comment explaining why the flag is required. |
| `.github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh` | Make the `rad` stub reject `env list` without `--preview` (mirroring the existing `resource list` guard), and assert the environment name reaches `deploy-controlplane.log` rather than only the header. |

### DIFF
--- a/.github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh
+++ b/.github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh
@@ -265,6 +265,10 @@ case "${1:-} ${2:-}" in
         echo "RELEASE VERSION 0.99.0-test"
         ;;
     "env list")
+        if [[ " $* " != *" --preview "* ]]; then
+            echo "rad: env list must use --preview" >&2
+            exit 64
+        fi
         printf '%s\n' '[{"name":"aks-dev"}]'
         ;;
     *)
@@ -561,7 +565,8 @@ progress_jq 'any(.resources[]; .name == "queue"
 
 assert_status_file_contains "deploy-activity.log" '"outcome":"success"'
 assert_status_file_contains "deploy-controlplane.log" "# rad version"
-assert_status_file_contains "deploy-controlplane.log" "# rad env list"
+assert_status_file_contains "deploy-controlplane.log" "# rad env list --preview"
+assert_status_file_contains "deploy-controlplane.log" "aks-dev"
 assert_status_file_contains "deploy-state.txt" "state=success"
 assert_status_file_contains "deploy-state.txt" "application=todolist"
 assert_status_file_contains "deploy-state.txt" "sha=deadbeefcafe"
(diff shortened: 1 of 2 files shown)


sk593 (approved):

.github/extension/actions/publish-deploy-status/publish-deploy-status_test.sh
```diff
@@ -561,7 +565,8 @@ progress_jq 'any(.resources[]; .name == "queue"
 
 assert_status_file_contains "deploy-activity.log" '"outcome":"success"'
 assert_status_file_contains "deploy-controlplane.log" "# rad version"
-assert_status_file_contains "deploy-controlplane.log" "# rad env list"
+assert_status_file_contains "deploy-controlplane.log" "# rad env list --preview"
+assert_status_file_contains "deploy-controlplane.log" "aks-dev"
```
sk593, line 569: should this be removed? the aks-dev line

lakshmimsft: it's needed. dropping  --preview from the command while leaving the header intact passes the header assert and fails only on  aks-dev, and with that line removed the suite goes green with the bug present

PR #12885, opened 2026-09-02
Support windowless background CLI execution on Windows

### LINKED ISSUE #702: [Bug]: Prevent managed Radius commands from flashing terminal windows on Windows
### What type of issue is this?

Bug report

### Summary

Managed Radius CLI commands launched by the extension can briefly display Windows Terminal or console windows. The extension needs a supported windowless subprocess path that preserves Radius/Bicep console semantics, output capture, exit status, cancellation, and process-tree cleanup.

### Steps to reproduce

1. Install and run the Radius extension on Windows.
2. Trigger an operation that invokes the managed `rad.exe`, such as application modeling or graph generation.
3. Observe a terminal window briefly open or flash while Radius or Bicep runs.

Flag-only approaches do not fully solve the problem: `CREATE_NO_WINDOW` causes Radius v0.60.1 to hang on Windows 11 ARM64, while `DETACHED_PROCESS` allows Radius to complete but can open Windows Terminal UI.

### Expected behavior

Managed Radius commands and descendants run without presenting any console UI while retaining exact argv, cwd, environment, stdout/stderr, exit status, timeout and abort behavior, and complete descendant cleanup.

### Actual behavior

Radius requires console-compatible process semantics on Windows. Existing Node `windowsHide` and Windows creation-flag combinations either allow visible terminal UI or cause the real Radius CLI to hang.

### Screenshots or logs

N/A

### Version or commit

Radius extension `main`; Radius CLI v0.60.1; implementation proposed in radius-project/ai-extensions#695.

### Environment

Windows 11 ARM64, Node.js 24.19.0. Windows x64 is covered by hosted CI.

### Additional context

PR #695 implements a GUI-subsystem helper that hosts Radius and Bicep in a headless ConPTY, assigns the process tree to a kill-on-close Job Object, and verifies that no visible root or descendant console window appears. The related Radius CLI investigation is radius-project/radius#12883.

### LINKED ISSUE #12883: Support windowless background CLI execution on Windows
## Problem

The Radius Copilot extension invokes `rad.exe` as a background subprocess on Windows. Node's normal `windowsHide: true` behavior is insufficient because Radius currently must be launched with `detached: true` to avoid a Windows Job Object hang. In libuv, detached execution adds `DETACHED_PROCESS`; Windows ignores `CREATE_NO_WINDOW` when that flag is present, leaving only `SW_HIDE` and allowing visible console-window flashes.

This affects automated callers that need to capture stdout/stderr and reliably terminate the complete `rad` + Bicep process tree without displaying terminal windows.

## Current behavior

The extension launches managed Radius commands with:

```js
spawn(radPath, args, {
  stdio: ["ignore", "pipe", "pipe"],
  windowsHide: true,
  detached: true
});
```

On Windows 11 ARM64 with Node.js 24 and Radius v0.60.1:

- the detached invocation completes successfully;
- the equivalent non-detached invocation produces no output and times out, including when all stdio is ignored;
- detached execution can display a brief console window because `CREATE_NO_WINDOW` is ignored with `DETACHED_PROCESS`.

This is distinct from #12516 / #12519. That change fixed Bicep inheriting Radius's stderr handle, but non-detached invocation from Node still hangs in the scenario above.

### 1. Investigate and fix non-detached execution under a Windows Job Object

Add a regression test that starts `rad version --cli --output json` from a Node/libuv-style parent without `detached`, using piped stdout/stderr and a bounded timeout. The command should complete without requiring callers to escape Node's Job Object.

Please include Windows ARM64 coverage because that is where the behavior was reproduced.

### 2. Provide a supported no-console automation mode for child processes

Radius directly launches console executables without a Windows no-window policy, including:

- Bicep in `pkg/cli/bicep/build.go`;
- Bicep in `pkg/cli/cmd/bicep/publishextension/publish.go`;
- kubectl in `pkg/cli/pgbackup/pgbackup.go`;
- Git in `pkg/statearchive/git/git.go`;
- the direct Azure CLI wrapper in `pkg/azure/azcli/azcli.go`.

Introduce a centralized command-construction helper and a Windows-specific automation/background policy that applies `SysProcAttr.HideWindow` and `CREATE_NO_WINDOW` where interaction is not required. An explicit environment variable or documented non-interactive mode would avoid changing interactive Git credential-helper behavior.

The Azure Identity SDK creates its own `cmd.exe /c az ...` process, so that path may require upstream SDK support or a credential strategy that does not invoke Azure CLI.

### 3. Consider a windowless Windows automation artifact

A command-line flag cannot prevent the initial `rad.exe` console from being created because flags are parsed after process startup. If non-detached operation cannot be made reliable, consider shipping a GUI-subsystem companion such as `radw.exe` that:

- runs the same CLI command implementation;
- preserves redirected stdin/stdout/stderr;
- starts descendants with no-console flags;
- supports process-tree cancellation;
- is published for Windows x64 and ARM64.

## Acceptance criteria

- `rad version --cli --output json` completes from a non-detached Node subprocess on Windows x64 and ARM64.
- Supported non-interactive commands can run without creating visible console windows for Radius or its direct children.
- stdout, stderr, exit codes, cancellation, and timeout behavior remain deterministic.
- Interactive terminal use remains unchanged.
- Windows regression tests cover Job Object behavior and window suppression.

brooke-hamilton:
## Summary

- Automatically apply `HideWindow` and `CREATE_NO_WINDOW` to Radius-owned Windows child processes when `rad.exe` has no attached console, without detaching them from the caller's Job Object.
- Replace eager terminal background detection with lazy, terminal-only detection so redirected `rad.exe` invocations reach command execution without hanging.
- Add native Windows amd64 and arm64 coverage for non-detached execution, piped output, child window suppression, and process-tree cancellation.

## Reason for change

Automated Windows callers need to capture `rad.exe` output and cancel its complete process tree without visible console windows. The previous detached workaround prevented `CREATE_NO_WINDOW` from taking effect, while eager terminal background detection could block before `main` when standard streams were redirected.

Fixes #12883

## How to use
Windows automation callers launch `rad.exe`. A Node.js caller should use `detached: false`, `windowsHide: true`, and piped stdout and stderr. The caller's `windowsHide` option suppresses the initial `rad.exe` console because Radius cannot change its own process creation flags after startup.

When `rad.exe` has no attached console, [pkg/process](../../pkg/process/) automatically applies `SysProcAttr.HideWindow` and `CREATE_NO_WINDOW` to Radius-owned child processes such as Bicep, kubectl, Git, and the direct Azure CLI wrapper. It does not set `DETACHED_PROCESS` or `CREATE_BREAKAWAY_FROM_JOB`, so `rad.exe` and its descendants remain in the automation caller's Windows Job Object for process-tree cancellation. Standard input, output, error, exit codes, and context cancellation retain their existing behavior.

When `rad.exe` has an attached console, child terminal access and interactive CLI behavior are unchanged. Azure Identity credentials create their own Azure CLI process and do not use `pkg/process`; automation that must guarantee windowless descendants should use a non-CLI authentication method such as `ServicePrincipal`, `ManagedIdentity`, or `UCPCredential`.

## How to test

```powershell
go test ./pkg/process ./pkg/cli/style ./test/windowless -count=1 -timeout=2m
go test ./pkg/cli/bicep -run '^TestVersion_ClosesCallerStderrPipePromptly$' -count=1 -timeout=30s
go test ./pkg/azure/azcli ./pkg/cli/cmd/bicep/publishextension ./pkg/cli/cmd/radinit/common ./pkg/cli/pgbackup ./pkg/cli/prompt/... ./pkg/statearchive/git ./cmd/rad/cmd -count=1 -timeout=3m
```

To run a local build from Windows (ARM or AMD64), run this from the root of the repo:

```powershell
go build -o "dist\windows_$(go env GOARCH)\release\rad.exe" ./cmd/rad
```

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `.github/workflows/unit-tests.yaml` | Run focused Windows CLI process tests on native amd64 and arm64 runners. |
| `docs/architecture/rad-cli.md` | Document automatic windowless child behavior and the Azure Identity limitation. |
| `docs/contributing/contributing-code/contributing-code-cli/README.md` | Document native Windows process test commands and coverage. |
| `go.mod` | Promote terminal and Windows syscall modules to direct dependencies. |
| `pkg/azure/azcli/azcli.go` | Construct the direct Azure CLI child through the shared process policy. |
| `pkg/cli/bicep/build.go` | Construct Bicep through the shared process policy while retaining private stderr piping. |
| `pkg/cli/cmd/bicep/publishextension/publish.go` | Construct Bicep publishing through the shared process policy. |
| `pkg/cli/cmd/radinit/common/display.go` | Replace eager compatibility color detection with Radius lazy styling. |
| `pkg/cli/pgbackup/pgbackup.go` | Construct kubectl children through the shared process policy. |
| `pkg/cli/prompt/list/list.go` | Use lazy adaptive colors for list prompts. |
| `pkg/cli/prompt/text/text.go` | Use lazy adaptive colors for text prompt errors. |
| `pkg/cli/style/color.go` | Add adaptive color detection that avoids terminal queries when no terminal is attached. |
| `pkg/cli/style/color_test.go` | Cover terminal-free and interactive adaptive color selection. |
| `pkg/process/process.go` | Add centralized command constructors. |
| `pkg/process/process_other.go` | Preserve default command behavior on non-Windows platforms. |
| `pkg/process/process_other_test.go` | Cover non-Windows no-op behavior. |
| `pkg/process/process_windows.go` | Detect console attachment and automatically configure windowless children when needed. |
| `pkg/process/process_windows_test.go` | Cover console detection, windowless flags, contexts, and existing flag preservation. |
| `pkg/statearchive/git/git.go` | Construct Git children through the shared process policy. |
| `test/windowless/windowless_windows_test.go` | Build the real `rad.exe` and test Job Object execution, output, automatic window suppression, and cancellation. |

### DIFF
--- a/pkg/process/process_windows.go
+++ b/pkg/process/process_windows.go
@@ -26,6 +26,9 @@ import (
 )
 
 var hasConsole = func() bool {
+	// GetConsoleWindow returns zero for Windows Terminal and other pseudoconsole hosts even when
+	// the process has an attached console. GetConsoleCP succeeds for both classic consoles and
+	// pseudoconsole hosts, so it preserves interactive child behavior in either environment.
 	_, err := windows.GetConsoleCP()
 	return err == nil
 }
--- a/pkg/process/process_windows.go
+++ b/pkg/process/process_windows.go
@@ -1,0 +1,41 @@
+//go:build windows
+
+/*
+Copyright 2023 The Radius Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package process
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+func configure(cmd *exec.Cmd) *exec.Cmd {
+	if !strings.EqualFold(os.Getenv(NoWindowEnvVar), "true") {
+		return cmd
+	}
+
+	if cmd.SysProcAttr == nil {
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	cmd.SysProcAttr.HideWindow = true
+	cmd.SysProcAttr.CreationFlags |= windows.CREATE_NO_WINDOW
+	return cmd
+}
--- a/pkg/process/process_windows.go
+++ b/pkg/process/process_windows.go
@@ -19,16 +19,19 @@ limitations under the License.
 package process
 
 import (
-	"os"
 	"os/exec"
-	"strings"
 	"syscall"
 
 	"golang.org/x/sys/windows"
 )
 
+var hasConsole = func() bool {
+	_, err := windows.GetConsoleCP()
+	return err == nil
+}
+
 func configure(cmd *exec.Cmd) *exec.Cmd {
-	if !strings.EqualFold(os.Getenv(NoWindowEnvVar), "true") {
+	if hasConsole() {
 		return cmd
 	}
 
--- a/test/windowless/windowless_windows_test.go
+++ b/test/windowless/windowless_windows_test.go
@@ -30,12 +30,12 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"syscall"
 	"testing"
 	"time"
 	"unsafe"
 
 	"github.com/radius-project/radius/pkg/cli/bicep"
+	"github.com/radius-project/radius/pkg/process"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/sys/windows"
 )
@@ -57,6 +57,7 @@ const (
 
 var (
 	getConsoleWindow = windows.NewLazySystemDLL("kernel32.dll").NewProc("GetConsoleWindow")
+	freeConsole      = windows.NewLazySystemDLL("kernel32.dll").NewProc("FreeConsole")
 	radBinaryPath    string
 )
 
@@ -181,6 +182,11 @@ func newAutomationParentCommand(ctx context.Context, statusPath string, bicepMod
 }
 
 func runAutomationParent() int {
+	if result, _, err := freeConsole.Call(); result == 0 {
+		fmt.Fprintln(os.Stderr, err)
+		return 1
+	}
+
 	job, err := createKillOnCloseJob()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
@@ -194,15 +200,11 @@ func runAutomationParent() int {
 	// Keep the only job handle open for the helper lifetime. Its automatic close when the helper
 	// exits is what terminates rad and every descendant during the cancellation test.
 
-	cmd := exec.Command(os.Getenv(testRadBinaryEnv), "version", "--cli", "--output", "json")
+	cmd := process.Command(os.Getenv(testRadBinaryEnv), "version", "--cli", "--output", "json")
 	cmd.Env = append(os.Environ(),
 		testHelperEnv+"="+testHelperBicep,
 		bicep.BicepEnvVar+"="+os.Args[0],
 	)
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		HideWindow:    true,
-		CreationFlags: windows.CREATE_NO_WINDOW,
-	}
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
--- a/test/windowless/windowless_windows_test.go
+++ b/test/windowless/windowless_windows_test.go
@@ -51,6 +51,7 @@ const (
 	testProcessTimeout     = 20 * time.Second
 	testJobExitCode        = 125
 	testBicepVersion       = "0.42.1"
+	testBicepStderr        = "Bicep diagnostic from test"
 	testStatusPollInterval = 25 * time.Millisecond
 )
 
@@ -105,6 +106,7 @@ func TestRadVersion_NonDetachedWindowlessProcessCompletesInsideJob(t *testing.T)
 	}
 	require.NoError(t, json.Unmarshal(stdout.Bytes(), &result), stdout.String())
 	require.Equal(t, testBicepVersion, result.Bicep)
+	require.Contains(t, stderr.String(), testBicepStderr)
 
 	status := readBicepStatus(t, statusPath)
 	require.Equal(t, "false", status["console"], "Bicep must not have an attached console")
@@ -284,6 +286,7 @@ func runFakeBicep() int {
 		}
 	}
 
+	fmt.Fprintln(os.Stderr, testBicepStderr)
 	fmt.Printf("Bicep CLI version %s (test)\n", testBicepVersion)
 	return 0
 }
--- a/test/windowless/windowless_windows_test.go
+++ b/test/windowless/windowless_windows_test.go
@@ -1,0 +1,322 @@
+//go:build windows
+
+/*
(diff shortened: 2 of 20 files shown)


test/windowless/windowless_windows_test.go
```diff
@@ -0,0 +1,322 @@
+//go:build windows
+
+/*
```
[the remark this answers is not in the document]
brooke-hamilton: This does not affect `make build`: Go's `./...` wildcard skips directories with no buildable files for the target platform. I verified this with `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build ./...`, which succeeds. The build-constraint error occurs only when `./test/windowless` is named explicitly, and CI names that package only on Windows. A non-Windows stub would therefore add an otherwise empty production package without fixing a build failure.

test/windowless/windowless_windows_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+	cmd := exec.Command(os.Getenv(testRadBinaryEnv), "version", "--cli", "--output", "json")
+	cmd.Env = append(os.Environ(),
+		testHelperEnv+"="+testHelperBicep,
+		bicep.BicepEnvVar+"="+os.Args[0],
+	)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		HideWindow:    true,
+		CreationFlags: windows.CREATE_NO_WINDOW,
+	}
```
nicolejms: This makes the test non-discriminating — it passes with the fix disabled.

Setting `CREATE_NO_WINDOW` here gives `rad.exe` a real console (just without a window). I probed this on Windows/arm64: a `CREATE_NO_WINDOW` child reports `GetConsoleCP = 437` (success); only `DETACHED_PROCESS`, or a parent with no console at all, yields an error. So `hasConsole()` returns true inside `rad.exe`, `configure()` is a no-op for the whole test run, and Bicep simply inherits the parent's windowless console — which is why `GetConsoleWindow()` is 0 at lines 112 and 129 regardless of what `pkg/process` does.

I verified this rather than inferring it: neutralizing `configure()` to return the command unmodified still yields `ok github.com/radius-project/radius/test/windowless`.

Suggested fix: call `kernel32!FreeConsole` at the top of `runAutomationParent` and drop this `SysProcAttr` block. That reproduces the real automation shape (a console-less parent) and makes the assertion meaningful — from a `FreeConsole`'d parent I measured a default-flag child getting a visible console window (`GetConsoleWindow() = 9375490`) versus 0 with `CREATE_NO_WINDOW`.

The rest of the test is valuable as-is: Job Object containment, piped output, and process-tree cancellation are all genuinely exercised.

pkg/process/process_windows.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"os/exec"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+var hasConsole = func() bool {
+	_, err := windows.GetConsoleCP()
```
nicolejms, line 32: Worth a comment explaining why `GetConsoleCP` is the right probe here, because the obvious alternative is wrong in a non-obvious way.

Under Windows Terminal / ConPTY, `GetConsoleWindow()` returns 0 even though a console is attached — I measured `CP=437, hwnd=0` for an ordinary PowerShell session. Had this used `GetConsoleWindow() != 0`, every modern terminal session would be classified as console-less and interactive children would lose terminal access. `GetConsoleCP` avoids that.

Without a note, someone is likely to "simplify" this later and reintroduce the bug.

nicolejms (dismissed):

pkg/process/process_windows.go
brooke-hamilton: Fixed — added a comment explaining that `GetConsoleWindow` returns zero for ConPTY hosts and that `GetConsoleCP` correctly detects both classic consoles and pseudoconsole environments.

test/windowless/windowless_windows_test.go
brooke-hamilton: Fixed — the automation parent now calls `FreeConsole` and launches `rad.exe` through the production process constructor instead of hard-coding `CREATE_NO_WINDOW`. I also verified the test is discriminating: neutralizing `configure` now fails because Bicep reports an attached window.

nicolejms (approved):

PR #12907, opened 2026-09-04
fix(ci): surface API response when CLI download test fails to parse version

### LINKED ISSUE #12906: CLI nightly test failed - darwin-arm64
Test failed on radius-project/radius for darwin-arm64. See [workflow logs](https://github.com/radius-project/radius/actions/runs/33819150355) for more details.

pujitha24:
## Summary

Fixes a silent-failure bug in `build/test-cli-download.sh` (used by the nightly rad CLI download workflow) where, under `set -euo pipefail`, a non-matching `grep` while extracting the release version from the GitHub API response would abort the script immediately -- before the script's own `[ -z "$RAD_VERSION" ]` error-handling branch (and its diagnostic message) could ever run. The fix (`|| true` on the extraction pipeline) lets that branch execute as intended, and it now also prints the raw API response body so a future failure is actually diagnosable.

## Reason for change

The nightly rad CLI download workflow has generated 130+ "CLI nightly test failed" issues over the past year, each closed by maintainers as an unexplained transient failure because the CI log always cuts off right after "GitHub API call successful" with no further output before `make` reports a bare "Error 1". The script's own error-reporting code for this exact case was unreachable dead code due to a `set -e`/`pipefail` interaction with a `grep` pipeline. This change does not claim to know the root cause of the underlying API failures (GitHub does not retain the raw API response from past runs, so it can't be confirmed to have been a rate-limit response specifically) -- only that the diagnostic dead-code bug is real, reproducible, and now fixed, so the next occurrence will actually be diagnosable.

Fixes #12906

## How to test

```
bash -n build/test-cli-download.sh
shellcheck --rcfile .github/linters/.shellcheckrc build/test-cli-download.sh
bash build/test-cli-download.sh linux amd64 rad "" ""   # happy path against the live GitHub API
```

To see the fix in action, mock `api_response` (e.g. `api_response='{"message":"API rate limit exceeded"}'`) in place of the `curl` call and re-run the script: before the fix it exits 1 with no output after "GitHub API call successful"; after the fix it prints the response body before exiting 1.

Note: `make lint-shell` could not be run directly in this sandbox because its Go-based tool installer needs disk space that isn't available here (an environment limitation, not a code issue); `shellcheck` was run directly against the repo's pinned `.shellcheckrc` as a faithful substitute for that specific CI check.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `build/test-cli-download.sh` | Add `\|\| true` so a non-matching `grep` in the version-extraction pipeline no longer silently kills the script under `set -e`; print the raw API response body when version extraction fails. |


pujitha24:
Thanks @lakshmimsft — both addressed in the latest commit.

The first one was a real bug. The script runs under `set -euo pipefail`, so a failing curl inside `api_response=$(curl ...)` exits at the assignment and the `curl_exit_code=$?` check below it never runs — against an unreachable endpoint the old code exits 7 silently. It is now a `|| { ... }` block, with `-sS` so curl surfaces its own error too.

Both error paths now use `printf ... >&2` rather than `echo`. Verified against a dead endpoint (curl error plus the exit code on stderr, exit 1) and against an unparseable body (response on stderr, nothing on stdout).

lakshmimsft (approved):

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

lakshmimsft:
@pujitha24 rebase is needed for it to merge.

build/test-cli-download.sh
```diff
@@ -65,21 +65,24 @@ echo "Fetching latest release version from GitHub API..."
...
 # Make API call
-api_response=$(curl -s "$radReleaseUrl")
-curl_exit_code=$?
-
-if [ $curl_exit_code -ne 0 ]; then
-    echo "GitHub API call failed with exit code: $curl_exit_code"
+# `|| { ... }` rather than a following `$?` check: the script runs under `set -e`, so
+# a failing curl inside the assignment exits immediately and a separate check below it
+# is never reached. -sS keeps curl quiet on success but lets its own error through.
+api_response=$(curl -sS "$radReleaseUrl") || {
```
kachawla, line 71: When parsing fails, the log does not show the HTTP status, which is the quickest way to tell a rate limit (403) from a server error (5xx). Can you capture it with `-w '%{http_code}'` and include it in the error message on line 85?

kachawla: There is no retry on this call, so a single transient 5xx or timeout still fails the nightly job and opens another issue. Can you add `--retry 3`?

build/test-cli-download.sh
```diff
@@ -65,21 +65,24 @@ echo "Fetching latest release version from GitHub API..."
...
-RAD_VERSION=$(echo "$api_response" | grep "tag_name" | grep -v rc | awk 'NR==1{print $2}' | sed -n 's/"\(.*\)",/\1/p')
+# Extract version from API response using grep, awk, and sed.
+# `|| true` stops a non-matching grep (e.g. an API error response with no
+# "tag_name") from tripping `set -e` here, before the empty-value check below
+# can report a useful error.
+RAD_VERSION=$(echo "$api_response" | grep "tag_name" | grep -v rc | awk 'NR==1{print $2}' | sed -n 's/"\(.*\)",/\1/p') || true
 
 if [ -z "$RAD_VERSION" ]; then
-    echo "Failed to extract RAD_VERSION from API response"
+    printf 'Failed to extract RAD_VERSION from API response:\n%s\n' "$api_response" >&2
```
kachawla: On success this endpoint returns about 1 MB (roughly 15k lines), so a parse failure on a valid response dumps all of it into the log. Can you cap it, for example with `${api_response:0:2000}`?

PR #12972, opened 2026-09-14
test(functional): retry post-deploy port-forward probe with backoff

### LINKED ISSUE #12935: Flaky test: Test_FirstApplicationSample exhausts HTTP retries before application starts listening
### Steps to reproduce

Run the scheduled noncloud functional tests, including `Test_FirstApplicationSample` in `test/functional-portable/samples/noncloud/tutorial_test.go`. The failure is intermittent and occurs when the tutorial application takes longer to start listening than the post-deployment verification allows.

Observed in the September 6, 2026 [workflow run 34003702190](https://github.com/radius-project/radius/actions/runs/34003702190), [samples-noncloud job](https://github.com/radius-project/radius/actions/runs/34003702190/job/101407680932), originally reported by #12913.

### Observed behavior

Deployment and Kubernetes validation succeed, but the post-deployment HTTP verification exhausts three immediate retries before the application starts listening on port 3000:

```text
[rad] Deployment Complete
Pod is ready tutorial:demo-tutorial-58f5675c5f-9fxcw
running post-deploy verification
01:35:39.530092 ... localhost:3000 ... connection refused
01:35:39.566572 ... localhost:3000 ... connection refused
01:35:39.594423 ... localhost:3000 ... connection refused
Get "http://localhost:34267/": EOF
tests failed after 3 retries
```

The [container-log artifact](https://github.com/radius-project/radius/actions/runs/34003702190/artifacts/9980451296) records:

```text
2026-09-06T01:35:40.846732302Z ... Server is running at http://localhost:3000
```

All three requests fail within approximately 64 milliseconds; the application starts listening approximately 1.25 seconds after the final attempt. Kubernetes pod readiness here does not establish that the application's HTTP endpoint is accepting connections.

### Desired behavior

Post-deployment verification should allow a bounded application-startup interval, using context-aware readiness polling or retry backoff rather than exhausting all attempts immediately. Failed port-forward sessions should be recreated as needed. A genuinely unavailable application must still fail within a clear timeout and retain actionable diagnostics.

Consider an application HTTP readiness probe as well, rather than treating pod readiness alone as sufficient for the HTTP check.

### Workaround

Rerun the failed job. The next scheduled [September 6, 13:07 UTC run](https://github.com/radius-project/radius/actions/runs/34035126893/job/101492490355) passed the samples job at the same Radius commit. This demonstrates intermittency, not a code fix.

### rad Version

CLI built from Radius commit `c8ad9211a25699c377c45268890e4f67070aa114` in the failing workflow; released version output was not captured in this triage.

### Operating system

Linux GitHub Actions runner with a local Kubernetes test cluster; see the linked job for its runner configuration.

### Additional context

- [Retry implementation at the failing commit](https://github.com/radius-project/radius/blob/c8ad9211a25699c377c45268890e4f67070aa114/test/functional-portable/samples/noncloud/tutorial_test.go#L108-L122).
- Original generated failure issue: #12913.
- Merged PR #12602 adds analogous retry handling to gateway tests, but does not modify this sample test.
- Merged PR #12697 addresses a different sample/resource-type mismatch and was already included in this failing run; it does not resolve this startup race.

pujitha24:
## Summary

`Test_FirstApplicationSample`'s post-deploy HTTP verification retried the
port-forward probe exactly 3 times with no delay between attempts. This
change replaces the fixed-count loop with one bounded by the already-declared
(but previously unwired) `retryTimeout` (1 minute), sleeping `retryBackoff`
(1 second) between failed attempts.

## Reason for change

A failing scheduled run showed all 3 retry attempts completing within ~64ms
of each other, while the container log showed the app's HTTP server didn't
start listening until ~1.25s after the last attempt. Kubernetes pod readiness
doesn't guarantee the app's HTTP endpoint is already accepting connections,
so the tight, backoff-free retry loop exhausted its budget before the app
came up. The file already had `retryTimeout` and `retryBackoff` constants
declared for exactly this kind of bounded retry, but only the separate
`retries = 3` constant was actually used by the loop — the backoff/timeout
constants were dead code.

Fixes #12935

## How to test

`Test_FirstApplicationSample` requires `RADIUS_SAMPLES_REPO_ROOT` (a checkout
of radius-project/samples) and a live Kubernetes cluster, so it is skipped
otherwise and could not be executed end-to-end in this environment. What was
run locally and passed:

```bash
go build ./test/functional-portable/samples/...
go test -vet=off -c -o /dev/null ./test/functional-portable/samples/noncloud/...
go vet ./test/functional-portable/samples/noncloud/...
gofmt -l test/functional-portable/samples/noncloud/tutorial_test.go
golangci-lint run ./test/functional-portable/samples/noncloud/...
```

`golangci-lint` reported "0 issues" and the full test binary compiled and
vetted cleanly, confirming the new loop type-checks correctly against the
existing `rp.RPTest`/`testWithPortForward` helpers. A reviewer can verify the
actual fix by running the samples functional job (`make
test-functional-samples-noncloud` with the prerequisites above) and checking
that the port-forward probe no longer exhausts all attempts within
milliseconds of pod readiness.

Note: this reduces the specific race demonstrated by the observed CI
failure; it doesn't guarantee no flakiness if the app ever takes longer than
a minute to start listening.

## File change summary

| File | Summary of change |
| ---- | ----------------- |
| `test/functional-portable/samples/noncloud/tutorial_test.go` | Removed the unused `retries` constant; replaced the count-based port-forward retry loop with one bounded by `retryTimeout` and paced by `retryBackoff` between attempts; the last error is now included in the failure message. |


test/functional-portable/samples/noncloud/tutorial_test.go
```diff
@@ -106,20 +105,27 @@ func Test_FirstApplicationSample(t *testing.T) {
...
-				for i := 1; i <= retries; i++ {
-					t.Logf("Setting up portforward (attempt %d/%d)", i, retries)
+				// Set up pod port-forwarding for the pod. Kubernetes pod readiness does not
+				// guarantee the application is already listening on remotePort, so retry
+				// across a bounded window with a backoff between attempts instead of
+				// exhausting all attempts within milliseconds of each other.
+				// See https://github.com/radius-project/radius/issues/12935.
+				deadline := time.Now().Add(retryTimeout)
+				var lastErr error
+				for attempt := 1; time.Now().Before(deadline); attempt++ {
```
lakshmimsft, line 122: The retry is not idempotent. testWithPortForward  creates a TODO item, so it can't be safely re-run. If an attempt fails after the POST at line 215 succeeds, the next attempt's empty-list check at line 204 sees the leftover item and fails the test outright instead of retrying. The wider retry window makes this more likely than before. Consider retrying only a read-only readiness check ( GET / ), then running the stateful verification once.

pujitha24:
Good catch, that's a real bug in the retry. I split the loop: it now retries only `waitForPodReady`, a new helper that just port-forwards and does a single idempotent `GET /`, and `testWithPortForward` (the stateful create/update/delete flow) runs exactly once, after the probe confirms the pod is up. So a leftover todo item can no longer trip the next attempt's empty-list check.

lakshmimsft:
@pujitha24 per note above suggest to update pr title: test(functional): retry post-deploy port-forward probe with backoff

pujitha24:
Updated the title to `test(functional): retry post-deploy port-forward probe with backoff` as suggested. Thanks for pointing at the exact wording.

lakshmimsft (approved):

PR #12987, opened 2026-09-16
Scope resourceNameHash to the application and environment

### LINKED ISSUE #750: postgreSqlDatabases: resource-name-scoped hash makes different apps share one Azure server (auth failure, cross-app data loss on delete)
## Summary

`Radius.Data/postgreSqlDatabases` provisions an Azure server named `pgsql-{{context.azure.resourceNameHash}}`. That hash is scoped to the **resource name**, so two different applications that both name their Postgres resource `postgres` resolve to the **same Azure server**, even across different environments.

The result is silent adoption of another application's database server, runtime auth failure, and — on delete — destruction of the other application's data.

## Observed

Two unrelated Radius applications in the same subscription:

| Application | Environment | Created |
|---|---|---|
| `chatbot` | `Chatbot-env` | 19:20:37Z |
| `posthog` | `posthog-env` | 19:53Z |

Both resolved to the Azure server `pgsql-f98a050c757a71d1`.

`chatbot` created it first, with `administratorLogin: myadmin`. The `posthog` model declared `username: 'posthog'`, but Azure Flexible Server cannot change `administratorLogin` after creation, so the value was silently ignored and the existing `myadmin` login was kept. The deploy reported success.

Every `posthog` workload then failed at runtime:

```
django.db.utils.OperationalError: connection failed:
connection to server at "20.59.23.70", port 5432 failed:
FATAL:  password authentication failed for user "posthog"
```

Confirmed by inspecting pod specs across the cluster: 17 pods referenced `pgsql-f98a050c757a71d1`, 16 belonging to `posthog` and one to `chatbot`.

## Data loss

Running the delete-application workflow for `posthog` removed that Azure Postgres server. Because the server was shared, **deleting one application destroyed a different application's database**, in a different environment, with no warning.

This is the most serious part of the issue: an application delete is expected to be scoped to that application.

## Why this is systemic, not a one-off

The `radius-app-bicep` skill's deterministic naming rules require a data store to be named after its engine:

> | Data store | Engine short name in kebab-case (`'mysql'`, `'postgres'`, `'neo4j'`, `'redis'`) |

So every application generated by that skill declares `name: 'postgres'` for its PostgreSQL resource. Combined with a resource-name-scoped hash, **every generated application collides on one Azure server per subscription**. The two applications above collided for exactly this reason, without either author doing anything unusual.

## Suggested fixes

Any of these would close it; the first is the most direct.

1. **Include application and environment in the hash input**, so `pgsql-<hash>` is unique per application/environment/resource rather than per resource name.
2. **Detect adoption of a pre-existing server and fail loudly.** If the server exists and was not created by this application, stop with a clear message instead of silently reusing it.
3. **Surface the effective `administratorLogin` as a read-only output.** When Azure ignores the requested login, the application should be able to wire the real value rather than a literal that will not authenticate.
4. **Scope delete to resources this application provisioned**, so a shared or adopted resource is never destroyed on another application's delete.

## Impact

- Silent cross-application coupling of a stateful resource.
- A deploy that reports success and cannot authenticate at runtime, with an error that points at the credential rather than the collision.
- Cross-application, cross-environment data loss on delete.

kachawla:
## Description

`context.azure.resourceNameHash` seeds from the Azure provider scope and the Radius resource ID. That resource ID is scoped to a UCP resource group and identifies the *resource*, not its *owner*. The application and environment are sibling resources referenced through properties, never segments of the ID. So the hash carries no application or environment identity.

Two applications in one Azure scope that reuse a common resource name therefore resolve to the same cloud resource, as does one application deployed to two environments.

Common resource names make this the default case rather than an edge case. Two applications that each name a database `postgres` get one Azure Flexible Server. The second deploy adopts the first's server and reports success, while Azure ignores the `administratorLogin` it asked for, because that login can only be set when the server is created. Every workload in the second application then fails at runtime:

```
django.db.utils.OperationalError: connection failed:
FATAL:  password authentication failed for user "posthog"
```

The error points at the credential rather than the collision. And because the server is shared, deleting either application runs its recipe's delete path against the other application's data and destroys it. That is a cross-application, cross-environment data loss from an operation users expect to be scoped to one application.

This was observed with two unrelated applications in different environments both resolving to `pgsql-f98a050c757a71d1`; 17 pods across the cluster referenced that one server.

## Change

Add the application and environment IDs to the seed:

```go
seed := strings.ToLower(azureScopeID) + "\x00" +
    strings.ToLower(recipeContext.Application.ID) + "\x00" +
    strings.ToLower(recipeContext.Environment.ID) + "\x00" +
    strings.ToLower(recipeContext.Resource.ID)
```

Both fields are already populated before the Azure block runs. A resource deployed without an application contributes an empty segment, which is a distinct seed rather than a collision with any named application.

The alternatives considered were worse. Making callers encode the application into the Radius resource name only helps models a given generator produces (hand-authored Bicep still collides) and pushes an infrastructure naming constraint into user-facing names. Varying the UCP resource group changes something callers reasonably treat as fixed, and still leaves two applications in one group colliding.

## Type of change

## Breaking change

This changes every name derived from `resourceNameHash`. On the first redeploy an existing deployment provisions a replacement and **the previous cloud resource is deleted**. The Bicep driver garbage-collects outputs absent from the new deployment (`pkg/recipes/driver/bicep/bicep.go:213-230`), and `getGCOutputResources` (`:564-585`) marks each one `RadiusManaged`, so the guard in `Delete` does not filter them. For stateful types the data goes with them.

Worth stating plainly, because the existing documentation understates it: **the previous revision of this feature behaved the same way.** Moving from `context.resource.name` to `resourceNameHash` also renamed resources and so also deleted the originals. `recipe-packs/azure/README.md` describes that only as deployments "provisioning replacement resources." This change is therefore no more destructive than the one already shipped in v0.60.0. I am surfacing the mechanism so the decision is made with it in view rather than around it.

Affected types in the Azure pack: Redis (`amr-`), OpenAI (`oai-`), Search (`srch-`), Cosmos (`cosmon-`), MySQL, PostgreSQL, SQL Server, Event Hubs (`evhns-`), Storage (`st`). Only the Bicep driver garbage-collects; the Terraform driver receives `PrevState` and ignores it.

Reasons to accept it regardless:

- **Compatibility here preserves a data-loss bug.** The current names are not a contract worth keeping: a name that is "stable" under today's rule is stable *onto a resource another application may own*.
- **A rename is unavoidable for affected users.** If two applications share a server, at least one must move.

One design consequence reviewers should weigh: with the owner in the seed, the name changes whenever the owner changes, so moving a resource between environments or changing its application association becomes a replace rather than an update. The pinning option below removes that too.

## If reviewers want backwards compatibility instead

This is implementable without much more work, so please say so in review rather than treating the above as load-bearing.

Pin the hash as durable state instead of recomputing it:

- Add `nameHash` and `nameHashVersion` to `ResourceStatus.Recipe`, beside the existing `TemplateKind` / `TemplatePath` / `TemplateVersion`.
- On execute: stored hash → use it; else evidence of a prior successful deployment → legacy seed, stored as v1; else → owner-scoped seed, stored as v2.
- Prefer durable recipe status over `len(PrevState) != 0` as that evidence, and fail closed when it is ambiguous, because misreading it re-provisions and then garbage-collects a live resource.

Existing resources keep their names, new ones are correctly scoped, owner changes stop being replacements, and pinning makes this the last time a seed change can rename anyone's infrastructure. Caveat: already-colliding deployments stay colliding until fixed by hand.

## Contributor checklist

- [x] Existing tests updated: four golden hash values in `context_test.go` and one in `paramresolver/resolver_test.go` changed, which is the intended behavior change.
- [x] New tests added: `changes with the application` (including the application-less case), `changes with the environment`, and `is stable for the same owner`. The last one pins that a redeploy of the same owner keeps its name, so this does not introduce churn.
- [x] Design note `eng/design-notes/recipes/2026-05-direct-recipe-modules.md` updated to describe the new seed and why the owner is part of it.
- [x] `go build`, `go vet`, `gofmt`, and `go test ./pkg/recipes/...` all clean locally.


pkg/recipes/recipecontext/context.go
```diff
@@ -109,7 +109,23 @@ func New(metadata *recipes.ResourceMetadata, config *recipes.Configuration) (*Co
...
+			// rather than an edge case: two applications that each call a database
+			// `postgres` resolve to one server, the second silently adopting the
+			// first's data with an administrator login Azure cannot apply to an
+			// existing server, and deleting either destroys the other's data.
+			// Including both IDs scopes the name to the owner the resource is
+			// actually deployed for.
+			seed := strings.ToLower(azureScopeID) + "\x00" +
+				strings.ToLower(recipeContext.Application.ID) + "\x00" +
+				strings.ToLower(recipeContext.Environment.ID) + "\x00" +
+				strings.ToLower(recipeContext.Resource.ID)
```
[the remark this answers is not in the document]
kachawla: Fixed in c4c756a. The field comment now covers all three: the lowercased seed components, the empty application segment for a resource deployed without an application, and the stability consequence.

```go
// ResourceNameHash is a short, stable hash used to generate globally unique
// resource names. It is derived from the lowercased Azure provider scope, application ID,
// environment ID, and resource ID. A resource deployed without an application contributes
// an empty application segment, which is a distinct seed rather than a collision with any
// named application. Because the owner is part of the seed, the hash changes if the
// resource's application or environment changes.
```

The last sentence matters most in the exported docs. The previous semantics implied a name was stable for the life of the resource, and it now tracks ownership instead. The PR description calls out the same consequence, since it means moving a resource between environments becomes a replace rather than an update.

nicolejms:
The copilot extension picks up the latest version of radius. If the plug-in is not updated and picks up v0.61 with this change will anything break?

pkg/recipes/recipecontext/context.go
```diff
@@ -109,7 +109,14 @@ func New(metadata *recipes.ResourceMetadata, config *recipes.Configuration) (*Co
 
 		// Use the configured Azure scope so subscription-scoped recipes also receive a stable hash.
 		if subID != "" && recipeContext.Resource.ID != "" {
-			seed := strings.ToLower(azureScopeID) + "\x00" + strings.ToLower(recipeContext.Resource.ID)
+			// The application and environment are part of the seed because the
+			// resource ID is scoped to a UCP resource group and names the resource,
+			// not its owner. Without them, two applications that reuse a resource
+			// name in one Azure scope resolve to the same cloud resource.
+			seed := strings.ToLower(azureScopeID) + "\x00" +
```
nicolejms, line 116: The four-line change is correct and the tests pin the right properties. My concern is the migration, which this PR leaves open.

Two things worth making explicit before merge:

1. **Rollback is destructive in both directions.** `getGCOutputResources` (bicep.go:628-649) stamps every previous resource absent from the new deployment with `RadiusManaged: true`, and the guard in `Delete` (bicep.go:268-271) only skips resources where that flag is nil or false. So the old server is deleted on upgrade, and reverting this commit deletes the *new* server on the next deploy. There is no safe revert once a user has redeployed, only a second replacement.

2. **The failure is silent.** The deploy reports success and the workload fails later against a resource that no longer exists. Nothing in the logs names the rename as the cause.

You offered the pinned-hash alternative (`nameHash` / `nameHashVersion` in `ResourceStatus.Recipe`) and asked reviewers to choose. I think that choice has to be made deliberately rather than by merging. If we take the rename, it needs a migration note; if we take the pinning, it also removes the owner-change-becomes-replace cost you call out. Either is defensible. Merging without answering is not.

pkg/recipes/recipecontext/context.go
```diff
@@ -109,7 +109,14 @@ func New(metadata *recipes.ResourceMetadata, config *recipes.Configuration) (*Co
...
+				strings.ToLower(recipeContext.Application.ID) + "\x00" +
+				strings.ToLower(recipeContext.Environment.ID) + "\x00" +
```
nicolejms, line 118: The seed takes these IDs as raw strings with only `strings.ToLower` applied. Any change to the ID *format* re-hashes and therefore replaces the cloud resource.

That matters right now because the repo carries both application namespaces: `cli/cmd/deploy/deploy_test.go:2236` expects `Applications.Core/applications/myapp` and `:2262` expects `Radius.Core/applications/myapp`. If a resource's stored application ID migrates from one namespace to the other, this hash changes and the Bicep driver garbage-collects the old resource, with no user-visible cause.

The hazard already existed for `Resource.ID`, so this widens an exposure rather than creating one. But it widens it during a live namespace migration. Either normalize the provider namespace before hashing, or document that callers own ID-format stability for these two fields.

pkg/recipes/recipecontext/context_test.go
```diff
@@ -532,6 +532,59 @@ func TestNewContext_ResourceNameHash(t *testing.T) {
...
+		}
+
+		const envPrefix = "/planes/radius/local/resourceGroups/test-group/providers/Applications.Core/environments/"
+		require.NotEqual(t, withEnvironment(envPrefix+"dev"), withEnvironment(envPrefix+"prod"))
+	})
+
+	// The same resource, application, and environment must keep its name across
+	// redeployments; otherwise every deploy would provision a replacement beside
+	// the resource it already owns.
+	t.Run("is stable for the same owner", func(t *testing.T) {
```
nicolejms, line 577: This is the most valuable of the three new subtests, and I would extend it one step.

Nothing here covers the transition that this PR actually causes: a resource whose name was computed under the old seed, redeployed under the new one. That path is the whole breaking change, and right now it rests on code reading rather than on a test. A case that asserts the legacy seed and the owner-scoped seed produce different hashes for the same metadata would pin the behavior and give the eventual pinning work (if we go that route) something to build against.

pkg/recipes/paramresolver/resolver_test.go
```diff
@@ -346,7 +346,7 @@ func Test_ResolveParameterExpressions_GeneratedAzureResourceNameHash(t *testing.
 		ctx,
 	)
 
-	require.Equal(t, map[string]any{"name": "redis-2ee7ced618d5dd8e"}, result)
+	require.Equal(t, map[string]any{"name": "redis-509212db4b1fe144"}, result)
```
nicolejms, line 349: This is the fifth hand-maintained golden hash across two packages (the other four are in `recipecontext/context_test.go` at lines 97, 190, 461, and 482). Any future seed change means editing all five by hand and hoping none is missed.

Consider deriving the expected value from the documented rule in a small test helper instead of hardcoding it, so the next seed change fails in one place with a clear message. Not worth holding the PR for, but it will bite whoever changes this next.

eng/design-notes/recipes/2026-05-direct-recipe-modules.md
```diff
@@ -224,7 +224,7 @@ Use any standard Bicep or Terraform module directly as a Recipe by pointing `sou
...
+The `{{context.*}}` expression system resolves Radius application context into Recipe parameters at deploy time. It supports resource metadata, application and Environment data, Kubernetes runtime data, Azure and AWS provider data, and single-level ternary expressions. For Azure resources whose names must be globally unique, `{{context.azure.resourceNameHash}}` returns the first 16 lowercase hexadecimal characters of the SHA-256 hash of the lowercased Azure provider scope ID, the application ID, the environment ID, and the Radius resource ID, NUL-separated. The provider scope is the resource-group ID when configured and otherwise the subscription ID. The application and environment are part of the seed because the resource ID is scoped to a UCP resource group and names the resource rather than its owner: hashing it alone gives two applications that reuse a common resource name in one Azure scope, or one application deployed to two environments, the same name for a globally named cloud resource. The application ID is empty for a resource deployed without one, which is itself a distinct seed.
```
nicolejms, line 227: This paragraph now describes the new seed accurately, but it says nothing about what happens to deployments that already exist. Given the change replaces and deletes live Azure resources on the first redeploy, the upgrade consequence belongs here rather than only in the PR description.

Two specifics to include:

- The affected types in the Azure pack (Redis, OpenAI, Search, Cosmos, MySQL, PostgreSQL, SQL Server, Event Hubs, Storage) and the fact that stateful data goes with the replaced resource.
- The Terraform case. The PR description says "only the Bicep driver garbage-collects," which is true about GC, but the recipe context reaches Terraform on both paths in `pkg/recipes/terraform/execute.go:382` and `:400`. Where the hash lands on a ForceNew attribute, Terraform destroys and recreates under its own plan. The data loss reaches Terraform recipes too, just by a different route, and the current wording reads as though it does not.

You are right that the v0.60.0 move to `resourceNameHash` did the same thing and that `recipe-packs/azure/README.md` understated it. Let us not repeat that understatement here.

nicolejms (commented):
Reviewed the change end to end and ran `go test ./pkg/recipes/recipecontext/... ./pkg/recipes/paramresolver/...` on c4c756a, which passes.

The diagnosis is good and the fix is correct. Worth noting that the collision was already visible in this repo's own fixtures: on `main`, `TestNewContext` (application `testApplication`) and `TestNewContext_ResourceNameHash` (no application) share a resource ID and Azure scope and both asserted `90af35ee55457d03`. The test file encoded the bug as expected behavior.

I am not asking for changes to the implementation. My comments are about the migration decision you put to reviewers, ID normalization, one test gap, and the docs. Details inline.

AzureMike:
LGTM.

Nit:
Existing resources can get new names on redeploy, including when an unchanged plug-in picks up the new Radius version. In my Kubernetes recipe tests, Bicep orphaned the old resource and Terraform replaced its resource. I think this is the right fix if we accept the name changes, but the PR description shouldn't promise that old resources are always deleted.

nicolejms (approved):
address all comments in followup

